PR #115, opened 2017-03-02
Fixed Protein constructor for FASTA reading. Writing XMLs and FASTAs.

acesnik:
Also, reading Ensembl pep.all.fasta files. Writing GoTerms with XMLs.


mzLib/Proteomics/Protein.cs
```diff
@@ -15,9 +15,12 @@ public Protein(string sequence, string accession, string name, string full_name,
             FullName = full_name;
             IsDecoy = isDecoy;
             IsContaminant = isContaminant;
+            ProteolysisProducts = new List<ProteolysisProduct>();
+            OneBasedPossibleLocalizedModifications = new Dictionary<int, List<Modification>>();
+            GoTerms = new List<GoTerm>();
```
acesnik: - This solves the problem of null reference exceptions in trying to write XMLs from data read from a FASTA originally. 
- These are not recorded in the FASTAs we read currently, but they might be in the future. 
- Intuitively, there is no reason why these proteins shouldn't contain these properties, so empty sets are more appropriate than null ones.

mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -288,13 +289,21 @@ public static List<Protein> LoadProteinDb<T>(string proteinDbLocation, bool onTh
...
                                 name = accession;
                                 full_name = output[3];
                             }
+                            else if (output_ensembl.Length > 2)
+                            {
+                                accession = output_ensembl[1];
+                                name = accession;
+                                full_name = String.Join(" ", Enumerable.Range(2, output_ensembl.Length - 2).Select(i => output_ensembl[i]));
+                            }
                             else
```
acesnik: Yes, and widely used by people in proteogenomics, peptidomics, and genomics in particular. ftp://ftp.ensembl.org/pub/release-87/fasta/homo_sapiens/pep/

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        #region Public Methods
+
+        public static void WriteXmlDatabase(Dictionary<string, HashSet<Tuple<int, ModificationWithMass>>> Mods, List<Protein> proteinList, string outputFileName)
+        {
+            var xmlWriterSettings = new XmlWriterSettings
+            {
+                Indent = true,
+                IndentChars = "  "
+            };
```
acesnik: Included in MetaMorpheus GPTMD code. This initialization makes the XML more human interpretable by indenting the tree structure.

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            using (XmlWriter writer = XmlWriter.Create(outputFileName, xmlWriterSettings))
+            {
+                writer.WriteStartDocument();
+                writer.WriteStartElement("uniprot");
```
acesnik: Standard for UniProt-XML structure.

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        public static void WriteFastaDatabase(List<Protein> proteinList, string outputFileName)
+        {
+            using (StreamWriter writer = new StreamWriter(outputFileName))
+            {
+                foreach(Protein protein in proteinList)
+                {
+                    string header = protein.FullName.Contains("|") ?
+                        protein.Accession + "|" + protein.FullName :
```
acesnik: Hack = only applicable to one circumstance? Maybe, but I've only ever seen the pipe used in UniProt FASTAs. Almost every other FASTA I've seen uses whitespace to separate fields. The only requirement of FASTA headers to my knowledge is the ">" starting symbol and some unique identifier following.

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
stefanks: Another way of thinking about this is that this is the name of the entity that generated the database. Let's change this to "mzLib" by default, and allow passing a string parameter that stands for the original source database, that would be concatenated with mzLib. 
So it could be something like:
"mzLib"  or "mzLib and uniprot"  or "mzLib and SomeOtherSource". 
We don't even know if we are using the real uniprot xml structure: we're not using an xsd file, and just writing stuff to a file that looks kindof like uniprot xml

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
stefanks: Ok

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
stefanks: So what I don't like about this is the idea that our code can output fasta headers in two different formats, depending on some minute detail in the protein full name. 
Why can't we pick one or the other? Space or | for outputting?

mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
stefanks: And you are comfortable determining the file type based on the number of elements in a split string? It looks a little arbitrary.

mzLib/Proteomics/Protein.cs
stefanks: ok

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
stefanks: On the other hand, "the first element to appear is called the "root element", which defines what kind of document the file will be. " http://www.tizag.com/xmlTutorial/xmldocument.php

Shouldn't we just call it mzLibProteinDb?

acesnik: Good point. Let's change it to mzLib, since people should expect to use our reader for these files anyway.

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
stefanks: So what I don't like about this is the idea that our code can output fasta headers in two different formats, depending on some minute detail in the protein full name.
Why can't we pick one or the other? Space or | for outputting?

PR #116, opened 2017-03-04
Writing the relevant PTM list at the beginning of the XML structure

acesnik:


acesnik:
The idea here is that information regarding PTMs referenced in the protein entries should be contained within the database. This would make it easier to create a database using mzLib in one program and read it in the next using mzLib without needing to carry along those text files. Also, this may make database searches more reproducible because the PTM list regarding that protein list will not change.

stefanks:
So is the idea that the new protein database format, mzLibProteinDb always includes the information about all modifications mentioned in the file? So this will make it unnecessary for proteofom suite to read in ptmlist formatted files? 

What about the passing into the protein db reader a list of known modifications, does that go away now? Some passing must be done for Uniprot XMLs, because they do not include modification lists.

What about the capability of selecting pre specified modifications of interest? Currently you just pass to the reader the modifications that are known. With the proposed change, how could you tell the protein reader that you are only interested in a certain kind of modifications? 

Well need to talk more about this when I'm back on Tuesday.

acesnik:
Sounds great about discussing this on Tuesday.

The dictionary of known modifications stays. I selected the modifications from the list of known modifications pertaining to proteins in the protein list, and then concatenated that with the modifications inside the protein objects. It writes the unique modifications from that combined list.

Preselecting modifications should probably be done before calling the writer, e.g. OfType<ModificationsWithMass>() prior to requesting they be written into protein entries. I didn't change the inputs of the writer.

stefanks:
I was thinking about the protein reader, which now would not need to have information from ptmlists, right?
I mean the whole purpose is for the reader to be able to read protein databases without the need to refer to any additional files, such as modification lists.

An option could be to always require and assume that the Uniprot ptmlist exists/is known, and these modifications would not need to be pre specified in XML, but the rest of mods, such as the new ones would need to be pre-specified. The alternative, which is to require every modification to be written in the file would be nice, but would necessarily require two different readers: one for Uniprot format which does not have mods in the for, and another for mzLibProteinDb files.

acesnik:
Summary of proposed additional changes:
1. GetPtmListFromXml method that retuns IEnumerable<Modification>
2. Method to compare two lists of Modifications and return a list with unique Modification elements, based on the fields of the Modifications, e.g. chemical formula, monoisotopic mass.
3. IEnumerable<string> mod_types input to decide which types of modifications to add into Protein objects

acesnik:
Implemented the proposed changes:
1. Implemented. Returns the same objects used to construct the protein list if they have already been generated. Returns a new list if a different valid XML is specified.
2. New comparer and hash functions for each of the Modification classes, so we can collapse the prespecified ones in the XML with any additional ones specified in the input.
3. Implemented modification_type in ModificationWithLocation and allowed the option to _exclude_ certain modification types.

PR #119, opened 2017-03-10
inline test

stefanks:


acesnik:
Cool, you're trying out some new .NET features?

PR #138, opened 2017-03-27
added mzidentml properties needed

lschaffer2:
small fix


stefanks:
Could you remove the unused method GetLastNumberFromString? Otherwise it looks good!

PR #139, opened 2017-03-28
Move modificationType to Modification, internal renaming

stefanks:


Proteomics/Modification.cs
```diff
@@ -8,14 +8,16 @@ public class Modification
         #region Public Fields
 
         public readonly string id;
+        public readonly string modificationType;
```
acesnik: Nice. I like this change.

PR #146, opened 2017-04-03
 Organize code a bit, make the default list of diagnostic ions to be an empty list instead of null. parse with invariant culture

stefanks:


Proteomics/ModificationWithLocation.cs
```diff
@@ -64,20 +64,11 @@ public override bool Equals(object o)
             return m == null ? false :
 
                base.Equals(m)
-               && (this.accession == null && m.accession == null
```
acesnik: Are there modifications that are the same in UniProt with different accession numbers?

Test/sampleModFileDouble2.txt
```diff
@@ -3,5 +3,6 @@ TG   Nxs or Nxt
 PP   Anywhere.
 MM   15.995
 NL   0 or 15
+OM   343
```
acesnik: What does this field represent?

acesnik: The masses seen in the histogram, it looks like. How is that different than the MM?

acesnik:
As another note, "uniprot"
is not capitalized in the modification fields of the search window and such. Is that due to the accession being labeled "uniprot" instead of "Uniprot" in PtmListLoader?

Proteomics/ModificationWithLocation.cs
stefanks: I'd like modifications to be equal if they have the same properties, i.e. mass, chemical formula, etc. So even with different accessions, modifications might be equal. 
The reason to make this change is if we writing a modification from uniprot and then read it, the way it was those would not be equal, since the uniprot accession is not written. If we don't compare that, then writing and reading modification back would be equal to the original.

Test/sampleModFileDouble2.txt
stefanks: Yea, it is different because modifications observed on C have a histogram peak at a value different than MM, due to the fixed modification. Now I'm thinking this should not be a property of a modification, but rather handled in the gptmd code in metamorpheus. I'll remove the OM field completely. What do you think?

Proteomics/ModificationWithLocation.cs
stefanks: Actually, following that logic, I should remove the comparison of id's as well...

stefanks:
Yeah, that's the reason. Do you think it should be capitalized?

Proteomics/ModificationWithLocation.cs
acesnik: Okay, that makes sense. We are considering modifications to be unique chemical entities.

Test/sampleModFileDouble2.txt
acesnik: Sounds good.

Issue #161, opened 2017-04-22
Serializability


acesnik:
I would like to serialize Proteomics objects (e.g. ModificationWithMass, Protein). Could we tag them with the `[Serializable]` attribute?

stefanks:
sure!

stefanks:
Done in https://github.com/smith-chem-wisc/mzLib/pull/162

PR #162, opened 2017-04-23
Added serialization attribute to classes associated with Protein and Modification objects

acesnik:


acesnik:
https://github.com/smith-chem-wisc/mzLib/issues/161

PR #164, opened 2017-04-24
Fix duplicate modification writing

stefanks:


stefanks:
Intended to fix https://github.com/smith-chem-wisc/MetaMorpheus/issues/367

PR #180, opened 2017-05-11
Read/Write sequence variations from XML

acesnik:


acesnik:
FilesLoading test is failing...
```
Message:   Expected string length 134 but was 130. Strings differ at index 93.
  Expected: "...0601\r\nDR   PSI-MOD; MOD:01956\r\nTG   R\r\nMM   15.994915\r\nCF   O"
  But was:  "...0601\r\nDR   PSI-MOD; 01956\r\nTG   R\r\nMM   15.994915\r\nCF   O"
  ------------------------------------^
```

rmillikin:
this test wouldn't pass for me yesterday when I submitted my changes, it expected the "MOD:" to *not* be there, so I added "MOD:" to the hardcoded string to test against. now it seems to be doing the opposite for you?

stefanks:
Could you test the writing and then reading with begin/end positions? It's not covered by your tests yet. Also reading generating decoy sequence variants if first amino acid is an M

PR #184, opened 2017-06-28
disulfide bonds read in xml

lschaffer2:
need this info for samples where we didn't reduce (insulin)

### DIFF
--- a/UsefulProteomicsDatabases/ProteinDbLoader.cs
+++ b/UsefulProteomicsDatabases/ProteinDbLoader.cs
@@ -245,7 +245,10 @@ namespace UsefulProteomicsDatabases
                                         }
                                         else if (feature_type == "disulfide bond")
                                         {
-                                            disulfideBonds.Add(new DisulfideBond((int)oneBasedbeginPosition, (int)oneBasedendPosition, feature_description));
+                                            if (oneBasedbeginPosition != null && oneBasedendPosition != null)
+                                                disulfideBonds.Add(new DisulfideBond((int)oneBasedbeginPosition, (int)oneBasedendPosition, feature_description));
+                                            else if (oneBasedfeature_position >= 1)
+                                                disulfideBonds.Add(new DisulfideBond(oneBasedfeature_position, feature_description));
                                         }
                                         oneBasedbeginPosition = null;
                                         oneBasedendPosition = null;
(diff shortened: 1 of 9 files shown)


UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -309,6 +315,11 @@ public static List<Protein> LoadProteinXML<T>(string proteinDbLocation, bool onT
                                                     decoyBigPeptideTypes[oneBasedBeginPositions.Count - i - 1] = peptideTypes[i];
                                                 }
 
+                                                foreach (DisulfideBond disulfideBond in disulfideBonds)
+                                                {
+                                                    decoy_disulfides.Add(new DisulfideBond(reversed_sequence.Length - disulfideBond.OneBasedBeginPosition + 1, reversed_sequence.Length - disulfideBond.OneBasedEndPosition + 1, disulfideBond.Description));
```
stefanks: Can you motivate this computation?

UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -242,6 +243,10 @@ public static List<Protein> LoadProteinXML<T>(string proteinDbLocation, bool onT
                                             else if (oneBasedfeature_position >= 1)
                                                 sequenceVariations.Add(new SequenceVariation(oneBasedfeature_position, original_value, variation_value, feature_description));
                                         }
+                                        else if (feature_type == "disulfide bond")
+                                        {
+                                            disulfideBonds.Add(new DisulfideBond((int)oneBasedbeginPosition, (int)oneBasedendPosition, feature_description));
```
stefanks: Can oneBasedbeginPosition or oneBasedendPosition ever be null?

lschaffer2: Thanks for catching this - yes. I'll account for this.

UsefulProteomicsDatabases/ProteinDbLoader.cs
lschaffer2: I thought for decoys the disulfides should be at the same cysteine positions in the reverse sequence

stefanks: Could you add this to a unit test?

PR #201, opened 2017-07-25
mzML Files changed to comply with mzML format specifications

dippman:
mzML Files changed to comply with mzML format specifications. Allows use of third party software to open and view mzML files generated by mzLib. 

Added Chromatogram generation


stefanks:
I don't think the changes in Mzml.cs and Test.csproj are necessary!

PR #220, opened 2017-08-04
Capability to merge protein entries

acesnik:


UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -515,6 +515,64 @@ public static List<Protein> LoadProteinFasta(string proteinDbLocation, bool onTh
             return result;
         }
 
+        /// <summary>
+        /// Merges proteins of the same accession and sequence. Considers contaminants separately.
```
stefanks: Merge proteins that have the same accession, sequence and contaminant designation

UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -515,6 +515,64 @@ public static List<Protein> LoadProteinFasta(string proteinDbLocation, bool onTh
...
+        /// </summary>
+        /// <param name="merge_these"></param>
+        /// <returns></returns>
+        public static IEnumerable<Protein> merge_proteins(IEnumerable<Protein> merge_these)
+        {
+            Dictionary<Tuple<string, string, bool, bool>, List<Protein>> proteins_by_accession_seq_cont_isdecoy = new Dictionary<Tuple<string, string, bool, bool>, List<Protein>>();
```
stefanks: the decoy in the name of this variable is confusing

UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -515,6 +515,64 @@ public static List<Protein> LoadProteinFasta(string proteinDbLocation, bool onTh
...
+            foreach (Protein p in merge_these)
+            {
+                Tuple<string, string, bool, bool> key = new Tuple<string, string, bool, bool>(p.Accession, p.BaseSequence, p.IsContaminant, p.IsDecoy);
+                if (!proteins_by_accession_seq_cont_isdecoy.TryGetValue(key, out List<Protein> bundled))
+                    proteins_by_accession_seq_cont_isdecoy.Add(key, new List<Protein> { p });
+                else
+                    bundled.Add(p);
+            }
```
stefanks: Should we throw an error if proteins with same accessions have different sequences?

UsefulProteomicsDatabases/ProteinDbLoader.cs
acesnik: 👍 changed this

UsefulProteomicsDatabases/ProteinDbLoader.cs
acesnik: How about just "proteinsByAccessionSequenceContaminant"

acesnik: I agree the decoy designation is a bit confusing... That can be inferred from the code.

UsefulProteomicsDatabases/ProteinDbLoader.cs
acesnik: Good question. I think this might be intended in some cases. In Rachel K's work, she is hard-coding in sequence variations for different HIV strains, but leaving the accessions the same. Leaving the accessions the same for different sequences is important for quantification, where they should be considered the same "protein."

PR #223, opened 2017-08-08
Remove Benchmark Project

stefanks:


acesnik:
Why do you want to get rid of the benchmark project? I would be fine with that.

PR #225, opened 2017-08-09
Target/decoy parameter names; Modification comparisons

acesnik:


Proteomics/Modification.cs
```diff
@@ -36,12 +36,14 @@ public override string ToString()
...
-            return o != null;
+            return o != null 
+                && m.id == this.id 
+                && m.modificationType == this.modificationType;
         }
 
         public override int GetHashCode()
         {
-            return 0;
+            return id.GetHashCode();
```
acesnik: Note, this won't work if we allow null IDs. Should we disallow null for ID in the constructor?

acesnik: We could also just return 0 here and leave it to the Equals method to iron out.

stefanks:
What's the reason for this change?

acesnik:
Leah's having instances where modifications with different IDs are being merged together, e.g. acetylation and http://pir0.georgetown.edu/cgi-bin/resid?id=AA0069

acesnik:
oops -- * methylation * and http://pir0.georgetown.edu/cgi-bin/resid?id=AA0069

acesnik:
I guess the chemical argument here is that it's possible to have two isobaric modifications with different names and the same motif, which we might want to distinguish.

stefanks:
I guess this makes equal behave as i would expect.

PR #271, opened 2017-10-28
accession fix

acesnik:


stefanks:
What does this fix?

acesnik:
Accessions with periods in them were only being read as the text preceding the period.

PR #290, opened 2018-02-08
use long for sequence variation positions

acesnik:


acesnik:
These are unbounded values in the UniProt XSD, and other packages (DotNetBio) are using long.

stefanks:
Does not compile

acesnik:
Oops, this field refers to the onebasedstart and end of the protein sequence, not the nucleotide sequence. No need for the long, then.

PR #294, opened 2018-02-14
Fix a place

XRSHEERAN:


MassSpectrometry/MsDataFile.cs
```diff
@@ -101,7 +101,7 @@ public static void WindowModeHelper(ref double[] intensities, ref double[] mArra
             //windowNum is the number of windows
```
rmillikin: maybe skip block copy if(filteringParams.NumberOfWindows == 1) ?

XRSHEERAN: Yea, I'll do it right away when I'm back around 4. Thanks!

PR #309, opened 2018-03-27
added chromatography class and CZE calculation of electrophoretic mobility

trishorts:


Chromatography.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+public class Class1
```
acesnik: Unused class. Consider removing.

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+                return (length / (60 * timeMin * voltsPerMeter) * 1e9);
+            else
+                return (-1);
+        }
+
+        public double PredictedElectrophoreticMobility(string peptideSequence, double observedMass)
+        {
+            double predictedMu = 0;
+
+            predictedMu = 3.069 + 386 * Math.Log(1d + 0.35 * PredictedChargeCorrected(peptideSequence)) / (Math.Pow(observedMass, 0.411) + Offset(PredictedChargeCorrected(peptideSequence), peptideSequence.Length));
```
acesnik: Consider splitting line and naming the values, so their purpose is clearer.

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+                case "D":
+                    runningSum -=0.26741;
+                    break;
+                case "E":
+                    runningSum -= 0.06852;
+                    break;
+                case "N":
+                    runningSum += 0.011699;
+                    break;
+                case "Q":
```
acesnik: Could delete this `case`, since it's the same as default.

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+            return (1d + (double)(peptideSequence.Length - substitutedString.Length));
+        }
+
+        private double PredictedChargeCorrected(string peptideSequence)
+        {
+            double runningSum = 0;
+            string internalString = peptideSequence.Substring(3, peptideSequence.Length - 5);
+
+
+            switch (peptideSequence.Substring(0,1))//first AA
```
acesnik: `peptideSequence[0]`

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+                    runningSum += 0.011699;
+                    break;
+                case "Q":
+                    runningSum += 0;
+                    break;
+                default:
+                    runningSum += 0;
+                    break;
+            }
+            switch (peptideSequence.Substring(1, 1))//second AA
```
acesnik: `peptideSequence[1]` is clearer

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+                    runningSum += 0.012535;
+                    break;
+                case "Q":
+                    runningSum += 0.011699;
+                    break;
+                default:
+                    runningSum += 0;
+                    break;
+            }
+            switch (peptideSequence.Substring(2, 1))//third AA
```
acesnik: `peptideSequence[2]` is clearer

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+                    runningSum += 0.016713;
+                    break;
+                case "Q":
+                    runningSum += 0.00585;
+                    break;
+                default:
+                    runningSum += 0;
+                    break;
+            }
+            switch (peptideSequence.Substring(peptideSequence.Length-2, 1))//second to last AA
```
acesnik: `peptideSequence[peptideSequence.Length-2]`

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+                    runningSum += 0.026741;
+                    break;
+                case "Q":
+                    runningSum -= 0.00084;
+                    break;
+                default:
+                    runningSum += 0;
+                    break;
+            }
+            switch (peptideSequence.Substring(peptideSequence.Length - 1, 1))//last AA
```
acesnik: `peptideSequence[peptideSequence.Length - 1]`

Chromatography/CZE.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        public double PredictedElectrophoreticMobility(string peptideSequence, double observedMass)
+        {
+            double predictedMu = 0;
+
+            predictedMu = 3.069 + 386 * Math.Log(1d + 0.35 * PredictedChargeCorrected(peptideSequence)) / (Math.Pow(observedMass, 0.411) + Offset(PredictedChargeCorrected(peptideSequence), peptideSequence.Length));
+            return predictedMu;
+        }
+
+        private double PredictedCharge(string peptideSequence)
```
acesnik: Consider making this return `int` and removing the `double` casting. You're adding it to a `double` below, which is fine.

mzLib.sln
```diff
@@ -1,8 +1,15 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
+<<<<<<< HEAD
```
acesnik: Remove merge conflict notes here

MzIdentML/MzidIdentifications.cs
```diff
@@ -241,7 +243,8 @@ public bool IsDecoy(int sirIndex, int siiIndex)
             {
                 try
                 {
-                    foreach (mzIdentML111.Generated.PeptideEvidenceRefType pe
+
```
acesnik: This class is pretty ugly with all the nested try-catch blocks. If I understand what's going on, each of the mzIdentML classes has the same properties:
```
        private readonly mzIdentML110.Generated.MzIdentMLType110 dd110;
        private readonly mzIdentML111.Generated.MzIdentMLType111 dd111;
        private readonly mzIdentML120.Generated.MzIdentMLType120 dd120;
```
Why not use inheritance to greatly simplify this class and improve our ability to maintain it? Bring them all into the same namespace and make an `abstract class` that they inherit with the same properties, so you can just store this information in one variable, rather than having to try-catch in the three.

acesnik:
Cool idea! Could you also put in class and method comments? You could note the publication you used for the prediction. (Type "///" above them to get the automatically generated comment blocks.)

Chromatography.cs
trishorts: i can't find this anymore

PR #313, opened 2018-03-30
fix window peak trimming

rmillikin:


MassSpectrometry/MsDataFile.cs
```diff
@@ -77,7 +77,7 @@ public int NumSpectra
 
         #region Public Methods
 
-        public static int TopNpeakHelper(double[] intensities, double[] mArray, IFilteringParams filteringParams)
+        public static int TopNpeakHelper(ref double[] intensities, ref double[] mArray, IFilteringParams filteringParams)
```
acesnik: Could you make method summary comments and make a brief comment before the for-loop in WindowModeHelper noting what's going on in the loop?

PR #322, opened 2018-05-21
Refactor datafiles

zrolfs:
-Remove redundancy
-Improve readability
-Facilitate incorporation of future input formats


acesnik:
Could you define your goal here in writing for us?

Also, it looks like AppVeyor was kicking back the "namespace System not found" error, so I tried the previous fix of changing the environment from "Visual Studio 2017" to "Previous Visual Studio 2017" or vice versa.

zrolfs:
Thanks Anthony. Keep me posted on if that fixes the issue.

There were two main things that the refactor addresses:
-There were several implementations of several interfaces that all did the same/similar things, with a unique implementation for each type of datafile (thermo/mzml/ms1/ms2). Many of the inherited methods were identical between methods. Thermo was the weirder one, which is why some of its classes remain. By simplifying IMzSpectra, MzmlSpectra, MzSpectra, and ThermoSpectra into just MzSpectra, it saves us from having to generate future iterations like "MgfSpectra", "PepXMLSpectra", etc...
-Interfaces were previously used to lock private fields in the same way that the MM parameters were used, which is unnecessary when all interfaces do identical things and makes downstream manipulation (like aggregating spectra) unyielding.

rmillikin:
I'm about halfway through, but I like this update. It greatly simplifies working with MS data. I'm all for simplicity.

I'll play devil's advocate and ask for some performance benchmarks comparing file loading, etc. Is this faster/slower/no change compared to the current code, for mzML and raw? And for different instrument types?

zrolfs:
Comparisons were done on Velos and QE-HF data.
Mzmls were loaded 1-2% faster (n=10)
Thermo were loaded 5-20% faster (n=10)

acesnik:
@zrolfs @rmillikin could you two make different avatars for your Github accounts? Your icons are nearly identical.

rmillikin:
I'll add another comment here for posterity.

Because this PR parses all spectra file formats into a single generic spectra file format, we only have to have one method to interact with the generic file instead of separate methods that all do the same thing for each of the specific file formats. The benefits are 1. we only have to fix bugs in one place, increasing code stability and 2. (as @zrolfs mentioned) this makes it far easier to interact with newly-added file formats. There could be bugs in parsing a particular format into the generic one of course, but that would be a bug with the parser, not the actual code to interact with the data.

The argument could be made that Thermo scans have some properties that mzML files do not contain (for example, noise arrays). That just means that these values will be null when reading in an mzML. MGF format, for example, often has very limited information besides the mass spectrum, so many values for the generic data file or scan will be null.

Issue #323, opened 2018-05-23
UnmanagedThermoHelperLayer is no longer a part of nuget package


acesnik:
# Problem
I wasn't able to run a test project in a new solution (running .NET Framework 4.71). It was complaining that it wasn't able to load ManagedThermoHelperLayer.dll. 
![image 2](https://user-images.githubusercontent.com/16342951/40400285-183ce792-5e07-11e8-91f6-6f89dcd32b31.png)

I'm pretty sure UnmanagedThermoHelperLayer.dll isn't getting carried along. It's one of ManagedThermoHelperLayer.dll's dependencies.

# Reproducing this problem
It turns out if you clean a solution right now (with mzLib v1.0.305), it doesn't delete UnmanagedThermoHelperLayer.dll, so it sneakily doesn't break. But if you delete that file manually, it isn't recreated.

I reverted back to some arbitrary earlier version (mzLib v1.0.285, I think), and UnmanagedThermoHelperLayer.dll was present in that release.

Another weird thing is that in the TestThermo project, `clean` is not deleting UnmanagedThermoHelperLayer.dll, but when I delete it manually and `build`, it's recreated.

# Possible solution
It looks like the nuspec file has some weird targets at the bottom, with UnmanagedThermoHelperLayer listed under a different target than the rest of the files. I don't know what the heck `mzLib.targets` is, either. I think this is what needs to be changed.
![image](https://user-images.githubusercontent.com/16342951/40400599-7d8da018-5e08-11e8-8735-d09f1daeef40.png)

rmillikin:
this is fixed now yeah?

acesnik:
yes

PR #324, opened 2018-05-23
updated packages and nuspec

acesnik:


acesnik:
Trying to address this https://github.com/smith-chem-wisc/mzLib/issues/323 issue. It might take some trial and error because it's a NuGet issue, and I'm not sure how to manually check that sort of stuff.

PR #330, opened 2018-05-25
fix another crash in getting precursor info

rmillikin:


acesnik:
```
1) Failed : Test.TestMzML.TestPeakTrimmingWithThreeWindows
  Expected: True
  But was:  False
   at Test.TestMzML.TestPeakTrimmingWithThreeWindows() in C:\projects\mzlib\Test\TestMzML.cs:line 176
```

PR #336, opened 2018-06-15
Ms scan reference fix w/ unit test

ianhirsch:
#1220 from Metamorpheus issues

### DIFF
--- a/Test/TestMzML.cs
+++ b/Test/TestMzML.cs
@@ -586,7 +586,7 @@ namespace Test
         [Test]
         public void LoadMzmlTest()
         {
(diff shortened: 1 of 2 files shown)


Test/TestMzML.cs
```diff
@@ -586,7 +586,7 @@ public void Setup()
         [Test]
         public void LoadMzmlTest()
         {
-            Assert.Throws<AggregateException>(() =>
```
rmillikin: the AppVeyor build is failing because of this unit test - try making it just Exception instead of MzLibException or AggregateException

PR #342, opened 2018-06-29
sort the list of modifications after reading from file

trishorts:


Test/ModificationTests/CommonBiological.txt
```diff
@@ -0,0 +1,171 @@
...
+TG   K
+PP   Anywhere.
+MT   Biological
+CF   H6 C3
+//
+ID   HexNAc
+TG   Nxs or Nxt or S or T
+PP   Anywhere.
+NL   C8H13NO5
+CF   C8H13NO5
```
acesnik: This isn't the same format as above...

PR #349, opened 2018-07-09
Ms file index fixes

zrolfs:
I have concerns that not having the scans indexed anymore will cause downstream issues. This might be quickly reverted once I test with MetaMorpheus.

This PR addresses https://github.com/smith-chem-wisc/MetaMorpheus/issues/1274


zrolfs:
I can't build this. It can't find Chemistry. Send help.

acesnik:
I looks like it built on AppVeyor.

MzML/Mzml.cs
```diff
@@ -195,6 +197,20 @@ public static Mzml LoadAllStaticData(string filePath, FilteringParams filterPara
                 }
             });
 
+            //Mzml sometimes have scan numbers specified, but usually not. In the event that they do, the iterator above unintentionally assigned them to an incorrect index.
+            //Do a quick check
```
acesnik: Make a HashSet<int> of the scan numbers, throw `new InvalidOperationException()` with an informative message if duplicate scan #s

zrolfs: I'm not sure InvalidOperationException makes the most sense here. I like using the MzlibExceptions so that we know we meant to throw it (instead of an unintentional exception).

acesnik: Seems fine. Here's the description for InvalidOperationException: "This exception should also be thrown when the combined state of a set of related objects is invalid for the operation." https://msdn.microsoft.com/en-us/library/ms229007%28v=vs.100%29.aspx?f=255&MSPPError=-2147217396

PR #355, opened 2018-07-14
fully updated mzlib for the new ModificationGeneral class

trishorts:


trishorts:
I don't think there are actually any conflicts. And, it builds and passes all tests on my machine.

trishorts:
I guess maybe there were and i was able to fix them.

Proteomics/Modifications/ModificationGeneral.cs
```diff
@@ -26,18 +39,19 @@ public class ModificationGeneral
...
             private set { ValidModification = value; }
         }
 
-        public ModificationGeneral(string _Id, string _Accession, string _ModificationType, string _FeatureType, ModificationMotif _Target, string _Position, ChemicalFormula _ChemicalFormula, double? _MonoisotopicMass, Dictionary<string, IList<string>> _DatabaseReference, Dictionary<string, IList<string>> _TaxonomicRange, List<string> _Keywords, Dictionary<DissociationType, List<double>> _NeutralLosses, Dictionary<DissociationType, List<double>> _DiagnosticIons, string _FileOrigin)
+        public ModificationGeneral(string _Id = null, string _Accession = null, string _ModificationType = null, string _FeatureType = null, ModificationMotif _Target = null, string _Position = "Unassigned.", ChemicalFormula _ChemicalFormula = null, double? _MonoisotopicMass = null, Dictionary<string, IList<string>> _DatabaseReference = null, Dictionary<string, IList<string>> _TaxonomicRange = null, List<string> _Keywords = null, Dictionary<DissociationType, List<double>> _NeutralLosses = null, Dictionary<DissociationType, List<double>> _DiagnosticIons = null, string _FileOrigin = null)
```
rmillikin: parameters should be camel case. like `id` and not `_Id`

Proteomics/Modifications/ModificationGeneral.cs
```diff
@@ -26,18 +39,19 @@ public class ModificationGeneral
 
         public bool ValidModification
         {
-            get { return (this.Id != null && (this.ChemicalFormula != null || this.MonoisotopicMass != null) && this.Position != null && this.ModificationType != null); }
+            get { return (this.Id != null && (this.ChemicalFormula != null || this.MonoisotopicMass != null) && this.Position != "Unassigned." && this.ModificationType != null && this.FeatureType != "CROSSLINK"); }
             private set { ValidModification = value; }
```
rmillikin: can you remove this `private set` or does visual studio complain?

Proteomics/Modifications/ModificationGeneral.cs
```diff
@@ -48,28 +62,33 @@ public ModificationGeneral(string _Id, string _Accession, string _ModificationTy
             this.FileOrigin = _FileOrigin;
         }
 
-        public static ModLocationOnPeptideOrProtein? Localization(string myLocalization)
+        public static string ModLocationOnPeptideOrProtein(string modLocation)
```
rmillikin: 1. the original switch statement is more efficient; just change the default return value from null to "Unassigned."
2. change to return an enumerated type instead of the string

Proteomics/Modifications/ModificationGeneral.cs
```diff
@@ -48,28 +62,33 @@ public ModificationGeneral(string _Id, string _Accession, string _ModificationTy
...
+            }
+            else
+            {
+                return "Unassigned.";
+            }
+        }
 
-                default:
-                    return null;
+        public double? RoundedDouble(double? myNumber)
```
rmillikin: this method may be more appropriate for a more global class if you plan to use this type of rounding elsewhere. up to you

Proteomics/Modifications/ModificationLocalization.cs
```diff
@@ -4,10 +4,10 @@ namespace Proteomics
 {
     public static class ModificationLocalization
     {
-        public static bool ModFits(ModificationWithMass attemptToLocalize, Protein protein, int peptideOneBasedIndex, int peptideLength, int proteinOneBasedIndex)
+        public static bool ModFits(ModificationGeneral attemptToLocalize, Protein protein, int peptideOneBasedIndex, int peptideLength, int proteinOneBasedIndex)
```
rmillikin: can we change the name of ModificationGeneral to just Modification

Proteomics/Modifications/ModificationLocalization.cs
```diff
@@ -16,29 +16,32 @@ public static bool ModFits(ModificationWithMass attemptToLocalize, Protein prote
...
             {
                 return false;
             }
-            if (attemptToLocalize.terminusLocalization == TerminusLocalization.ProtC && proteinOneBasedIndex < protein.Length)
+            if (attemptToLocalize.Position == "Peptide C-terminal." && proteinOneBasedIndex < protein.Length)
             {
                 return false;
             }
+
+            // I guess Anywhere. and Unassigned. are true since how do you localize anywhere or unassigned.
```
rmillikin: I realize this is nitpicky but this comment is confusing because we're using "Anywhere." as "not specific to a terminus". It is completely reasonable to localize a modification to a phosphorylation on a serine in the middle of a peptide.

Proteomics/ProteolyticDigestion/CompactPeptideBase.cs
```diff
@@ -190,16 +190,23 @@ protected static IEnumerable<double> ComputeFollowingFragmentMasses(PeptideWithS
                 // If modification exists
                 if (yyy.AllModsOneIsNterminus.TryGetValue(oneBasedIndexToLookAt + 1, out currentModification))
                 {
-                    if (currentModification.neutralLosses.Count == 1 && oneBasedIndexToLookAt != 0 && oneBasedIndexToLookAt != yyy.Length + 1)
+                    List<double> neutralLossValuesList = new List<double>();
```
rmillikin: I don't really understand the changes here, can you explain this?

Creating these lists for each compact peptide (actually I think it's making new lists for each fragment in the peptide...) may cause serious RAM problems

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -59,14 +61,19 @@ public double MonoisotopicMass
...
+                        _monoisotopicMass += mod.MonoisotopicMass;
                     }
                     _monoisotopicMass += BaseSequence.Select(b => Residue.ResidueMonoisotopicMass[b]).Sum();
                 }
-                return _monoisotopicMass.Value;
+                return RoundedDouble(_monoisotopicMass.Value);//rounding this at 1e-9 tolerance
             }
         }
 
+        public double RoundedDouble(double myNumber)
```
rmillikin: delete this, move the other method to a more accessible location, and call that method. we don't want duplicate methods doing the same work

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -260,15 +267,15 @@ public CompactPeptide CompactPeptide(TerminusType terminusType)
...
-            if (vvv.TryGetValue(j + 2, out ModificationWithMass modToReplace))
+            if (vvv.TryGetValue(j + 2, out ModificationGeneral modToReplace))
             {
-                massOfExistingMod = modToReplace.monoisotopicMass;
+                massOfExistingMod = (double)modToReplace.MonoisotopicMass;
                 vvv.Remove(j + 2);
             }
 
-            vvv.Add(j + 2, new ModificationWithMass(null, null, null, TerminusLocalization.Any, massToLocalize + massOfExistingMod));
+            vvv.Add(j + 2, new ModificationGeneral(null, null, null, null, null, "Anywhere.", null, massToLocalize + massOfExistingMod, null, null, null, null, null, null));
```
rmillikin: clean this constructor up with the parameter-specifying syntax. i.e. `new Modification(MonoisotopicMass: massToLocalize + massOfExistingMod, ....... );`

Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
```diff
@@ -14,6 +14,7 @@ static ProteaseDictionary()
                     && !AppDomain.CurrentDomain.BaseDirectory.Contains("Jenkins") ?
                 Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "MetaMorpheus") :
                 AppDomain.CurrentDomain.BaseDirectory;
+            string p = Path.Combine(dataDirectory, "ProteolyticDigestion", "proteases.tsv");
```
rmillikin: delete unnecessary?

UsefulProteomicsDatabases/ProteinXmlEntry.cs
```diff
@@ -234,7 +234,7 @@ public List<Protein> ParseEntryEndElement(XmlReader xml,  bool isContaminant, st
                 else
                 {
                     // Not known and not seen
-                    unknownModifications[FeatureDescription] = new Modification(FeatureDescription, "unknown");
+                    unknownModifications[FeatureDescription] = new ModificationGeneral(FeatureDescription, null, "unknown", null, null, null, null, null, null, null, null, null, null, null);
```
rmillikin: clean up constructor

MassSpectrometry/DissassociationTypeCollection.cs
```diff
@@ -0,0 +1,127 @@
...
+        HCD,
+        ECD,
+        ETD,
+        CID,
+        EThCD,
+        Any,
+        Custom
+    }
+
+    public class DissassociationTypeCollection
```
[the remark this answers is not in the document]
trishorts: that seems good. where would you put it?

Proteomics/Modifications/ModificationGeneral.cs
```diff
@@ -14,9 +14,22 @@ public class ModificationGeneral
         public string ModificationType { get; private set; }
         public string FeatureType { get; private set; }
         public ModificationMotif Target { get; private set; }
-        public ModLocationOnPeptideOrProtein? Position { get; private set; }
+        public string Position { get; private set; }
```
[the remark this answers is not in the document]
trishorts: I changed it because it's not always on a terminus. the anywhere means anywhere. I was looking for a term that was more encompassing. I also don't like position. can you think of another option?

Proteomics/Modifications/ModificationGeneral.cs
trishorts: it does not complain. I don't think it should be set publicaly. maybe i'm understanding it wrong.

MassSpectrometry/DissassociationTypeCollection.cs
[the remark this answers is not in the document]
rmillikin: I would put it in mzLib/MassSpectrometry/Enums/DissociationType.cs but maybe Anthony would object, since maybe that file is only intended to have the Enum in it...

Proteomics/Modifications/ModificationGeneral.cs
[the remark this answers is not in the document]
rmillikin: yeah I agree.. maybe like LocationRestriction?

Proteomics/Modifications/ModificationGeneral.cs
rmillikin: I think you're not ever setting it. Every time you're Getting it, it reevaluates whether it's valid or not and returns that, without storing the value. The evaluation not very computationally expensive so I think this is fine. If you remove the Set then there is no way of setting it, which is fine, because you're never actually trying to store it

Proteomics/Modifications/ModificationGeneral.cs
trishorts: switched to switch but using an enumerated type ends up being a pain in the neck. i did that for a while then changed everything back. easier to explain in person.

Proteomics/Modifications/ModificationGeneral.cs
trishorts: you're right. can you suggest a place?

Proteomics/Modifications/ModificationLocalization.cs
trishorts: that's my plan once everything is merged. problem is that "modification" is in MM and PS. I was worried that if I used the same name, some errors might get in accidentally. It would probably be okay to do it. I was just being cautious

Proteomics/Modifications/ModificationGeneral.cs
rmillikin: mzLib/Chemistry/ClassExtensions.cs

rmillikin: then you can call it like:
double myNumber = 1000.29398730491820981;
double myNumberRounded = myNumber.RoundedDouble();

Proteomics/ProteolyticDigestion/CompactPeptideBase.cs
trishorts: Here is a place where we need to have access to or pass the dissociation type. Then we can access the values in the dictionary key. This code is wrong and needs to be fixed.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
trishorts: same as before. do we have a utility class?

MassSpectrometry/DissassociationTypeCollection.cs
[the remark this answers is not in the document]
rmillikin: on second thought you'd have to make a class there anyway. so I would just delete the enum declared here and use the one in mzLib/MassSpectrometry/Enums/DissociationType.cs. Then here in this class, make a static dictionary which is populated appropriately which you can reference from any project using MassSpectrometry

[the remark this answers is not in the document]
trishorts: I deleted the enum. discovered that it was redundant with something dbailey created.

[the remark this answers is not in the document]
trishorts: What about the custom set? How would you deal with that? That's the case where the user picks all the fragmentation ion series that they want.

Proteomics/Modifications/ModificationGeneral.cs
trishorts: i removed the set. i didn't know i could do that. good to knwo.

MassSpectrometry/DissassociationTypeCollection.cs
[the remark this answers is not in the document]
rmillikin: I think at that level, the client (e.g. MetaMorpheus) would create a new Value for the "Custom" dissociation type key and populate it, because there's no way to know what the user wants to search for until it's set in MetaMorpheus

Proteomics/Modifications/ModificationGeneral.cs
trishorts: should we default a number of digits or pass it 
double myNumberRounded = myNumber.RoundedDouble(9)

Proteomics/Modifications/ModificationLocalization.cs
trishorts: that comment is because I don't understnad the whole code block and what it's trying to do. would you help me? I would like to add the description to the top of the code block then handle this correctly.

Proteomics/Modifications/ModificationGeneral.cs
```diff
@@ -14,9 +14,22 @@ public class ModificationGeneral
...
         public ChemicalFormula ChemicalFormula { get; private set; }
-        public double? MonoisotopicMass { get; private set; }
+        private double? monoisotopicMass = null;
```
acesnik: Could you use `_MonoisotopicMass` as the tracking field name instead?

trishorts: What is a tracking field?

acesnik: Private field used in the getter or setter methods of a property.

acesnik: You can type CTRL+R+R to do a quick refactor of everywhere the variable name occurs, btw.

trishorts: Try explaining again and pretend I'm in kindergarten. Also, thanks for the tip. That will be handy.

trishorts: doesn't "this" do that? 
image

MassSpectrometry/MassSpectrometry.csproj
```diff
@@ -19,4 +19,8 @@
     <ProjectReference Include="..\Spectra\Spectra.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
rmillikin: ?

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -172,12 +176,14 @@ public string SequenceWithChemicalFormulas
...
         /// Generates theoretical fragment ions for given product types for this peptide
         /// </summary>
-        public List<TheoreticalFragmentIon> GetTheoreticalFragments(List<ProductType> productTypes)
+        public List<TheoreticalFragmentIon> GetTheoreticalFragments(DissociationType dissociationType, FragmentationTerminus fragmentationTerminus)
         {
             // TODO: make this method more self-contained... right now it makes a CompactPeptide (deprecated) and fragments it
             List<TheoreticalFragmentIon> theoreticalFragmentIons = new List<TheoreticalFragmentIon>();
 
-            foreach (var productType in productTypes)
+            var productCollection = TerminusSpecificProductTypes.ProductIonTypesFromSpecifiedTerminus[fragmentationTerminus].Intersect(DissociationTypeCollection.ProductsFromDissociationType[dissociationType]);
```
rmillikin: don't need to do anything yet, but let's check to see if this causes RAM problems

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -187,13 +193,12 @@ public List<TheoreticalFragmentIon> GetTheoreticalFragments(List<ProductType> pr
...
-                TerminusType terminusType = ProductTypeMethods.IdentifyTerminusType(temp);
-
-                var productMasses = new CompactPeptide(this, terminusType).ProductMassesMightHaveDuplicatesAndNaNs(temp);
+                
+                var productMasses = new CompactPeptide(this, fragmentationTerminus, dissociationType).ProductMassesMightHaveDuplicatesAndNaNs(temp);
 
                 for (int i = 0; i < productMasses.Length; i++)
                 {
-                    theoreticalFragmentIons.Add(new TheoreticalFragmentIon(productMasses[i], double.NaN, 1, productType, i + ionNumberAdd));
+                    theoreticalFragmentIons.Add(  new TheoreticalFragmentIon(productMasses[i], double.NaN, 1, productType, i + ionNumberAdd));
```
rmillikin: need to make an issue for this; if neutral losses are part of this modified peptide, the fragment numbering will be wrong (e.g., should be y1 and y1, not y1 and y2)

UsefulProteomicsDatabases/UnimodLoader.cs
```diff
@@ -54,24 +63,49 @@ internal static IEnumerable<ModificationWithLocation> ReadMods(string unimodLoca
...
                             if (nl.mono_mass == 0)
-                                neutralLosses.Add(0);
+                            {
+                                if (neutralLosses == null)
+                                {
+                                    neutralLosses = new Dictionary<MassSpectrometry.DissociationType, List<double>>();
+                                    neutralLosses.Add(MassSpectrometry.DissociationType.AnyActivationType, new List<double> { 0 });
+                                }
+                                else
+                                {
```
rmillikin: unimod doesn't have neutral losses but we need to work on this section if it ever does

MassSpectrometry/MsDataScan.cs
```diff
@@ -16,6 +16,7 @@
 // You should have received a copy of the GNU Lesser General Public
 // License along with MassSpectrometry. If not, see <http://www.gnu.org/licenses/>.
 
+using MassSpectrometry;
```
acesnik: This shouldn't be necessary. Please use codemaid to clean this one.

Proteomics/Fragmentation/FragmentationTerminus.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Proteomics.Fragmentation
+{
+    public enum FragmentationTerminus
+    {
+        Both, //N- and C-terminus
+        N, //N-terminus only
+        C, //C-terminus only
+        None //could be used for top down intact mass
+    }
+
```
acesnik: codemaid

UsefulProteomicsDatabases/PtmListLoaderGeneral.cs
```diff
@@ -66,7 +66,7 @@ public static IEnumerable<ModificationGeneral> ReadModsFromFile(string ptmListLo
                     modification_specification.Add(line);
                     if (line.StartsWith("//"))
                     {
-                        foreach (var mod in ReadMod(ptmListLocation, modification_specification))
+                        foreach (var mod in ReadMod(ptmListLocation, modification_specification, formalChargesDictionary))
```
acesnik: Todo after merge: also change `PtmListLoaderGeneral` to just `PtmListLoader`.

PR #361, opened 2018-07-18
Public set mz spectrum

zrolfs:
Alright, I know I'm going to be burned at the stake for this, but I need a public set for MsDataScan's MzSpectrum. 

Calibration already modifies MzSpectrum by using a function, but averaging/clustering removes some peaks, not just shifts them. Additionally, I need to change intensities when averaging/clustering Ms2 scans.

I'm open to other ideas, but I don't see a better way of doing this than having a public set method.


zrolfs:
A possible work around is generating a new MsDataFile filled with new MsDataScans containing the modified spectra and copies of all the other fields, but that's pretty ugly.

rmillikin:
I think it's a bad idea to have this publicly settable. And IMO, philosophically, when you merge/average two scans, you create a new scan (though you could argue the same for calibrating a spectrum). How difficult is it to create a new list of spectra you want? It is really that slow/complicated? I think it wouldn't be much worse than trying to modify an existing one... maybe even cleaner

zrolfs:
I was thinking about this and I agree. I'll just make a clone method in the AggregationEngine.

zrolfs:
It takes 6 ms to write a new one, which feels significant when a single ms file has 60K spectra 
(6 ms*60000 spectra = 6 min increase in time).

rmillikin:
is it really faster to create the new MzSpectrum and attach it to an existing scan, compared to creating a new scan with the MzSpectrum? like what's the difference in speed between these two methods? 6min per file is not too bad IMO. But you need to compare that to the alternative. We need to weigh that time savings (if any) against the danger of a settable spectrum.

zrolfs:
6 minutes is pretty bad just for modifying a variable, but I don't think it's because of the constructor. I'm doing diagnostics and it seems like it's an issue with my method.

PR #371, opened 2018-08-03
Remove 2/3 PWSM constructors

zrolfs:
There were separate PWSM constructors made to shorthand nonspecific searches and tests. The single constructor can accomplish what the other two did.
When digestion params was added to PWSM, these two constructors were not updated, resulting in parsimony crashes because digestion params was null.


rmillikin:
I included this in #374

PR #374, opened 2018-08-09
Serializable pep with set mods

rmillikin:
I did a rough performance benchmark to test serializing PepWithSetMods (PWSM) vs. CompactPeptide. The benchmark was to digest a yeast fasta with no fixed/variable mods, make a HashSet of the respective objects (PWSM or CompactPeptide), serialize it, and deserialize it.

capture

We added a bunch of stuff to CompactPeptide annotation-wise, and now it's pretty slow and takes up a lot of space to serialize, though it's still serializable if we need it to be for some reason.

PWSM file size is much smaller. Not sure why RAM is higher. It's about the same speed. We can work on making it more efficient - but PWSM is much easier to work with, and we can skip the whole "matching compact peptides to PWSM" bit in MetaMorpheus, so that will be faster.

4 new unit tests check to make sure properties are the same before+after serialization. 

You do need to run a method, SetNonSerializedPeptideInfo(), after deserializing to get the mods, Protein, and DigestionParams objects back since these are not serialized.

### DIFF
--- a/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
+++ b/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
@@ -12,7 +12,6 @@ namespace Proteomics.ProteolyticDigestion
     [Serializable]
     public class PeptideWithSetModifications : ProteolyticPeptide
     {
-        [NonSerialized] public readonly DigestionParams DigestionParams;
         public string Sequence { get; private set; }
         public readonly int NumFixedMods;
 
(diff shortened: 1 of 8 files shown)


acesnik:
Great benchmarking. Thanks for including that!

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -9,57 +9,47 @@
...
-        }
-
-        public PeptideWithSetModifications(int numFixedMods, Protein protein, int proteinOneBasedStart, int proteinOneBasedEnd,
-            Dictionary<int, Modification> allModsOneIsNterminus = null, int missedCleavages = 0)
-            : base(protein, proteinOneBasedStart, proteinOneBasedEnd, missedCleavages, null)
-        {
-            NumFixedMods = numFixedMods;
-            AllModsOneIsNterminus = allModsOneIsNterminus ?? new Dictionary<int, Modification>();
-        }
+        private readonly string DigestionParamString; // used to get digestion param object after deserialization
```
acesnik: nice

PR #382, opened 2018-08-28
Adjust proteolysis products and modification indices with applying variants

acesnik:
- Use tab text rather than the tab character, since XML parsing is inconsistent with whitespace parsing
- Adjust proteolysis product and modification indices.


acesnik:
* Fix hash and equals methods for invalid modifications

acesnik:
@rmillikin, could you refresh your approval of this one?

PR #392, opened 2018-09-05
getmassshiftfromproducttype made public method

trishorts:


acesnik:
whoa, that's a lot of commits for one word

PR #393, opened 2018-09-05
Create a method for finding the product ions to be used for searching

zrolfs:


rmillikin:
do we still need this?

zrolfs:
I don't think so

PR #398, opened 2018-09-08
Filter the same way for XML modifications and PTM lists; also return the errors and filtered modifications in an `out` variable

acesnik:
currently, xml-listed modifications aren't validated... expectations of ptmlists should be the same


acesnik:
Now, this outputs modifications with their error messages directly from the ptmlist loader.

These could be displayed in MM and PS in warning message boxes.

acesnik:
I also refactored the ptmlist loader tests, since there was redundancy left over from when there was a separate `PtmListloaderGeneral` and `PtmListLoader`

PR #399, opened 2018-09-10
modification equals and gethashcode speedup

lschaffer2:


Proteomics/Modifications/Modification.cs
```diff
@@ -127,12 +128,16 @@ public override bool Equals(object o)
         {
             Modification m = o as Modification;
             return o != null
-                && m.ToString() == this.ToString();
+                   && m.IdWithMotif != null && IdWithMotif != null && m.ModificationType != null && ModificationType != null && m.MonoisotopicMass.HasValue 
+                   && MonoisotopicMass.HasValue
```
acesnik: If two modifications are the same but have both have null `MonoisotopicMass` values, this will return false inaccurately.

lschaffer2: Yeah @rmillikin and I talked about this. I think we decided that those are invalid mods anyways

acesnik: Noted :) I have an open pull request to return invalid modifications with their error messages, though, so we might want to be careful about considering how invalid mods might be used.

https://github.com/smith-chem-wisc/mzLib/pull/398

lschaffer2: Yes that's true. I wonder if there is a case where we would want invalid mods to be equal to return true. If so we can modify this in the future to account for that!

PR #401, opened 2018-09-14
better mod parsing

rmillikin:


rmillikin:
@acesnik re-review please

acesnik:
Looks good. Should we put a check in `Modification` for the colons? Not sure if we want to throw an exception or just put it in the validation and warning methods.

acesnik:
In other words, could you add a unit test for where `Modification.ModificationType` has a colon to test that behavior?

PR #403, opened 2018-09-20
fix parallel peptide bugs

rmillikin:


Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -101,12 +101,15 @@ public double MonoisotopicMass
...
+
                     foreach (var mod in AllModsOneIsNterminus.Values)
                     {
-                        _monoisotopicMass += mod.MonoisotopicMass;
+                        monoMass += mod.MonoisotopicMass.Value;
                     }
-                    _monoisotopicMass += BaseSequence.Select(b => Residue.ResidueMonoisotopicMass[b]).Sum();
+                    monoMass += BaseSequence.Select(b => Residue.ResidueMonoisotopicMass[b]).Sum();
+
+                    _monoisotopicMass = monoMass;
```
acesnik: Don't you want to `lock (this) {}` for this set?

rmillikin: no - it's fine to overwrite the mass with parallel threads. 

the problem was that the old code added residue masses to the global variable _monoisotopicMass (not a local variable).  so multiple threads saw that the global variable was null, then added residue masses to it.

here, the threads will see that the mass is null, create a local variable, add residue masses to that, then assign the local variable value to the global variable. it's ok to assign it twice on accident. locking it would just slow it down

acesnik: Hmm, sounds reasonable. I haven't tackled a parallel bug like that before. Effectively, it will set it to the same value if this method is triggered again...

PR #405, opened 2018-09-24
Update digestion params

zrolfs:


zrolfs:
This isn't ready yet, sorry.

zrolfs:
Ready!

Test/DatabaseTests/DatabaseLoaderTests.cs
```diff
@@ -107,7 +107,7 @@ public void FilesLoading()
             Loaders.LoadElements(Path.Combine(TestContext.CurrentContext.TestDirectory, "elements2.dat"));
 
             var unimodMods = Loaders.LoadUnimod(Path.Combine(TestContext.CurrentContext.TestDirectory, "unimod_tables2.xml")).ToList();
-            Assert.AreEqual(2639, unimodMods.Count); // UniMod PTM list may be updated at some point, causing the unit test to fail
+            Assert.AreEqual(2633, unimodMods.Count); // UniMod PTM list may be updated at some point, causing the unit test to fail
```
rmillikin: this will probably fail on appveyor because it re-downloads the latest version of unimod mods and you probably have an older version of the mods list on your machine (same with below unit tests)

rmillikin: try deleting mzLib\Test\bin\x64\Debug on your machine and see if the unit tests pass

rmillikin:
some comments that don't necessarily need immediate changing but I think we should talk about them:
1) why are there two different bools for non-specific and semi-specific protease cleavage? would it make more sense to have an enum with Full, Semi, and Non specific cleavage? You can't have semi and non-specific both be true or both be false, right? Just seems like an odd choice of data structure to represent the cleavage specificity
2) Fragmentation terminus seems like the wrong structure to represent the proteolytic cleavage terminus - I think we used to have a cleavage terminus enum and it got removed at some point. I think we should bring it back. Also, what is this used for in a protease context? I guess I'm thinking of the collagenase GPX|GPX type of motif. I guess you could view it as C-terminus of X or N-terminus of G, I'm not sure what the actual reaction is. Is this used for the non-specific search algorithm?

PR #415, opened 2018-10-09
new protease digestion motifs

bintiibrahim:


Proteomics/ProteolyticDigestion/Protease.cs
```diff
@@ -172,21 +170,84 @@ public CleavageSpecificity GetCleavageSpecificity(string proteinSequence, int st
         internal List<int> GetDigestionSiteIndices(string proteinSequence)
         {
             var indices = new List<int>();
-            for (int i = 0; i < proteinSequence.Length - 1; i++)
+            string proteinSubstring = proteinSequence;
```
rmillikin: I'd like to see a lot of this code moved into DigestionMotif (I've added a few shells of methods that need completing there). Right now this class is kind of hard for me to follow. If we break this up into smaller methods in more appropriate places I think it'll be easier to understand

Test/DoubleProtease.tsv
```diff
@@ -1,4 +1,4 @@
 Name	Sequences Inducing Cleavage	Sequences Preventing Cleavage	Cleavage Terminus	Cleavage Specificity	PSI-MS Accession Number	PSI-MS Name	Site Regular Expression	Notes
-Test1	K|R|D	"P,D|P|"	C|N|N	full				
-Test2	K|D		C|C	full				
-Test3	K|		C|N	full				
+Test1	"K|,|R,|D[D]"			full
```
rmillikin: is this the same protease as before? shouldn't it be
`K|[P],K|[D],|R[P],|D` ?

bintiibrahim: so i took a look at the test and the protein OKAREDY would be digested into: OK, A, REDY under Test 1.
so maybe the protease would be K|[P],K|[D],|R[P],|D**[D]** ?

Proteomics/ProteolyticDigestion/Protease.cs
```diff
@@ -172,21 +167,19 @@ public CleavageSpecificity GetCleavageSpecificity(string proteinSequence, int st
         internal List<int> GetDigestionSiteIndices(string proteinSequence)
         {
             var indices = new List<int>();
-            for (int i = 0; i < proteinSequence.Length - 1; i++)
+            for (int i = 0; i < proteinSequence.Length; ++i)
             {
-                foreach (var c in SequencesInducingCleavage)
+                foreach(var c in MotifList)
```
rmillikin: much cleaner. thx. this looks good. I have some minor code style comments we should talk about but this is good

rmillikin: also need to redo the benchmarks for time and number of peptides

bintiibrahim: I did the benchmark test yesterday for 4 different protease and it all looks good

rmillikin: awesome 👍

PR #428, opened 2018-11-07
Update PWSM Hashcode

zrolfs:
Getting crashes when DigestionParams are null.
The previous addition didn't guarantee a unique number.


Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -370,7 +370,15 @@ public override bool Equals(object obj)
 
         public override int GetHashCode()
         {
-            return FullSequence.GetHashCode() + DigestionParams.Protease.GetHashCode();
+            unchecked
```
acesnik: why unchecked?

MassSpectrometry/MsDataFile.cs
```diff
@@ -108,27 +108,27 @@ public static void WindowModeHelper(ref double[] intensities, ref double[] mArra
...
-                    }
+                    ranges.Add(scanMin, (scanMin + mzRangeInOneWindow));
                     scanMin += mzRangeInOneWindow;
                 }
+
+                //add the last range
+                if (numberOfWindows > 1) //don't add a last if there's only one window
+                {
+                    ranges.Add(scanMin, (scanMin + mzRangeInOneWindow) + shiftToMakeRangeInclusive);
+                }
```
acesnik: Is this just refactoring? I don't really see a logical change...

rmillikin:
hashcode does not need to generate a unique number and it should never be assumed to generate a unique number

https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=netframework-4.7.2

> Two objects that are equal return hash codes that are equal. However, the reverse is not true: equal hash codes do not imply object equality, because different (unequal) objects can have identical hash codes.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
acesnik: Good catch on the null exception. I'd suggest this instead:
```
int digestionParamsHash = DigestionParams == null ? 0 : DigestionParams.Protease == null ? 0 : DigestionParams.Protease.GetHashCode();
return (FullSequence ?? "").GetHashCode() ^ digestionParamsHash;
```

rmillikin:
`&& q.DigestionParams.Protease == this.DigestionParams.Protease;`

this seems like a problem too if digestionparams is null

btw when is digestionparams ever null?

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
acesnik: If you need to update the DigestionParams GetHashCode() method, you should go ahead and do that.

acesnik:
Lots of commits here for a small set of changes... Remember that you can squash merge on your machine 
https://stackoverflow.com/questions/5308816/how-to-use-git-merge-squash

acesnik:
Is this still active, or should we close the PR?

PR #429, opened 2018-11-07
Digestionmotif fix and test refactoring

acesnik:


Proteomics/ProteolyticDigestion/Protease.cs
```diff
@@ -169,17 +172,17 @@ internal List<int> GetDigestionSiteIndices(string proteinSequence)
             var indices = new List<int>();
             for (int r = 0; r < proteinSequence.Length; r++)
             {
-                foreach (DigestionMotif motif in MotifList)
+                foreach (DigestionMotif motif in DigestionMotifs)
                 {
-                    if (motif.Fits(proteinSequence, r))
+                    if (motif.Fits(proteinSequence, r) && r + motif.CutIndex < proteinSequence.Length)
```
acesnik: this is the only change in logic

PR #434, opened 2018-11-19
peak filtering

trishorts:


TestThermo/TestThermo.cs
```diff
@@ -5,42 +5,29 @@
...
 using System.Linq;
-using Stopwatch = System.Diagnostics.Stopwatch;
 
 namespace TestThermo
 {
     [TestFixture]
-    [Ignore("Takes 20 mins on appveyor")]
     public sealed class TestThermo
     {
-        private static Stopwatch Stopwatch { get; set; }
```
acesnik: Could you put back the timing methods?

acesnik: They help illustrate how very long these tests take on AppVeyor.

acesnik: If you want to test these changes before the end of the day, I'd suggest making a local nugget package with `nuget pack` in your mzLib repository. You can then load that into MetaMorpheus. I believe we did that on your machine before, so it should be mostly set up.

acesnik: Using the `-Version` flag after `nuget pack` to specify and increment the version of the local nuget package is really efficient for successive builds, since MetaMorpheus project sees it as an update.

PR #446, opened 2019-01-07
raw file reader

rmillikin:


acesnik:
What's going on here?

acesnik:
Oh, this is cool. Have you tried this out in MM locally?

rmillikin:
there are some problems reading some ITMS data, I think it's a bug in rawfilereader but I haven't resolved it

rmillikin:
I think we should merge this in soon and try to get all the NuGet stuff working and try it out in MetaMorpheus. The licence that we're required to bundle with this might not pack into NuGet...

RawFileReader has some issues centroiding one of the test datafiles but I don't know how common this is.

It's much faster at file loading than the old system and doesn't require the user to install anything.

rmillikin:
We can use RawFileReader for opening most files, and if it crashes then we can use MSFileReader as a backup

ThermoRawFileReader/ThermoRawFileReaderData.cs
```diff
@@ -0,0 +1,405 @@
...
+        {
+            if (!File.Exists(filePath))
+            {
+                throw new FileNotFoundException();
+            }
+
+            Loaders.LoadElements();
+
+            // I don't know why this line needs to be here, but it does...
+            var temp = RawFileReaderAdapter.FileFactory(filePath);
```
zrolfs: weird

ThermoRawFileReader/ThermoRawFileReaderData.cs
```diff
@@ -0,0 +1,405 @@
...
+            {
+                return null;
+            }
+
+            return GetOneBasedScan(dynamicConnection, filterParams, oneBasedScanNumber);
+        }
+
+        private static MsDataScan GetOneBasedScan(IRawDataPlus rawFile, IFilteringParams filteringParams, int scanNumber)
+        {
+            var scan = Scan.FromFile(rawFile, scanNumber);
```
zrolfs: is this used for anything?

rmillikin: I don't think so

rmillikin:
don't merge this yet, I fixed the problem with reading ITMS files

rmillikin:
I fixed reading ITMS data. just to be safe, I used MSConvert to convert the three .raw files to .mzML, read in that .mzML along with the .raw, and compared the peak count, intensities, and m/z values. All looks good

PR #459, opened 2019-03-13
multipurpose spectrum filtering

trishorts:


MassSpectrometry/IFilteringParams.cs
```diff
@@ -23,6 +23,8 @@ public interface IFilteringParams
         double? MinimumAllowedIntensityRatioToBasePeakM { get; }
         int? NumberOfPeaksToKeepPerWindow { get; }
         int? NominalWindowWidthDaltons { get; }
+        int? NumberOfWindows { get; }
```
zrolfs: Does this need to be nullable? If you didn't want any windows, couldn't it just be 1?

MassSpectrometry/IFilteringParams.cs
```diff
@@ -23,6 +23,8 @@ public interface IFilteringParams
...
+        double? WindowMaxNormalizationValue { get; }
```
zrolfs: Same thing here, where 1 is the same as null.

MassSpectrometry/MsDataFile.cs
```diff
@@ -127,6 +127,27 @@ public static void WindowModeHelper(ref double[] intensities, ref double[] mArra
                     }
                 }
             }
+            else if (filteringParams.NumberOfWindows.HasValue && filteringParams.NumberOfWindows.Value > 0)
+            {
+                double mzRangeInOneWindow = (scanRangeMaxMz - scanRangeMinMz) / filteringParams.NumberOfWindows.Value;
+
+                for (int i = 1; i <= filteringParams.NumberOfWindows.Value; i++)
+                {
```
zrolfs: Remind me to refactor this at some point

MzML/Mzml.cs
```diff
@@ -378,11 +378,13 @@ private static MsDataScan GetMsDataOneBasedScanFromConnection(Generated.mzMLType
                 }
             }
 
-            if (filterParams != null && intensities.Length > 0 && (filterParams.MinimumAllowedIntensityRatioToBasePeakM.HasValue || filterParams.NumberOfPeaksToKeepPerWindow.HasValue)
-                && ((filterParams.ApplyTrimmingToMs1 && msOrder.Value == 1) || (filterParams.ApplyTrimmingToMsMs && msOrder.Value > 1)))
+            //if (filterParams != null && intensities.Length > 0 && (filterParams.MinimumAllowedIntensityRatioToBasePeakM.HasValue || filterParams.NumberOfPeaksToKeepPerWindow.HasValue)
```
zrolfs: Remove commented code

MassSpectrometry/IFilteringParams.cs
trishorts: I'm not sure to be honest. You can imagine that setting both window width and number of windows could cause a collision. Lot's depends on how downstream programs use filtering params and window helper. In window helper, I have an order of operations 1) if window width defined, do something else 2) if window number defined, else 3) apply filtering to the entire spectrum.  Allowing the values to be null (but handling the null assuming conventional parameters) allows crash-free experience. 
image

trishorts: Happy to get your further thoughts on this

MzML/Mzml.cs
trishorts: done. but nervously.

MassSpectrometry/IFilteringParams.cs
zrolfs: That makes sense to me. I like to stay away from nullable variables when possible, though, because they're slow and can cause issues with tomls. Can we use <1 instead of null? That will prevent issues with negative values, too. I like the order of operations, but how about moving it from the window helper to the filteringParams constructor, so that it will be universal? If it goes into 1) (window width defined), then it can set the number of windows to 0. If it goes into the number of windows, then it sets the window width to 0.

zrolfs: Should windowWidthInDaltons be a double? That would make more sense to me.

zrolfs: When using window widths, how do you handle the edges of the spectra? I imagine there will be cases where the experimental spectra range divided by the window width is not an integer.

trishorts: Window width can be double (it gets converted). I was worried about "fractional widths). Like if someone made it 0.01Th or something stupid. I'll have a look.

trishorts: as for edges. Two things. One, increment always begins at zero mz but skips empty windows. So if you pick 100mz window and your spectrum begins at 400, the first bin is 400-499.99. if your spectrum starts @ 350, the first bin is 300-399.99 (and you get  a partial bin). Since almost all spectra have varying width, I defaulted to fixed and reproducible behavior. that way all the widows are comparable accross all the scans. probably should take fractional peaks. So, if you take 100 peaks per bin and you have a short bin (like 50 mz out of 100 mz), we could take only 50 peaks. Last thing is that I add 0.000000001 to the edges so nothing gets clipped.

trishorts: forced to give up on the non nullable b/c IFilteringParams doesn't have a constructor, and collides with FilteringParams..... chasing my tail all over mzlib

PR #467, opened 2019-03-21
unit test to keep most intense peaks after filtering

trishorts:


MassSpectrometry/MsDataFile.cs
```diff
@@ -210,6 +211,7 @@ public static void WindowModeHelper(ref double[] intensities, ref double[] mArra
 
             intensities = reducedIntensityList.ToArray();
             mArray = reducedMzList.ToArray();
+            Array.Sort(mArray, intensities);
```
zrolfs: were these just accidentally removed?

trishorts: i removed them intentionally yesterday, then i got UT failure in MM. Tracked it back to this. Added them back today.

trishorts: also added new unit test to keep me from making same foolish mistake in the future

PR #471, opened 2019-04-02
Fix Isoform reading bug for FASTA and XML

rmmiller22:


Proteomics/Protein/VariantApplication.cs
```diff
@@ -13,9 +13,15 @@ public static class VariantApplication
         /// <param name="sequenceVariation"></param>
         public static string GetAccession(Protein protein, IEnumerable<SequenceVariation> appliedSequenceVariations)
         {
-            string[] accSplit = protein.Accession.Split('_');
-            string baseAcc = protein.Accession.StartsWith("DECOY_") ? $"{accSplit[0]}_{accSplit[1]}" : accSplit[0];
-            return baseAcc + 
+            if (protein.Accession.Contains("DECOY_"))
```
zrolfs: It looks like previously we were removing everything after a "_". I understand you don't want to do that, though, since the isoforms use "_"? (Even though they're "-" in the database?). Your change doesn't really make a lot of sense, though, since now it's removing everything after a "_" if it's a decoy but not if it's a target. There's also two checks for if the protein accession has "DECOY_" in it, which seems unnecessary. Why not just keep the protein accession as is? I guess I don't understand the prupose of filtering out the "_".

UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -84,12 +84,31 @@ public static class ProteinDbLoader
...
                             if (newProtein != null)
                             {
-                                targets.Add(newProtein);
+                                potentialTargets.Add(newProtein);
                             }
                         }
                     }
                 }
             }
+            HashSet<string> unique_accessions = new HashSet<string>();
```
zrolfs: Can you include a comment that this is for isoform numbering? You could probably refactor from foreach() to for() and remove the "potentialTargets" variable, too.

Proteomics/Protein/VariantApplication.cs
zrolfs: Alright, the underscore just makes things italic. Neat.

rmmiller22: This is all of Anthony's Spritz variant code and I was just trying to fix problem with minimizing my impact in the code area.  The double check for decoy is definitely redundant and I can hopefully remove that.

UsefulProteomicsDatabases/ProteinDbLoader.cs
rmmiller22: Will do.

Proteomics/Protein/VariantApplication.cs
zrolfs: Does a unit test fail if we just wipe lines 16-23?

rmmiller22: All tests still pass. This now essentially does the same thing as all my excess code did before.

UsefulProteomicsDatabases/ProteinDbLoader.cs
rmmiller22: With fix in variant application lines 93-112 can be removed. The main issue was in line 116 squashing together isoforms.

PR #475, opened 2019-04-23
MBR update +bugfixes

rmillikin:


zrolfs:
Uff da

FlashLFQ/FlashLfqEngine.cs
```diff
@@ -519,55 +526,241 @@ private void QuantifyMatchBetweenRunsPeaks(SpectraFileInfo idAcceptorFile)
...
+                }
+
+                // calculate FDR
+                foreach (var peak in matchedPeaks.Where(p => p != null))
+                {
+                    int cumulativeTrue = res.Count(p => p.isTruePositive && p.bestAcceptorPeak.MbrScore >= peak.MbrScore);
+                    int cumulativeFalse = res.Count(p => !p.isTruePositive && p.bestAcceptorPeak.MbrScore >= peak.MbrScore);
+                    double qValue = ((double)cumulativeFalse) / (cumulativeTrue + cumulativeFalse);
+
+                    if (cumulativeFalse == 0)
```
zrolfs: Is this needed? If cumulativeFalse is zero, then q-value should also be zero (unless cumulativeTrue is also zero, but then we don't have anything and we wouldn't be in this loop...)

FlashLFQ/FlashLfqEngine.cs
```diff
@@ -519,55 +526,241 @@ private void QuantifyMatchBetweenRunsPeaks(SpectraFileInfo idAcceptorFile)
...
+                    else
+                    {
+                        cumulativeFalsePositives--;
+                    }
+                }
+
+                // calculate FDR
+                foreach (var peak in matchedPeaks.Where(p => p != null))
+                {
+                    int cumulativeTrue = res.Count(p => p.isTruePositive && p.bestAcceptorPeak.MbrScore >= peak.MbrScore);
```
zrolfs: This linq function feels pretty inefficient for determining target/decoy counts. I feel like the above loop and this loop could be aggregated with the -- operations.

FlashLFQ/FlashLfqEngine.cs
```diff
@@ -519,55 +526,241 @@ private void QuantifyMatchBetweenRunsPeaks(SpectraFileInfo idAcceptorFile)
...
+
+                    if (cumulativeFalse == 0)
+                    {
+                        qValue = 0;
+                    }
+
+                    peak.MbrQValue = qValue;
+                }
+
+                foreach (var peak in matchedPeaks.Where(p => p != null && p.MbrScore > scoreCutoff))
```
zrolfs: Same thing here where this linq function seems unnecessary

FlashLFQ/IntensityNormalizationEngine.cs
```diff
@@ -188,7 +187,7 @@ private void NormalizeFractions()
                     }
 
                     // we're only normalizing on a subset of data here because it might take too long to do the whole set
-                    seenInBothBioreps = SubsetData(seenInBothBioreps, filesForThisBiorep.Concat(filesForCond1Biorep1).ToList());
+                    //seenInBothBioreps = SubsetData(seenInBothBioreps, filesForThisBiorep.Concat(filesForCond1Biorep1).ToList());
```
zrolfs: meant to be commented out?

FlashLFQ/IntensityNormalizationEngine.cs
```diff
@@ -321,201 +320,193 @@ private void NormalizeBioreps()
...
-                        // this peptide is in the return list regardless of whether or not it was actually just added;
-                        // we just want to guarantee this fraction has 500 peptides in the return list to normalize with
-                        peptidesAddedPerFraction[fraction]++;
-                    }
-                }
-            }
-
-            return subsetList.ToList();
-        }
+        //private List<Peptide> SubsetData(List<Peptide> initialList, List<SpectraFileInfo> spectraFiles)
```
zrolfs: Same here

FlashLFQ/ChromatographicPeak.cs
```diff
@@ -176,7 +178,6 @@ public override string ToString()
             sb.Append("" + NumIdentificationsByFullSeq + "\t");
             sb.Append("" + SplitRT + "\t");
             sb.Append("" + MassError + "\t");
-            //sb.Append(string.Join(",", IsotopicEnvelopes.OrderBy(p => p.ChargeState).ThenBy(p => p.IndexedPeak.ZeroBasedMs1ScanIndex).Select(p => p.ToString())));
```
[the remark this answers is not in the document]
rmillikin: it's useful debug code, I'd rather keep it commented rather than regenerating it every time I want to use it

PR #477, opened 2019-05-07
Fix get cleavage specificity when deserializing

zrolfs:


zrolfs:
Yep. The DigestionParam object isn't serialized, only the string is. From that string, the DigestionParam object is recreated. The protease is selected from a Dictionary, where the protease name is the key. The unit test is just checking if the key was written.

zrolfs:
I could check for ",Arg-C," or I could split the string and check a specific index, if you would prefer that.

trishorts:
h

PR #478, opened 2019-05-08
Filter peptides contain N-glycan motif in Protein.digestion

lonelu:
This will reduce the database size when searching N-glycopeptides.


rmillikin:
maybe it would be better and more flexible to include a "not" character in the modification motif, like we do with digestion motif? Something like Nx[P]t to mean NXT, but X can't be P?

Maybe we can somehow merge the digestion motif and modification motif objects... that's probably a cleaner end result, but that seems a little more complicated and bug-prone...

lonelu:
I agree with this. I like the digestion motif combination one.

PR #479, opened 2019-05-08
Fix bug with OneBasedVariantModifications

rmmiller22:
the OneBasedVariantModifications dictionary was not being reset for each variant sequence making it so that all applied variants had the exact same modifications regardless of whether they had an modifications or not. This was fixed by resetting the dictionary after each sequence variation object was created. A new unit test and text xml database were added.


acesnik:
Good work! Thanks for working on this.

acesnik:
@rmillikin I merged in this small pull request on variant stuff.

PR #482, opened 2019-05-20
Fix proteolysis product digestion (remove redundancy, save full length products)

zrolfs:


zrolfs:
Sorry for all of the old commits. This PR starts at https://github.com/smith-chem-wisc/mzLib/pull/482/commits/ab858d78473bbbe7c71a99a8bb535ace796601c0

zrolfs:
All lines are hit by unit tests, but the multi-boolean "if" statements are only counted as "partial" coverage for CodeCov, unless every possible path is executed (which is a lot). I would like to appeal not to suffer through 50+ tests, because we're planning on removing proteolysis products anyway in favor of separate protein entries. If the appeal is denied, it would be faster for me to generate separate protein entries than to generate all of the needed unit tests.

Proteomics/ProteolyticDigestion/Protease.cs
```diff
@@ -37,23 +37,27 @@ public override int GetHashCode()
             return (Name ?? "").GetHashCode();
         }
 
-        public CleavageSpecificity GetCleavageSpecificity(string proteinSequence, int startIndex, int endIndex)
+        public CleavageSpecificity GetCleavageSpecificity(Protein protein, int startIndex, int endIndex, bool retainMethionine)
```
trishorts: add a comment that this method is used only for semi- and non-specific searches

PR #486, opened 2019-05-30
Fix for naming of variant containing proteins

rmmiller22:
Currently variant names are repeatedly stacked causing a protein name to have the same variant multiple times within the accession. This was die to each variant being applied one at a time, and the base name for the protein not being kept as the original accession. A simple adjustment fixed this issue and all variant protein names are now correct.


Test/DatabaseTests/TestVariantProtein.cs
```diff
@@ -489,11 +489,15 @@ public void VariantModificationTest()
...
-            bool homozygousVariant = targets.Select(p => p.Accession).Contains("Q6P6B1");
+            bool homozygousVariant = targets.Select(p => p.Accession).Contains("Q6P6B1");           
+
             var variantMods = targets.SelectMany(p => p.AppliedSequenceVariations.Where(x=>x.OneBasedModifications.Count>= 1)).ToList();
             var decoyMods = decoys.SelectMany(p => p.AppliedSequenceVariations.Where(x => x.OneBasedModifications.Count >= 1)).ToList();
             var negativeResidues = decoyMods.SelectMany(x => x.OneBasedModifications.Where(w => w.Key < 0)).ToList();
-
+            bool namingWrong = targets.Select(p => p.Accession).Contains("Q8N865_H300R_A158T_H300R");
+            bool namingRight = targets.Select(p => p.Accession).Contains("Q8N865_A158T_H300R");
+            Assert.AreEqual(false, namingWrong);
```
zrolfs: Assert.IsTrue() and Assert.IsFalse() works for these

Proteomics/Protein/VariantApplication.cs
```diff
@@ -11,10 +11,9 @@ public static class VariantApplication
         /// </summary>
         /// <param name="protein"></param>
         /// <param name="sequenceVariation"></param>
-        public static string GetAccession(Protein protein, IEnumerable<SequenceVariation> appliedSequenceVariations)
+        public static string GetAccession(Protein protein,IEnumerable<SequenceVariation> appliedSequenceVariations)
         {
-                        
-            return protein.Accession + 
+            return protein.NonVariantProtein.Accession +
```
zrolfs: How does the protein's accession vary from the nonvariant protein's accession?

rmmiller22: in a protein with multiple variants the variants are applied one at a time. So for example the non variant protein is A and A has variants 1,2,3 and 4. When going through variant application first, variant 1 is applied and the protein. Accession becomes A_1 and the non variant protein.Accession remains A. Then variant 2 is applied to A_1. At this time all variants listed are applied. This means the protein becomes A_1_1_2 with previous code because the list of variant contains 1 and 2 and the protein. Accession is A_1. With the new code the non variant accession is what is added to since the list of variants are cumulative.

Test/DatabaseTests/TestVariantProtein.cs
rmmiller22: k with adjust

zrolfs: You don't have to, just an FYI

zrolfs: You don't have to, just an FYI

PR #487, opened 2019-05-30
Fix SNES min pep length bug

zrolfs:


zrolfs:
Impacts speedy nonspecific searches of specific proteases in the center of the protein, where the minimum peptide length was effectively one greater than specified.

PR #490, opened 2019-06-17
FlashLFQ now uses RawFileReader instead of MSFileReader

rmillikin:


rmillikin:
Capture

tech rep vs. tech rep intensities seem ok

Test/TestRawFileReader.cs
```diff
@@ -47,12 +47,14 @@ public static void TestDynamicConnectionRawFileReader()
 
             ThermoRawFileReaderData.InitiateDynamicConnection(path);
 
+            var msOrders = ThermoRawFileReaderData.GetMsOrdersByScanInDynamicConnection();
```
zrolfs: Is there a reason for this? If code coverage, include an assert. If just checking to make sure it doesn't crash, add a comment.

PR #493, opened 2019-06-27
bayesian estimation of mean/stddev

rmillikin:
initial commit. not ready for review yet


BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+        private double exp; // exponent of prior prob exponential distribution
+
+        /// <summary>
+        /// Construct the adaptive Metropolis within Gibbs sampler. Leave data2 null for one-sample data.
+        /// </summary>
+        public AdaptiveMetropolisWithinGibbs(List<double> data1, List<double> data2, int batch_size = 3, int? seed = null)
+        {
+            if (seed != null)
+            {
+                random = new MersenneTwister(seed.Value, true);
```
trishorts: don't need to change anything. might be good to have a "good" generic seed generator in mzlib that was accessible to each method that needed one. allows for more consistency.

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+            
+            SetUpAcceptableParameterRanges();
+            SetUpPriorProbabilityDistributions();
+            SetUpPosteriorFunction(isTwoSample);
+        }
+
+        /// <summary>
+        /// Burn in and then sample the MCMC chain.
+        /// </summary>
+        public void Run(int burnin = 20000, int n = 20000)
```
trishorts: what's special about 20000. i guess depending on set size, particularly small sets, you get redundancy?

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+            if (acceptableParameterRanges == null)
+            {
+                if (isTwoSample)
+                {
+                    acceptableParameterRanges = new List<Tuple<double, double>>[5];
+                    acceptableParameterRanges[0] = new List<Tuple<double, double>> { new Tuple<double, double>(double.NegativeInfinity, double.PositiveInfinity) }; // mu1 (can take any value)
+                    acceptableParameterRanges[1] = new List<Tuple<double, double>> { new Tuple<double, double>(double.NegativeInfinity, double.PositiveInfinity) }; // mu2 (can take any value)
+                    acceptableParameterRanges[2] = new List<Tuple<double, double>> { new Tuple<double, double>(0, double.PositiveInfinity) }; // sd1 (must be positive)
+                    acceptableParameterRanges[3] = new List<Tuple<double, double>> { new Tuple<double, double>(0, double.PositiveInfinity) }; // sd2 (must be positive)
+                    acceptableParameterRanges[4] = new List<Tuple<double, double>> { new Tuple<double, double>(0, double.PositiveInfinity) }; // nu (must be positive)
```
trishorts: what is "nu"?

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            var pooled = Data.SelectMany(p => p.Select(v => v)).ToList();
+
+            meanMu = pooled.Mean();
+            sdMu = pooled.StandardDeviation() * 1000000;
+
+            sigmaLow = pooled.StandardDeviation() / 1000.0;
+            sigmaHigh = pooled.StandardDeviation() * 1000.0;
+
+            exp = 1.0 / 29.0;
```
trishorts: what's this?

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// <summary>
+        /// Samples the parameter distributions via MCMC.
+        /// </summary>
+        private double[] next_sample()
+        {
+            markovChain.Add(currentState.ToArray());
+
+            for (var param_i = 0; param_i < numParams; param_i++)
+            {
+                var parameterProposal = Normal.Sample(random, currentState[param_i], Math.Exp(logSd[param_i]));
```
trishorts: is the assumption from everything that distributions are normal. maybe that should be stated at the top?

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+        
+        /// <summary>
+        /// Sets up the prior probability distributions for each parameter.
+        /// </summary>
+        protected void SetUpPriorProbabilityDistributions()
+        {
+            var pooled = Data.SelectMany(p => p.Select(v => v)).ToList();
+
+            meanMu = pooled.Mean();
+            sdMu = pooled.StandardDeviation() * 1000000;
```
trishorts: for low numbers of values, can you use standard deviation? maybe variance is better?

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
```diff
@@ -0,0 +1,285 @@
...
+            sigmaLow = pooled.StandardDeviation() / 1000.0;
+            sigmaHigh = pooled.StandardDeviation() * 1000.0;
+
+            exp = 1.0 / 29.0;
+        }
+
+        /// <summary>
+        /// Sets up the posterior function.
+        /// </summary>
+        protected void SetUpPosteriorFunction(bool isTwoSample)
```
trishorts: again, should this be variance and not standard deviation?

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
rmillikin: nothing really - it's generally enough iterations to converge on a reasonable answer in most cases. more iterations of the MCMC algorithm produce more accurate answers, for the point estimate and uncertainty for each parameter

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
rmillikin: nu is the degree of normality of the t-distribution. "degrees of freedom" refers to nu.

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
rmillikin: I changed the location of this, but in this MCMC method (described by Kruschke) the "nu" parameter of the t-distribution has an exponential prior probability distribution. 1/29 is used as the exponent of this prior probability distribution in this example.

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
rmillikin: this is just setting the prior prob distribution standard deviation of the "mu" parameter. basically it's saying the range of possible values is very large - we have a very fuzzy idea of what value "mu" might actually be, but it's most probably near the arithmetic mean.

as a separate thought, the variance can be estimated by squaring the std dev estimate.

BayesianEstimation/AdaptiveMetropolisWithinGibbs.cs
rmillikin: you could use variance over sd but I don't know what the advantage of that would be - it's just std dev squared

PR #494, opened 2019-07-08
FlashLFQ uses Bayesian protein quant

rmillikin:


FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,604 @@
...
+        private readonly MersenneTwister Rng;
+        private bool conditionsAreDefined;
+        private Dictionary<ProteinGroup, List<Peptide>> ProteinsWithConstituentPeptides;
+        private Dictionary<(Peptide, string, int), double> PeptideToSampleQuantity;
+        private Dictionary<ProteinGroup, double> ProteinToBaseConditionIntensity;
+
+        /// <summary>
+        /// Constructs the protein quantification engine. 
+        /// 
+        /// "baseCondition" refers to the condition that the protein fold-changes could be calculated relative to.
```
zrolfs: Isn't this kind of arbitrary? Particularly if you're doing multiple comparisons (i.e. there are four different groups (A, B, C, D) and you want to compare against each of them (AvsB, AvsC, AvsD, BvsC, BvsD, CvsD). Wouldn't this clump them together so that it'd be comparing AvsBCD?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,604 @@
...
+        /// 
+        /// If "foldChangeCutoff" is null, the fold-change cutoff will be estimated from the data ("experimental null").
+        /// 
+        /// </summary>
+        public ProteinQuantificationEngine(FlashLfqResults results, int maxThreads, string baseCondition, bool useSharedPeptides = false,
+            double? foldChangeCutoff = null, int? randomSeed = null, int mcmcBurninSteps = 1000, int mcmcSteps = 3000, bool pairedSamples = false)
+        {
+            if (string.IsNullOrWhiteSpace(baseCondition))
+            {
```
zrolfs: Similar to the last comment. Is a baseCondition necessary? Couldn't you arbitrarily grab one of the conditions, treat it as the "base"/WT, and get similar results?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,604 @@
...
+
+            // calculate FDR for the condition
+            foreach (var condition in conditions)
+            {
+                var res = proteinList.Select(p => p.Key.conditionToQuantificationResults[condition]).ToList();
+                CalculateFalseDiscoveryRates(res);
+            }
+
+            // remove temporary conditions
+            if (!conditionsAreDefined)
```
zrolfs: So if a single file doesn't have a condition defined, then all files have their condition wiped?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,604 @@
...
+        /// Pairs up proteins with constituent peptides and calculates peptide biorep intensities for easy lookup.
+        /// </summary>
+        private void Setup()
+        {
+            // temporarily set conditions if they aren't defined
+            if (!conditionsAreDefined)
+            {
+                foreach (var file in results.SpectraFiles)
+                {
+                    file.Condition = file.FullFilePathWithExtension;
```
zrolfs: This just assigns a placeholder that ensures no files share a condition, which makes comparisons impossible?

FlashLFQ/FlashLfqEngine.cs
```diff
@@ -44,19 +51,30 @@ public class FlashLfqEngine
...
+            double maxMbrWindow = 2.5,
+            int maxThreads = -1,
+
+            // settings for the Bayesian protein quantification engine
+            string proteinQuantBaseCondition = null,
+            double? proteinQuantFoldChangeCutoff = null,
+            int mcmcSteps = 3000,
+            int mcmcBurninSteps = 1000,
+            bool useSharedPeptidesForProteinQuant = false,
+            int? randomSeed = null)
```
zrolfs: Is there a reason to make this nullable? Is it here for people that prefer their data to not be reproducible?

FlashLFQ/FlashLfqEngine.cs
```diff
@@ -44,19 +51,30 @@ public class FlashLfqEngine
...
         {
-            if (optionalPeriodicTablePath == null)
```
zrolfs: Did you mean to delete?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
```diff
@@ -0,0 +1,92 @@
...
+            this.condition2 = condition2;
+            this.foldChangeMeasurements = fcs;
+            this.FoldChangePointEstimate = mus.Average();
+            this.StandardDeviationPointEstimate = sds.Average();
+            this.NuPointEstimate = nus.Average();
+            this.hdi_95 = Util.GetHighestDensityInterval(mus, 0.95);
+            this.ConditionIntensityPointEstimate = Math.Pow(2, FoldChangePointEstimate) * referenceIntensity;
+        }
+
+        public void CalculatePosteriorErrorProbability(double[] skepticalMus)
```
zrolfs: Comments here would be super appreciated

FlashLFQ/SpectraFileInfo.cs
```diff
@@ -4,7 +4,7 @@ public class SpectraFileInfo
     {
         public readonly string FullFilePathWithExtension;
         public readonly string FilenameWithoutExtension;
-        public readonly string Condition; // condition/biorep/techrep/fraction info
+        public string Condition; // condition/biorep/techrep/fraction info
```
zrolfs: Still beating on the condition thing here

zrolfs:
How are you handling missing values? Since there are no intensities output, the user is unable to impute downstream.

FlashLFQ/Bayesian Protein Quantification/ProteinFoldChangeEstimationModel.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public class ProteinFoldChangeEstimationModel : Model
+    {
+        public ProteinFoldChangeEstimationModel(double priorMuMean, double priorMuSd, double muInitialGuess,
+            double priorSdStart, double priorSdEnd, double sdInitialGuess, double priorNuExponent, double nuInitialGuess) : base()
+        {
+            modelParameters = new Parameter[3];
+
+            // mu (mean)
+            modelParameters[0] = new Parameter(
```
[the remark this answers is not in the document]
rmillikin: you mean how I decided on those distributions for the priors? the last two are from Kruschke (). 

I used a Student's t-distribution for mu's prior instead of a normal distribution because I wanted long tails. Intuitively, the protein's fold change is likely around some value (the arithmetic mean, the median, zero, etc. depending on assumptions). So it makes sense to use either a normal or Student's t for the prior, because I want the prior probability density to be high around that initial value. I chose Student's t instead of normal because there is a fair amount of uncertainty around what the fold-change actually is. 

In other words, I want it to be likely around a certain value, but it should be plausible to be a different value, especially given enough data. Student's t is good for this purpose.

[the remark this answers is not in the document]
zrolfs: Makes sense. Can you comment that here?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: sort of. it depends on the experimental design. typically there's some condition you're calculating intensities relative to, like "normal vs. tumor". the engine is designed with this in mind. You could potentially construct an all vs. all statistical analysis, but it is not necessary to change the engine to accomplish this. you could just call this engine with different base conditions as needed.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: no. the fold-changes are relative to something. if you choose a different reference, you'll get a different answer

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: MM and FlashLFQ (the clients that call this) do not allow blank conditions, specifically for this reason

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: no - comparisons are made, but just on a one file vs. one file basis

FlashLFQ/FlashLfqEngine.cs
rmillikin: a random seed is automatically generated if one isn't provided. the seed should be `public` instead of `private`, though, for it to be reported along with the output, so that you can reproduce someone else's work. but it isn't necessary to generate your own random seed

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
zrolfs: Can you clarify really quick: If there are two conditions, A and B, do you get different results when A is the base vs when B is the base, even though you're comparing the same conditions?

FlashLFQ/FlashLfqEngine.cs
rmillikin: yeah. this is outdated code. there's no elements.dat file anymore; it's read directly from the string stored in `PeriodicTableLoader.cs`

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
zrolfs: ...then why support it? If there's a blank condition, shouldn't you just return?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
zrolfs: I don't understand how comparisons can be made with just two files. There's not a standard deviation at that point, so if the values are not exactly identical (1.0 vs 1.00000001), then they're different...?

rmillikin:
a point estimate of protein intensity is one of the outputs

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: sorry, yeah, that was unclear. if you do A vs. B and then B vs. A, you'll get the same answer (approximately, given that there is inherent randomness in the MCMC sampler). if you do:

A vs. B
C vs. B

you'll get a different estimate for the fold-change (and intensity) of B, because the B condition has different references.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: I want to be able to do quantification without defining the conditions. but the code gets a lot messier if you have to check for conditions constantly. so it's just easier to set the conditions temporarily and then do the analysis, then undo the temporary conditions

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: a comparison can be made if there are multiple peptides per protein. the N isn't the number of bioreps - it's the number of peptide fold-change measurements. each of which is an independent measurement

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
zrolfs: I see. If you wanted to do an all-all comparison, then you would end up doing (A vs B, A vs C) and (B vs A, B vs C), which works but is computationally redundant.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
zrolfs: Is that useful, though? It's not really measuring biological variance (which is what we're interested in), but rather how much peptides from a shared protein vary against one another. I'm not feeling super good about that comparison.

rmillikin: I think I see the hangup. if, let's say, N=1, a t-test is not possible. In Bayesian statistics, it's possible to estimate a probability of a change (though IMO it's not especially useful, so I require at least N=2). in this method, thousands of t-distributions are generated by Monte Carlo sampling. if you have one datapoint, a wide range of very different distributions seem plausible. thus, there is great uncertainty in the mean. as the number of datapoints increases, and the datapoints become more precise, the MCMC sampling generates t distributions that look very similar, because if it generated a t distribution with a vastly different mean/stdev, the likelihood of that distribution representing the data is very low. thus, the sampled t-distributions have a relatively precise estimate of the mean.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: yeah. I suppose eventually I could implement a Bayesian ANOVA. that's outside the scope of this PR though

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,608 @@
...
+
+                        ProteinQuantificationEngineResult result;
+
+                        if (peptideFoldChanges.Count > 1)
+                        {
+                            // run the Bayesian analysis
+                            result = RunBayesianProteinQuant(peptideFoldChanges, protein, condition,
+                                randomSeedsForEachProtein[protein][0], null, false, BurnInSteps, NSteps, out var mus);
+
+                            var skepticalResult = RunBayesianProteinQuant(peptideFoldChanges, protein, condition,
```
zrolfs: This variable isn't used for anything.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,608 @@
...
+                    else
+                    {
+                        ProteinsWithConstituentPeptides.Add(protein, new List<Peptide> { peptide.Value });
+                    }
+                }
+            }
+
+            // calculate peptide biorep intensities
+            var allpeptides = ProteinsWithConstituentPeptides.Values.SelectMany(p => p).Distinct().ToList();
+            PeptideToSampleQuantity = new Dictionary<(Peptide, string, int), double>();
```
zrolfs: This dictionary isn't used for anything.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,608 @@
...
+            // this is a weird case where there are very few proteins in the data...
+            if (experimentalNullResults.Count < 10)
+            {
+                if (experimentalNullResults.Any())
+                {
+                    stdDev = experimentalNullResults.SelectMany(p => p.Value).Select(p => p.StandardDeviationPointEstimate).Average();
+                }
+                else
+                {
+                    // TODO: not sure what to do here.. there were no protein quant results for this condition
```
zrolfs: ...did you ever get to this? Can you remove the TODO and explain what your solution was, or elaborate on the TODO and  what your work around currently is?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Numerically:
+        /// F = fold change
+        /// C = cutoff
+        /// 
+        /// if F > C, then P(H0|D) = P(F < C) / P (F > C)
+        /// if F < -C, then P(H0|D) = P(F > -C) / P (F < -C)
+        /// </summary>
+        public void CalculatePosteriorErrorProbability(double[] skepticalMus)
+        {
+            double pep = 1.0;
```
zrolfs: What's the point of assigning a value to pep here and at ln 71 if you just overwrite it at ln 101? Delete these assignments.

rmillikin: it's mostly to be robust against bugs. if there is some problem calculating the PEP, the protein will be marked as rejecting the alternative hypothesis, rather than accepting it. in other words, I want the default value of PEP to be 1.0 and not 0.0, which is the default assignment for `double`

zrolfs: The statement "pep = nullHypothesisCount / alternativeHypothesisCount" will always produce a value that will overwrite the 1.0 or the value double.NaN. If there is a bug and one of those counts are null, an exception will be thrown that is not caught, which is not robust. If you suspect a bug could occur, use an exception and have the catch set the pep value to 1.0/NaN. The logic doesn't make sense as it currently reads.

rmillikin: there can't be a crash because `double` can't be null but I see your point with the NaN, I'll fix that

zrolfs: I'm still lost on how the pep = 1.0 would ever make it to the end of the method.

rmillikin: > I see your point with the NaN, I'll fix that

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: I changed the wording of this comment a bit

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -568,7 +568,7 @@ private double DetermineExperimentalNull(string treatmentCondition)
                 else
                 {
                     // TODO: not sure what to do here.. there were no protein quant results for this condition
-                    // throw an exception... or just make the cutoff 1.0?
+                    // for now the cutoff is just made to be a fold-change of 1.0, but maybe an exception should be thrown...
                     stdDev = 0.3333;
```
zrolfs: Where does this value come from?

zrolfs:
Can you add another unit test to bump the code coverage up?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
rmillikin: it's arbitrary. if there are no proteins to determine a fold-change, 1.0 is used to continue with the algorithm without crashing. the user can choose to re-run with a manually input cutoff if they want

zrolfs: My math's a little rusty, but 0.3333 != 1.0

rmillikin: the fold-change cutoff is 3 * stdDev

zrolfs: Can you add that info in a comment here?

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -0,0 +1,617 @@
...
+
+                    //throw new MzLibException("Could not determine experimental null for condition " + treatmentCondition);
+                }
+            }
+            else
+            {
+                stdDev = experimentalNullResults.SelectMany(p => p.Value).Select(p => p.FoldChangePointEstimate).InterquartileRange() / 1.34896;
+            }
+
+            experimentalNull = 3.0 * stdDev;
```
zrolfs: This is the first time "experimentalNull" is used, so there's no reason to initialize it back at ln 497

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// if F > C, then P(H0|D) = P(F < C) / P (F > C)
+        /// if F < -C, then P(H0|D) = P(F > -C) / P (F < -C)
+        /// </summary>
+        public void CalculatePosteriorErrorProbability(double[] skepticalMus)
+        {
+            double pep;
+
+            if (cutoff == null || skepticalMus == null)
+            {
+                pep = double.NaN;
```
zrolfs: This is still useless assignment, since it will always be overwritten at ln 103

PR #495, opened 2019-07-12
Allow Silac Label Mixtures

zrolfs:
Needed for turnover experiments because of amino acid recycling: https://www.hindawi.com/journals/ijcb/2015/798936/

### DIFF
--- a/Proteomics/Modifications/SilacLabel.cs
+++ b/Proteomics/Modifications/SilacLabel.cs
@@ -39,6 +39,22 @@ namespace Proteomics
             }
         }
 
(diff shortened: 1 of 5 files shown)


rmillikin:
can you digest a human database and make sure you get the same set of peptides (same #, same modified sequences, etc.) and measure RAM usage, time, etc.? Just want to make sure this doesn't affect "normal" digestion negatively

zrolfs:
I'm not sure why it would; everything is contained within the "GetSilacPeptides" method.

PR #500, opened 2019-07-18
replace unrecognized amino acid characters in protein databases with 'X'

rmillikin:


rmillikin:
see metamorpheus issue [1697](https://github.com/smith-chem-wisc/MetaMorpheus/issues/1697)

PR #502, opened 2019-07-25
correct PEP calculation in Bayesian quant

rmillikin:


FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
```diff
@@ -8,125 +8,105 @@ namespace FlashLFQ
     public class ProteinQuantificationEngineResult
     {
         public readonly ProteinGroup protein;
-        public readonly string condition1;
-        public readonly string condition2;
+        public readonly string BaseCondition;
```
trishorts: Does this only work for binary? Can it be extended to more conditions? In drug trials on cells one usually has control, vehicle and treatment.

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
```diff
@@ -8,125 +8,105 @@ namespace FlashLFQ
...
+        public readonly string TreatmentCondition;
         public readonly double FoldChangePointEstimate;
         public readonly double ConditionIntensityPointEstimate;
         public readonly double StandardDeviationPointEstimate;
         public readonly double NuPointEstimate;
-        public readonly (double, double) hdi_95;
-        public readonly List<(Peptide peptide, List<double> foldChanges)> peptideFoldChangeMeasurements;
+        public readonly (double, double) HDI_95;
```
trishorts: HDI_95? maybe explain

TestFlashLFQ/TestFlashLFQ.cs
```diff
@@ -965,19 +965,19 @@ public static void TestBayesianProteinQuantification()
 
             var quantResult = pg.conditionToQuantificationResults["b"];
 
-            Assert.That(Math.Round(quantResult.cutoff.Value, 3) == 0.366);
-            Assert.That(Math.Round(quantResult.PosteriorErrorProbability, 3) == 0.089);
+            Assert.That(Math.Round(quantResult.NullHypothesisCutoff.Value, 3) == 0.366);
```
trishorts: there aren't (m)any comments in the flashlfq unit tests. Maybe at least add comments to these explain what is being test

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngine.cs
```diff
@@ -447,7 +447,7 @@ private List<(Peptide, List<double>)> GetPeptideFoldChangeMeasurements(ProteinGr
             double mean = foldChanges.Mean();
 
             // the Math.max is here for cases where the cutoff is very small (e.g., 0)
-            double priorMuSd = Math.Max(0.05, skepticalPrior ? nullHypothesisCutoff.Value : sd * 100);
+            double priorMuSd = Math.Max(0.05, skepticalPrior ? (nullHypothesisCutoff.Value * (2.0 / 3.0)) : sd * 100);
```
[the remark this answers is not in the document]
rmillikin: it was a heuristic but as we talked about this morning I changed it to something that is (in my view) very theoretically sound. I also added a comment about why I chose the prior that I did

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
rmillikin: right now it's only for binary comparisons. you can call the engine multiple times with different base conditions if you want an all vs. all comparison. at some point it would be cool to do a bayesian ANOVA but that's probably down the line a ways

FlashLFQ/Bayesian Protein Quantification/ProteinQuantificationEngineResult.cs
rmillikin: I added a comment here

TestFlashLFQ/TestFlashLFQ.cs
rmillikin: I added a comment at the top of the test

PR #503, opened 2019-07-25
Change protein quant to use peptides instead of peaks

zrolfs:


zrolfs:
Too many conflicts, just going to update and redo

PR #513, opened 2019-09-04
stop gain bug fix

rmmiller22:
Bug in accounting for stop gain variant that cause peptides was causing a crash. Simple fix. No more crashes.

### DIFF
--- a/Test/TestPeptideWithSetMods.cs
+++ b/Test/TestPeptideWithSetMods.cs
@@ -625,9 +625,11 @@ namespace Test
                 new Protein("MPEPTIDE", "protein8",sequenceVariations: new List<SequenceVariation> { new SequenceVariation(4, 4, "P", "PPP", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", new Dictionary<int, List<Modification>> {{ 5, new[] { mp }.ToList() } }) }),
                 new Protein("MPEPTIDEPEPTIDE", "protein9", sequenceVariations: new List<SequenceVariation> { new SequenceVariation(4, 15, "PTIDEPEPTIDE", "PPP", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", null) }),
                 new Protein("MPEPTIDE", "protein10", oneBasedModifications: proteinPMods ,sequenceVariations: new List<SequenceVariation> { new SequenceVariation(4, 4, "P", "V", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", null) }),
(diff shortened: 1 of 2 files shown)


Test/TestPeptideWithSetMods.cs
```diff
@@ -625,9 +625,14 @@ public static void TestIdentifyandStringMethods()
                 new Protein("MPEPTIDE", "protein8",sequenceVariations: new List<SequenceVariation> { new SequenceVariation(4, 4, "P", "PPP", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", new Dictionary<int, List<Modification>> {{ 5, new[] { mp }.ToList() } }) }),
                 new Protein("MPEPTIDEPEPTIDE", "protein9", sequenceVariations: new List<SequenceVariation> { new SequenceVariation(4, 15, "PTIDEPEPTIDE", "PPP", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", null) }),
                 new Protein("MPEPTIDE", "protein10", oneBasedModifications: proteinPMods ,sequenceVariations: new List<SequenceVariation> { new SequenceVariation(4, 4, "P", "V", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", null) }),
+                new Protein("MPEPTIDE", "protein11", sequenceVariations: new List<SequenceVariation> { new SequenceVariation(5, 5, "T", "*", @"1\t50000000\t.\tA\tG\t.\tPASS\tANN=G||||||||||||||||\tGT:AD:DP\t1/1:30,30:30", null) }), //stop-gain (can identify)
```
acesnik: These will actually look like `SequenceVariation(4, 8, "PTIDE", "P",` instead of `SequenceVariation(5, 5, "T", "*",`

acesnik: I did this, so that we can use the same replace methods for all different types of variations. Otherwise, we'd end up with `*` all over, and we'd have to make sure we handled that correctly everywhere.

acesnik: @rmmiller22 It seems like this works, but could you add the same test with the `SequenceVariation(4, 8, "PTIDE", "P",` stop gain notation, so that we make sure that works, too?

rmmiller22: There was a previously existing test that handled a scenario like SequenceVariation(4,8, "PTIDE", "P"). This variant is identified without issue because it intersects the peptide sequence. However, while testing the code with real data, stop gain variations from the database do show up as T5* not as PTIDE4P. The star format previously was not being handled correctly, since the variant residue does not intersect with the peptide, making these stop gain variants unidentified. When I added code this weekend (yesterday's PR) to identify variants that caused "cleavage sites" or caused a peptide to be formed that would not previously have existed, this issue with stop gain variants came to light and are now properly handled with this PR. Now truncations and stop gain variants are both handled appropriately by the code

acesnik: Thanks for clarifying!!

rmmiller22: line 543  of TestIntersectsSequenceVariations() is the test for the truncation/stop gain format (for future reference)

acesnik:
@rmillikin I merged this, FYI. It only affects the sequence variant intersection code.

PR #517, opened 2019-09-16
nterminal protease bug fix

rmmiller22:
if the variant occurs at the previous end of the protein, and the variant adds on additional amino acids that cause N terminal cleavage. The original protein will be shorter than peptide +1. These peptides cannot be identified as variant because there is no way to tell if the protein ends  therefore causing the peptide or if the variant exists causing the same peptide sequence.


acesnik:
Good catch; that's a tough one.

rmmiller22:
> Good catch; that's a tough one.

Thanks. If only I would have caught it before it caused the crash haha!

PR #523, opened 2019-11-20
Memory efficient fragmentation

rmillikin:
this PR looks more extensive than it really is. the major changes are in the `PeptideWithSetModifications.Fragment()` method. the syntax to use that method have changed, so unit tests had to be updated, though their content is the same. the `CompactPeptide` and `CompactPeptideBase` classes have been removed, along with their corresponding unit tests. Additional helper methods have been added to `DissociationTypeCollection` for computational efficiency. Unused code has also been removed.


Proteomics/Fragmentation/DissociationTypeCollection.cs
```diff
@@ -41,19 +65,31 @@ public class DissociationTypeCollection
             { ProductType.D, null},// diagnostic ions are not shifted but added sumarily
         };
 
-        //From http://www.matrixscience.com/help/fragmentation_help.html
-        //Low Energy CID -- In low energy CID(i.e.collision induced dissociation in a triple quadrupole or an ion trap) a peptide carrying a positive charge fragments mainly along its backbone, 
-        //generating predominantly b and y ions. In addition, for fragments containing RKNQ, peaks are seen for ions that have lost ammonia (-17 Da) denoted a*, b* and y*. For fragments containing 
-        //STED, loss of water(-18 Da) is denoted a°, b° and y°. Satellite ions from side chain cleavage are not observed.
-        public static Dictionary<ProductType, List<char>> ProductTypeAminoAcidSpecificites = new Dictionary<ProductType, List<char>>
+        private static Dictionary<DissociationType, (double[], double[])> DissociationTypeToNTermMassCaps = new Dictionary<DissociationType, (double[], double[])>();
```
zrolfs: I can't figure out what a "mass cap" is.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -175,174 +175,336 @@ public string SequenceWithChemicalFormulas
...
+                    // add side-chain mod
+                    if (AllModsOneIsNterminus.TryGetValue(r + 2, out Modification mod))
+                    {
+                        nTermMass += mod.MonoisotopicMass.Value;
+                    }
+
+                    // handle star and degree ions for low-res CID
+                    if (dissociationType == DissociationType.LowCID)
+                    {
+                        if (BaseSequence[r] == 'R' || BaseSequence[r] == 'K' || BaseSequence[r] == 'N' || BaseSequence[r] == 'Q')
```
zrolfs: I understand we're in a "don't allocate memory" stance right now, but it was my understanding that if you have to look up an index (e.g. BaseSequence[r]) more than twice, then it was faster to store it (e.g. char currentResidue = BaseSequence[r]). Just wanted to confirm.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -175,174 +175,336 @@ public string SequenceWithChemicalFormulas
...
 
-                for (int f = 0; f < terminalMasses.Length; f++)
-                {
-                    if (completeFragmentationSeries)
+                    // skip first N-terminal fragment (b1, aDegree1, ...) for CID
+                    if (r == 0 && (dissociationType == DissociationType.CID || dissociationType == DissociationType.LowCID))
                     {
-                        // fragments with neutral loss
-                        if (AllModsOneIsNterminus.TryGetValue(terminalMasses[f].AminoAcidPosition + 1, out Modification mod) && mod.NeutralLosses != null
+                        goto CTerminusFragments;
```
zrolfs: image

trishorts: wow. i didn't even know that goto still existed. when i learned to program in the 80s, we called program with gotos "spaghetti code"

zrolfs:
In general this looks like a great PR. There are several nuances involved for code like fragmentation and digestion which pose challenges for overhauls like this. We hope the existing unit tests are effective at preventing unexpected issues, but we should remain vigilant for a bit following the merge of this PR.

I want a commented description or a change in nomenclature for "MassCaps", but otherwise I'm ready to approve.

Proteomics/Fragmentation/DissociationTypeCollection.cs
rmillikin: I can change this name, I suppose the current nomenclature is "mass shift"

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
rmillikin: Not sure, I can benchmark this

rmillikin:
I built this in release mode and packaged it into a NuGet package, which I put into MetaMorpheus. The current master branch and the proposed fragmentation changes produce identical classic search results (17345 PSMs, same PSM full sequences). The classic search time changes from 1:24 to 1:14 (MM built in debug mode for both).

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
rmillikin: it's about 3% faster to save the char rather than retrieving the char from the string each time according to my benchmarking

PR #525, opened 2019-12-02
191126 fix slide decoy

zrolfs:
There was an infinite while loop while trying to "slide" a single amino acid variant. I fixed/removed nonsense code. Some try-catches were replaced with more efficient if-else


Test/DatabaseTests/TestProteinReader.cs
```diff
@@ -388,6 +390,10 @@ public static void TestSlideDecoyXML()
                 Assert.AreEqual(ok2[1].BaseSequence[bond.OneBasedBeginPosition - 1], 'C');
                 Assert.AreEqual(ok2[1].BaseSequence[bond.OneBasedEndPosition - 1], 'C');
             }
+
```
rmmiller22: Can you expand on this test to make sure that results of decoy generation are as expected, not just that it does not loop indefinitely.

zrolfs: Good call. Caught some bugs.

UsefulProteomicsDatabases/DecoyProteinGenerator.cs
```diff
@@ -393,6 +360,36 @@ private static List<Protein> GenerateSlideDecoys(List<Protein> proteins, int max
             return decoyProteins;
         }
 
+        private static Dictionary<int, List<Modification>> SlideProteinSequenceWithMods (char[] sequenceArraySlided, char[] sequenceArrayUnslided, bool initiatorMethionine, int numSlides, Protein protein)
```
rmmiller22: Would also be good to have a unit test for slide modifications since there is also not one previously existing.

PR #530, opened 2019-12-18
Dramatically Improve Peptide Deserialization Speed

zrolfs:
For a nonspecific search, peptide deserialization took up 6 minutes of an 11 minute run. Now it takes 2 minutes of a 7 minute run.


zrolfs:
Large time sinks were the mods (~1.5 min) and the digestion params (~3.5 min). Mods are now <1 min and digestion params are negligible. There's still ~1 min for reading the data.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -851,116 +844,71 @@ public override int GetHashCode()
...
+                {
+                    currentModificationLocation++;
+                }
+                //else do nothing
             }
-
-            baseSequence = baseSequenceSb.ToString();
-        }
-
-        private void GetProteinAfterDeserialization(Dictionary<string, Protein> idToProtein)
```
rmillikin: does removing this method really save a lot of time? the exception message thrown in the deleted code is more informative than a generic keynotfound exception

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -851,116 +844,71 @@ public override int GetHashCode()
         /// <summary>
         /// This should be run after deserialization of a PeptideWithSetModifications, in order to set its Protein and Modification objects, which were not serialized
         /// </summary>
-        public void SetNonSerializedPeptideInfo(Dictionary<string, Modification> idToMod, Dictionary<string, Protein> accessionToProtein)
+        public void SetNonSerializedPeptideInfo(Dictionary<string, Modification> idToMod, Dictionary<string, Protein> accessionToProtein, DigestionParams dp)
```
rmillikin: are the digestionparams known after deserialization? i.e., where does the "dp" parameter come from?

zrolfs:
There are other methods using the "GetModsAfterDeserialization" method, so I still need to address those.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
zrolfs: The dp is known from the search task that's being done. Currently, each digestion parameter needs its own index, such that all peptides within a given index contain the same digestion parameter object. This is output in the index params file.

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
zrolfs: We can keep this method. It only added a few seconds.

zrolfs:
Odd that appveyor passed while my visual studio unit tests are failing...

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -62,18 +59,14 @@ public class PeptideWithSetModifications : ProteolyticPeptide
             }
 
             FullSequence = sequence;
-            GetModsAfterDeserialization(allKnownMods, out _baseSequence);
+            GetModsAfterDeserialization(allKnownMods);
```
rmillikin: I believe the `out _baseSequence` part is so the base sequence gets set if the protein is not known (for example if you construct a peptide with the `PeptideWithSetModifications(string sequence ...)` constructor). can you check that you can serialize/deserialize these types of peptides and that their base sequences are set correctly?

zrolfs: Yep, that's what was causing the unit test failures.

zrolfs:
Alright, this is ready again!

PR #538, opened 2020-03-20
Fix a bug for ETD fragmentation

lonelu:
Fix a bug for ETD fragmentation. 
Nick found this bug for O-glyco peptide
 "T[O-Glycosylation:H1N1 on T]VYLGAS[O-Glycosylation:H1N1A1 on S]K", 
basicly the zDot8 ion of the peptide contain a glycan, the previous zDot8 ion didn't add the mass of the modification.


trishorts:
is this the sequence that is being checked, including the glycan? i'm looking up fragment masses in protein prospector. 
image

lonelu:
Yes, I also checked at prospector. They don't include zDot8 ion as we did.

PR #540, opened 2020-04-02
return empty data file instead of crashing if there are 0 scans

rmillikin:
https://github.com/smith-chem-wisc/MetaMorpheus/issues/1843

I couldn't unit test the .raw files because I can't create fake files with 0 scans, but I did do that for mzML and mgf


zrolfs:
Is there a reason to throw besides just handling it as a zero scan datafile? It would suck for somebody to be searching 50 files and then 40 files in have a throw.

zrolfs:
If this is related to the low res crashes, I thought those were caused by processing after loading?

rmillikin:
well right now it throws an index out of bounds exception, this is just an improvement of the error message. at some point we could try/catch or check # scans or something downstream but it's a more difficult solution, this one at least provides meaningful feedback

rmillikin:
@lschaffer2 got this message when transferring a file from bison to her computer and it didn't transfer completely or something. so it's probably a super rare problem

zrolfs:
Where does it throw? In MetaMorpheus? I don't love that we throw an error upstream when we can handle it downstream.

rmillikin:
i agree but `Parallel.ForEach(Partitioner.Create(0, msDataScans.Length)` throws an exception and the only way to handle that is to catch the exception. i don't want to catch the exception in MM because it's a generic ArgumentOutOfRangeException which could be thrown from something else

rmillikin:
nvm I think I see what you're saying, instead of throwing an exception it should just return a file with 0 MSDataScan objects. i think that's a better solution than this, you're right. we'd have to handle it in metamorpheus but it would be easier than dealing with an exception

ThermoRawFileReader/ThermoDynamicData.cs
```diff
@@ -76,7 +76,6 @@ private void InitiateDynamicConnection(string filePath)
             }
 
             dynamicConnection.SelectInstrument(Device.MS, 1);
-            GetMsOrdersByScanInDynamicConnection();
```
zrolfs, line 79: ?

rmillikin: GetMsOrdersByScanInDynamicConnection is already called on line 20 when creating a dynamic connection. i just deleted this line because it's a duplicate call and it's useless anyway (the returned data is not saved)

PR #545, opened 2020-09-04
WebClient To Download UniProt Proteome XML or FASTA

trishorts:


Test/DatabaseTests/TestDatabaseLoaders.cs
```diff
@@ -505,5 +507,28 @@ public static void IsoformReadTest()
 
             Assert.AreNotEqual(proteinList[2].Accession, proteinList[4].Accession);
         }
+
+        [Test]
+        public static void TestListOfAvailableUniProtProteomes()
+        {
+            Dictionary<string,string> proteomesDictionary = ProteinDbRetriever.UniprotProteomesList();
+            Assert.AreEqual(292464, proteomesDictionary.Keys.Count);
```
acesnik, line 515: Do we need to test the exact number? Could we just test that it's greater than 200k entries?

Test/DatabaseTests/TestDatabaseLoaders.cs
```diff
@@ -197,7 +199,7 @@ public void FilesEqualHash()
 
             // UniProt PTMs
             var uniprotPtms = Loaders.LoadUniprot(uniProtPath, formalChargesDictionary).ToList();
-            Assert.AreEqual(347, uniprotPtms.Count()); // UniProt PTM list may be updated at some point, causing the unit test to fail
+            Assert.AreEqual(359, uniprotPtms.Count()); // UniProt PTM list may be updated at some point, causing the unit test to fail
```
acesnik, line 202: Do we need to test the exact number? Could we test that it's greater than 300 so we don't have to update it?

Test/DatabaseTests/TestDatabaseLoaders.cs
trishorts: no. and you're right not to include a specific number. not sure this really counts as a unit test. actually could use ideas for real unit tests. these are pretty weak.

rmillikin:
i'd probably just try to find a very small proteome (<100 kilobyte download size) and check to make sure the organism name is correct, at least N proteins, maybe check for 1 or 2 particular protein base sequences are correct, etc. not sure about mods. maybe just check that a few known ones exist on one or two proteins

acesnik (changes requested):

UsefulProteomicsDatabases/ProteinDbRetriever.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// The list of available UniProt Proteomes was downloaded 09/03/2020
+        /// This is returned as a dictionary
+        /// key = Proteome ID
+        /// value = Organism
+        /// </summary>
+        /// <returns></returns>
+        public static Dictionary<string,string> UniprotProteomesList()
+        {
+            string currentDirectory = Directory.GetCurrentDirectory();
+            string filePath = Path.Combine(currentDirectory, "UniProtProteomes.txt");
```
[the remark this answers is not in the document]
trishorts: I believe the answer is yes, and i have added a skeleton for that. For instance, we can request different file formats (text/tab, etc.). And there is a list of known columns. So we can submit protein list and get back specific metadata.

However, there is also a way to search with key word (see below). We just have to format the query correctly. And, i'd need a constructor for that. If you have an example in mind, i'll try to make the constructor and a unit test. Otherwise, i prefer just ot let this dog lie. We really have immediate use for proteomes and no specific use for other types of queries currently

https://www.uniprot.org/help/text-search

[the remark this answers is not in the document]
acesnik: Sounds good. The reason I ask is that we have to duplicate the text files in mzLib currently, e.g. https://github.com/smith-chem-wisc/mzLib/blob/master/Proteomics/ProteolyticDigestion/proteases.tsv and https://github.com/smith-chem-wisc/MetaMorpheus/blob/master/EngineLayer/ProteolyticDigestion/proteases.tsv. Maybe that's been resolved, but at the time, we had trouble getting the nuget file to package the proteases.tsv file, so the same may be true for these text files.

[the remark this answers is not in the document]
acesnik: On that note, these text files should be added to the [nuspec file](https://github.com/smith-chem-wisc/mzLib/blob/master/mzLib.nuspec) if you do end up keeping them.

zrolfs (approved):

UsefulProteomicsDatabases/ProteinDbRetriever.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// A poorly formed query will not return anything, but it won't crash.
+        /// </summary>
+        /// <param name="proteomeID">valid proteome ID corresponding to a specific organism (e.g. UP000005640)</param>
+        /// <param name="formt">format of retrieved proteome (e.g. fasta or xml)</param>
+        /// <param name="reviewed">if yes file contains only reviewd proteins</param>
+        /// <param name="compress">if yes file is saved as .gz</param>
+        /// <param name="absolutePathToStorageDirectory"></param>
+        public static void RetrieveProteome(string proteomeID, string absolutePathToStorageDirectory, ProteomeFormat format, Reviewed reviewed, Compress compress, Include include)
+        {
+            if (Directory.Exists(absolutePathToStorageDirectory))
```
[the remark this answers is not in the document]
acesnik: Looks like you defined the expected behavior differently, so I'll mark this resolved.

UsefulProteomicsDatabases/UsefulProteomicsDatabases.csproj
```diff
@@ -19,4 +19,10 @@
     <ProjectReference Include="..\Proteomics\Proteomics.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="UniProtKB_columnNamesForProgrammaticAccess.txt">
```
acesnik, line 23: Need to add this to the nuspec file

acesnik (changes requested):

UsefulProteomicsDatabases/UsefulProteomicsDatabases.csproj
acesnik: Like this: https://github.com/smith-chem-wisc/mzLib/blob/master/mzLib.nuspec#L36

trishorts: does this look right? hard to tell outside of metamorpheus:
image

acesnik: I guess I'm a little confused because I'd expect it to copy to the directory `UsefulProteomicsDatabases`, in which case I think it should be `netcoreapp3.0\UsefulProteomicsDatabases\UniProtKB_column...` and the target should be `lib/netcoreapp3.0/UsefulProteomicsDatabases`. Does it copy to the base directory when you build it?

acesnik: Just tested the nuget packaging, and it works.

acesnik (approved):

rmmiller22 (approved):

PR #546, opened 2020-09-15
memory-efficient indexed mzml writing

rmillikin:
- Index is valid according to OpenMS validating tool ( https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/release/latest/html/TOPP_FileInfo.html )
- It's currently a bit slow but is much more memory-efficient than the previous method


rmillikin:
Regarding the checksum:
OpenMS calculates the checksum by reading the whole file sans checksum and computing the checksum from that: https://github.com/OpenMS/OpenMS/blob/cfaa2fddfff1b22f07f475a33931596aa34a3979/src/openms/source/FORMAT/FileHandler.cpp#L507

Technically the checksum should be:
"SHA-1 checksum from beginning of file to end of 'fileChecksum' open tag." see https://raw.githubusercontent.com/HUPO-PSI/mzML/master/schema/schema_1.1/mzML1.1.1_idx.xsd

So I think it will be difficult to validate if we've actually got the checksum correct..

See also this OpenMS thread:
https://github.com/OpenMS/OpenMS/issues/2418

trishorts (approved):

rmmiller22 (approved):

PR #547, opened 2020-09-16
easier charting (mzPlot)

rmillikin:


rmillikin:
you can make a plot with a single line of code w/ this change.


```cs
ScatterPlot plot = new ScatterPlot(examplePlotView, new List<Datum> { new Datum(0, 0), new Datum(1, 1) });
```

MzLibUtil/Datum.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            this.X = x;
+            this.Y = y;
+            this.Z = z;
+
+            this.Label = label;
+            this.Weight = weight;
+        }
+
+        public override string ToString()
```
trishorts, line 23: Should this class have an "Equals" methods?

MassSpectrometry/Chromatography.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Text;
+
+namespace MassSpectrometry
+{
+    public class Chromatography
+    {
+        /// <summary>
+        /// Extracts an ion chromatogram from a spectra file, given a mass, charge, retention time, and mass tolerance.
+        /// </summary>
+        public static ExtractedIonChromatogram ExtractIonChromatogram(MsDataFile file, double mass, int charge, Tolerance massTolerance,
```
trishorts, line 15: rename mass as neutralMass

MassSpectrometry/Chromatography.cs
```diff
@@ -0,0 +1,44 @@
...
+            double retentionTime, int msOrder = 1, double retentionTimeWindowWidth = 5)
```
trishorts, line 16: for times i think we need to specify units (either min or sec)

MassSpectrometry/Chromatography.cs
```diff
@@ -0,0 +1,44 @@
...
+
+                double expMz = scan.MassSpectrum.XArray[ind];
+
+                if (massTolerance.Within(expMz.ToMass(charge), mass))
+                {
+                    xicData.Add(new Datum(scan.RetentionTime, scan.MassSpectrum.YArray[ind]));
+                }
+                else
+                {
+                    xicData.Add(new Datum(scan.RetentionTime, 0));
```
trishorts, line 37: i guess we should zero fill but I am not certain about this. return arrays would be smaller returning only data points with a recorded intensity

MassSpectrometry/ExtractedIonChromatogram.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        /// <summary>
+        /// Returns the XIC data, ordered by retention time. 
+        /// Each datum's X field is the m/z value, Y is the intensity, and Z is the retention time.
+        /// </summary>
+        public IEnumerable<Datum> Data
+        {
+            // return the data, ordered by retention time
+            get { return XicData.OrderBy(p => p.Z).AsEnumerable(); }
```
trishorts, line 26: should this be x? i thought that x was time

Test/TestMsDataFile.cs
```diff
@@ -180,6 +184,18 @@ public void TestAMoreRealFile()
...
+        [Test]
+        public static void TestXicExtraction()
+        {
+            string dataFilePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "DataFiles", "SmallCalibratibleYeast.mzml");
+            var data = Mzml.LoadAllStaticData(dataFilePath);
+            
+            var peptide = new PeptideWithSetModifications("KAPAGGAADAAAK", new Dictionary<string, Modification>());
+
+            var xic = Chromatography.ExtractIonChromatogram(data, peptide.MonoisotopicMass, 2, new PpmTolerance(10), 24.806);
+            Assert.That(xic.Data.Count(p => p.Y > 0) == 4);
```
trishorts, line 196: again wondering about the utility of including zero values.

mzPlot/Annotations/PlotTextAnnotation.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public class PlotTextAnnotation : Annotation
+    {
+        public PlotTextAnnotation()
+        { 
+
+        }
+
+        public string Text { get; set; }
+        public double X { get; set; }
```
trishorts, line 17: it's not clear to me if this is a data label or just floating text in the graph with offsets. If offsets, what is the reference point from which the offsets are measured.

mzPlot/Plots/BarPlot.cs
```diff
@@ -0,0 +1,27 @@
...
+using OxyPlot.Axes;
+
+namespace mzPlot
+{
+    public class BarPlot : Plot
+    {
+        /// <summary>
+        /// Creates a bar plot. The data X value is the height of the bar, and the data label is the X-axis label under the bar.
+        /// </summary>
+        public BarPlot(OxyPlot.Wpf.PlotView oxyPlotView, IEnumerable<Datum> data, OxyColor? borderColor = null, OxyColor? fillColor = null,
```
trishorts, line 18: is there an option for data labels? value? what about error bars?

mzPlot/Plots/ScatterPlot.cs
```diff
@@ -0,0 +1,26 @@
...
+using OxyPlot.Axes;
+
+namespace mzPlot
+{
+    public class ScatterPlot : Plot
+    {
+        /// <summary>
+        /// Creates a scatter plot. The data X value is the X-coordinate, the data Y value is the Y-coordinate.
+        /// </summary>
+        public ScatterPlot(OxyPlot.Wpf.PlotView oxyPlotView, IEnumerable<Datum> data, OxyColor? markerColor = null, double markerSize = 3,
```
trishorts, line 18: what about x- and y-error bars?

can axes be on log scale?

trishorts (dismissed):
This is a good PR. I'll leave it to you to address my comments in a way that seems appropriate to you.

MassSpectrometry/ExtractedIonChromatogram.cs
rmillikin: good catch, will fix

MassSpectrometry/Chromatography.cs
rmillikin: I prefer minutes just because 4000 seconds is not as intuitive to me as 66.6 minutes but maybe that's just a personal thing. i'll rename it though

MassSpectrometry/Chromatography.cs
rmillikin: I'd like to add options for 1. padding w/ zeroes (current), 2. not padding w/ zeroes, or 3. ignoring all zeroes. prob will be a separate PR though

mzPlot/Annotations/PlotTextAnnotation.cs
rmillikin: this one is a floating text in the graph, not coupled to any data. I think it's in reference to the upper left pixel in the chart but I'll check and make a comment.

I added it because we had a really hard time adding the peptide sequence in metadraw, and right now we use a hacky workaround where we generate a separate image and tack it onto the graph. it's not great. I want to add a series of annotation options for both data-coupled and non-data-coupled, but in a separate PR.

mzPlot/Plots/BarPlot.cs
rmillikin: you can add this stuff by accessing the oxyplot model but it's kind of hard. I do want to add these features in a more intuitive way (probably adding xError, yError, zError in the Datum object) in a separate PR

mzPlot/Plots/ScatterPlot.cs
rmillikin: both probably in a future PR

MzLibUtil/Datum.cs
rmillikin: this one i'm not sure about... part of me says they're distinct data points even if they happen to have equal data values, another part says that if they have the same data they're equal to each other..

trishorts: having an equals doens't require the user to do anything in response. its simply a comparer.

rmillikin: it could affect downstream usage somewhat weirdly though. like if you do:

```cs
List<Datum> myListOfDataPoints = new List<Datum> { new Datum(1, 0), new Datum(1, 0) };
var distinctDataPoints = myListOfData.Distinct();

bool test = distinctDataPoints.Contains(new Datum(1, 0));
```

`distinctDataPoints` would only have 1 item in it, since the Datum objects in myListOfDataPoints are considered duplicates. `test` would also return `true`. Which I'm not sure is "correct" behavior. I'm not really sure of a situation where you'd ever want to get distinct data points though, so I'm not sure what the "correct" behavior should be

trishorts (approved):

jgpavek (approved):
Very convenient. My one comment about zero filling has been addressed by Shortreed, so I see nothing to change. I look forward to having these capabilities.

PR #549, opened 2020-10-14
new default protein quant in FlashLFQ

rmillikin:
switches default protein quant in FlashLFQ from top3 to median polish.

median polish is as good as maxLFQ but faster and easier. see https://academic.oup.com/bioinformatics/article/36/8/2611/5697917 supp figure 3b.

see https://mgimond.github.io/ES218/Week11a.html for an example of the median polish algorithm.

in my hands, median polish improved protein coefficients of variation (CVs) from ~11% to ~7%.


rmillikin:
I should note, this PR also summarizes fractionated data into sample-level data for the default protein quant output. peptide quant is still reported by individual MS file.

trishorts:
I'm content changing condition to sample group. 
But, I'm not supportive of changing biological replicate to sample OR changing technical replicate to replicate. Biological replicate and technical replicate have specific meanings in a statistical context whereas sample and replicate do not. It is also too easy to confuse replicate and biological replicate, each of which needs to be treated differently.

FlashLFQ/FlashLFQResults.cs
```diff
@@ -299,6 +300,181 @@ public void CalculateProteinResultsTop3(bool useSharedPeptides)
...
+                        {
+                            double max = 0;
+
+                            foreach (var sample in group)
+                            {
+                                double peptideIntensity = peptide.GetIntensity(sample);
+
+                                if (!double.IsNaN(peptideIntensity))
+                                {
+                                    max = Math.Max(max, peptideIntensity);
```
trishorts, line 363: a comment here would be helpful. i'm not sure whats happening in this step.

FlashLFQ/FlashLFQResults.cs
```diff
@@ -299,6 +300,181 @@ public void CalculateProteinResultsTop3(bool useSharedPeptides)
...
+                            {
+                                double sampleIntensity = 0;
+                                double highestFractionIntensity = 0;
+
+                                foreach (var fraction in sample.GroupBy(p => p.Fraction))
+                                {
+                                    double fractionIntensity = 0;
+                                    int replicatesWithValidValues = 0;
+
+                                    foreach (SpectraFileInfo replicate in fraction.OrderBy(p => p.Replicate))
```
trishorts, line 404: i'd like comments for this secition also. i'm not understanding how fraction intensites are being used.

FlashLFQ/FlashLFQResults.cs
```diff
@@ -299,6 +300,181 @@ public void CalculateProteinResultsTop3(bool useSharedPeptides)
...
+                    // do median polish protein quantification
+                    MedianPolish(peptideIntensityMatrix, 10, 0.0001, out var rowEffects, out var columnEffects, out var overallEffect);
+
+                    // set the sample protein intensities
+                    sampleN = 0;
+                    foreach (var group in SpectraFiles.GroupBy(p => p.SampleGroup).OrderBy(p => p.Key))
+                    {
+                        foreach (var sample in group.GroupBy(p => p.Sample).OrderBy(p => p.Key))
+                        {
+                            double sampleProteinIntensity = Math.Pow(2, columnEffects[sampleN]) * referenceProteinIntensity;
```
trishorts, line 451: why square the column effect? i haven't seen that in examples but maybe it makes mathematical sense.

FlashLFQ/FlashLFQResults.cs
```diff
@@ -415,5 +591,169 @@ public void WriteResults(string peaksOutputPath, string modPeptideOutputPath, st
                 Console.WriteLine("Finished writing output");
             }
         }
+
+        private static void MedianPolish(double[,] table, int maxIter, double tol, out double[] rowEffects, out double[] columnEffects, out double overallEffect)
```
trishorts, line 595: is this called iteratively? i only see one call, and I don't see a cut-off for vanishing returns. maybe i'm missing it?

rmillikin (commented):
I somehow can't reply to your other comments because github says they're "outdated" but I also added a comment to line 407. the median polish algorithm is iterative and has a tolerance for improvement; line 652 is the iteration and I've clarified the improvement variable name to "improvementCutoff", which is used on line 667 to check whether or not to continue the iteration procedure

FlashLFQ/FlashLFQResults.cs
rmillikin: added a comment to line 407

FlashLFQ/FlashLfqEngine.cs
```diff
@@ -201,7 +201,7 @@ public FlashLfqResults Run()
             _results.CalculatePeptideResults();
 
             // do top3 protein quantification
```
lonelu, line 203: Comment need to be updated.

lonelu (approved):

trishorts (approved):

FlashLFQ/FlashLfqEngine.cs
rmillikin: thx, good catch. it's not necessary to keep both. I think median polish would be better than top3 in the vast majority of scenarios. There's no particular reason to keep both methods around and I will probably eventually delete the top3 code but I'd rather be careful right now

PR #552, opened 2020-11-09
load periodic table automatically on startup

rmillikin:
- Move periodic table loading from the UsefulProteomicsDatabases project to the Chemistry project
- Loading the periodic table is called from a static constructor in the PeriodicTable class, meaning the periodic table will be loaded the first time the PeriodicTable class is referenced

So this PR will make coding new projects w/ mzLib more convenient because you won't get random error messages about elements being missing. But it does sort of mess up the idea of keeping the "idea of the periodic table" separate from "loading the elements in the periodic table". They're sort of mixed together now but I couldn't figure out another way to do it. So from a "theoretical" angle it's a bit janky but from a practical angle it's much more convenient.


acesnik (approved):
This looks like a good idea to me. Tracking down how to load the periodic table has been inconvenient at times.

PR #553, opened 2020-11-16
ability to write QuantifiedPsms file in FlashLFQ

rmillikin:
a user requested a "Quantified PSMs" file and this PR implements that


acesnik:
Looks good, and it looks like the tests cover making the new file, which is also good.

PR #554, opened 2020-12-01
autodetect fasta header type and use appropriate regex

rmillikin:
this PR allows passing a fasta file without prespecified regex to parse accession/name/gene/etc and mzLib will try to figure out how to parse the headers as appropriate.


trishorts (approved):
I downloaded to the latest gencode data base and it loaded just fine. This seems to be covered by unit tests. This is a good PR.

rmmiller22 (approved):

PR #557, opened 2020-12-09
Update proteases.tsv to match the same file in MetaMorpheus

trishorts:
Discovered today that the mzlib file was missing elastase, peptidomics and StcE-trypsin. I added those three proteases from the MetaMorpheus file to maintain consistency.


zrolfs (approved):

rmmiller22 (commented):
Good to go!

PR #558, opened 2020-12-09
[WIP] Decoy peptide from target peptide

trishorts:
Working towards decoy on the fly. This pull request adds a function to PeptideWithSetMods to create a related decoy. The new function identifies proteolytic sites in the peptide and keeps those amino acids and attached mods in the same position in the decoy. N-terminal mods are also preserved. All other amino acids in the peptide are reversed. Mods on reversed amino acids go with the amino acid so they appear in a new position on the decoy. 

Sometimes the above process returns the identical sequence. In this case, I added a new method that provides the mirror image. This reduces the incidence. But still there are some cases where a decoy still matches a target. This is from total palindrome (including cleavage motif) for example (RLEEEEELR). Oh well.

Traditional MetaMorpheus Protein reverse results in a higher number of decoys matching targets than this method by 5X.

Things to deal with:

- decoy protein name and sequence with subsequent effect on inference if any.


trishorts:
Closing — decoy-from-target already landed in master (`PeptideWithSetModifications.GetReverseDecoyFromTarget` + `PeptideWithSetModsDecoyTests`). 2022 [WIP], conflicting.

PR #561, opened 2021-01-06
enable proteolytic cleavage events to change the mass of the product peptide

rmmiller22:
For certain proteases, such as the chemical protease CNBr, proteolytic cleavage events alter the mass of the peptide so that it is not just the mass of the amino acid residues. With CNBr cleavage the C-terminal Met residue is converted into a N-acyl homoserine lactone. Previously this was not reflected in our peptide mass calculations. Now proteases can list specific cleavage residues that also have a mass shift associated with them, and these are applied during the digestion process. This application is generalized so that custom proteases can also incorporate this information.


Proteomics/Protein/Protein.cs
```diff
@@ -275,6 +278,73 @@ internal static IEnumerable<PeptideWithSetModifications> RemoveTerminalModificat
...
+                                {
+                                    var newMass = tempPeptide.MonoisotopicMass + massShift;
+                                    tempPeptide.SetMonoisotopicMass(newMass);
+                                }
+                            }
+
+                        }
+                        else if (cleavageIndex == 1) //look for residue at end of peptide
+                        {
+                            if (peptide.NextAminoAcid != '-')
```
rmillikin, line 330: use the protein.Length variable to determine the location in the protein sequence instead of '-'

Proteomics/Protein/Protein.cs
```diff
@@ -275,6 +278,73 @@ internal static IEnumerable<PeptideWithSetModifications> RemoveTerminalModificat
...
+                    var tempPeptide = peptide;
+                    foreach (var modEvent in residueSpecifics)
+                    {
+                        var cleavageIndex = modEvent.Value.Item1;
+                        var massShift = modEvent.Value.Item2;
+                        var residue = modEvent.Key;
+
+                        if (cleavageIndex == 0) //look for residue at begining of peptide 
+                        {
+                            if (peptide.PreviousAminoAcid != '-')
```
rmillikin, line 317: use the protein.Length variable to determine the location in the protein sequence instead of '-'

Proteomics/Protein/Protein.cs
```diff
@@ -275,6 +278,73 @@ internal static IEnumerable<PeptideWithSetModifications> RemoveTerminalModificat
...
+                var massShifts = digestionParams.Protease.CleavageMassShifts;
+                var digestionMotifs = digestionParams.Protease.DigestionMotifs;
+                foreach (var motif in digestionMotifs)
+                {
+                    var cleavageResidue = motif.InducingCleavage;
+                    // cutIndex of 0 is N and 1 is C
+                    var cleavageIndex = motif.CutIndex;
+                    if (massShifts.ContainsKey(cleavageResidue))
+                    {
+                        residueSpecifics.Add(cleavageResidue, (cleavageIndex, massShifts[cleavageResidue]));
```
rmillikin, line 302: would this work if 'X' was the motif residue? if you implemented this as a mod you could just use `ModificationLocalization.ModFits` to determine if the mod (CNBr mass shift) fits at this location.

actually on second thought I think all you'd have to do is in the `Protein.Digest` method, just add the mod that comes from CNBr cleavage to fixed mods. I think none of the subsequent code would have to change?

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -108,6 +108,12 @@ public double MonoisotopicMass
                 }
                 return (double)ClassExtensions.RoundedDouble(_monoisotopicMass.Value);
             }
+            
+        }
+
+        public void SetMonoisotopicMass(double mass)
+        {
+            _monoisotopicMass = mass;
```
rmillikin, line 116: seems kind of problematic because the correct mass can't be reconstructed from the sequence alone, you would need protease info. might be better to add a modification to the C term rather than an "invisible" mass shift? it also might be easier to implement

rmillikin (changes requested):

Proteomics/Protein/Protein.cs
```diff
@@ -227,33 +228,38 @@ public string GetEnsemblFastaHeader()
         /// </summary>
         public IEnumerable<PeptideWithSetModifications> Digest(DigestionParams digestionParams, IEnumerable<Modification> allKnownFixedModifications,
             List<Modification> variableModifications, List<SilacLabel> silacLabels = null, (SilacLabel startLabel, SilacLabel endLabel)? turnoverLabels = null)
-        {
+        {            
             //can't be null
             allKnownFixedModifications = allKnownFixedModifications ?? new List<Modification>();
+            List<Modification> allFixedModsList = allKnownFixedModifications.ToList();
```
rmillikin, line 234: creating a new list for each time `Digest` is called is going to create memory/garbage collection problems. I would just change `IEnumerable<Modification> allKnownFixedModifications` in the parameters to `List<Modification> allKnownFixedModifications`. then check `if (digestionParams.Protease.CleavageMod != null && allKnownFixedModifications.Contains(digestionParams.Protease.CleavageMod))`

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -96,7 +99,41 @@ public double MonoisotopicMass
             {
                 if (!_monoisotopicMass.HasValue)
                 {
-                    double monoMass = WaterMonoisotopicMass;
+                    double monoMass = 0;
```
rmillikin, line 102: can you explain what's going on here? if there are mods annotated in the sequence then you shouldn't need special code to calculate the mono mass right?

Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
```diff
@@ -314,7 +314,14 @@ private bool CanBeCTerminalMod(Modification variableModification, int peptideLen
                 {
                     case "N-terminal.":
                     case "Peptide N-terminal.":
-                        if (ModificationLocalization.ModFits(mod, Protein.BaseSequence, 1, peptideLength, OneBasedStartResidueInProtein))
+                        if (mod.ModificationType == "Protease" && ModificationLocalization.ModFits(mod, Protein.BaseSequence, 1, peptideLength, OneBasedStartResidueInProtein))
```
rmillikin, line 317: is the purpose of this section to put the mod on peptide N-term while avoiding protein N-Term?

Proteomics/Modifications/Modification.cs
```diff
@@ -127,6 +129,43 @@ public static string ModLocationOnPeptideOrProtein(string _locationRestriction)
             }
         }
 
+        public static  Modification ParseProteaseModificationsFromString(string input)
```
rmillikin, line 132: is it possible to use a "normal" mod annotation structure to avoid writing a second mod parser here?

Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
rmmiller22: So for traditional peptide generation we add the monoisotopic mass of water to the mass of all the amino acid residues because there is an -OH that caps the c-terminus and an -OH that caps the n-terminus.
Modifications that occur because of proteolytic cleavage can affect this. the homo-serine lactone formed by CNBr cleavage is a great example of this in that it would be incorrect to have the OH group on the C-terminus and that would make the mass incorrect. To remedy this. The protease modification takes in information as to whether or not the N-terminal H and C-terminal OH should be maintained for peptide cleavage. The changes here ensure that the mass calculations are handled properly.  
Its not per say about the mod itself its about the affect the mod has on the "finishing" of the peptide sequence. 
It would also not be entirely correct to have the calculated into the PTM since it is not actually the mass of the modification.

Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
rmmiller22: Yes. same for C-term.

Proteomics/Modifications/Modification.cs
rmmiller22: To have the modifications added to the protease.tsv file it does not work to have the modification provided in the multi-line format that is utilized in the ReadModsFromString() method that reads in the mods from PTM lists.

Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
```diff
@@ -37,18 +38,30 @@ static ProteaseDictionary()
...
                     List<DigestionMotif> motifList = DigestionMotif.ParseDigestionMotifsFromString(fields[1]);
-
                     string name = fields[0];
                     var cleavageSpecificity = ((CleavageSpecificity)Enum.Parse(typeof(CleavageSpecificity), fields[4], true));
                     string psiMsAccessionNumber = fields[5];
                     string psiMsName = fields[6];
-                    var protease = new Protease(name, cleavageSpecificity, psiMsAccessionNumber, psiMsName, motifList);
-                    dict.Add(protease.Name, protease);
+                    string proteaseModDetails = fields[8];                    
+                    if (proteaseModDetails != "" && proteaseMods != null)
```
trishorts, line 46: do we need a check here to see if the dictionary already contains the mod?

trishorts (commented):
This is looking good to me. Please see if you need to check if key exists in dictionary at the position indicated or not.

The only remaining thing is to improve unit test coverage of this PR. mzlib coverage is falling. Could you try to get 90% of the code in this pr covered? I think that will help put us in the right direction.

Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
rmmiller22: We never have before, but would be a good check in case someone adds two proteases with the same name to the file. I will make that adjustment.

trishorts (approved):

Proteomics/Protein/Protein.cs
```diff
@@ -225,11 +226,15 @@ public string GetEnsemblFastaHeader()
...
         /// </summary>
-        public IEnumerable<PeptideWithSetModifications> Digest(DigestionParams digestionParams, IEnumerable<Modification> allKnownFixedModifications,
+        public IEnumerable<PeptideWithSetModifications> Digest(DigestionParams digestionParams, List<Modification> allKnownFixedModifications,
             List<Modification> variableModifications, List<SilacLabel> silacLabels = null, (SilacLabel startLabel, SilacLabel endLabel)? turnoverLabels = null)
-        {
+        {            
             //can't be null
-            allKnownFixedModifications = allKnownFixedModifications ?? new List<Modification>();
+            allKnownFixedModifications = allKnownFixedModifications ?? new List<Modification>();            
+            if (digestionParams.Protease.CleavageMod!= null)
```
rmillikin, line 234: this will add `digestionParams.Protease.CleavageMod` to `allKnownFixedModifications` every time `Digest` is called, meaning once for each protein being digested. change line 234 to include a `&& !allKnownFixedModifications.Contains(digestionParams.Protease.CleavageMod)` and then I think the PR is good to go

rmillikin (changes requested):

Proteomics/Protein/Protein.cs
rmmiller22: change is made and comments have been added at some key places to hopefully help with understanding

rmillikin (approved):

PR #563, opened 2021-02-16
Fragmentation for internal fragments

zrolfs:
I created a new method that generates internal fragment ions. I felt it was important to make a separate method, rather than add functionality to the existing .Fragment(), because the existing method is called millions of times in a MM search and internal fragments will rarely be used. If users want to create terminal and internal fragment ions, they can simply call both methods and concat the two outputs.
I added two new parameters to Product to keep track of the additional terminus type and location. This seemed cleaner than creating a new class and shouldn't substantially impact the memory requirements of terminal fragments (storing null and zero).


zrolfs:
I created a new method that generates internal fragment ions. I felt it was important to make a separate method, rather than add functionality to the existing .Fragment(), because the existing method is called millions of times in a MM search and internal fragments will rarely be used. If users want to create terminal and internal fragment ions, they can simply call both methods and concat the two outputs.
I added two new parameters to Product to keep track of the additional terminus type and location. This seemed cleaner than creating a new class and shouldn't substantially impact the memory requirements of terminal fragments (storing null and zero).

rmillikin (dismissed):
neutral loss and "special" ions like star/dot/etc might not work but it's OK w/ me if you plan to implement that stuff later

rmillikin (approved):

PR #564, opened 2021-02-18
Read compressed and uncompressed uniprot proteomes

trishorts:
The UniProt proteome files can be quite large, particularly the xml versions. It is convenient to store them in compressed format. Two common formats for compression are .zip and .gz. This PR enables reading of uniprot proteomes compressed in either of these two formats.


acesnik (approved):
This looks good to me. Looks like there's a WIP tag; what's there left to do?

UsefulProteomicsDatabases/ProteinDbRetriever.cs
```diff
@@ -128,17 +128,38 @@ public static string DownloadAvailableUniProtProteomes(string filepath)
...
-                    return dictionaryOfAvailableProteomes;
-                }
-                return null;
+                    case ".gz":
+                        idNameList = ReadAllGZippedLines(completePathToAvailableUniProtProteomes).ToList();
+                        foreach (string item in idNameList)
+                        {
+                            var lineValuesArray = item.Split("\t");
+                            dictionaryOfAvailableProteomes.Add(lineValuesArray[0], lineValuesArray[1]);
+                        }
```
avcarr2, line 142: I noticed this set of lines is copy-pasted many times. I think it may be better if a method is instead added to the ProteinDbRetriever class.

avcarr2 (approved):
Overall looks good. Could use a new method in ProteinDbRetriever to handle some repeated code, but doesn't necessarily need it.

acesnik (approved):

PR #565, opened 2021-03-01
Increase buffer so write mzML doesn't crash on files with line length…

trishorts:
… exceed 1000000

### DIFF
--- a/MzML/MzmlMethods.cs
+++ b/MzML/MzmlMethods.cs
@@ -1155,7 +1155,7 @@ namespace IO.MzML
                 using (StreamReader reader = new StreamReader(stream, bufferSize: 4096))
                 {
                     string line;
-                    byte[] buffer = new byte[1000000];
+                    byte[] buffer = new byte[2000000];
                     bool foundChecksumField = false;
 
                     while (reader.Peek() > 0 && !foundChecksumField)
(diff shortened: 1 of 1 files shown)


MzML/MzmlMethods.cs
```diff
@@ -1155,7 +1155,7 @@ public static SHA1 GetSHA1Hash(string filePath)
                 using (StreamReader reader = new StreamReader(stream, bufferSize: 4096))
                 {
                     string line;
-                    byte[] buffer = new byte[1000000];
+                    byte[] buffer = new byte[2000000];
```
zrolfs, line 1158: These can be up to 2 billion, if we want to further guard against future crashes. I'm sure there's a trade-off on RAM, though.

zrolfs (approved):

MzML/MzmlMethods.cs
trishorts: i guess these aren't too big. size increased from 1MB to 2 MB. Maybe we can just go to 10 Mb. Seems like the array is just re-used so that's not a big deal i guess.

zrolfs (approved):

rmmiller22 (approved):

trishorts:
small update to handle writing mzML files with high line length

PR #567, opened 2021-03-16
correctly interpret EThcD dissociation type

rmillikin:
mzML and Thermo spectra files do not have an EThcD dissociation type, but mzLib does. this PR adds code to properly interpret .mzML and .raw scan headers if EThcD was performed.


MzML/Mzml.cs
```diff
@@ -477,11 +477,32 @@ private static MsDataScan GetMsDataOneBasedScanFromConnection(Generated.mzMLType
...
+                    }
+                }
+
+                if (scanDissociationTypes.Contains(DissociationType.ETD) && scanDissociationTypes.Contains(DissociationType.HCD))
+                {
+                    dissociationType = DissociationType.EThcD;
+                }
+                else if (scanDissociationTypes.Any())
+                {
+                    dissociationType = scanDissociationTypes.First();
```
trishorts, line 501: I think this is fine but it seems a little klugey.  I don't want to recommend a change but i feel like we could always allow a list of dissociation types here. For now ETHCD is fine b/c it is well known. But, int he future, mabye we call it ETD/HCD and then we can just do any combination ETC/ECD, ETC/CID HCD/CID.  etc. etc.

Test/TestMzML.cs
```diff
@@ -1563,6 +1563,15 @@ public static void TestDynamicMzmlWithPeakFiltering()
...
 
+        [Test]
+        public static void TestEthcdReading()
+        {
+            string filePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "DataFiles", "sliced_ethcd.mzML");
+            Mzml mzml = Mzml.LoadAllStaticData(filePath, null, 1);
+            var ethcdScan = mzml.GetOneBasedScan(6);
+            Assert.That(ethcdScan.DissociationType == DissociationType.EThcD);
+        }
+
```
trishorts, line 1574: I do feel like we need unit tests for "dissociationType = scanDissociationTypes.First();" and "dissociationType = DissociationType.Unknown;"

trishorts (commented):
two unit tests and I think we're good.

Test/TestMzML.cs
rmillikin: I don't know how to write one for DissociationType.Unknown. This would mean that the scan has a dissociation type that we don't know how to interpret. I don't know if that exists

trishorts: can you edit your mzml slide to have dissociation type xyz which would just be a place holder for any that get added in the future?

I'm not hung up on this if its a pain. I just like high unit test coverage.

rmillikin: ok, done

trishorts (approved):

PR #570, opened 2021-04-01
Use C#9 syntactics and only x64 platform

acesnik:
This PR is stacked on the .NET 5.0 one, so that's first.

This PR change the platform for all the projects to be x64, and only x64, rather than forcing it to build with x64 despite AnyCPU being specified (which usually allows x86 builds, which isn't the case here and could lead to confusion). This also eliminates the abundant architecture mismatch warnings.

Also, this PR updates some of the syntax to C# 9, including:
* Using basic index operators (https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0056).
* Note, I didn't update to using range operators to avoid making off-by-one errors (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#range-operator-)
* Simplified some default expressions (https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0034)
* Updated pattern matching, including shortening equals operations (https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0019).
* Using simplifed `new()` expressions where possible. This shortens a lot of initializations and gets them onto one screen :)


acesnik:
Closing this for now, and will split it so it's easier to review.

PR #571, opened 2021-04-02
Only built with x64, so explicitly only use the x64 platform

acesnik:
* Use only the x64 platform. This eliminates the MSIL/AMD64 warnings
* Eliminates changing the target platform to x64 within AnyCPU builds by eliminating the AnyCPU builds, which aren't actually used anywhere
* Allows Visual Studio to give code quality recommendations again


acesnik:
One potential issue is that if/when a new project is added from a template (which is usually the case), the AnyCPU builds will come back and should be deleted in the Visual Studio Configuration Manager (found in platforms dropdown or otherwise search for Configuration Manager in the help bar).
image
image
image

acesnik:
For an illustration of what changes here, these are the build logs before (left) and after these changes (right). The abundant MSB3270 warning is eliminated.

image

trishorts:
thanks for working on this. before we approve and merge this change, I think we need more than the usual "single review". I this this will impact all programs using mzlib (metamorpheus, proteoform suite, flashlfq, proteaseguru, etc.)

acesnik:
Sounds good! Should we change the reviewing policy for the repository? It's not matched with the 3 required for MM. I was wondering if I jumped the gun with hitting that merge button after Rob's review of the net5.0 changes earlier this week... 😅

trishorts:
that has been on my mind and we haven't officially changed the policy. we don't have a "minder" like we do with MM. Maybe a quick convo between interested parties to come to an agreement???

acesnik:
For this PR, it's worth mentioning that this isn't a breaking change downstream. Building MM will throw a similar MSIL/AMD64 warning because it's building binaries tagged as AnyCPU (usually targeting x86) but forcing them to build with x64. See https://github.com/smith-chem-wisc/MetaMorpheus/blob/master/TaskLayer/TaskLayer.csproj#L10 for an example. It will still build successfully, as has been the case with mzLib up to now.

zrolfs (approved):

PR #576, opened 2021-05-06
[WIP] baseline imputation implemented in flashlfq

rmillikin:
- added option to impute missing peptide values from baseline estimation
- made the median polish intensity scaling more stable (issue reported by @blfrey )
- protein quant output reports number of peptides used to quantify that protein

### DIFF
--- a/FlashLFQ/ChromatographicPeak.cs
+++ b/FlashLFQ/ChromatographicPeak.cs
@@ -6,17 +6,19 @@ using System.Text;
 
 namespace FlashLFQ
 {
+    public enum PeakType { MSMS, MBR, Imputed }
+
     public class ChromatographicPeak
     {
         public double Intensity;
         public readonly SpectraFileInfo SpectraFileInfo;
         public List<IsotopicEnvelope> IsotopicEnvelopes;
         public double SplitRT;
-        public readonly bool IsMbrPeak;
+        public readonly PeakType PeakType;
         public double MbrScore;
         public double PosteriorErrorProbability { get { return NumIdentificationsByFullSeq > 1 ? 1 : Identifications.Min(p => p.PosteriorErrorProbability); } }
 
-        public ChromatographicPeak(Identification id, bool isMbrPeak, SpectraFileInfo fileInfo)
+        public ChromatographicPeak(Identification id, PeakType peakType, SpectraFileInfo fileInfo)
         {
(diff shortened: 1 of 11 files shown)


rmillikin:
image

MzLibUtil/DoubleRange.cs
```diff
@@ -127,5 +127,22 @@ public bool Contains(double item)
         {
             return CompareTo(item).Equals(0);
         }
+
+        public override int GetHashCode()
+        {
+            return Minimum.GetHashCode();
```
acesnik, line 133: Should this be `return Minimum.GetHashCode() ^ Maximum.GetHashCode();` instead?

MzLibUtil/DoubleRange.cs
```diff
@@ -127,5 +127,22 @@ public bool Contains(double item)
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = (DoubleRange)obj;
+
+            if(other.Minimum == this.Minimum && other.Maximum == this.Maximum)
```
acesnik, line 140: Possibly this:
`return other is DoubleRange && (other as DoubleRange).Minimum.Equals(this.Minimum) && (other as DoubleRange).Maximum.Equals(this.Maximum);`

MzLibUtil/DoubleRange.cs
acesnik: i.e. seems like it's missing the maximum

FlashLFQ/ChromatographicPeak.cs
```diff
@@ -6,17 +6,19 @@
 
 namespace FlashLFQ
 {
+    public enum PeakType { MSMS, MBR, Imputed }
```
acesnik, line 9: Break this out into another `.cs` file? This might make it easier to find, kind of like `Proteomics.Fragmentation.FragmentationTerminus` or something.

acesnik (changes requested):
Nice testing to this. Looks good overall, but had a few questions regarding DoubleRange.cs.

FlashLFQ/ChromatographicPeak.cs
acesnik: This is an improvement over the previous boolean value, too 👍🏼

PR #577, opened 2021-05-20
Restructure base directory & add GitHub Actions builds/releases

acesnik:


acesnik:
This PR simplifies the landing page for mzLib. It will need some adjustment to AppVeyor before merging if we decide this is something we'd like to do.

Before:
image

After:
image

trishorts:
do you know how to adjust appveyor?
does this update change links to objects in programs like MM that use mzlib?

acesnik:
This would aim to move mzLib over to using github actions instead of appveyor, so that the whole pipeline could be housed in github rather than two places. (Easier maintenance)

I've drafted the github action files in this PR, as well.

The links won't change, since the nuspec file is packaged the same way, and the nupkg would be released into GitHub releases.

acesnik:
The builds will also be reported and kept here https://github.com/smith-chem-wisc/mzLib/actions
The secrets for pushing releases are also kept secure on GitHub, which makes them easier to find

acesnik:
I'll note I haven't figured out codecov in GitHub Actions just yet, but I'd be happy to put a little more work in to figure that out if this continuous integration change is something the group is interested in. :)

acesnik:
Codecov is now working under Github Actions in this PR. Should obviate the need for AppVeyor.

image

acesnik:
I think the appveyor build failure is signaling to Codecov not to post this, but here's what it'll look like with the codecov report:
image

trishorts:
I think this PR makes good sense. The one concern I have, which may be totally unfounded, is how programs that use mzlib (like MM) will find existing classes. Do their paths change?

acesnik:
Their paths do not change for other programs. You can see this in the empty diff for mzlib.nuspec: https://github.com/smith-chem-wisc/mzLib/pull/577/files#diff-4678f0a7d73f29ffd80f0d7ad24ddd30beef7f96f1498431fe78723d6e4b479a. That nuspec is used to package the nupkg package that other programs use.

trishorts (dismissed):
I'm down to try this. I never liked the verbose directory on the landing page.

acesnik:
Updated the release branch and the settings for the repository. Should be good to go!

Also, the codecov feedback is working now with GitHub Actions!

acesnik:
Should be ready to go @rmmiller22 & @trishorts!

trishorts (approved):

PR #578, opened 2021-05-25
Update pepXML to version 122 from 120

acesnik:


acesnik (commented):
@trishorts I fixed the merge conflicts here on the smith-chem-wisc/mzLib/pepxml122 branch that you can also push to. 

Since you are updating `pepXML/pepXML_v120.cs` to be version 122, unless you want to maintain different versions of pepXML here (like before?), I'd recommend changing the name of the file to just `pepXML/pepXML.cs` and note the versioning in the header of the file.

trishorts:
how safe do you think this is? on the surface it seems totally fine. but, we only use pepxml for xlink output. and we only do that so that it can be ready in using one particular kind of software.

We could try to merge this PR, then run xlink and try upload to that visualizer. but that's not in my wheelhouse. i wonder if lei would help

acesnik:
How about just adding this as pepXML_v122.cs instead of overwriting the v120 one?

acesnik:
I wonder if that's enough separation, though. It looks like the MM code just uses `pepXML...` not `pepXML_120...`.

acesnik:
@lonelu Any thoughts here?

PR #583, opened 2021-06-16
handle proteoform classification edge case

zrolfs:
Edge case with ambiguity between the following two proteoforms:
…K[UniProt:N6-glutaryllysine on K]…K[UniProt:N6-succinyllysine on K]…
…K[UniProt:N6-succinyllysine on K]…K[UniProt:N6-glutaryllysine on K]…

This should be a level 2A, but is currently a level 1. This PR catches this edge case.


trishorts (approved):
Good catch. Excellent PR.

PR #584, opened 2021-07-01
32bit dynamic mzml fix

rmillikin:
- fixes reading 32-bit .mzML data through the dynamic connection
- fixes/tests mixed-bit scans (e.g., where the m/z values are encoded in 64 bit and the intensities are encoded in 32 bit)


trishorts:
trying your pr on my file and got this
image

trishorts:
they are in fact not equal (1.117 and 0.0816)

trishorts:
btw, this was the converted to x64, not the original

trishorts:
interestingly it was the static scan that was wrong...

rmillikin:
actually it looks like the static one is correct. the RT is in seconds in the .mzML, which the static correctly converts to minutes, but the dynamic doesn't

rmillikin:
ok, i fixed that error too. there was another issue which at this point i'm not sure is easy to fix. if the precursor scan number is not listed in the .mzML, the static .mzML assumes it's the previous MS1 (which may or may not be correct). the dynamic reader leaves it as null

trishorts (approved):

PR #588, opened 2021-11-10
eliminate zero intensity values when reading mzML either static or dy…

trishorts:
mass spectra with zero intensity values can cause issues downstream, particularly with calibration in MetaMorpheus. This PR eliminates all zero intensity values (with corresponding mz) when the file is read statically or dynamically.


acesnik (dismissed):
Looks good!

acesnik:
You're getting this error:
image

This comes from `net5.0-windows` missing the platform version. Not sure exactly why but it seems like `net5.0-windows7.0` is what you need to use.

Giving it a quick test here: https://github.com/smith-chem-wisc/mzLib/pull/589

acesnik:
MetaMorpheus may need to update these, as well, but maybe it can cast the TFMs implicitly.

See here for example: https://github.com/smith-chem-wisc/MetaMorpheus/blob/14016405afb808d357b4956d2809c4a8767c343a/GuiFunctions/GuiFunctions.csproj#L4

acesnik (approved):

PR #594, opened 2022-01-11
LoadPsiMod changed to close FileStream after reading in data

Alexander-Sol:
TestLoading test in TestDatabaseLoaders.cs wasn't passing on my build. The FileStream to the PsiMod file was left open when LoadPsiMod was called. This prevented the deletion of the PsiMod file.

I changed LoadPsiMod to close the FileStream before returning the data. The test now passes.

### DIFF
--- a/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
+++ b/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
@@ -8,31 +8,31 @@ namespace MassSpectrometry.MzSpectra
 {
     public class SpectralSimilarity
     {
-        public SpectralSimilarity(MzSpectrum primary, MzSpectrum secondary, SpectrumNormalizationScheme scheme, double toleranceInPpm, bool allPeaks)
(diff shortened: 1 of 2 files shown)


kyp4:
I am able to pass the FilesLoading() test both with and without this change (not sure why), but did need to update the number of PSMs coming from Unimod: Assert.AreEqual(2723, unimodMods.Count);

kyp4 (changes requested):
I am able to pass the FilesLoading() test both with and without this change (not sure why), but did need to update the number of PSMs coming from Unimod: Assert.AreEqual(2723, unimodMods.Count);

trishorts:
number changes frequently, so this number needs updating when that happens. we've talked about adapting the unit test before but haven't yet. maybe there is another way to compute the number or do the check.

Alexander-Sol:
Currently we're just checking that the # of PSMs is greater than 2700:  Assert.Greater(unimodsMods.Count, 2700)

trishorts:
# of mods not number of psms but i get your point.

PR #596, opened 2022-01-19
Flash lfq for percolator

trishorts:
I'm updating FlashLFQ for percolator input. There are three issues to deal with:
1. there is no filename, only a file integer. We're hoping the percolator people will update their output as a solution. otherwise the user will have to manually change this.
2. there is no retention time. We're modifying FlashLFQ to read retention time from the raw data file using the scan number supplied in the percolator output.
3. many modified sequences are unreadable because they contain characters with no meaning to our sequence parser. These sequences will be used as unique peptide/proteoform identifiers BUT NOT BE USED TO COMPUTE PEPTIDE/PROTEOFORM MASS.

### DIFF
--- a/FlashLFQ/Identification.cs
+++ b/FlashLFQ/Identification.cs
@@ -33,6 +33,25 @@ namespace FlashLFQ
             UseForProteinQuant = useForProteinQuant;
             PosteriorErrorProbability = posteriorErrorProbability;
         }
+        /// <summary>
+        /// This Identification object was created for instances when the modified sequence isn't parsable into a formula
+        /// for calcuation of the peptide/proteoform mass. 
+        /// </summary>
+        public Identification(SpectraFileInfo fileInfo, string ModifiedSequence, double monoisotopicMass, 
+            double ms2RetentionTimeInMinutes, int chargeState, List<ProteinGroup> proteinGroups, 
+            bool useForProteinQuant = true, double posteriorErrorProbability = 0)
+        {
+            this.FileInfo = fileInfo;
+            this.BaseSequence = null;
+            this.ModifiedSequence = ModifiedSequence;
+            this.MonoisotopicMass = monoisotopicMass;
+            this.Ms2RetentionTimeInMinutes = ms2RetentionTimeInMinutes;
+            this.PrecursorChargeState = chargeState;
+            this.ProteinGroups = new HashSet<ProteinGroup>(proteinGroups);
+            this.OptionalChemicalFormula = null;
+            UseForProteinQuant = useForProteinQuant;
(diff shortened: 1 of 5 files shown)


trishorts:
ChromatographicPeak has me worried because for percolator the base sequence is null:
image

trishorts:
This PR only handles the necessary changes to the "Identification" object and the use of averagine to return a chemical formula appropriate to the reported monoisotopic mass

percolator:
Hi Michael,

There are two upcomming updates to percolator that that might help you. We plan to introduce optional columns for
1. filename of the psms.
2. retention time.
Cheers --Lukas

trishorts:
> Hi Michael,
> 
> There are two upcomming updates to percolator that that might help you. We plan to introduce optional columns for
> 
>     1. filename of the psms.
> 
>     2. retention time.
>        Cheers --Lukas

Thanks Lukas,
Filename was the really big one. I coded a solution to pull RT from index, but I guess I can scrap that (unless we want to keep it around for compatibility to old data). I'll think about it. 

A bigger issue was dealing w/ the protein sequence column, which has mass shifts. I also coded a solution for that that uses the percolator monoisotopic mass to create the proper isotope envelopes using averagine instead of the actual chemical formula of the peptide (which can't be determined from numerical mass shifts). I hope that code can get reviewed and approved soon and incorporated into the release version of flash lfq.

Changes take some time on our end b/c of a stringent review process. Thanks for your help.
-michael

wsnoble:
I would vote for keeping the RT stuff in FlashLFQ, if possible.  The RT feature in Percolator will only work if the upstream search engine reports that in its output, and I know that some of them do not do so.

Regarding the representation of mass shifts, I don't think we can provide chemical formulas in general, because that's not possible in the case of open search results. Averagine seems like the right solution for this.  Thanks!

FlashLFQ/Identification.cs
```diff
@@ -33,6 +33,25 @@ public class Identification
             UseForProteinQuant = useForProteinQuant;
             PosteriorErrorProbability = posteriorErrorProbability;
         }
+        /// <summary>
+        /// This Identification object was created for instances when the modified sequence isn't parsable into a formula
+        /// for calcuation of the peptide/proteoform mass. 
+        /// </summary>
+        public Identification(SpectraFileInfo fileInfo, string ModifiedSequence, double monoisotopicMass,
```
rmillikin, line 40: don't need this second `Identification` constructor; you can use the original constructor and simply pass in `null` for the `BaseSequence` parameter

rmillikin (changes requested):

kyp4 (approved):
I trust Rob implicitly

PR #598, opened 2022-01-24
add library search mz filter(>300)

Yuling1996:


MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -161,8 +176,12 @@ private double[] NormalizeSpectrumSum(double[] spectrum)
         #region similarityMethods
 
         //The cosine similarity returns values between 1 and -1 with 1 being closes and -1 being opposite and 0 being orthoganal
-        public double CosineSimilarity()
+        public double? CosineSimilarity()
```
trishorts, line 179: do we really want null?

trishorts:
we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.

MassSpectrometry/MzSpectra/SpectralSimilarity.cs
Yuling1996: we don't want null but it happens when all mz are less than 300 in theoretical(library) peaks

Yuling1996: I mean it happens that there is no intensity pair when all mz are less than 300 in theoretical(library) peaks. I need to assign a value to it

Yuling1996:
> we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.

For null test, does line 308-318 in Test/TestSpectralSimilarity.cs make sense?

Yuling1996:
> we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.

There is new test TestAllSpectrumSimilaritiesWithDefaultedMzFilter() with cut off used. Did I miss something?

trishorts:
> > we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.
> 
> For null test, does line 308-318 in Test/TestSpectralSimilarity.cs make sense?

im not sure. definitely line 300 intensity pairs is zero, not null. After that seems maybe okay. But, the secondary has mz values all above 300, so i guess it depends if we use secondary array to make intensity pairs or not. The

trishorts:
> > we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.
> 
> There is new test TestAllSpectrumSimilaritiesWithDefaultedMzFilter() with cut off used. Did I miss something?

is default 0 or 300? if 300, then that is correct. i didn't see that.

Yuling1996:
> > > we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.
> > 
> > 
> > There is new test TestAllSpectrumSimilaritiesWithDefaultedMzFilter() with cut off used. Did I miss something?
> 
> is default 0 or 300? if 300, then that is correct. i didn't see that.

default is 300

Yuling1996:
> > > we need a series of unit tests that return values when a cut-off is used. it should be different than when no cutoff is used. the number of pairs should change when there are mz values below cuttoff. if you are going to return null, then we have to check that null value is returned for each instance in a unit test.
> > 
> > 
> > For null test, does line 308-318 in Test/TestSpectralSimilarity.cs make sense?
> 
> im not sure. definitely line 300 intensity pairs is zero, not null. After that seems maybe okay. But, the secondary has mz values all above 300, so i guess it depends if we use secondary array to make intensity pairs or not. The

what do you think that I should add?

MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -105,7 +117,11 @@ private List<(double, double)> IntensityPairs(bool allPeaks)
                     }
                     index++;
                 }
-                if(index > 0)
+                if (experimental.Count == 0)
```
trishorts, line 120: please explain this change

trishorts (approved):

PR #601, opened 2022-02-04
[WIP]Top down biomarker

trishorts:
working toward a top-down biomarker search co-opting existing code.

there is a new protease (top-down biomarker) that uses semi-specific cleavage from both ends.

Digestion params is updated on the fly to get drop up to 8 AAs from both ends. Mods are added accordingly. Uses old school digestion rather than speedy cuz i can't understand speedy.

### DIFF
--- a/FlashLFQ/FlashLfqEngine.cs
+++ b/FlashLFQ/FlashLfqEngine.cs
@@ -16,6 +16,7 @@ namespace FlashLFQ
     {
         // settings
         public readonly bool Silent;
+
         public readonly int MaxThreads;
         public readonly double PeakfindingPpmTolerance;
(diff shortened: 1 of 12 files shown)


trishorts:
Not ready for review. Please hold for more unit tests

trishorts:
This is getting closer. I chose to add biomarkers as proteolysis products, which require only adding to the existing protein.proteolysisproduct list using only begin and end positions indexed to the full protein. This method allows for retention or deletion of n-terminal methionine. Importantly, I think this allows all new biomarker proteolysis products to have PTMs from the XML, which should make it a more complete solution for top-down analysis

rmmiller22 (approved):

PR #603, opened 2022-02-09
generation of truncated proteoforms for biomarker searches

trishorts:
Truncated and post-translationally processed proteoforms are more challenging to identify than full length proteoforms. This is due to incomplete knowledge a priori of post translational processing, by which we mean cleavage of the translated proteoform along the amino acid backbone into one or more functional products. This PR takes input of the primary amino acid sequence (and known truncation products) and returns a series of n- and c-terminal degradation sequences (default length 7). These sequences, in conjunction with annotated PTMs can be used in a top-down search with metamorpheus to ID unknown proteoforms.


Yuling1996 (approved):

Proteomics/Protein/Protein.cs
```diff
@@ -585,60 +585,236 @@ public void RestoreUnfilteredModifications()
             return validModDictionary;
         }
 
-        public void AddBiomarkersToProteolysisProducts(int fullProteinOneBasedBegin, int fullProteinOneBasedEnd, bool addNterminalDigestionBiomarkers, bool addCterminalDigestionBiomarkers, bool retainNterminalMethionine, int minProductBaseSequenceLength, int lengthOfProteolysis, string proteolyisisProductName)
+        public void AddBiomarkersToProteolysisProducts(int fullProteinOneBasedBegin, int fullProteinOneBasedEnd, bool addNterminalDigestionBiomarkers, bool addCterminalDigestionBiomarkers, InitiatorMethionineBehavior initiatorMethionineBehavior, int minProductBaseSequenceLength, int lengthOfProteolysis, string proteolyisisProductName)
```
avcarr2, line 588: It took me a minute to figure out how initiator methionine are being dealt with by the logic. Could you add a comment specifically pointing out that the fullProteinOneBasedBegin + 1 is the part that's skipping the methionine?

Proteomics/Protein/Protein.cs
```diff
@@ -585,60 +585,236 @@ public void RestoreUnfilteredModifications()
...
         {
             bool sequenceContainsNterminus = (fullProteinOneBasedBegin == 1);
```
avcarr2, line 590: Could you add comment for posterity on choice to use one-based protein sequence indexing versus zero-based indexing?

avcarr2 (approved):
Only changes requested are to add comments. Successfully built on my machine and all unit tests passed and appear to adequately test cases.

Proteomics/Protein/Protein.cs
trishorts: ha. it is not skipping the methionine, the n-terminus all by itself is index position 0, regardless of if ther is methionine at pos 1 or not

avcarr2: Yeah that makes complete sense to me now.

Issue #610, opened 2022-02-22
MGF peak list parser expects field separator to be SPACE


andzajan:
Hi, we had issues reading some MGF files with MetaMorpheus.

I did some debugging and problem is that [ParsePeakLine](https://github.com/smith-chem-wisc/mzLib/blob/master/Mgf/Mgf.cs#L167) function expects fields to be separated by a single white space. In our case values are TAB separated.

Matrix sciences don't specify use of single space, TAB or multiple white spaces in the peak list: http://www.matrixscience.com/help/data_file_help.html

I have tested locally and changing line `169` to something like example below will work with all types of MGF peak lists:

`var sArray = line.Split((char[])null, StringSplitOptions.RemoveEmptyEntries);`

I didn't want to open pull request because of such a small change, but I hope you will accept my suggestion. All Unit tests in your pipeline are parsing with this change.

trishorts:
can you provide an mgf file to use in our unit test?

trishorts:
I submitted a pull request that should solve the problem. I need two reviews to get approval for merge. then I will make a PR to MetaMorpheus w/ the update. This may take a couple days. If it's urgent, I may be able to get you a pre-release. 
https://github.com/smith-chem-wisc/mzLib/pull/611

andzajan:
Thank you for such a quick turnaround. On our side it's not extremely urgent, as we are using our local build for now.

PR #611, opened 2022-02-22
Allow both tab a space separators when reading mgf

trishorts:
Some mgf file writers output tab separated lines (e.g. mascot). This PR enables reading these lines along with the formerly allowed space separator. The problem is described in issue #610


avcarr2 (approved):
Good to go

PR #620, opened 2022-03-19
Use multiline commands in build-test actions to get proper exit codes for tests

acesnik:


acesnik:
Tests failed here in the `test` step, https://github.com/smith-chem-wisc/mzLib/runs/5575438259?check_suite_focus=true, but it's still reporting a passing exit code because I had semicolon-concatenated the commands, rather than executing them in order. This PR makes that workflow all use multiline commands, rather than semicolon concatenation.

trishorts (approved):

trishorts (approved):

rmmiller22 (approved):

PR #624, opened 2022-03-29
fix median polish bug when proteins cannot be quantified

rmillikin:
this PR fixes a bug in FlashLFQ's default protein quantification algorithm, "median polish".

Median polish sometimes introduces an artifact where protein quantities are mistakenly marked as identical in multiple files, if there is not enough information to quantify a protein in a particular sample. As a result, I check to see if protein quantities are exactly the same in multiple files and mark their quantities as NaN. This change introduces a bug where if multiple samples do in fact have exactly the same intensity, they are marked as NaN by mistake.

In real life, this is extremely rare, but this change did break several unit tests, because in those, several proteins did in fact have exactly the same intensity in multiple "samples". I commented those out and added an explainer comment on the first one.


trishorts (approved):
we'll have to test this in MM.

rmmiller22 (approved):

PR #625, opened 2022-03-31
[WIP] Deconvolution using UniDec

avcarr2:
Low-resolution, top-down mass spectra are difficult to deconvolute due to their lack of isotopically resolved peaks. [UniDec](http://unidec.chem.ox.ac.uk/) is the current standard for the deconvolution of low-resolution protein mass spectra. 

UniDec's primary algorithm is written in C, while mzLib is written in C#, requiring me to pass data back and forth between managed C# code and unmanaged C code. I created a dynamic library in C from which mzLib can call UniDec functions. I had to heavily rely on unsafe C# code (pointers and such) in order to efficiently pass data to the unmanaged C code functions. 

There are significant issues with the initial commit which I will be working on in the coming weeks. The primary issues are 1) further optimizations of the UniDec algorithm and 2) memory management in both C and C#. There are several points in this code that risk int overflow issues, which the unmanaged C code will not tell you are occurring. I would also like to get rid of most of the unsafe C# code, but I'm not sure yet how viable that is. 

Unit test coverage for the functions called from the C unmanaged code are forthcoming. Unit tests will also cover passing data structures back and forth between the managed and unmanaged code.


acesnik:
Cool PR! 

Could you please relocate the new files into the mzLib subfolder, e.g. `mzLib.sln` is now at `mzLib/mzLib.sln`, when you have the chance?

mzLib.sln
```diff
@@ -33,103 +33,267 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test", "Test\Test.csproj",
...
+		{707DB7D7-3C7F-4E2C-91D9-A933D4FCAE2F}.TestAndRelease|x86.ActiveCfg = Release|x64
+		{707DB7D7-3C7F-4E2C-91D9-A933D4FCAE2F}.TestAndRelease|x86.Build.0 = Release|x64
+		{D2B8F696-1172-405B-B788-3251B27CC71A}.Debug|Any CPU.ActiveCfg = Debug|x64
 		{D2B8F696-1172-405B-B788-3251B27CC71A}.Debug|x64.ActiveCfg = Debug|x64
 		{D2B8F696-1172-405B-B788-3251B27CC71A}.Debug|x64.Build.0 = Debug|x64
+		{D2B8F696-1172-405B-B788-3251B27CC71A}.Debug|x86.ActiveCfg = Debug|x64
+		{D2B8F696-1172-405B-B788-3251B27CC71A}.Release|Any CPU.ActiveCfg = Release|x64
 		{D2B8F696-1172-405B-B788-3251B27CC71A}.Release|x64.ActiveCfg = Release|x64
 		{D2B8F696-1172-405B-B788-3251B27CC71A}.Release|x64.Build.0 = Release|x64
+		{D2B8F696-1172-405B-B788-3251B27CC71A}.Release|x86.ActiveCfg = Release|x64
```
acesnik, line 130: Do you need these x86 and AnyCPU configurations? I think x64 is the only one that's used currently to allow > 2 GB RAM usage, so consider removing these.

acesnik:
🤖 Closing this. Not because the idea was wrong — deconvolution from the UniDec family is in mzLib today — but because the route it takes has been replaced by a better one, and the branch it targets is not master.

**What supersedes it.** `MassSpectrometry/Deconvolution/Algorithms/IsoDecAlgorithm.cs` integrates the Marty lab's deconvolution as a native dependency (`isodeclib`, referenced from `MassSpectrometry.csproj:29`) rather than by vendoring C sources into this repository. `Deconvolution/Scoring/` — `DeconvolutionScorer`, `DeconvolutionQValueCalculator`, `EnvelopeScoreFeatures` — covers the scoring this PR's `UniDecAPI/Scoring.cs` was for, with q-values on top, and `Deconvolution/Consensus/` adds a mass-trace layer that did not exist when this was opened. Consuming an upstream library beats carrying a copy of it, and that is now the established pattern here.

**One thing it aimed at is not superseded, and I have opened an issue for it rather than let it close with the PR.** IsoDec works from isotopically resolved envelopes; this branch's fixture is a *low-resolution* MS1 scan, and charge deconvolution where the isotopes are not resolved is UniDec's distinctive job. I searched `Deconvolution/` and found nothing addressing it. That gap is real and worth tracking on its own terms — see #1266.

**Reopening signal**, so this is not a judgement that closes the door: if someone wants low-resolution charge deconvolution in mzLib, the way in is a native dependency on UniDec the way `IsoDecAlgorithm` depends on `isodeclib` — not this branch. If that turns out to be impossible and vendoring is the only option, reopen and this becomes the starting point again.

Also deleting the `UniDecPort` branch, which this targeted rather than master.

PR #628, opened 2022-05-02
Ignore empty mzml scans during read

trishorts:
Seems like fast thermo instruments can sometimes stow away an empty scan. These scans are inserted into mzMLs, presumably by msconvert. When we try to read this in mzlib (for programs like MM and FlashLFQ), there is a crash, which is annoying. We could simply ignore these empty scans during the read and get on w/ life. Maybe we need to return an error message listing that scans that were skipped. Not sure how to do this but I think its a good option.


mzLib/MzML/Mzml.cs
```diff
@@ -353,7 +356,8 @@ private static MsDataScan GetMsDataOneBasedScanFromConnection(Generated.mzMLType
             }
 
             if (!msOrder.HasValue || !isCentroid.HasValue)
-                throw new MzLibException("!msOrder.HasValue || !isCentroid.HasValue");
+                return null;
+                //throw new MzLibException("!msOrder.HasValue || !isCentroid.HasValue");
```
acesnik, line 360: Is there a way to pass a warning message? Perhaps in an `out` variable?

mzLib/MzML/Mzml.cs
```diff
@@ -373,16 +377,24 @@ private static MsDataScan GetMsDataOneBasedScanFromConnection(Generated.mzMLType
                     intensityArray |= cv.accession.Equals(_intensityArray);
                 }
 
-                double[] data = ConvertBase64ToDoubles(binaryData.binary, compressed, is32bit);
-                if (mzArray)
+                try
```
acesnik, line 380: Fair enough! Could you comment the `try` and `catch` blocks with their intended purposes?

mzLib/Test/TestMsDataFile.cs
```diff
@@ -215,7 +215,27 @@ private MzSpectrum CreateMS2spectrum(IEnumerable<Fragment> fragments, int v1, in
             Array.Sort(allMassesArray, allIntensitiessArray);
             return new MzSpectrum(allMassesArray, allIntensitiessArray, false);
         }
+        [Test]
+        public static void SkipEmptyScansWhenReadingMzml()
+        {
+            //this original mzML has four scans including 1 MS1 and three MS2s. The second MS2 scan does
+            //not have mz or intensity values. We skip this scan when reading.
+            string dataFilePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "DataFiles", @"badScan7192.mzml");
```
acesnik, line 223: 👍🏼

acesnik (approved):

mzLib/MzML/Mzml.cs
trishorts: i would like to do something like that. be good to report a list of scans that were ignored. i'm not sure how.

acesnik:
Approved, but it would be nice to have the warning `out` result, and some comments of the try-catch :)

mzLib/MzML/Mzml.cs
trishorts: yes. of course

mzLib/MzML/Mzml.cs
acesnik: https://github.com/smith-chem-wisc/ProteoformSuite/blob/e94716ce9b76395a3ebc1b6a0864090d3e3f3865/ProteoformSuiteInternal/Sweet.cs#L106

acesnik: Something like this? Can use `out` string variables to pass warnings back from method

acesnik (approved):

trishorts:
i plan to do a future update where I supply the list of skipped scans. but I can't think of a way to do that where I don't have to edit the method call w/ an extra variable. I don't want to break all the programs using it. so, I'd like to push this through and follow it up w/ a proper warning message if that is cool w/ everyone.

rmmiller22 (approved):

nbollis (approved):

PR #630, opened 2022-05-17
Kolmogorov–Zurbenko filtering for MzSpectrum objects

avcarr2:
My new deconvolution algorithm idea relies on high quality smoothing of the data. I implemented the Kolmogorov-Zurbenko filter, directly porting to C# most of the C code written originally for R found here: https://github.com/cran/kza/blob/master/src/kz.c.


acesnik (commented):
Looks good. Are all the changes in `mzLib/Test/Test.csproj`, e.g. removing mzPlot, intended for this PR?

PR #631, opened 2022-05-19
Add a Kolmogorov-Zurbenko smoothing filter for profile spectra

avcarr2:
My new deconvolution algorithm idea relies on high quality smoothing of the data. I implemented the Kolmogorov-Zurbenko filter, directly porting to C# most of the C code written originally for R found here: https://github.com/cran/kza/blob/master/src/kz.c.

This is a re-do of my previous commit that included some changes that I didn't want to include in this pull request.


acesnik (dismissed):

nbollis (dismissed):

Alexander-Sol (dismissed):

trishorts (changes requested):
tidy up those spots and we're good to go

acesnik:
🤖 **What this solves.** A profile-mode spectrum carries high-frequency noise between its peaks, and anything trying to locate peaks or isotopic envelopes in one wants a low-pass filter first — otherwise a local maximum in the noise is indistinguishable from a small real peak. mzLib has nowhere to go for that today: no moving average, no Savitzky-Golay, no smoothing primitive of any kind. `SpectralAveraging` is the nearest thing and answers a different question, reducing noise *across* scans and needing several of them. This works inside a single scan.

**Being straight about the scope:** nothing in the tree consumes it. `TofSpectraMerger` is the only code that turns raw profile-ish data into centroids, and it collapses within a ppm tolerance rather than smoothing first; `IsCentroid` is propagated everywhere and branched on nowhere. So this lands as a documented building block, not as a fix to a live failure, and it is worth saying that plainly rather than dressing it up.

**Why this filter is a good block to have.** A KZ filter is *k* successive moving averages of width 2*w*+1, converging toward a Gaussian kernel — two numbers to tune, no dependencies, about forty lines. It also has a canonical reference to check against, which is how everything below was found.

**Three behaviours changed, each the port diverging from the reference it cites** (`cran/kza`, `src/kz.c`). In all three the reference is right:

- The end index was clamped to `Length - 1` where the reference uses `Length`, so the final channel never entered any window, including its own. `[0,0,0,0,100]` smoothed to `[0,0,0,0,0]` — a spectrum's entire intensity discarded in silence. Now `[0,0,0,33.3,50]`.
- `iterations = 0` returned the freshly allocated result buffer instead of the input, so asking for no smoothing zeroed the spectrum. The reference returns an unmodified copy.
- Non-finite values were summed rather than skipped. The reference tests `R_FINITE`, so one NaN costs its own channel; here it spread to every window touching it — `[1,1,NaN,1,1]` gave `[1,NaN,NaN,NaN,1]`, now `[1,1,1,1,1]`.

A half-window of 0 also produced NaN at the last point, and a single-channel spectrum returned NaN rather than itself; both follow from the first fix.

@trishorts — your five notes, including one I could not do as written:

- The `ArgumentException` on unequal lengths is **removed rather than converted**. It cannot fire: `KZ1D` always returns `new double[x.Length]`, so no test could make it happen (swept n=1..40, window 0..10, iterations 0..4). Reachable validation replaces it — null intensities, negative half-window, negative iteration count — each an `MzLibException` naming the offending value, each with a test.
- Your NaN-coverage ask **became possible**. With non-finite values skipped, the reference's own NaN path is reachable (a window holding nothing finite) rather than depending on an empty loop, so it is now covered.
- `window` → `halfWindowSize` since it is a half-width, `v`/`w` → `values`/`halfWidth`, `Mavg1D` → `MovingAverage`, with recommended ranges on the parameters.
- "smoothering" and the malformed `<param></param><summary>` pairs fixed.

Two properties are **documented rather than changed**, being properties of the filter itself. It indexes by data point, so it assumes evenly spaced samples: meaningful for profile spectra, not for centroided ones where adjacent peaks can be any distance apart. And it buys smoothness with peak height — a unit square wave's peak falls to 0.92 at half-window 3 and 0.70 at 5 over two iterations, so where height must survive, Savitzky-Golay is the better tool. Total intensity is redistributed rather than conserved: exact while the smeared signal clears both ends, drifting once the window is clamped there (+0.24% and −0.70% in two measured cases). All of it is pinned by tests so none of it can shift unnoticed.

The doc comment no longer calls this the first step of a deconvolution algorithm. Nothing consumes the method, that algorithm did not follow, and smoothing is not deconvolution — better to describe what it does.

Test file rewritten: it asserted output length and little else, and its own comment said a better test would cross-correlate. 19 tests now, **14 of which fail against the implementation as it stood**, so they are load-bearing. The compile error that was breaking the build — NUnit 3's `Assert.AreEqual`/`IsTrue`, only in that file — is fixed; CI will confirm the job.

Not verified by me: `Test.csproj` is `net10.0-windows` via `mzPlot` and cannot run on macOS, so I built `MassSpectrometry` and compiled `Test` with `EnableWindowsTargeting` (0 errors) and ran these 19 against the real `MzSpectrum` through a throwaway net10.0 harness. CI is the authority on the full suite.

Mechanics: pushed to the branch directly as `0d41588d` (`maintainer_can_modify`). That dismissed the three 2022 approvals, mine included — which is right, since mine would otherwise be vouching for code I just wrote. Your changes-requested is the only thing left on it, and only you can clear it.

PR #633, opened 2022-06-09
SpectralEntropy as measure of SpectralSimilarity 

Alexander-Sol:
Implementing Spectral Entropy comparison, as seen here: https://www.nature.com/articles/s41592-021-01331-z

### DIFF
--- a/mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
+++ b/mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
@@ -323,10 +323,6 @@ namespace MassSpectrometry.MzSpectra
         // This method should only be used when allPeaks is set to true
         public double? SpectralEntropy()
         {
-            if (_intensityPairs.First().Item1 == -1)
-            {
-                return null;
-            }
             double theoreticalEntropy = 0;
             foreach (double intensity in TheoreticalYArray)
             {
--- a/mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
+++ b/mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
@@ -10,22 +10,22 @@ namespace MassSpectrometry.MzSpectra
     {
         public SpectralSimilarity(MzSpectrum experimentalSpectrum, MzSpectrum theoreticalSpectrum, SpectrumNormalizationScheme scheme, double toleranceInPpm, bool allPeaks, double filterOutBelowThisMz = 300)
         {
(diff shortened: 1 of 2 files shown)


trishorts (changes requested):
please create a unit test where both arrays are null so that first line is tested.

Also add a check for the value. So use some dummy spectra that return an actual entropy value.

mzLib/Test/TestSpectralSimilarity.cs
```diff
@@ -47,6 +47,8 @@ public void TestAllSpectrumSimilaritiesWithoutMzFilter()
             Assert.That(s.PearsonsCorrelation(), Is.EqualTo(-0.87).Within(0.01));
             Assert.That(s.DotProduct(), Is.EqualTo(0.17).Within(0.01));
 
+            double? test = s.SpectralEntropy();
+
```
Yuling1996, line 51: I didn't see a real test for SpectralEntropy here.  Did I miss something? How about something like "Assert.That(test, Is.EqualTo(??).Within(??))".

Alexander-Sol: Good idea Yuling, thank you. Implemented as suggested.

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -315,6 +319,37 @@ private double[] NormalizeSpectrumSum(double[] spectrum)
             return sum;
         }
 
+        // This method should only be used with the SpectrumSum normalization method
```
trishorts, line 322: which normalization method?

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -315,6 +319,37 @@ private double[] NormalizeSpectrumSum(double[] spectrum)
...
+        // This method should only be used when allPeaks is set to true
+        public double? SpectralEntropy()
+        {
+            if (_intensityPairs.First().Item1 == -1)
```
trishorts, line 326: can delete this first check. add comment that the test for some valid pairs is performed higher in the food chain

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -315,6 +319,33 @@ private double[] NormalizeSpectrumSum(double[] spectrum)
             return sum;
         }
 
+        // This method should only be used with the SpectrumSum normalization method
+        // This method should only be used when allPeaks is set to true
+        public double? SpectralEntropy()
```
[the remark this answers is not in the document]
Alexander-Sol: It's possible that other normalization methods could be employed, so I don't want to throw an exception. If I could figure out how to implement a custom warning, that would be ideal.

nbollis (approved):

PR #652, opened 2022-09-09
Period tolerant filename without extension

trishorts:
Some users foolishly have extra periods in their filenames. This can confuse the C# Path.GetFilenameWithoutExtension function.

This utility returns the filename after stripping away the path and the bit after the last period. It tolerates extra periods in the filename and in the filepath. It can also handle windows or linux file paths.

if files happen to be compressed, this method strips away the .gz or .zip before stripping away the file extension.

filname.xml.gz returns filename


mzLib/MzLibUtil/PeriodTolerantFilenameWithoutExtension.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Linq;
+
+namespace MzLibUtil
+{
+    public class PeriodTolerantFilenameWithoutExtension
+    {
+        public static string GetPeriodTolerantFilenameWithoutExtension(string filepathAndOrName)
+        {
+            string windowsSplit = filepathAndOrName.Split('\\').Last();
+            string linuxSplit = filepathAndOrName.Split('/').Last();
```
rmillikin, line 10: I would use Path.GetFileName to get the name with extension; this will simplify the Linux/Windows path issues. then use LastIndexOf to get the last period. I suppose you could check to see if the extension is .gz, then use LastIndexOf again in that case.

rmillikin (approved):

acesnik (approved):

PR #659, opened 2022-09-27
Obo parsing bug fixes and updates

avcarr2:
* Fixed bug that prevented negative formal charges from being read as negative values. 

* Fixed Obo parsing for mods.


avcarr2:
Note that the +82k lines of code is purely due to me adding a file.

mzLib/Test/TestMzML.cs
```diff
@@ -9,8 +9,11 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Xml.Serialization;
+using TopDownProteomics.IO;
+using TopDownProteomics.IO.Obo;
```
nbollis, line 16: Should these new usings be there?

mzLib/MzML/OboReader.cs
```diff
@@ -0,0 +1,16 @@
...
+using TopDownProteomics.IO.Obo;
+using UsefulProteomicsDatabases.Generated;
+
+namespace MzML
+{
+    public class OboReader
+    {
+
+    }
+}
```
nbollis, line 16: Should this class be removed?

mzLib/Test/Test.csproj
```diff
@@ -132,6 +133,9 @@
     <None Update="DatabaseTests\ProteoformClassificationUnitTest.csv">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Update="DatabaseTests\PSI-MOD.obo">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
acesnik, line 137: We've usually used "Always;" is this intentional?

acesnik (approved):

mzLib/Test/TestMzML.cs
avcarr2: Yes.

mzLib/Test/Test.csproj
avcarr2: I did it because the PSI-MOD.obo file shouldn't need to change unless there's a new version from the mzml people. I suppose I could make it dynamically fetch it from GitHub, but I don't have time to write that method.

YulingDai (approved):

trishorts (approved):

PR #660, opened 2022-10-03
Reader refactoring

avcarr2:
The purpose of this pull request is to unify the file readers into a single, easy-to-use file reader structure. 

File reading is now greatly simplified. The new file reader can be called like this: 
`var reader = ReaderCreator.CreateReader(filePath)`
`reader.LoadAllStaticData(); `

To interact with the MsDataScans[] and the SourceFile reader contains public properties that contain them. 

To implement this, I converted the MsDataFile to an abstract class and gave it three properties: MsDataScan[], SourceFile, and FilePath. These properties are accessed and manipulated using the virtual methods that are defined in the abstract class, keeping with the "write it once" principal. I implemented this because while I was reworking MsDataFile, I also noticed that each derived class was overriding methods that should've been virtually implemented in the base class, and/or re-implementing methods with very similar functionality. With this update, all derived classes now share the same methods, making it easier to maintain and use. 

Each type of file (.mgf, .mzml, and .raw) all read files in completely different ways. Therefore, the MsDataFile abstract class defines five abstract methods that each derived class  must implement: LoadAllStaticData, GetSourceFile, GetOneBasedScanFromDynamicConnection, InitiateDynamicConnection, and CloseDynamicConnection. Any reader class derived from MsDataFile will need to implement these methods. 

To read a specific file, I implemented an abstract factory design architecture to create the file type-specific reader class. This can be seen in ReaderFactoryClasses.cs and ReaderCreator.cs. 

Errata:

- I removed the dynamic data connection classes for each file type. The same functionality is now achieved using the methods containing "Dynamic" in classes derived from MsDataFile. 

- Because MsDataFile is now an abstract class, the minimal implementation of a concrete MsDataFile is now GenericMsDataFile.


avcarr2:
I know that there are 64 files changed, but the vast majority of those changes are method renames, file locations moving, files being deleted, etc.

acesnik:
Just took a glance, and it looks like testing these things might be a quick way to get codecov back up:
* loading mzMLs
* “DynamicDataConnection"
* exceptions testing in loading files

PR #664, opened 2022-10-20
Read one based precursor scan number from mzml

trishorts:
user wants to process mzMl files where precursor scans have been removed. mzLib errors when reading these files because it looks for those scans by default. Also, it renumbers all scans in order beginning with 1. 

In this PR, we maintain the default behavior but let users opt out and instead simply use the scan numbers recorded in the supplied mzml.

I have no idea what will come of this approach.

caveat emptor.


lonelu (approved):

lonelu:
Please merge this, thank you @trishorts

Alexander-Sol:
Couple of dependency issues that need to be fixed

image

lonelu:
Not sure what is happening, it doesn't seem like related to this pull request.

> Couple of dependency issues that need to be fixed
> 
> image

acesnik (dismissed):

lonelu (approved):

trishorts:
> Couple of dependency issues that need to be fixed
> 
> image

nuget package dependencies are now resolved.

Alexander-Sol (approved):

acesnik (approved):

lonelu (approved):

PR #667, opened 2022-10-24
Added RtHypothesis and RtStdDev fields to ChromatographicPeak in Flas…

Alexander-Sol:
We like to enable re-analysis of an MBR Chromatographic peak once it is created & passed to MetaMorpheus. We add retention time hypothesis and retention time std dev, which can be used to define the region used to search for the peak.

This PR adds three new files (nullable doubles) to ChromatographicPeak: RtHypothesis and RtStdDev, and RtInterquartile Range. RtStdDev and RtInterquartileRange are mutually exclusive. These fields are only applied to MBR peaks, for other peak types they remain null

These fields are assigned in FlashLFQEngine method QuantifyMatchBetweenRunsPeaks.


mzLib/FlashLFQ/ChromatographicPeak.cs
```diff
@@ -35,6 +34,9 @@ public ChromatographicPeak(Identification id, bool isMbrPeak, SpectraFileInfo fi
         public int NumIdentificationsByBaseSeq { get; private set; }
         public int NumIdentificationsByFullSeq { get; private set; }
         public double MassError { get; private set; }
+        public double? RtHypothesis { get; private set; }
```
acesnik, line 37: Could you add summaries for these new properties?

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -761,9 +763,19 @@ private void QuantifyMatchBetweenRunsPeaks(SpectraFileInfo idAcceptorFile)
 
                                     var xic = Peakfind(seedEnv.IndexedPeak.RetentionTime, donorIdentification.PeakfindingMass, z, idAcceptorFile, mbrTol);
                                     List<IsotopicEnvelope> bestChargeEnvelopes = GetIsotopicEnvelopes(xic, donorIdentification, z);
-
                                     acceptorPeak.IsotopicEnvelopes.AddRange(bestChargeEnvelopes);
                                     acceptorPeak.CalculateIntensityForThisFeature(Integrate);
+                                    if (interquartileRangeUsed)
+                                    {
+                                        acceptorPeak.SetRtWindow(acceptorFileRtHypothesis, rtStdDev: null,
+                                            rtInterquartileRange: rtRange / 4.5);
```
acesnik, line 771: Could you add a comment explaining this value and  `/ 6` below?

acesnik (changes requested):
The math isn't obvious to me, but it looks well tested. Some more comments would probably help make it clearer.

Cool that you're working on FlashLFQ!

acesnik (approved):

mzLib/FlashLFQ/ChromatographicPeak.cs
```diff
@@ -68,6 +73,13 @@ public static string TabSeparatedHeader
             }
         }
 
+        public void SetRtWindow(double rtHypothesis, double? rtStdDev, double? rtInterquartileRange)
```
trishorts, line 76: I don't know why this is public or why you need to set all three values at once. Seems like it should be a private method and there should be interdepencies or whatever.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -685,6 +685,7 @@ private void QuantifyMatchBetweenRunsPeaks(SpectraFileInfo idAcceptorFile)
 
                             // default range (if only 1 datapoint, or SD is 0, range is very high, etc)
                             double rtRange = MbrRtWindow;
+                            bool interquartileRangeUsed = false;
 
                             if (nearbyCalibrationPoints.Count < 6 && nearbyCalibrationPoints.Count > 1 && rtDiffs.StandardDeviation() > 0)
                             {
```
acesnik, line 691: Why stdev * 6 and IQR * 4.5 here? I'm not sure I understand those constants.

Alexander-Sol: I honestly don't know. That's how FlashLFQ is written, and those values aren't explained. Rob might remember

acesnik (dismissed):

nbollis (dismissed):

nbollis (approved):

trishorts (approved):
not a big fan of externally changing values for variables

PR #674, opened 2022-11-10
Change CompareTo method within DoubleRange to adhere to conventions

Alexander-Sol:
In C#, the convention for CompareTo methods is to compare the instance to the item passed as an argument. For example

`Double five = 5;`
`double ten = 10;`
`five.CompareTo(ten);`

Would return -1, because five is less than ten. The current implementation of DoubleRange.CompareTo() has this reversed. This PR fixes that (admittedly pedantic) issue.

https://learn.microsoft.com/en-us/dotnet/api/system.int32.compareto?view=net-7.0
https://learn.microsoft.com/en-us/dotnet/api/system.double.compareto?view=net-7.0


avcarr2 (approved):
LGTM

nbollis (approved):

YulingDai (approved):

PR #681, opened 2023-01-20
Mrs noise estimation

avcarr2:
Added spectra weighting by MRS noise estimation. 

MRS noise estimation uses a wavelet transform to discriminate which parts of a signal are noise versus signal. The standard deviation of the noise is then calculated using only the noise values, giving a more robust and accurate noise estimation than just taking the standard deviation of the signal. The noise estimate is subsequently used in calculating the weight of a spectra when averaging. 

MRS noise estimation algorithm is as follows: 

1. Estimate the standard deviation of the noise in the original signal using the regular standard deviation. 
2. Compute the maximal overlap discrete wavelet transform of the image. 
3. Set n = 0; (this is done implicitly while in the do while loop.)
4. Compute the multiresolution support M that is derived from the wavelet coefficients and the standard deviation of the noise at each level. The multiresolution support creates a boolean array where a true indicates at the location in the array, there is a value higher than the noise estimate, and false means there is not a value higher than the noise estimate. 
5. Select all points that belong to the noise; these points don't have significant wavelet coefficients above noise. Significant is defined as 3*estimated standard deviation. 
6. For the selected pixels, calculate original array - smoothed array (where the smoothed array is the original signal array - the sum of the wavelet coefficients across all scales) and compute the standard deviation for only those values.
7. n = n + l. 
8. start again at 4 if (sigma_I^n - sigma_I^(n-1)) / sigma_I^(n) > epsilon. 

image

### DIFF
--- a/mzLib/SpectralAveraging/MrsNoiseEstimation/ModWtOutput.cs
+++ b/mzLib/SpectralAveraging/MrsNoiseEstimation/ModWtOutput.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SpectralAveraging;
+
+public class ModWtOutput
+{
+    public ModWtOutput(int maxScale, WaveletType waveletType, BoundaryType boundaryType)
+    {
+        Levels = new List<Level>();
+        MaxScale = maxScale;
+        WaveletType = waveletType;
+        BoundaryType = boundaryType;
+    }
+
+    public List<Level> Levels { get; private set; }
+    public int MaxScale { get; private set; }
+    public WaveletType WaveletType { get; }
+    public BoundaryType BoundaryType { get; }
+
+    public void AddLevel(double[] waveletCoeff, double[] scalingCoeff, int scale,
+        BoundaryType boundaryType, int originalSignalLength, int filterLength)
+    {
+        if (boundaryType == BoundaryType.Reflection)
+        {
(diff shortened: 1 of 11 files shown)


mzLib/SpectralAveraging/Util/BasicStatistics.cs
```diff
@@ -7,8 +7,18 @@ namespace SpectralAveraging;
...
+        IEnumerable<double> sortedValues = toCalc.OrderByDescending(p => p).ToList();
+        double median;
+        int count = sortedValues.Count();
+        if (count % 2 == 0)
+            median = sortedValues.Skip(count / 2 - 1).Take(2).Average();
+        else
+            median = sortedValues.ElementAt(count / 2);
+        return median;
+    }
     public static double CalculateNonZeroMedian(IEnumerable<double> toCalc)
```
nbollis, line 22: This method is unnecessary, can use .Median()

nbollis (approved):

acesnik (approved):

mzLib/SpectralAveraging/MrsNoiseEstimation/ModWtOutput.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SpectralAveraging;
+
+public class ModWtOutput
```
elaboy, line 7: Does this class need to be public or can it be changed to internal?

elaboy (approved):

PR #683, opened 2023-01-26
Averaging Updates 

nbollis:
Averaging Changes
Fixed bug that would cause a crash if all values within a bin got rejectect
Added method to pass in a ton of possible parameters and output all valid parameters for bulk testing
Added option to output averaged spectra to custom location and with a custom name. Previous iterations placed them in the same location as the original with the filename $"{OrignalName}-averaged.mzML"
Removed option to export averaged spectra as csv. Not necessary for production, but useful when constructing algorithm.


mzLib/SpectralAveraging/Algorithms/OutlierRejection.cs
```diff
@@ -153,18 +153,20 @@ private static double[] WinsorizedSigmaClipping(double[] initialValues, double s
...
-            var toProcess = values.ToArray();
+            var median = values.Median();
+            var standardDeviation = values.Where(p => p != 0).StandardDeviation();
+            double[] toProcess = new double[values.Count];
+            values.CopyTo(toProcess, 0);
             double winsorizedStandardDeviation;
+
             do // calculates a new median and standard deviation based on the values to do sigma clipping with (Huber loop)
             {
                 var medianLeftBound = median - 1.5 * standardDeviation;
```
trishorts, line 164: can we get some definition on these numbers (1.5 and 1.1.34)?

trishorts (commented):
very close

mzLib/MzLibUtil/MrsNoiseEstimation/Level.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace SpectralAveraging;
```
acesnik, line 1: Consider removing the "MRS" parts of these namespaces

mzLib/mzLib.nuspec
```diff
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>mzLib</id>
-    <version>5.0.538</version>
+    <version>5.0.540</version>
```
acesnik, line 5: I think this is supposed to be the version of the package. Should this be `1.0.540`?

acesnik (commented):
Looks good to me. Would be happy to approve if ready.

mzLib/MzLibUtil/MrsNoiseEstimation/Level.cs
nbollis: The thought here was that theses are methods and classes that will not be used frequently, but could be useful in many places. Adding them to MzLibUtil would allow them to be called form anywhere, but relegating them to their own namespace would prevent them from being accessible unless actually needed. Should this be changed?

acesnik (dismissed):

mzLib/MzLibUtil/MrsNoiseEstimation/Level.cs
acesnik: Hmm, this is just a naming comment regarding `MrsNoiseEstimation` and possibly changing it to `NoiseEstimation` or something. No strong preference, just a thought.

acesnik (dismissed):

mzLib/MzLibUtil/MrsNoiseEstimation/MRSNoiseEstimator.cs
```diff
@@ -25,7 +23,7 @@ public static class MRSNoiseEstimator
         {
             int iterations = 0; 
             // 1. Estimate the standard deviation of the noise in the original signal. 
-            double stdevPrevious = BasicStatistics.CalculateStandardDeviation(signal);
+            double stdevPrevious = signal.StandardDeviation();
```
trishorts, line 26: how about naming this EstimatedStandardDeviation if it is not actually the standard deviation

mzLib/SpectralAveraging/Algorithms/AveragedSpectraWriter.cs
```diff
@@ -14,82 +15,93 @@ public static class AveragedSpectraWriter
...
+    /// specified by the optional destination directory and destination file name fields
     /// </summary>
     /// <param name="averagedScans"></param>
     /// <param name="originalSpectraPath"></param>
-    private static void WriteAveragedSpectraAsMzML(MsDataScan[] averagedScans,
-        string originalSpectraPath)
+    /// <param name="destinationDirectory"></param>
+    /// <param name="averagedFileName"></param>
+    private static void WriteAveragedSpectraAsMzMl(MsDataScan[] averagedScans,
+        string originalSpectraPath, string? destinationDirectory = null, string? averagedFileName = null)
```
trishorts, line 43: why leave filename and path as null?

mzLib/SpectralAveraging/Algorithms/OutlierRejection.cs
```diff
@@ -191,18 +194,26 @@ private static double[] WinsorizedSigmaClipping(double[] initialValues, double s
     private static double[] AveragedSigmaClipping(double[] initialValues, double sValueMin, double sValueMax)
```
trishorts, line 194: is there a reference you can supply for averaged sigma clipping and winsorize?

mzLib/SpectralAveraging/Algorithms/SpectraAveraging.cs
```diff
@@ -65,11 +65,10 @@ public static class SpectraAveraging
             for (; binIndex < keys.Count; binIndex += maxThreadsPerFile)
             {
                 bins[keys[binIndex]] = OutlierRejection.RejectOutliers(bins[keys[binIndex]], parameters);
```
trishorts, line 67: variable names?

trishorts (changes requested):
some small details to attend to

mzLib/SpectralAveraging/Algorithms/AveragedSpectraWriter.cs
nbollis: By default, those can be unspecified. This results in the averaged spectra being placed in the same location as your original spectra with the file name being "{original file name}-averaged".

This allows more flexibility for developers using the writer in the future. I can add that to the comments

trishorts (approved):

acesnik (approved):

avcarr2 (approved):

mzLib/MzLibUtil/MrsNoiseEstimation/MRSNoiseEstimator.cs
```diff
@@ -48,7 +46,7 @@ public static class MRSNoiseEstimator
                 // and the standard deviation of the noise at each level. 
                 // 5. Select all points that belong to the noise; they don't have an significant coefficients above noise 
                 var booleanizedLevels = BooleanizeLevels(wtOutput,
-                    stdevPrevious, 1.97); 
+                    estimatedStandardDeviation, 1.97);
```
Alexander-Sol, line 49: Why 1.97? Seems like a magic number at first glance

avcarr2: Added comment for magic number in latest commit.

avcarr2 (approved):

trishorts (approved):

mzLib/MzLibUtil/MrsNoiseEstimation/MRSNoiseEstimator.cs
```diff
@@ -47,8 +45,11 @@ public static class MRSNoiseEstimator
                 // 4. Compute the multiresolution support M that is derived from the wavelet coefficients
                 // and the standard deviation of the noise at each level. 
                 // 5. Select all points that belong to the noise; they don't have an significant coefficients above noise 
+                // 1.97 is the z-score representing a value that is two standard deviations away from the mean.
```
Alexander-Sol, line 48: This is incorrect. Z-score represents the number of standard deviations separating a measurement from the mean. Measurments with a z-score < 1.97 represent 97.6% of a normal distribution.

Alexander-Sol (approved):

PR #691, opened 2023-03-02
Test Median Polish

trishorts:
We never really tested the output of Median Polish and FlashLFQ. This PR adds some unit tests of the sort that will allow us to make sure that the results of median polish don't change unexpectedly.


avcarr2 (approved):

mzLib/TestFlashLFQ/TestFlashLFQ.cs
```diff
@@ -14,6 +14,8 @@
 using UsefulProteomicsDatabases;
 using ChromatographicPeak = FlashLFQ.ChromatographicPeak;
 using Stopwatch = System.Diagnostics.Stopwatch;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using OxyPlot;
```
acesnik, line 18: What's oxyplot used for here?

mzLib/TestFlashLFQ/TestFlashLFQ.cs
```diff
@@ -916,26 +918,153 @@ public static void TestBayesianProteinQuantification()
...
+                            proteinGroups.Add(allProteinGroups[protein]);
+                        }
+                    }
+
+                    Identification id = new Identification(file, baseSequence, fullSequence, monoMass, rt, z, proteinGroups);
+                    ids.Add(id);
+                }
+            }
+
+            var engine = new FlashLfqEngine(ids, matchBetweenRuns: true, requireMsmsIdInCondition: false, useSharedPeptidesForProteinQuant: true, maxThreads: 1);
```
acesnik, line 988: Could squeeze more lines of tested code out if you set maxthreads to -1 here. Should reset it to 1 in the code.

acesnik (dismissed):
Looks good! It looks like the Bayesian statistics in the FlashLFQ engine are also not tested, in case it'd be easy to extend these tests for that purpose.

There are also some small bits of code you could test here, perhaps, e.g. maxthreads <= 0.

mzLib/TestFlashLFQ/TestFlashLFQ.cs
trishorts: eliminated

mzLib/TestFlashLFQ/TestFlashLFQ.cs
trishorts: changed

avcarr2 (approved):

Alexander-Sol (approved):
In the future, we should move to test data with some known ground truth. But this is better than what we have now.

acesnik (approved):

PR #695, opened 2023-03-15
Zero column effects result in identical intensity values

Alexander-Sol:
This reverts the change to FlashLFQ which caused proteins with 0 column effects after median/mean polish to have no reported protein group intensity values for samples corresponding to the 0 column effect columns.

Now, overall effect of median/mean polish is reported as the intensity in cases where the column effect is 0.

Currently testing this with the E. coli vignette data, will update with the results

### DIFF
--- a/mzLib/FlashLFQ/FlashLFQResults.cs
+++ b/mzLib/FlashLFQ/FlashLFQResults.cs
@@ -434,13 +434,15 @@ namespace FlashLFQ
                     }
 
                     // Checks to see if no peptides are shared across samples. If true
-                    // This protein should not be quantified
+                    // This protein will not be quantified
                     if (CheckForMutuallyExclusivePeptides(peptideIntensityMatrix))
                     {
                         foreach (var sample in SpectraFiles)
                         {
-                            proteinGroup.SetIntensity(sample, Double.NaN);
+                            proteinGroup.SetIntensity(sample, double.NaN);
                         }
+
+                        continue;
                     }
 
                     // do median polish protein quantification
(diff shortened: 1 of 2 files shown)


rmillikin:
not sure this is a good idea... for example check out the unit test "TestMedianPolish_UnquantifiableProtein". no intensity values should be reported in that case in my opinion. it doesn't make sense to report identical intensities for the two samples in that test. 

for example, consider performing a t-test on protein intensities, which uses standard deviation as part of its calculation. if there are proteins with identical intensities, the standard deviation will be artificially low and will be more likely to reject the null hypothesis regardless of whether or not it should be rejected.

Alexander-Sol:
> not sure this is a good idea... for example check out the unit test "TestMedianPolish_UnquantifiableProtein". no intensity values should be reported in that case in my opinion. it doesn't make sense to report identical intensities for the two samples in that test.
> 
> for example, consider performing a t-test on protein intensities, which uses standard deviation as part of its calculation. if there are proteins with identical intensities, the standard deviation will be artificially low and will be more likely to reject the null hypothesis regardless of whether or not it should be rejected.

I definitely understand your point here. I don't envision this as a long-term fix, I just want to revert to what we were doing before to assuage collaborator concerns over missing intensity values. 

It feels wrong to report no intensity values in situations where multiple unique peptides have been observed. 

An alternate, easy-to-implement way to address this would be to add a scaling factor  (e.g., mean protein intensity) for each sample. However, that has the potential to cause unanticipated changes.

trishorts:
> not sure this is a good idea... for example check out the unit test "TestMedianPolish_UnquantifiableProtein". no intensity values should be reported in that case in my opinion. it doesn't make sense to report identical intensities for the two samples in that test.
> 
> for example, consider performing a t-test on protein intensities, which uses standard deviation as part of its calculation. if there are proteins with identical intensities, the standard deviation will be artificially low and will be more likely to reject the null hypothesis regardless of whether or not it should be rejected.

if no change is observed, it might be "significant" via a t-test. But, I think most people are looking for at least some minimal fold change. Even 10 or 20% but more comonly 2-fold. Do you feel like such an occurence (0-fold change but passable t-test) is really problematic?

It is hard to explain to users that peptides with intensities result in a protein with no intensities. We are among the very few people that are down this rabbit hole.

rmillikin:
you're assuming that the samples with identical intensities are in different conditions. I don't think it's much of a problem in that case (conservatively biased, but if you're going to pick a bias, better to be conservative than anticonservative IMO). 

I'm positing a case in which samples with identical intensities are in the same condition, which when you calculate standard deviation for that condition, would give you an artificially low value.

The change from median to mean polish should make 0 column effects pretty rare, and getting two 0 column effects either means the protein intensities are exactly identical in two runs which is extremely rare, or the protein is unquantifiable. I actually get fewer missing values in the FlashLFQ vignette with the current MM release + mzLib 1.0.538 compared to MM 0.0.320. That's because unquantifiable proteins are easier to distinguish from quantifiable proteins with ~0 column effects due to the change from median to mean polish.

PR #696, opened 2023-03-16
Bug in ChemicalFormula.

avcarr2:
This is to fix a bug that I found. 

The bug occurs when multiple subtractive chemical formulas are added to ChemicalFormula such that the number of Elements in the Chemical Formula becomes less than zero. 

This is a problem when there are multiple modifications like deamidations occurring on a peptide. Below is a picture of what I found in a PSMS.psmtsv file: 
image

The correct answer should be H-2N-2O2. 

Original code was written by Stefan; Rob made a commit adding lines 290-291 in ChemicalFormula.cs indicating that this was a bug fix. These lines added a check that removes elements with less than 1 element present. Indeed, when I commented out Rob's bug fix, two tests in TestChemicalFormula.cs failed. 

If anybody has any thoughts on how to fix this, I would be all ears. My plan right now is to modify the ChemicalFormula properties so that they calculate for only positive values of elements. Using  TestChemicalFormula TestAddLargeNegative() as an example, I can get the tests to pass by modifying the MonoisotopicMass property to only count the elements with values greater than zero. That way, all the masses and element counts are correct, but MetaMorpheus will output the correct string when dealing with subtractive modifications. 

Initial commit demonstrates the issue with the solution of just removing the control flow Rob added.

@rmillikin @Alexander-Sol @trishorts @nbollis @elaboy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -250,3 +250,9 @@ ModelManifest.xml
 # Macintosh files
 **/.DS_Store
 
+/mzLib/Test/DataFiles/WT_ESO_+_B_B5_01_4035.mzML
+/mzLib/.idea/.idea.mzLib/.idea/workspace.xml
+/mzLib/.idea/.idea.mzLib/.idea/vcs.xml
+/mzLib/.idea/.idea.mzLib/.idea/projectSettingsUpdater.xml
+/mzLib/.idea/.idea.mzLib/.idea/indexLayout.xml
+/mzLib/.idea/.idea.mzLib/.idea/encodings.xml
--- a/mzLib/Chemistry/ChemicalFormula.cs
+++ b/mzLib/Chemistry/ChemicalFormula.cs
@@ -287,8 +287,8 @@ namespace Chemistry
             else
             {
                 Elements[element] += count;
-                if (Elements[element] <= 0)
-                    Elements.Remove(element);
+                //if (Elements[element] <= 0)
+                //    Elements.Remove(element);
             }
             formulaString = null;
         }
--- a/mzLib/Chemistry/ChemicalFormula.cs
+++ b/mzLib/Chemistry/ChemicalFormula.cs
@@ -72,7 +72,10 @@ namespace Chemistry
         {
             get
             {
-                return Isotopes.Sum(b => b.Key.AtomicMass * b.Value) + Elements.Sum(b => b.Key.AverageMass * b.Value);
+                return Isotopes.Where(b => b.Value > 0)
+                           .Sum(b => b.Key.AtomicMass * b.Value) 
+                       + Elements.Where(b => b.Value > 0)
+                           .Sum(b => b.Key.AverageMass * b.Value);
             }
         }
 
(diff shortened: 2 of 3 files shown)


avcarr2:
Turns out you can't just exclude all negative values form the calculated properties because apparently there are things relying on those downstream. So now I'm pretty stuck.

rmillikin:
I guess you could say if the element count is zero then remove it from the formula, if it is less than or greater than 0 then include it? Is it 'wrong' to have a negative mass for example? Essentially you're treating it as a relative mass or relative formula, not an absolute mass/formula.

trishorts:
agree with rob. we definitely need the negatives.

trishorts:
`        [Test]
        public static void Bubba()
        {
            IDictionary<int, List<Modification>> mods = new Dictionary<int, List<Modification>>();
            ModificationMotif.TryGetMotif("N", out ModificationMotif motif);
            Modification mod = new Modification(_originalId: "Deamidation", _modificationType: "Common Artifact", _target: motif, _chemicalFormula: ChemicalFormula.ParseFormula("H-1N-1O1"), _locationRestriction: "Anywhere.");

            mods.Add(1, new List<Modification> { mod });
            mods.Add(3, new List<Modification> { mod });
            var prot = new Protein("NPNP", null, null, null, mods);
            DigestionParams digestionParams = new DigestionParams(minPeptideLength: 1);
            List<PeptideWithSetModifications> peptides = prot.Digest(digestionParams, new List<Modification>(), new List<Modification>()).ToList();

            PeptideWithSetModifications ye = peptides[3];

            Assert.IsTrue(ye.AllModsOneIsNterminus.Count== 2);

            var frags = new List<Product>();
            ye.Fragment(DissociationType.HCD, FragmentationTerminus.Both, frags);
            var massArray = frags.Select(p => p.NeutralMass).ToArray();
            Array.Sort(massArray);
            double[] intensities = new double[] { 1, 1, 1, 1 };
            double[] mz = new double[] { massArray[0].ToMz(1), massArray[2].ToMz(1), massArray[4].ToMz(1), 10000 };
            MzSpectrum massSpectrum = new MzSpectrum(mz, intensities, false);
            MsDataScan scan = new MsDataScan(massSpectrum, 1, 1, true, Polarity.Positive, 1, new MzRange(300, 2000), "", MZAnalyzerType.Unknown, massSpectrum.SumOfAllY, null, null, "scan=1", 0, null, null, 0, null, DissociationType.Unknown, 1, null);

            PeptideSpectralMatch[] globalPsms = new PeptideSpectralMatch[1];
            Ms2ScanWithSpecificMass[] arrayOfSortedMS2Scans = { new Ms2ScanWithSpecificMass(scan, 0, 1, null, new CommonParameters()) };
            CommonParameters CommonParameters = new CommonParameters(
                productMassTolerance: new PpmTolerance(5),
                scoreCutoff: 1,
                digestionParams: new DigestionParams(
                    maxMissedCleavages: 0,
                    minPeptideLength: 1,
                    initiatorMethionineBehavior: InitiatorMethionineBehavior.Retain));
            bool writeSpectralLibrary = false;
            ClassicSearchEngine cse = new ClassicSearchEngine(globalPsms, arrayOfSortedMS2Scans, new List<Modification>(), new List<Modification>(), null, null, null,
                new List<Protein> { prot }, new OpenSearchMode(), CommonParameters, null, null, new List<string>(), writeSpectralLibrary);

            cse.Run();
            Assert.AreEqual(ChemicalFormula.ParseFormula("H-2N-2O2"), globalPsms[0].ModsChemicalFormula.Formula);
        }`

trishorts:
That unit test in MM generates a PSM with the incorrect chemical formula for mods of O2, which austin reported. I'll try to track

trishorts:
Fixed here: https://github.com/smith-chem-wisc/mzLib/pull/697

PR #697, opened 2023-03-17
Fix add chemical formula with negative values

trishorts:
Austin found a bug wherein adding two chemical formulas with negative isotopes gave the wrong answer.

The example was deamidation (H-1N-1O) plus deamidation. The expected answer is H-2N-2O2 but the original gave O2.

This PR fixes the problem


trishorts:
I'm gonna solve the isotope thing in another PR. This solves a known bug and should have immediate positive impact. the other stuff may need some discussion.

acesnik (approved):

nbollis (approved):

PR #698, opened 2023-04-04
Reader refactoring, take 2

avcarr2:
I did the initial refactoring of the code base to unify the file readers. Nic refactored my refactoring a little more to make the user interface a little easier. 

Summary: 

This pull request unifies the three file readers we currently support into a single, unified user interface in the same project. We made MsDataFile a public abstract class that the individual readers inherit from. MsDataFile defines the abstract members that we previously expected each project to have. We have also managed to put all shared functionality that was implemented in each file reader project into virtual methods within MsDataFile, cutting down on the code re-use present in the original file reader system. Each file type has a corresponding class derived from MsDataFile that implements the specific data handling protocols for importing data from the underlying file.

Any derived class of MsDataFile can be exported as an MzML
Note that the original unit tests for file reading are completely intact; they have only been moved to a new folder in the test project. 

**IMPORTANT: backwards compatibility with previous syntax of the specific readers is retained, old code and projects that use mzLib will not need any adjustments. This PR was packaged and added to MetaMorpheus, all unit tests passed.** 

How it works: 

Nic implemented an abstract factory design pattern to handle the different file formats in the static class MsDataFileReader. To read a file's static data, you can simply write a one liner. 
`var staticScansList = MsDataFileReader.GetDataFile(filePath).LoadAllStaticData([filterParams], [maxThreads]).GetAllScansList()`

Dynamic connections work in a similar manner: 
`MsDataFile msReader = MsDataFileReader.GetDataFile(filePath); 
msReader.InitiateDynamicConnection;`

The user doesn't need to be aware of the extension of the file to be loaded. 

One final note: FakeMsDataFile, which was used to test generic abilities of the original MsDataFile class has been replaced with GenericMsDataFile in all unit tests and methods that reference FakeMsDataFile. The functionality is exactly the same.


mzLib/mzLib.sln
```diff
@@ -35,109 +29,186 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestFlashLFQ", "TestFlashLF
...
+		Debug|x86 = Debug|x86
 		Release|x64 = Release|x64
+		Release|x86 = Release|x86
 		TestAndRelease|x64 = TestAndRelease|x64
+		TestAndRelease|x86 = TestAndRelease|x86
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{4658AE10-B660-46DB-B56B-7F4701339FD0}.Debug|x64.ActiveCfg = Debug|x64
 		{4658AE10-B660-46DB-B56B-7F4701339FD0}.Debug|x64.Build.0 = Debug|x64
+		{4658AE10-B660-46DB-B56B-7F4701339FD0}.Debug|x86.ActiveCfg = Debug|x64
```
acesnik, line 46: Intentional changes? I think mzLib is only used in x64, unless that's changed.

acesnik: Otherwise consider removing the x86 "AnyCPU" configuration

acesnik (dismissed):
Cool, this is much clearer syntax and organization.

Looks like mzLib.sln changed, so maybe check in on whether those are intentional changes.

acesnik (approved):

nbollis (approved):

trishorts (approved):

PR #699, opened 2023-04-17
Prepare Release - Readers backward compatibility improvement && Remove Develpment Dependency

nbollis:
Issues this PR addresses:
- Development dependency issues
- Readers packaging issue
- Readers backwards compatibility

**Development Dependency Issues**
- Development was a dependency of test, but was not getting packaged. This PR should resolve that issue by removing development as a dependency of test
- Combined like classes into a folder

**Readers Packaging Issue**
- Removed Mzml.dll from nuspec
- Removed Mgf.dll from nuspec
- Changed file location for thermo dependencies in nuspec

**Readers backwards Compatibility**
Main issue here is that while the functionality remains the same, the namespaces changed. This would force the manual conversion of namespaces when updating to the new version of mzlib
- Moved MsDataFile back to MassSpectrometry
- Moved SourceFile back to MassSpectrometry
- To (ThermoRawFileReader/MzML/Mgf), added the old namespace (IO.ThermoRawFileReader, IO.MzML/IO.Mgf) which points to the new locations in Readers
- MzmlMethods could not  use this workaround as it is a static class and cannot be inherited from

**General Cleanup**
- Individual reader factory classes are small will not be expanded or inherited from. The classes were combined into a single .cs file for cleanliness sake and the keyword 'sealed' was added
- Uncommented some lines of code in VariantApplication that were needed in MetaMorpheus

**Verification**
- MzLib was packaged locally and used in a new MetaMorpheus branch, changes were made to account for new syntax where necessary
-- Instantiation of MsDataFiles were changed to GenericMsDataFiles (only occurred in tests)
-- Changed IO.MzmlMethods to Readers.MzmlMethods

- All tests passed
- Vignette data was ran with MM 1.0.2 with MzLib 5.0.38 (old) and MM 1.0.2 with MzLib 5.0.39 (this PR)
-- Data was comparable with small changes likely accounted for with Austin's correction for PTM mass and is shown below

image


Alexander-Sol (changes requested):
You should edit the nuspec file to include Readers and remove/change ThermoReader. 

There's something else going on too. Even with Readers included in the nuspec, MetaMorpheus doesn't load Readers.dll. Could be due to an issue with Reader.csproj, which has a couple of main differences compared to other .csproj files.

mzLib/mzLib.nuspec
```diff
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>mzLib</id>
-    <version>5.0.538</version>
```
acesnik, line 5: Shouldn't this be `1.0.539`?

mzLib/Test/FileReadingTests/TestRawFileReader.cs
```diff
@@ -6,27 +6,13 @@
...
 using Readers;
-using Readers.ReaderFactories;
 
 namespace Test.FileReadingTests
 {
     [TestFixture]
     [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
     public sealed class TestRawFileReader
     {
-        [Test]
```
acesnik, line 17: Nice work on structuring this below

acesnik (dismissed):

Alexander-Sol (approved):

mzLib/MassSpectrometry/MsDataFile.cs
```diff
@@ -23,9 +23,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
-using MassSpectrometry; 
 
-namespace Readers
+namespace MassSpectrometry
 {
     // TODO: Define scope of class
```
avcarr2, line 29: TODO is still here.

avcarr2 (approved):

trishorts (approved):

PR #705, opened 2023-05-05
removed the zero-equivalent intensity filtering from mzMl

avcarr2:
Removes the pre-filtering of "zero-equivalent values" from loading an mzML file. I noticed this behavior when trying to load in simulated data where most values were under 0.01.

### DIFF
--- a/mzLib/Readers/MzML/Mzml.cs
+++ b/mzLib/Readers/MzML/Mzml.cs
@@ -621,22 +621,7 @@ namespace Readers
                                     {
                                         throw new MzLibException("Could not determine the MS order or centroid/profile status");
                                     }
-
-                                    //Remove Zero Intensity Peaks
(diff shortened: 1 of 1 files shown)


nbollis:
Do you think we should just set an option to do this instead of just removing the code?

avcarr2:
I think we shouldn't mess with the data on read, and we already have a peak filtering option using the FilteringParameters object when reading files.

nbollis:
> 

That is a good point, did you package this into MM and run all the test data in there to ensure it does not break anything?

trishorts:
> > 
> 
> That is a good point, did you package this into MM and run all the test data in there to ensure it does not break anything?

I think we need to be able to show that we can filter zeros for mgf/mzml/and raw so that there is no crash in MM.

PR #711, opened 2023-06-13
External Software File Reading

nbollis:
This PR enables the reading and writing of external result files. 
Each result file will need two layers, one to describe the file as a whole, and one to describe the individual result entries within the file. 

NOTE: The scope of this PR is to establish a framework for reading and writing files, not how they are utilized once read in

**Use**
Each file can be constructed directly from its specific type, or be instatiated through the FileReader generic static factory class. 
image

**Code Structure**
image

- ResultFile<TResult> where TResult is IResult **(Top)**
  - abstract generic class all result files inherit from. 
  - Handles the logic of loading results only if explicitly called, reducing overhead at instantiation
  - The list of results will contain whichever specific individual result entry (IResult) the derived class contains. 
  - Implements IEnumerable on the list of results and IEquatable with the file path and software used. 

- IResultFile interface  **(Middle)**
  - defines behavior for file reading and writing
  - All Derived classes from ResultFile implement this interface  
  - e.g. Ms1FeatureFile

- Specific Results **(Bottom)**
  - Individual classes representing a specific data type
  - e.g. Ms1Feature

- MsDataScanToResultileAdapter
  - Unifies the interfaces of MsDataScan and IResultFile to allow MsDataFiles to also be read through the FileReader factory class
image

**Other Things**
Two New Enums
- SupportedFileType Enum
  - Contains all currently supported file types
  - Extension method to get the file extension from the enum value 
image
   - Extension method to parse a string for its file type
image
   - This enables the file type extensions to be stored in a singular location and not hard-coded throughout each place switching on file type occurs (alot in MM)
image

- Software Enum
  - A list of software in which files can be read in from
  - Derived ResultFiles can either hard code a software the file came from, or leave it as unspecified

**How to Expand**
To create a new data type within the structure. 
1. Define the individual result type in a new class
- For things in the csv/tsv format, csv helper is a really easy way to map the properties of a file to that of the object
- See FlashDeconvMs1TsvRecord.PerChargeIntensity for an example of how to perform specific text to object data conversions

2. Define a new class derived from ResultFile<TResult> where TResult is the new data type created in step one. Also implement the IResultFile interface.
- you will then need to define methods for reading and writing the specific file. 
image
- If applicable, the software used and supported file type can be hard coded. In the case of Ms1Feature (above), this is not hardcoded as the files can arise from both FlashDeconv and TopFD. However, for a FlashDeconvMs1TsvFile (below), only one software can give rise to this data type.
image

3. Add the new data types to Readers.ExternalResults.SupportedVersions.txt with the date and version of the software

4. Your new type should now be fully integrated into the FileReader factory infrastructure


mzLib/Readers/ExternalResults/BaseClasses/ResultFile.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != this.GetType()) return false;
+            return Equals((ResultFile<TResult>)obj);
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(FilePath, Software, FileType);
```
Alexander-Sol, line 102: Could this just be FilePath?

nbollis: Change has been made

Alexander-Sol (dismissed):

elaboy (approved):

trishorts (approved):

Alexander-Sol (approved):

avcarr2 (approved):

PR #715, opened 2023-06-29
fix gz decompress so that we get the complete xml database

trishorts:
There has been intermittent trouble decompressing an xml file that is compressed as .gz. I think the trouble is that we were simultaneously streaming the decompression and the xml to yield protein entries. In this PR, i separate this into two steps. First I completely decompress the file into a temp.xml. Then i stream this temp file into protein entries. Then I delete the temp file.


Alexander-Sol (changes requested):
I saw this issue with a .gz compressed fasta file. Can we add in a test for this situtaion

nbollis (approved):

elaboy (approved):

nbollis (approved):

PR #726, opened 2023-08-14
Encyclopedia formated peptide full sequence including mass shift in brackets

trishorts:
some external tools require that peptide sequence is formatted with mass shifts rather than PTM names for operation.
This PR adds a new method to peptidesWithSetModifications that yields the required output.


nbollis (approved):

mzLib/Test/TestPeptideWithSetMods.cs
kyp4, line 1: Please add a comment regarding what this tests

kyp4 (approved):
Looks good, passes tests

acesnik (approved):

PR #727, opened 2023-08-23
Reduced compexity of GetOneBasedScan from O(logN) to O(1)

nbollis:
This change reduced time to calibrate and search two raw files from 350 minutes to 77 minutes. MM Version 101 had the time at 51 minutes. While this change does not get us to previous performance, it is a step in the right direction

Big change is in the MsDataFile.GetOneBasedScan() method. 
Previous implementation had it indexing the scan array. 
Reader refactor had it perform a linq statement. 

All other data file types will leave a null value at the specific index when a scan is not present, but mgf does not. To account for this, and indexed scans array is created for mgf files when loaded. The indexed array and override of GetOneBasedScan will allow all data files to call scans with O(1) complexity.


Alexander-Sol (approved):
<3

trishorts (approved):

PR #749, opened 2023-12-11
Eliminate extra tab from chromatographic peak

trishorts:
in dealing with the extra tab in the flashlfq output https://github.com/smith-chem-wisc/FlashLFQ/issues/134
I solved the peaks problem previously but ignored the peptides and proteins output problem. this pr tests and fixes these last two cases.


Alexander-Sol:
Can you also fix the issue with the AllPeptides header in this PR?

acesnik (approved):

Alexander-Sol (approved):

nbollis (approved):

PR #753, opened 2024-01-03
Proteomics repository retriever

trishorts:


trishorts:
Closing — this is a 2024 work-in-progress scaffold that isn't functional (tip commit was titled "bunch of stuff that doesnt work"). A review confirms the core paths don't work: `PrideFtp()` has its entire transfer body commented out (and references an undefined variable), the file-list method hits the authenticated `/private/` endpoint with no credentials, `RetrieveProjectFileByFilename` writes JSON metadata under the binary file's name, and the input-validation guard is always-true dead code; the tests use hardcoded `E:\junk` paths, pass an empty accession instead of the declared one, and several have no assertions.

PRIDE dataset retrieval is genuinely missing from master (which only retrieves UniProt proteomes), so this is worth revisiting — but as a fresh, properly-scoped implementation rather than salvaging this stub. Closing to keep the open-PR set to review-ready work.

PR #755, opened 2024-01-09
Usi downloader

Alexander-Sol:
Implemented UsiLoader class that downloads spectra from various databases as JSON, then converts them into a lightweight UsiSpectrumFromJSON class. Conversion from JSON object is done using Newtonsoft, which is a nuget package that this PR adds to the UsefulProteomicsDatabases dependencies.

MzSpectrum objects can be extracted from UsiSpectrumFromJSON objects.

UsiSpectrumFromJSON spectra could have many possible attributes, of which and isolation window and peptide base sequence can currently be accessed.


trishorts:
Closing this as superseded — the capability landed in #1099 (merged `ab4dd2af`), which you reviewed.

`PrideArchiveClient.GetProxiSpectrumAsync(usi)` downloads a spectrum by USI over PROXI and returns `PrideProxiSpectrum`, which *derives from* `MzSpectrum` — your review comment on #1099 is why, so there is no extraction step to call. `GetSpectrumAsync(usi)` hands back the bare `MzSpectrum` when that is all you want.

On the two attributes this PR calls out specifically: the isolation window is read in `PrideArchiveExtensions.ToMsDataScan`, which maps `MS:1000827`/`828`/`829` onto `MsDataScan.IsolationMz` and `IsolationWidth`; and the raw `List<CvParam> Attributes` is public on the DTO, so the peptidoform / base sequence is reachable without a dedicated accessor. The Newtonsoft dependency this PR would have added to `UsefulProteomicsDatabases` is already there from #1088.

Reopen if something here is still uncovered — I would rather add it to the merged client than lose it.

PR #757, opened 2024-01-19
Nucleotide, Rnase, Digestion Params Clone Method, and Interface Modificaiton

nbollis:
Adjusted interfaces to allow for abstraction of the SpectralMatchClass in MetaMorpheus

Added Clone method to DigestionParams
  There are many many times in MetaMorpheus where we need to copy digestion parameters to file specific parameters and they are done manually each time. Once incorporated, will allow all cloning to take place in one location

Added Nucleotide class that is modeled off of the Residue Class

Added Rnase without digestion functionality. 
This will make subsequent PR's with digestion easier

### DIFF
--- a/mzLib/Omics/IBioPolymer.cs
+++ b/mzLib/Omics/IBioPolymer.cs
@@ -15,9 +15,13 @@ namespace Omics
         string Organism { get; }
         string Accession { get; }
         /// <summary>
-        /// The list of gene names consists of tuples, where Item1 is the type of gene name, and Item2 is the nameof the specific molecule.
-        /// There may be many genes and names of a certain type produced when reading an XML protein database.
-        /// For RNA there may be a gene name and a specific transcript name due to splice variants.
+        /// This tuple originates from reading UniProt XML protein entries. It consists of two parts.
+        /// The second part is the gene name, which is also often referred to as the gene symbol.
+        /// The first part is a descriptor for the second part. For example, "primary" is a word used for the
+        /// first part to indicate that the gene symbol/name that follows is the name most widely accepted
+        /// by the community. The word "synonym" in the first part would indicate that the gene name in the
+        /// second part is an alternative. The word "ORF", which stands for open reading frame in the first part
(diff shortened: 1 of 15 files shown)


mzLib/Omics/IBioPolymer.cs
```diff
@@ -1,24 +1,25 @@
...
+        string FullName { get; }
         string BaseSequence { get; }
         int Length { get; }
         string DatabaseFilePath { get; }
         bool IsDecoy { get; }
         bool IsContaminant { get; }
         string Organism { get; }
         string Accession { get; }
+        /// <summary>
+        /// The list of gene names consists of tuples, where Item1 is the type of gene name, and Item2 is the nameof the specific molecule.
```
trishorts, line 18: https://www.uniprot.org/help/gene_name

trishorts: image

trishorts: when reading uniprot xml, the two fields that go in the tuple in this case are "primary" and "dpiA", which is the most common gene symbol. I believe that the other pairs of words are added as the tuple in the list. I'm not sure what is done with those values.

trishorts: Here is my attempt at a comment:
//This tuple originates from reading UniProt XML protein entries. It consists of two parts.
//The second part is the gene name, which is also often referred to as the gene symbol.
//The first part is a descriptor for the second part. For example, "primary" is a word used for the 
//first part to indicate that the gene symbol/name that follows is the name most widely accepted
//by the community. The word "synonym" in the first part would indicate that the gene name in the
//second part is an alternative. The word "ORF", which stands for open reading frame in the first part
//is also acceptable. Here, the second part will include a range, for example "AO055_05240"

trishorts (approved):
check the comments

trishorts (approved):

elaboy (approved):

Alexander-Sol (approved):

PR #760, opened 2024-02-07
Changed Deconvoluter to have a static call

nbollis:
The Deconvoluter class was altered to have a static call instead of an instance. 
This enables MM to parallelize deconvolution as each call is through a static method. 

An error in the MsDataScan.GetIsolatedMassesAndCharges() was corrected. A precursor spectrum must now be passed in.

### DIFF
--- a/mzLib/Development/Deconvolution/StandardDeconvolutionTest.cs
+++ b/mzLib/Development/Deconvolution/StandardDeconvolutionTest.cs
@@ -46,85 +46,83 @@ namespace Development.Deconvolution
                 "Averaged_221110_CytoOnly.mzML");
 
             // set up deconvoluters to be utilized by test cases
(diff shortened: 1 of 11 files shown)


mzLib/MzLibUtil/DoubleRange.cs
```diff
@@ -135,5 +138,15 @@ public bool Contains(double item)
         {
             return CompareTo(item).Equals(0);
         }
+
+        /// <summary>
+        /// Determines if a majority of the items are within the range of values
```
trishorts, line 143: items?

nbollis: This is the convention used with the contains method above

trishorts (approved):

elaboy (approved):

Alexander-Sol (approved):

PR #768, opened 2024-03-07
Changed PWSM GetHashCode and Equals methods

Alexander-Sol:
GetHashCode and Equals now only care about the full sequence of the PWSM. 

Previously, non-equal PWSMs could have identical hash codes. Now, any equal PWSM will have identical hash codes and vice versa.

### DIFF
--- a/mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
+++ b/mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
@@ -888,28 +888,13 @@ namespace Proteomics.ProteolyticDigestion
         {
             var q = obj as PeptideWithSetModifications;
 
-            if (Protein == null && q.Protein == null)
-            {
-                return q.FullSequence.Equals(this.FullSequence);
-            }
-
             return q != null
-                && q.FullSequence.Equals(this.FullSequence)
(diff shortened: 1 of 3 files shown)


mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -888,28 +888,13 @@ public override bool Equals(object obj)
...
-                return q.FullSequence.Equals(this.FullSequence);
-            }
-
             return q != null
-                && q.FullSequence.Equals(this.FullSequence)
-                && q.OneBasedStartResidue == this.OneBasedStartResidue
-                && (q.Protein.Accession == null && this.Protein.Accession == null || q.Protein.Accession.Equals(this.Protein.Accession))
-                && q.DigestionParams.DigestionAgent.Equals(this.DigestionParams.DigestionAgent);
+                && q.FullSequence.Equals(this.FullSequence);
         }
```
nbollis, line 893: This new equals method will make two Peptides from different proteins, but having the same full sequence equal. Which differs from the previous functionality requiring them to have the same full sequence and originate from the same protein. I believe this makes more sense for peptide equality. My only concern is if this method was used somewhere downstream with the previous functionality.

nbollis (dismissed):

PR #776, opened 2024-04-17
Tims tof reader

Alexander-Sol:
Contains the basic implementation of a reader for timsTOF data files


Alexander-Sol:
This is not ready for review, I think there's still local file references in the tests

PR #785, opened 2024-06-04
[WIP] Edited FlashLFQ to include option for verbose peaks

Alexander-Sol:
FlashLFQ now has the option to encode the entire isotopic envelope in the quantified peaks outputs

This is done using three new columns

Isotope Peak Intensity: Report the intensity of every isotope for every charge state
ex. - "[i0z2: 483938.28125, 2493913, 3555188, 3072931, 804185.4375, 141336.234375, 97940.609375, 147924.265625, 74724.625];[i1z2: 309911.4375, 1654543.625, 2229853, 2131625.75, 523685.3125, 121983.21875, 86103.21875, 36135.3828125, 23342.609375]"
Here, i0 refers to the first monoisotopic peak, and z2 indicates that it is charge state 2. Each intensity value is taken from a sequential MS1 scan. Information for each isotope is enclosed in brackets, different isotopes are delimited using a semi-colon

Isotope Peak m/z: Reports the exact m/z of every isotope peak for every charge state
IsotopePeakRTs: Reports the retention time of each MS1 scan in which an isotopic envelope was observed
[peaks.xlsx](https://github.com/user-attachments/files/15570391/peaks.xlsx)


avcarr2 (dismissed):

trishorts:
Rebuilt this on current master and pushed. It was conflicting in four files against a single commit from 2024-06-04, so rather than resolve hunk by hunk I took master's side everywhere and re-applied the feature on top. **5546 passed, 0 failed, 29 skipped** locally (`TestCategory!=ExternalService`); the feature is now **+443/−5 across 6 files**, including a new test file.

## What master had moved

`IsotopicEnvelope` now takes a Pearson correlation and an `IIndexedPeak`; the engine takes a `FlashLfqParameters` instead of sixteen constructor arguments; the test project moved from `TestFlashLFQ/` to `Test/FlashLFQ/`. So the option is now `FlashLfqParameters.WriteVerbosePeaks` (defaulted off) rather than an engine constructor argument, carried through to `FlashLfqResults` so the writer knows which header to emit.

## Master had already half-landed this

`FlashLfqEngine` builds an `isotopologuePeaks` list — declared at line 1574, populated at 1636, and **never read**. That is exactly the collection step this PR added as `allPeaks`, landed separately at some point and left dangling. I wired the feature to it instead of adding a second list.

It also had a latent bug: declared outside the loop over the XIC and never cleared, so it accumulated peaks across every scan. Harmless while nothing read it; it is cleared per peak now that something does.

## Two defects fixed

**The verbose rows had one more column than the verbose header.** `GetIsotopeInformation()` returned a trailing tab, and `ToString(bool)` already joined with one. I verified this rather than assuming it — reverting the fix makes the new column test fail with `Expected: 29 / But was: 30`.

**`WritePeakDictionary` could recurse without bound.** It walked observed peaks and expected isotope positions in lockstep and, on a miss, called itself with a 5 ppm wider tolerance. For a peak sitting between two isotope positions there is no tolerance that resolves it, so the widening never stops. It now computes the isotope number directly from the m/z in a single pass — no widening, no recursion, no way to fail to terminate — and drops peaks that match no isotope position rather than forcing them to fit.

Both of those method bodies were decompiler output (explicit generic arguments on `OrderBy<,>`, casts to `IEnumerable<T>` of the element type already in hand), so they are rewritten as ordinary code. Numbers now format with `InvariantCulture`, which they did not before — the TSV was previously dependent on the machine's decimal separator. Also dropped from the branch: `int placeholder = 0;`, a stray `using System.Net;`, and an unused `Easy.Common.Extensions`.

## Tests

`Test/FlashLFQ/TestVerbosePeaks.cs`, 5 tests:

- off by default, and off leaves envelopes, header and rows exactly as before
- summed intensities are identical with the option on, so this adds output rather than changing quantification
- every retained peak sits at the m/z its isotope number implies, in the scan its envelope belongs to
- every row has exactly as many columns as the header, and the three columns really do carry the envelope over its elution
- the dictionary drops unmatched peaks, keeps the more intense of two peaks landing on one isotope, and returns empty for degenerate input

## One thing to decide

This overlaps #1122, which exports strictly more MS1 data — every peak in a window around each quantified peak, not just the isotope peaks the envelope was summed from. The difference is really format rather than substance: this appends three human-readable columns to `QuantifiedPeaks.tsv`, #1122 writes a separate JSON Lines file.

Worth keeping both only if the TSV columns are wanted alongside the JSONL. If not, this is the one to drop — it is the smaller dataset. I have not assumed either way.

PR #788, opened 2024-07-11
All Quantified Peaks Reader

nbollis:
Added in a class to read in our all quantified peaks output from FlashLFQ


trishorts (dismissed):

elaboy (approved):

mzLib/Readers/QuantificationResults/QuantifiedPeak.cs
```diff
@@ -0,0 +1,94 @@
...
+        public string FullSequence { get; set; }
+
+        [Name("Protein Group")]
+        public string ProteinGroup { get; set; }
+
+        [Name("Peptide Monoisotopic Mass")]
+        public double PeptideMonoisotopicMass { get; set; }
+
+        [Name("MS2 Retention Time")]
+        public double MS2RetentionTime { get; set; }
```
Alexander-Sol, line 37: This should be nullable, otherwise MBR peaks will throw an exception

nbollis: fixed this

elaboy (approved):

Alexander-Sol (approved):

trishorts (approved):

PR #792, opened 2024-08-19
MSFragger Results Folder Reader

mzhastings:
Implemented method to more efficiently access MsFragger results using the full results folder rather than accessing each psm.tsv file within the folder individually.

**Main**

- MsFraggerCombinedResults uses the path to the folder generated by MsFragger to collectively access the results from every psm.tsv file in the folder and concatenate them into one complete list of results. 

**Result Files**

- ExperimentAnnotation and ExperimentAnnotationFile enable reading of the expriment_annotation.tsv file produced in the MsFragger results folder. This file contains the file path to each raw result file and their corresponding sample names, which are both necessary for locating the psm.tsv files.


mzLib/Readers/ExternalResults/ResultFiles/ExperimentAnnotationFile.cs
nbollis, line 1: How is this different than our current experimental design file? Is it a fragger output?

mzLib/Readers/ExternalResults/IndividualResultRecords/ExperimentAnnotation.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public class ExperimentAnnotation
+    {
+        public static CsvConfiguration CsvConfiguration = new CsvConfiguration(CultureInfo.InvariantCulture)
+        {
+            Delimiter = "\t",
+            HasHeaderRecord = true,
+            IgnoreBlankLines = true,
+            TrimOptions = TrimOptions.Trim,
+            BadDataFound = null,
```
nbollis, line 23: BadDataFound = null will skip all lines it cannot understand, instead of crashing. Is this the intended behavior?

mzLib/Readers/Util/SupportedFileTypes.cs
```diff
@@ -64,6 +65,7 @@ public static string GetFileExtension(this SupportedFileType type)
                 SupportedFileType.MsPathFinderTDecoys => "_IcDecoy.tsv",
                 SupportedFileType.MsPathFinderTAllResults => "_IcTDA.tsv",
                 SupportedFileType.CruxResult => ".txt",
+                SupportedFileType.ExperimentAnnotation => "experiment_annotation.tsv",
                 _ => throw new MzLibException("File type not supported")
             };
         }
```
nbollis, line 71: Need to add a line to ParseFileType for the annotation file

mzLib/Test/TestMsFraggerCombinedResults.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                Assert.That(allFiles.TryGetValue(fileName, out var output));
+                Assert.That(filePaths.Contains(output));
+            }
+        }
+
+        [Test]
+        [TestCase(@"FileReadingTests\ExternalFileTypes\EditedMSFraggerResults\experiment_annotation.tsv")]
+        public void TestExperimentAnnotationFile(string path)
+        {
```
nbollis, line 91: This test should have a check at the beginning to delete the file if it already exists. It should also delete the file at the end

mzLib/Test/FileReadingTests/TestSupportedFileExtensions.cs
nbollis, line 1: You should add a test case to this method for your new file type

nbollis (changes requested):
Great idea. Working with all the fragger individual files is very annoying, this will  be a lovely addition to our ever-growing collection of readers

mzLib/Readers/ExternalResults/ResultFiles/ExperimentAnnotationFile.cs
mzhastings: It is a fragger output within the results folder containing a summary of the samples including their names and the full file path to their raw files. This is necessary for finding the directories within the results folder with the psm.tsv files since the names of the directories are the sample names. The raw file paths are used to build the dictionary linking file names to their file paths.

nbollis (approved):

Alexander-Sol (approved):

PR #795, opened 2024-08-21
DecoyScrambler now accounts for modifications

Alexander-Sol:
Now, when creating scrambled decoy proteins, modifications come along for the ride. 

Rearranges the OneBasedPossibleLocalizedModifications dictionary to correspond to the new scrambled protein sequenc

### DIFF
--- a/mzLib/Proteomics/Protein/Protein.cs
+++ b/mzLib/Proteomics/Protein/Protein.cs
@@ -857,7 +857,7 @@ namespace Proteomics
 
                     scrambledProteinSequence = scrambledProteinSequence.Replace(peptideSequence, scrambledPeptideSequence);
 
-                    if (!scrambledModificationDictionary.IsNotNullOrEmpty()) continue;
+                    if (!scrambledModificationDictionary.Any()) continue;
 
                     // rearrange the modifications 
                     foreach (int index in scrambledProteinSequence.IndexOfAll(scrambledPeptideSequence))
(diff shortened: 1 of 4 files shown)


mzLib/Proteomics/Protein/Protein.cs
```diff
@@ -832,33 +834,75 @@ public static Protein ScrambleDecoyProteinSequence(
             }
 
             string scrambledProteinSequence = originalDecoyProtein.BaseSequence;
+            // Clone the original protein's modifications
+            var scrambledModificationDictionary = originalDecoyProtein.OriginalNonVariantModifications.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+
```
nbollis, line 839: Instead of cloning, then checking if anything is contained later, could you scramble, check if there are mods, then clone?

mzLib/Test/TestProteinDigestion.cs
```diff
@@ -393,6 +394,61 @@ public static void TestDecoyScramblerReplacesPeptides()
...
+            ModificationMotif.TryGetMotif("F", out ModificationMotif motifF);
+            Modification modG = new Modification("myMod", null, "myModType", null, motifG, "Anywhere.", null, 10, null, null, null, null, null, null);
+            Modification modF = new Modification("myMod", null, "myModType", null, motifF, "Anywhere.", null, 10, null, null, null, null, null, null);
+
+            //IDictionary<int, List<Modification>> modDictTarget = new Dictionary<int, List<Modification>>
+            //{
+            //    {9, new List<Modification> { modG } },
+            //    {7, new List<Modification> { modF } }
+            //};
+
```
nbollis, line 415: simple cleanup can be done here

nbollis (changes requested):
What happens with terminal modifications? We have Modifications that are designated as Protein N/C terminal as well as Peptide N/C terminal. How are these cases handled?

trishorts (approved):

nbollis (approved):

PR #797, opened 2024-08-27
PTM Stoichiometry

pcruzparri:
Creating a mzLib method to calculate the stoichiometry (or site-occupancy) of PTMs using the intensity of each quantified peak. The current inputs are the _protein database(s) file(.xml) paths_ and the _AllQuantifiedPeaks.tsv_ file path. The output, `occupancyDict`, is currently a dictionary of nested dictionaries with the following structure:
```
{{string PROTEIN1, {{int MAA1, {{string MODNAME1, double INTENSITY}, 
                                {string MODNAME2, double INTENSITY},
                                ...,
                                {string "Total", double INTENSITY}}} 
                    {int MAA2, {...}}, 
                   ...}},
 {string PROTEIN2, {...}},
 ...}
```
where `PROTEINX` is the protein accession, `MAAX` is the modified amino acid at protein position X, and `MODNAME1` is the full label of the modification. For each `MAAX`, there is a `"Total"` key (instead of a modification name) that holds the total intensity of that amino acid measured in the quantified peaks file, including modified and unmodified peptides with that specific residue. 

The general approach is to first get all of the modification intensities and record those in `occupancyDict` while storing in `proteinSeqRangesSeen` a dictionary with protein accession keys and values stored as a list of `(STARTINDEX, ENDINDEX, INTENSITY)` tuples. This helps keep track of the index ranges seen for each protein. Once we have parsed all of the mods, for every amino acid falling into any of those ranges, we increase its `"Total"` intensity by that amount. 

From our discussion, I've added below some of the items I'd like to get some opinions about. Imade them a task list primarily for me to keep track of what I've figured out. 

- [ ] Where should this code live in mzLib. The most reasonable suggestions so far are in `FlashLFQResults` and `Readers/QuantificationResults`. 
- [ ] To interface this nicely with MetaMorpheus, what should the inputs be? My goal now is to look into how/where this will be integrated into MM, but any suggestions on places to look to figure this out are appreciated. 
- [ ] I have some ideas on making the code more efficient/succinct, especially foreseeing a lot more information about the peaks being readily available in MM (like the exact protein index for a peptide/peak). Any new ideas are welcomed. 

Thanks in advance!


mzLib/Test/FileReadingTests/TestPsmFromTsv.cs
```diff
@@ -76,7 +91,104 @@ public static void ReadOGlycoPsmsLocalizedGlycans()
...
+                        {
+                            occupancyDict.Add(protein, new Dictionary<int, Dictionary<string, double>>());
+                            proteinSeqRangesSeen.Add(protein, new List<Tuple<int, int, double>>());
+                        }
+
+                        proteinSeqRangesSeen[protein].Add(new Tuple<int, int, double>(offset, offset + peptidesBase[i].Length, peak.PeakIntensity));
+
+                        // get the localized modifications from the peptide full sequence and add any amino acid/modification combination not
+                        // seen yet to the occupancy dictionary
+                        foreach (KeyValuePair<int, List<string>> aaWithModList in peptideMods)
```
Alexander-Sol, line 162: In situations like this, you can use "var aaWithModList" instead of specifying the actual class

nbollis (commented):
I think readers/Quant... is the best place for it. That way it can be used to find occupancy of the results from another software should that be desired. 

In order to optimize your inputs and outputs of the function, you should break your test method into two. One test method with reads in all the data you need. Another method (not a test method) that gets called to calculate the occupancy. This will help you to better understand what is needed for the method, and for use to help make recommendations

pcruzparri:
Requesting a second round of reviews! The second to last commit contains a little more in detail most changes. Currently pending work is to create a small enough subset of the raw data to create a test similar to the `TestFlashLFQoutputRealData()` test. More rigorous testing can be done with some of the identifications in the vignette data, since some base sequences have enough variations in fullSequence mods and positions to have better case coverage. 

I'd be happy to hear about 1) code optimization, 2) currently written tests, and 3) clarifications on code commenting. In a conversation, Nic suggested using objects for my main ptm calculation code rather than the 5-level deep dictionary, thoughts on that would be useful as well. Ofc, anything else is useful. TIA!

mzLib/MassSpectrometry/Enums/DissociationType.cs
```diff
@@ -109,6 +109,11 @@ public enum DissociationType
         /// </summary>
         LowCID,
 
+        /// <summary>
```
trishorts, line 112: if this is not needed in the PR, let's put it in another and see that such a mechanism has all the fragmentation type associate with it.

mzLib/MzLibUtil/ClassExtensions.cs
```diff
@@ -19,12 +19,96 @@
...
     public static class ClassExtensions
     {
+        /// <summary>
+        /// Parses the full sequence to identify mods.
+        /// </summary>
+        /// <param name="fullSeq"> Full sequence of the peptide in question</param>
+        /// <param name="IncludeNTerminus"> If true, the index of modifications at the N-terminus will be 0 (zero-based indexing). Otherwise, it is the index of the first amino acid (one-based indexing).</param>
+        /// <param name="IncludeCTerminus"> If true, the index of modifications at the C-terminus will be one more than the index of the last amino acid. Otherwise, it is the index of the last amino acid.</param>
+        /// <returns> Dictionary with the key being the amino acid position of the mod and the value being the string representing the mod</returns>
+        public static Dictionary<int, List<string>> ParseModifications(this string fullSeq, bool IncludeNTerminus=false, bool IncludeCTerminus=false)
```
trishorts, line 36: for one full sequence, there can be only 1 mod at each position, so there is no need for the value to be a list.
fullSeq => fullSequence
IncludeNTerminus => modOnNTerminus
=>modOnCTerminus

mzLib/MzLibUtil/ClassExtensions.cs
```diff
@@ -19,12 +19,96 @@
...
+
+            // remove each match after adding to the dict. Otherwise, getting positions
+            // of the modifications will be rather difficult.
+            //int patternMatches = regex.Matches(fullSeq).Count;
+            Dictionary<int, List<string>> modDict = new();
+
+
+            // If there is a missed cleavage, then there will be a label on K and a Label on X modification.
+            // It'll be like [label]|[label] which complicates the positional stuff a little bit. Therefore, 
+            // RemoveSpecialCharacters will remove the "|", to ease things later on.
```
trishorts, line 50: I think we need to handle ambiguity outside of this method. There is no way to ensure that the base sequence in an ambiguous pair is the same, which will complicate matters. for example
PEPT[phospo]IDE | PET[phosph]IDE

mzLib/MzLibUtil/ClassExtensions.cs
```diff
@@ -19,12 +19,96 @@
...
+        /// <summary>
+        /// Fixes an issue where the | appears and throws off the numbering if there are multiple mods on a single amino acid.
+        /// </summary>
+        /// <param name="fullSeq"></param>
+        /// <param name="replacement"></param>
+        /// <param name="specialCharacter"></param>
+        /// <returns></returns>
+        public static void RemoveSpecialCharacters(ref string fullSeq, string replacement = @"", string specialCharacter = @"\|")
+        {
+            // next regex is used in the event that multiple modifications are on a missed cleavage Lysine (K)
```
trishorts, line 107: mods with transition metals will be problematic. suggest:
                          sequence = sequence.Replace("[I]", "");
                            sequence = sequence.Replace("[II]", "");
                            sequence = sequence.Replace("[III]", "");

mzLib/MzLibUtil/ClassExtensions.cs
```diff
@@ -19,12 +19,96 @@
...
+        /// </summary>
+        /// <param name="fullSeq"> Full sequence of the peptide in question</param>
+        /// <param name="IncludeNTerminus"> If true, the index of modifications at the N-terminus will be 0 (zero-based indexing). Otherwise, it is the index of the first amino acid (one-based indexing).</param>
+        /// <param name="IncludeCTerminus"> If true, the index of modifications at the C-terminus will be one more than the index of the last amino acid. Otherwise, it is the index of the last amino acid.</param>
+        /// <returns> Dictionary with the key being the amino acid position of the mod and the value being the string representing the mod</returns>
+        public static Dictionary<int, List<string>> ParseModifications(this string fullSeq, bool IncludeNTerminus=false, bool IncludeCTerminus=false)
+        {
+            // use a regex to get all modifications
+            string pattern = @"\[(.+?)\]";
+            Regex regex = new(pattern);
```
trishorts, line 40: we need to make sure that this method never thinks that
[hydroxylation]EPT[phospho] is accidentaly identified as a mod for P[hydroxylation]EPT[phospho]IDE
I'm not sure that ]EPT[ won't be ignored by your regex

mzLib/MzLibUtil/PositionFrequencyAnalysis.cs
```diff
@@ -0,0 +1,166 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Easy.Common.Extensions;
+
+namespace MzLibUtil
+{
+    // Should this have all of the parent data (i.e. protein group, protein, peptide, peptide position)? Unnecessary for now, but probably useful later.
+    public class UtilModification
```
trishorts, line 9: UtilModification => LocalizedModificationFromTsv
modName => IdWithMotif
position =>PeptidePositionZeroIsNterminus

mzLib/MzLibUtil/PositionFrequencyAnalysis.cs
```diff
@@ -0,0 +1,166 @@
...
+            Position = position;
+            Intensity = intensity;
+        }
+
+    }
+    public class UtilPeptide
+    {
+        public string FullSequence { get; set; }
+        public string BaseSequence { get; set; }
+        public UtilProtein ParentProtein { get; set; }
```
trishorts, line 27: maybe this should be ProteinGroup?

trishorts:
it's possible that "Identification" should be the currency of this realm as it is what is passed into flashlfq by MM and it is what is generated by FlashLFQ when it is run alone on any acceptable input.
image

mzLib/MzLibUtil/ClassExtensions.cs
pcruzparri: removed RemoveSpecialCharacters use in the ParseModifications logic since obsolete.

mzLib/MzLibUtil/ClassExtensions.cs
pcruzparri: Updated the bracket match logic to ignore a closing bracket if it matches the pattern for a metal ion modification, namely `"\[I+\]"`. New mod search pattern is  `@"\[(.+?)\](?<!\[I+\])"`

mzLib/MzLibUtil/ClassExtensions.cs
pcruzparri: After finding an opening bracket, regex will always find the next closing bracket, except (updated now) in the case where the closing bracket belongs to an ion charge state.

mzLib/MzLibUtil/ClassExtensions.cs
pcruzparri: renamed

pcruzparri:
Some pending changes:
Potentially modify the input to ProteinGroupsOccupancyByPeptide() be some object and create an interface rather than a list. 
Remove occupancy additions to FlashLFQ. These additions have been moved to MetaMorpheus.
Create new package for modification quantification within MzLib. 

Side Note: 
This branch/pr is a mess. I made a new clean branch on my fork from the master branch and cherry-picked the important, recent commits from this branch. I can change the pull request to track that branch unless there is opposition.

PR #800, opened 2024-09-15
Tims tof reader

Alexander-Sol:
Added support for timsTOF data files (.d)

TimsTofFileReader is a child class of MsDataFile. Currently, only DDA PASEF data is supported.

Dynamic loading is not supported at this time


Alexander-Sol:
replace by pr #812

PR #803, opened 2024-10-08
Eliminate a potential crash in ResultFile

nbollis:
#### PR Classification
Code enhancement to improve null handling and file existence checks in the readers base ResultFile class.

#### PR Motivation
If the Results is not populated at time of calling, the LoadResults method is called.  This would lead to a FileNotFoundException when the file does not exist.

#### PR Summary
Improved null handling in `ResultFile` and added file existence checks in `Ms1FeatureFile`, along with new tests.
- `ResultFile.cs`: Made `_results` nullable, updated `Results` property logic, and added comments.
- `Ms1FeatureFile.cs`: Added file existence check and set `Software` property to `Unspecified` if file is missing.
- `TestResultFile.cs`: Removed unnecessary `using` directives and added tests for file existence scenarios.


mzLib/Readers/ExternalResults/BaseClasses/ResultFile.cs
```diff
@@ -12,14 +12,26 @@ public abstract class ResultFile<TResult> : IResultFile, IEquatable<ResultFile<T
...
         {
             get
             {
-                if (!_results.Any())
+                // if the results have already been read in or set
+                if (_results is not null) 
+                    return _results;
+
+                // if the results have not been read, and the file exists, read the results
+                if (File.Exists(FilePath))
```
Alexander-Sol, line 25: It seems like just adding this line would be enough?

Alexander-Sol: Like, if (!_results.Any() & File.Exists(FilePath) ...
That would accomplish the same thing with fewer changes

mzLib/Readers/ExternalResults/BaseClasses/ResultFile.cs
```diff
@@ -31,7 +43,7 @@ public List<TResult> Results
         protected ResultFile(string filePath, Software software = Software.Unspecified)
         {
             FilePath = filePath;
-            _results = new List<TResult>();
+            _results = null;
```
Alexander-Sol, line 46: I don't understand the rationale behind this change

mzLib/Readers/ExternalResults/BaseClasses/ResultFile.cs
nbollis: This is because of a specific case. The file exists, but there is nothing besides the header. In that case, the file will exist, the _results will be empty, and the LoadResults() method would be called, leading to the same crash.

mzLib/Readers/ExternalResults/BaseClasses/ResultFile.cs
nbollis: see above

PR #824, opened 2025-01-14
Test digest two xmls for same protein

trishorts:
the order of peptides digested from a protein depend on the order in which localized modifications are encoded in the xml. This is particularly problematic when the number of modified forms of a peptide is limited to a specific number (e.g. take the first 1024). 

This PR adds code that sorts all protein onebasedlocalizedmodifications first by the position and then by the modification. This prevents variations in the xml encoded order from influencing the resultant peptides in the digest.

This PR also includes a test, where the modifications are present in different orders. This test will fail without the sort. But since the sort is added, the failure is eliminated and the test passes.

### DIFF
--- a/mzLib/Omics/Modifications/Modification.cs
+++ b/mzLib/Omics/Modifications/Modification.cs
@@ -13,7 +13,7 @@ namespace Omics.Modifications
     /// Represents a modification
     /// Mods.txt format was taken from https://ftp.uniprot.org/pub/databases/uniprot/current_release/knowledgebase/complete/docs/ptmlist.txt
     /// </summary>
-    public class Modification : IComparable<Modification>
+    public class Modification
     {
         public string IdWithMotif { get; private set; }
         public string OriginalId { get; private set; }
@@ -298,18 +298,6 @@ namespace Omics.Modifications
             sb.Append("#This modification can be found in file " + this.FileOrigin);
 
             return sb.ToString();
-        }
-        public int CompareTo(Modification other)
-        {
-            if (other == null) return 1;
-
-            int idComparison = string.Compare(this.IdWithMotif, other.IdWithMotif, StringComparison.Ordinal);
-            if (idComparison != 0)
-            {
-                return idComparison;
-            }
-
-            return Nullable.Compare(this.MonoisotopicMass, other.MonoisotopicMass);
         }
     }
 }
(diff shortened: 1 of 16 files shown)


mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
```diff
@@ -97,48 +97,51 @@ internal IEnumerable<PeptideWithSetModifications> GetModifiedPeptides(IEnumerabl
             }
 
             // LOCALIZED MODS
-            foreach (var kvp in Protein.OneBasedPossibleLocalizedModifications)
+            var localizedModsOrderedByPositionThenByModification = Protein.OneBasedPossibleLocalizedModifications
+                .SelectMany(kvp => kvp.Value.Select(value => (kvp.Key, value)))
+                .OrderBy(tuple => tuple.Key)
+                .ThenBy(tuple => tuple.value)
```
Alexander-Sol, line 103: Given how many times this gets called, I think it would be better to use a custom comparer, that way we only have to iterate through the list once.

Additionally, I'm not sure that it makes sense to store this as a list, since we're only iterating through it once. We could just put the linq expression in the foreach

trishorts: updated according to your comments

mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
```diff
@@ -97,12 +97,10 @@ internal IEnumerable<PeptideWithSetModifications> GetModifiedPeptides(IEnumerabl
             }
 
             // LOCALIZED MODS
-            var localizedModsOrderedByPositionThenByModification = Protein.OneBasedPossibleLocalizedModifications
+            foreach (var localizedMod in Protein.OneBasedPossibleLocalizedModifications
                 .SelectMany(kvp => kvp.Value.Select(value => (kvp.Key, value)))
                 .OrderBy(tuple => tuple.Key)
-                .ThenBy(tuple => tuple.value)
-                .ToList();
-            foreach (var localizedMod in localizedModsOrderedByPositionThenByModification)
```
Alexander-Sol: Sorry, I meant a custom comparer that considers the tuple's key and value simultaneously. So there would only be an OrderBy statement with a custom comparer, as opposed to an OrderBy().ThenBy().

mzLib/Omics/Modifications/Modification.cs
```diff
@@ -299,17 +299,5 @@ public string ModificationErrorsToString() //reports errors in required fields.
...
-            if (other == null) return 1;
-
-            int idComparison = string.Compare(this.IdWithMotif, other.IdWithMotif, StringComparison.Ordinal);
-            if (idComparison != 0)
-            {
-                return idComparison;
-            }
-
-            return Nullable.Compare(this.MonoisotopicMass, other.MonoisotopicMass);
-        }
```
Alexander-Sol: I thought this was fine

PR #826, opened 2025-01-21
Changed rounding method

Alexander-Sol:
As it turns out, there is a very good reason that Math.Round() uses MidpointRounding.RoundToEven as the default method of rounding. For more information, see here:n [https://learn.microsoft.com/en-us/dotnet/api/system.math.round?view=net-9.0](https://learn.microsoft.com/en-us/dotnet/api/system.math.round?view=net-9.0)

This PR amends the mzML writer rounding PR (#821) to use the default rounding method.


trishorts (approved):

nbollis (approved):

trishorts (changes requested):
some itraq failures.

nbollis:
> some itraq failures.

Those are my bad. They are fixed by my PR

PR #827, opened 2025-01-21
Put those mods back where they came from or so help me

nbollis:
My last PR made an incorrect assumption: modification localization restriction "peptide N-terminal" is meant to refer to ANY N-term, not just those that are generated during digestion. 

In the last digestion PR, I made peptide N-terminal mod and N-terminal modification localization mutually exclusive.
This incorrect change results in no isobaric quantification labels being added to the peptide which contains the protein terminus. 

**The only integration test failing is due to the calibration change. This PR is safe**

#### Changes
- `DigestionProduct.cs`: Ensured correct application and management of protease-associated and terminal modifications. By maintining current overwriting behavior, but ensuring that protein n-term overwrites peptide and not the other way around when applicable. 
- `ModificationLocalization.cs`: "Peptide N-terminal." mods are now valid on ALL N-termini. 
- `TestProteinDigestion.cs`: Updated and added test cases to reflect new terminal modification logic.
- `TestDigestion.cs`: Adjusted tests for expected digestion products and sequences with terminal modifications.


Alexander-Sol:
Does this only apply to the behaviour of fixed mods? If you have peptide-N-term and protein-N-term variable mods, does one take precedence?

trishorts (approved):

Alexander-Sol (approved):

PR #835, opened 2025-02-12
Changed a few methods in FlashLFQ to work with DIA

zhuoxinshi:
Added an overload GetIndexedPeak method that uses mz instead of neutral mass and charge as input to find a peak from a specific scan
Added GetXIC in FlashLfqEngine as a static method to find XIC for a spectral peak


mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -1760,45 +1758,64 @@ public List<IndexedMassSpectralPeak> Peakfind(double idRetentionTime, double mas
...
                 {
                     missedScans++;
                 }
                 else if (peak != null)
                 {
                     missedScans = 0;
                     xic.Add(peak);
+
+                    if (xic.First().RetentionTime - peak.RetentionTime > maxRT)
+                    {
```
nbollis, line 1813: Is maxRT a limit to the distance from the point we start looking or maximum retention time? If the first, use a more informative name. If the second, this case is impossible.

mzLib/FlashLFQ/PeakIndexingEngine.cs
```diff
@@ -162,10 +166,7 @@ public IndexedMassSpectralPeak GetIndexedPeak(double theorMass, int zeroBasedSca
                             break;
                         }
 
-                        double expMass = peak.Mz.ToMass(chargeState);
-
-                        if (tolerance.Within(expMass, theorMass) && peak.ZeroBasedMs1ScanIndex == zeroBasedScanIndex
-                            && (bestPeak == null || Math.Abs(expMass - theorMass) < Math.Abs(bestPeak.Mz.ToMass(chargeState) - theorMass)))
+                        if (tolerance.Within(peak.Mz, mz) && peak.ZeroBasedMs1ScanIndex == zeroBasedScanIndex && (bestPeak == null || Math.Abs(peak.Mz - mz) < Math.Abs(bestPeak.Mz - mz)))
```
nbollis, line 169: Is this a stable change? Difference in mass have a different magnitude in mz

mzLib/TestFlashLFQ/TestFlashLFQ.cs
```diff
@@ -1387,8 +1387,8 @@ public static void RealDataMbrTest()
             // Any change to ML.NET or the PEP Analysis engine will cause these to change.
             Console.WriteLine("r1 PIP event count: " + f1r1MbrResults.Count);
             Console.WriteLine("r2 PIP event count: " + f1r2MbrResults.Count);
-            Assert.AreEqual(138, f1r1MbrResults.Count);
-            Assert.AreEqual(70, f1r2MbrResults.Count);
+            Assert.AreEqual(137, f1r1MbrResults.Count);
+            Assert.AreEqual(76, f1r2MbrResults.Count);
```
nbollis, line 1392: I am scared. Why did changing no functional parts of the code, but only the entry points, change the results?

nbollis (changes requested):

mzLib/MassSpectrometry/MsDataScan.cs
```diff
@@ -270,6 +270,11 @@ public void SetOneBasedScanNumber(int value)
             this.OneBasedScanNumber = value;
         }
 
+        public void SetMsnOrder(int value)
```
Alexander-Sol, line 273: What's the reason for this change? I think it makes sense that the order should only be set on initialization

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -1743,8 +1743,6 @@ public bool CheckIsotopicEnvelopeCorrelation(
         /// <returns></returns>
         public List<IndexedMassSpectralPeak> Peakfind(double idRetentionTime, double mass, int charge, SpectraFileInfo spectraFileInfo, Tolerance tolerance)
```
Alexander-Sol, line 1744: Should we rename this to GetXIC as well?

Alexander-Sol (changes requested):
Great changes! Now you just need some tests, especially for the maxRt argument of GetXIC

mzLib/FlashLFQ/FlashLfqEngine.cs
Alexander-Sol: I think the second case is intended, and it is possible here. In this second loop, we're iterating backwards in time. So the clause subtracts RT of the current point from the RT of the middle point of the XIC

Alexander-Sol (approved):

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -1758,45 +1753,74 @@ public List<IndexedMassSpectralPeak> Peakfind(double idRetentionTime, double mas
...
+
+        /// <summary>
+        /// A generic method of peak tracing across the retention time. Finds peaks with a given mz that occur on either side of a given
+        /// retention time. Peak searching iterates backwards through the scans until the peak 
+        /// is no longer observed (i.e., is absent in more scans than allowed, as defined by the
+        /// missedScansAllowed parameter. Missed scans don't have to be sequential. The same procedure
+        /// is then repeated in the forward direction.
+        /// </summary>
+        /// <param name="zeroBasedStartIndex"> the scan where peak searching behaviour begins </param>
+        /// <param name="maxRT"> the maximum distance from the apex RT of the XIC to both start RT and end RT </param>
```
trishorts, line 1769: this sounds like peak width. i don't know what maximum retention time means.

mzLib/FlashLFQ/PeakIndexingEngine.cs
```diff
@@ -37,25 +42,49 @@ public bool IndexMassSpectralPeaks(SpectraFileInfo fileInfo, bool silent, Dictio
...
+                }
+
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Read in all spectral peaks from scans, index the peaks and store them in a list ordered by m/z
```
trishorts, line 77: this explanation is opaque to me. I can help you rephrase it.

mzLib/FlashLFQ/PeakIndexingEngine.cs
```diff
@@ -37,25 +42,49 @@ public bool IndexMassSpectralPeaks(SpectraFileInfo fileInfo, bool silent, Dictio
...
             {
                 _indexedPeaks = new List<IndexedMassSpectralPeak>[0];
                 return false;
             }
 
+            PeakIndexing(msDataScans, out List<Ms1ScanInfo> scanInfo);
+
+            _ms1Scans.Add(fileInfo, scanInfo.ToArray());
+
+            if (_indexedPeaks == null || _indexedPeaks.Length == 0)
```
nbollis, line 63: None of the error handling is tested, should I be concerned?

trishorts (approved):

Alexander-Sol (approved):

PR #839, opened 2025-03-10
Updates to Full Sequence Writing and Parsing of Mods

pcruzparri:
Purpose:
1) ParseModifications was bugged in that it could output negative indices. Updated the regex pattern also to explicitly ignore cation brackets (i.e. [Metal: Calcium[II] on E] will ignore the first closing bracket, since it is not the end of the mod representation). 
2) Our current string representation of full sequences does not differentiate between a C-terminus mod and a mod on the side chain of the C-terminus amino acid. 

Once these changes are positively reviewed, I'll update some of the remaining tests not passing, since they are more tedious to update.

First commit message:
Changed the BioPolymerWithSetModsExtensions class to write full sequences separating the C-terminus with a dash. Updated some of the tests that failed because of the new notation of C-terminus mods. Some tests are still failing, and will be updated once happy with this general change.


pcruzparri:
The only failing integration error is the following:
```Error: D:\a\mzLib\mzLib\MetaMorpheus\MetaMorpheus\EngineLayer\Silac\SilacConversions.cs(659,130): error CS1061: 'IIndexedMzPeak' does not contain a definition for 'ZeroBasedMs1ScanIndex' and no accessible extension method 'ZeroBasedMs1ScanIndex' accepting a first argument of type 'IIndexedMzPeak' could be found (are you missing a using directive or an assembly reference?)```

The IIndexedMzPeak interface renamed the field `ZeroBasedMs1ScanIndex` to `ZeroBasedScanIndex` but the change has not been made on MetaMorpheus. Will open a small PR to fix that.

nbollis (dismissed):

Alexander-Sol:
> The only failing integration error is the following: `Error: D:\a\mzLib\mzLib\MetaMorpheus\MetaMorpheus\EngineLayer\Silac\SilacConversions.cs(659,130): error CS1061: 'IIndexedMzPeak' does not contain a definition for 'ZeroBasedMs1ScanIndex' and no accessible extension method 'ZeroBasedMs1ScanIndex' accepting a first argument of type 'IIndexedMzPeak' could be found (are you missing a using directive or an assembly reference?)`
> 
> The IIndexedMzPeak interface renamed the field `ZeroBasedMs1ScanIndex` to `ZeroBasedScanIndex` but the change has not been made on MetaMorpheus. Will open a small PR to fix that. Edit: PR #2494 on MetaMorpheus.

I'm responsible for that test breaking. We need to put out a mzLib release before we fix it in MetaMorpheus. When we release, I'll fix that issue in MM

trishorts (dismissed):

nbollis (approved):
lgtm

trishorts (approved):
Re-approving. My approval on `d7f8c4f4` was dismissed as stale; the only commit since is `6678f541`, which deletes the unused WIP `ParseModificationsWithAmbiguity` and the two usings it needed. That strictly narrows what I approved — nothing was added or changed — so the approval still stands on the current head.

PR #857, opened 2025-04-03
Results reading FlashLQ

mzhastings:
Edited the MzLibExtensions class so that the spectra file info used to make an identification is no longer arbitrary. 

**Classes**

- MzLibExtensions contains a MakeIdentification method that takes in a IQuantifiableResultFile and a list of spectra files to return an identification
    - new addition of a helper method that takes in the same arguments and returns a dictionary that maps each spectra file to the file name associated with each identified protein


mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
```diff
@@ -53,12 +51,51 @@ public static List<Identification> MakeIdentifications(this IQuantifiableResultF
...
+        private static Dictionary<string, SpectraFileInfo> makeSpectraFileDict(this IQuantifiableResultFile quantifiable, List<SpectraFileInfo> spectraFiles)
+        {
+            Dictionary<string, SpectraFileInfo> allSpectraFiles = new Dictionary<string, SpectraFileInfo>();
+
+            // 1. from list of SFIs create a list of strings that contains each full file path w/ extension
+            List<string> fullFilePaths = new List<string>();
+            foreach (SpectraFileInfo spectraFileInfo in spectraFiles)
+            {
+                fullFilePaths.Add(spectraFileInfo.FullFilePathWithExtension);
+            }
```
Alexander-Sol, line 70: It looks like this is unneccesary/unused

mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
```diff
@@ -53,12 +51,51 @@ public static List<Identification> MakeIdentifications(this IQuantifiableResultF
...
+            foreach (var file in allFiles)
+            {
+                string key = file.Key;
+                string filePath = file.Value;
+                // FirstOrDefault matches the 1st elt from spectraFiles w/ specified filePath
+                SpectraFileInfo? matchingSpectraFile = spectraFiles.FirstOrDefault(spectraFileInfo => spectraFileInfo.FullFilePathWithExtension == filePath);
+
+                if (matchingSpectraFile != null)
+                {
+                    if (allSpectraFiles.TryGetValue(key, out var existingSpectraFile))
```
Alexander-Sol, line 86: This can be simplified. Just check if allSpectraFiles contains the key. If not, add the key/value. If it already contains it, don't do anything. Can be combined into one if statement

mzLib/TestFlashLFQ/TestIdentificationAdapter.cs
```diff
@@ -86,5 +87,130 @@ public void TestFileNametoFilePathLocalPath(string path)
...
+
+            // Assert
+            Assert.IsEmpty(result);
+        }
+
+        [Test]
+        public void MakeIdentifications_ShouldReturnIdentifications_WhenQuantifiableRecordsExist()
+        {
+            // Arrange
+            var quantifiableRecords = new List<IQuantifiableRecord>
```
Alexander-Sol, line 109: This test doesn't actually check the SpectraFileInfo

Alexander-Sol (changes requested):
Address comments and add more tests

mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
mzhastings: This is needed to create the allFiles dictionary, which takes in a list of fullFilePaths. The allFiles dictionary is used to create the allSpectraFiles Dictionary since the key needs to be the same between the two

Alexander-Sol (approved):

RayMSMS (approved):

PR #861, opened 2025-04-08
FlashLFQ Refactor: Part 2

Alexander-Sol:
This PR is part 2 of a broader refactor of FlashLFQ.

### New Abstract Class
 - IndexingEngine: IndexingEngine contains generic functions designed to index peaks, retrieve individual indexed peaks, and retrieve Xics, or lists of indexed peaks over time.
      - GetIndexedPeak() Method to enable the retrieval of an IIndexedMzPeak with a specific mass and retention time
      - GetXic() Method to enable the retrieval of a list of IIndexedMzPeak. Previously, this function was a static method in the FlashLFQEngine. This method was refactored slightly to improve performance

### PeakIndexingEngine Changes
- PeakIndexingEngines can now only be constructed via one factory methods. Three new factory methods were implemented, all of which are called InitializeIndexingEngine. They take different arguments: SpectraFileInfo, MsDataFile, or MsDataScan[]

### New Interface
- IFlashLfqIndexingEngine: FlashLFQ requires that indexed peaks be serializable, and there is a spectraFileInfo associated with the IndexingEngine. This interface guarantees that behaviour. When timsTof support is added, the new timsIndexinEngine will also implement this interface

### Other changes
 - GetXIC methods removed from FlashLFQEngine, as they're now redundant
 - IIndexingEngine interface removed from FlashLFQ as it was made reundant


mzLib/FlashLFQ/PeakIndexing/Interfaces/IFlashLfqIndexingEngine.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FlashLFQ.Interfaces
+{
+    /// <summary>
+    /// FlashLFQ requires indexing engines to IndexPeaks, serialize and deserialize indexed peaks,
+    /// and retrieve specific peaks (m/z, index pair) or xics (m/z at and around a given retention time
+    /// </summary>
+    public interface IFlashLfqIndexingEngine
```
nbollis, line 15: Comments on methods would be lovely. Summary comments on interface methods will pop up in Visual Studio when mousing over the method called on a concrete type. This ensure you do not need to make comments on every implementing IFlashLfqIndexingEngine

nbollis (approved):

nbollis (dismissed):

nbollis (approved):

PR #864, opened 2025-04-08
Reader extensions

Alexander-Sol:
This PR adds two new factory methods to the FileReader class
- ReadResultFile: Takes in a file path and returns an IResultFile or throws an exception if the file can't be converted to the IResultFile interface
- ReadQuantifiableResultFile: Takes in a file path and returns an IResultFile or throws an exception if the file can't be converted to the IQuantifiableResultFile interface


mzLib/Readers/ExternalResults/ResultFiles/MsPathFinderTResultFile.cs
```diff
@@ -10,18 +10,12 @@ namespace Readers
 {
     public class MsPathFinderTResultFile : ResultFile<MsPathFinderTResult>, IResultFile
     {
-        public override SupportedFileType FileType { get; }
+        public override SupportedFileType FileType => FilePath.ParseFileType();
```
nbollis, line 13: Problematic. In any solution which uses mzlib, we could construct a ResutlFile<T> from any record whatsoever. This would not have a supported file type, but retain all other result file operations. 

With this change, we would get an exception exception thrown frequently

nbollis (approved):

mzhastings (approved):

PR #866, opened 2025-04-10
Added intensity field to Sm from TSV

nbollis:
This was added in the MM version, but not the Mzlib


Alexander-Sol (dismissed):

mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
```diff
@@ -29,12 +29,11 @@ public override MsDataScan GetOneBasedScanFromDynamicConnection(int oneBasedScan
         #region IResultFile Members
```
nbollis, line 29: Changes in this class enable MsDataFiles to be loaded through Alex's new FileReader methods

RayMSMS (approved):

PR #868, opened 2025-04-11
MsAlign Mass Spec File Reading

nbollis:
Added support to read MsAlign files. They contain neutral mass and charge of MS2 ions and the mass that was selected for fragmentation, just like a .mgf mass spec file.


mzLib/Readers/MsAlign/MsAlign.cs
```diff
@@ -0,0 +1,561 @@
...
+            charges[i] = int.Parse(splits[2]);
+            monoMasses[i] = double.Parse(splits[0]);
+            intensities[i] = double.Parse(splits[1]);
+            mzs[i] = monoMasses[i].ToMz(charges[i]);
+        }
+
+
+
+        double minmz = mzs.Length == 0 ? 0 : mzs.Min();
+        double maxmz = mzs.Length == 0 ? 2000 : mzs.Max();
```
Alexander-Sol, line 473: 0 for minmz seems too low for the default value

Alexander-Sol (approved):

RayMSMS (approved):

PR #871, opened 2025-04-22
Edited MSFragger Reader

mzhastings:
Made changes to MSFraggerPsm class to fix errors when reading psm.tsv files returned by MSFragger. 
- Added a new property RetentionTimeSeconds to map the raw "Retention" column from the psm.tsv file (reported in seconds).
- Modified the existing RetentionTime property to be computed from RetentionTimeSeconds, converting it to minutes
- Changed FileName property to use FileNameWithoutExtension instead of SpectrumFilePath (points to a .pep.xml file created by MSFragger, which appends the file name with interact-). The list of SpectrumFilePaths was thus inconsistent with the raw file names used in the dictionary mapping file names to their paths

Made changes to SpectrumMatchFromTsv class
- Added support for optional PEP and PEP_QValue columns for compatibility with different versions of MetaMorpheus that may or may not include these columns


Alexander-Sol (changes requested):
You can get rid of the SmallCalibratibleYeast.mzml file, as it's not actually used for the tests

Alexander-Sol (dismissed):

mzLib/mzLib.nuspec
```diff
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>mzLib</id>
-    <version>1.0.559</version>
+    <version>2.0.7</version>
```
nbollis, line 5: Fix versioning in nuspec

nbollis (changes requested):

trishorts (dismissed):

PR #872, opened 2025-04-24
Restricted mods

Avni2000:
Implementation-less (in protein, probably) method of generating the number of possible proteoforms given a max number of residue sites. Draft mostly because it's untested.

### DIFF
--- a/mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
+++ b/mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
@@ -87,13 +87,16 @@ namespace Proteomics.ProteolyticDigestion
 
         internal long numPossibleProteoforms(List<Modification> allKnownFixedModifications,
         DigestionParams digestionParams, List<Modification> variableModifications)
+
         {
-            var iterator = GetModifiedPeptides(allKnownFixedModifications, digestionParams, variableModifications);
+            int peptideLength = OneBasedEndResidue - OneBasedStartResidue + 1;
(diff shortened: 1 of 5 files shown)


mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
```diff
@@ -83,5 +84,22 @@ internal IEnumerable<PeptideWithSetModifications> GetModifiedPeptides(List<Modif
                 DictionaryPool.Return(twoBasedPossibleVariableAndLocalizeableModifications);
             }
         }
+
+        internal long numPossibleProteoforms(List<Modification> allKnownFixedModifications,
+        DigestionParams digestionParams, List<Modification> variableModifications)
+
```
Alexander-Sol, line 90: Instead of taking digestion params, it can just take an int paramater called "MaxNumberOfModifiedSites". DigestionParams.MaxModsForPeptide() and DigestionParams.MaxMods are equal

mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
```diff
@@ -83,5 +84,22 @@ internal IEnumerable<PeptideWithSetModifications> GetModifiedPeptides(List<Modif
                 DictionaryPool.Return(twoBasedPossibleVariableAndLocalizeableModifications);
             }
         }
+
+        internal long numPossibleProteoforms(List<Modification> allKnownFixedModifications,
```
Alexander-Sol, line 88: This feels like it will be the entry point. This method should be public and well commented

Avni2000:
Added new commit!

PR #873, opened 2025-04-28
IsoTracker: Change searching target

RayMSMS:
The origin searching target is **"the peptide with same base seq and same mass**"
But the definition of an Isobaric peptide is "**whatever peptide with the same mass**"
We changed the search target

### DIFF
--- a/mzLib/FlashLFQ/FlashLfqEngine.cs
+++ b/mzLib/FlashLFQ/FlashLfqEngine.cs
@@ -1959,20 +1959,39 @@ namespace FlashLFQ
             int isoGroupsSearched = 0;
             double lastReportedProgress = 0;
             double currentProgress = 0;
+            List<IsobaricPeptide> isobaricPeptides = new List<IsobaricPeptide>();
+            var ids = _allIdentifications.OrderBy(p => p.PeakfindingMass).ToList();
+            for (int i = 0; i < ids.Count; i++)
+            {
+                Identification id = ids[i];
+                IsobaricPeptide isobaricPeptide = new IsobaricPeptide(id, "10 ppm");
+                ids.Remove(id);
 
-            var idGroupedBySeq = _allIdentifications
-                .Where(p => p.BaseSequence != p.ModifiedSequence && !p.IsDecoy)
-                .GroupBy(p => new
-                    { p.BaseSequence, MonoisotopicMassGroup = Math.Round(p.MonoisotopicMass / 0.0001) })
-                .ToList();
+                for (int j = 0; j < ids.Count; j++)
+                {
+                    Identification id2 = ids[j];
+                    if (id2.PeakfindingMass < isobaricPeptide.MaxMass)
+                    {
+                        isobaricPeptide.Ids.Add(id2);
+                    }
+                    else break;
+                }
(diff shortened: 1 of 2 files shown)


Alexander-Sol:
A few comments:
 - Can you change IsobaricPeptide to inherit from Peptide? Then you can move the PeakOrder and IsoGroupIndex properties to IsobaricPeptide
 - If you make that change, I don't like having the ppm argument in the constructor. That property feels unneccessary, and it means we're creating and disposing a Tolerance object every time an IsobaricPeptide is created
 - If you use any object as a key in a dictionary, you have to make sure that class overrides Equals and GetHashCode

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -17,7 +17,7 @@ public class FlashLfqResults
         public readonly Dictionary<string, ProteinGroup> ProteinGroups;
         public readonly Dictionary<SpectraFileInfo, List<ChromatographicPeak>> Peaks;
         private readonly HashSet<string> _peptideModifiedSequencesToQuantify;
-        public IDictionary<string, Dictionary<PeakRegion, List<ChromatographicPeak>>> IsobaricPeptideDict = null;
+        public IDictionary<PeptideMassBin, Dictionary<PeakRegion, List<ChromatographicPeak>>> IsobaricPeptideDict = null;
```
trishorts, line 20: I think this should be PeptideMzBin unless it is deconvoluted neutral mass. In that case, Mass is fine.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -2061,6 +2059,38 @@ private void QuantifyIsobaricPeaks()
                 Console.WriteLine("Finished quantifying isobaric species!");
         }
 
+        /// <summary>
+        /// Sort ids in the IDList and classify into isobaricPeptide by their mass.
+        /// </summary>
+        internal List<PeptideMassBin> SortIsobaricPeptide(List<Identification> _allIdentifications, Tolerance massTolerance)
+        {
+            List<PeptideMassBin> isobaricPeptides = new List<PeptideMassBin>();
+            var ids = _allIdentifications.Where(p => p.BaseSequence != p.ModifiedSequence).OrderBy(p => p.PeakfindingMass).ToList();
```
trishorts, line 2068: can you add a comment explaining this restriction

mzLib/FlashLFQ/IsoTracker/PeptideMassBin.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// This class is used to group isobaric peptides together based on their mass.
+    /// </summary>
+    public class PeptideMassBin
+    {
+        public double MaxMass { get; private set; } // The mass of the isobaric peptide
+        public List<Identification> Ids { get; set; } // The identification of the isobaric peptide
+
+        public PeptideMassBin(Identification id, Tolerance tolerance)
```
trishorts, line 15: i think you should specify the kind of tolerance (ppm or mass)

trishorts (changes requested):

PR #880, opened 2025-05-07
Isotracker: Acceptable peak width for searching

RayMSMS:
In Isotracker, we will start searching from the peak region. However, too many peaks are identified and reported in the output. In order to eliminate the noise report, we create two parameters to set up the acceptable peak width to control the resulting quality.

-  **AcceptablePeakWidth :** any peak region over the value can be searched for isobaric peaks.


nbollis (dismissed):

trishorts:
Merged master and pushed — mergeable again, and the FlashLFQ + IsoTracker suite is 117/117 locally.

**The conflicts** were all adjacency. Master added `IsoTrackerIdFilter` and `RequireMultipleIdsInOneFiles` to the same constructor parameter list, the same initializer and the same property block this branch added `AcceptablePeakWidth` to. Kept both sides throughout, and gave `AcceptablePeakWidth` the XML summary its neighbours have.

**One thing you should know before this merges.** The change this PR exists to make is not covered by any test.

It raises the shared-peak filter from `peakWindow > 0.001` to `peakWindow > FlashParams.AcceptablePeakWidth`, default `0.4` — a 400x stricter threshold that will drop narrow shared peaks from isobaric searching. That is a deliberate, results-affecting change, and it is the right kind of thing to make configurable.

But I verified it is untested: with the default set back to `0.001`, the IsoTracker suite still passes **27/27**, identically. So no fixture exercises a shared peak with a width between 0.001 and 0.4 min, which means nothing here would catch a regression in either direction, and nothing demonstrates the filter does what it is meant to.

Worth adding a test that builds an XIC group with a deliberately narrow shared peak and asserts it is searched at `0.001` and skipped at `0.4`. Happy to write it if you'd like.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -1891,7 +1893,7 @@ private void QuantifyIsobaricPeaks()
                             foreach (var peak in xICGroups.SharedPeaks)
                             {
                                 double peakWindow = peak.Width;
-                                if (peakWindow > 0.001) //make sure we have enough length of the window (0.001 min) for peak searching
+                                if (peakWindow > FlashParams.AcceptablePeakWidth) //make sure we have enough length of the window (default is 0.4 min) for peak searching
```
trishorts, line 1896: **The new setting can only ever tighten the filter â€” it can never loosen it below 0.3 min.**

`QuantifyIsobaricPeaks` still builds the group with `new XICGroups(xicGroup)` (line 1884), and the `XICGroups` constructor calls `SharedPeaks = BuildSharedPeaks();` with no argument (`XICGroups.cs:62`). That runs `sharedPeaks.RemoveAll(p => p.Width < windowLimit)` with `windowLimit`'s hard-coded default of `0.3` (`XICGroups.cs:208`, `:247`) *before* the engine ever sees a region. `PeakRegion.Width` is a computed `EndRT - StartRT` with `private set` on both fields, so nothing between those two points can shrink a region.

So the effective threshold is `max(0.3, AcceptablePeakWidth)`, not the configured value:

- `AcceptablePeakWidth = 0.1` (deliberately widening the search) â†’ still loses every region below 0.3 min; the setting looks broken.
- `AcceptablePeakWidth = 0` (opting out to recover pre-PR behaviour) â†’ same; there is no way back to the base `> 0.001` behaviour through the parameter this PR adds.

Note also that the `XICGroups` constructor (`XICGroups.cs:44`) exposes `sharedPeakThreshold`, `tolerance` and `cutOff` but deliberately *not* `windowLimit`, so a caller cannot reach it even by hand.

Suggested fix: thread the value end to end â€” add an acceptable-width parameter to the `XICGroups` constructor (defaulted to today's `0.3` so existing call sites are unaffected), forward it explicitly to `BuildSharedPeaks(...)`, pass `FlashParams.AcceptablePeakWidth` at the `new XICGroups(xicGroup)` site, and then drop this duplicate engine-side check so one threshold governs the filter.

mzLib/FlashLFQ/FlashLfqParameters.cs
```diff
@@ -59,6 +60,11 @@ public FlashLfqParameters()
         //IsoTracker settings
         public bool IsoTracker { get; set; }
         /// <summary>
+        /// The shortest a shared peak may be, in minutes, and still be searched for isobaric species. Peaks
+        /// narrower than this carry too little of the elution to align across runs.
+        /// </summary>
+        public double AcceptablePeakWidth { get; set; }
```
trishorts, line 66: Two small things on the doc comment, both worth fixing while the property is still new and has no serialized consumers:

**1. The stated bound is inclusive; the code is strict.** "The shortest a shared peak may be â€¦ and still be searched" reads as `>=`, but the filter is `peakWindow > FlashParams.AcceptablePeakWidth` (`FlashLfqEngine.cs:1896`), so a region exactly this wide is rejected. The PR description's own wording â€” "any peak region **over** the value" â€” is the accurate one.

Worth noting the two width filters currently disagree on the boundary in opposite directions: `BuildSharedPeaks` uses `RemoveAll(p => p.Width < windowLimit)`, which *keeps* a region exactly at the limit. Unifying them (see my note on `FlashLfqEngine.cs:1896`) resolves this too.

**2. The doc doesn't mention the 0.3 floor.** Until the value is actually threaded through, anything at or below 0.3 is a no-op, and nothing here tells the reader that.

Optional, take it or leave it: the siblings on this class carry their subsystem and unit in the name (`MaxMbrRtWindow`, `MbrPpmTolerance`, `IsoTrackerIdFilter`). `AcceptablePeakWidth` reads like a general FlashLFQ chromatographic-peak setting that would affect ordinary MS1/MBR quantification, which it doesn't â€” it's only ever read from `QuantifyIsobaricPeaks()`. Something like `IsoTrackerMinSharedPeakRtWidth` would be self-describing in a TOML file.

mzLib/FlashLFQ/FlashLfqParameters.cs
```diff
@@ -23,6 +23,7 @@ public FlashLfqParameters()
 
             // IsoTracker settings
             IsoTracker = false;
+            AcceptablePeakWidth = 0.4;
```
trishorts, line 26: Could we pin this default in a test? `Test/FlashLFQ/TestFlashLfqParameters.cs` already round-trips the parameters through TOML and hand-asserts the defaults, so a one-liner there (`Assert.That(readToml.AcceptablePeakWidth, Is.EqualTo(0.4));`) covers both the default and the serialization.

More importantly, nothing currently exercises the filter as a *threshold*. The default moves the effective cutoff from 0.3 to 0.4 min, so regions in (0.3, 0.4] that master keeps are now discarded for every existing IsoTracker user â€” that's a real output change arriving without an opt-in. A focused test over synthetic XICs with known region widths straddling the threshold (say 0.35 and 0.5 with the setting at 0.4, asserting exactly which regions survive) would demonstrate the filter discriminates as intended rather than deleting real signal.

(Codecov reports the added lines as covered, but that's line coverage from existing tests incidentally executing them â€” no assertion depends on the new behaviour.)

mzLib/FlashLFQ/IsoTracker/XICGroups.cs
```diff
@@ -244,6 +244,7 @@ public List<PeakRegion> BuildSharedPeaks(double windowLimit = 0.3)
             }
 
             // remove the peaks with the time window less than the window limit
+            // We already have a limit peak width for the peak region.
```
trishorts, line 247: This comment isn't accurate in the merged state, and it's the kind of inaccuracy that costs someone an hour later. It says a peak-width limit is already established, but the only limit here is this method's own hard-coded `windowLimit = 0.3` default â€” the new `AcceptablePeakWidth` never reaches this line (see my note on `FlashLfqEngine.cs:1896`).

So it tells the next reader not to go looking for exactly the plumbing that's missing. Either drop it, or once the parameter is actually threaded through, rewrite it to say concretely where the limit comes from and in what units.

mzLib/FlashLFQ/IsoTracker/XICGroups.cs
```diff
@@ -91,7 +91,7 @@ public void FindSharedExtrema(double count_threshold, double tolerance, double c
 
             SharedExtrema = group_min.Concat(group_max).ToList();
             SharedExtrema.Sort((p1, p2) => p1.RetentionTime.CompareTo(p2.RetentionTime)); // sort the shared extrema by the retention time
-            SharePeakTrimming();
+            ExtremaTrimming(); // Just rename the function to be more readable
```
trishorts, line 94: Nit: `// Just rename the function to be more readable` describes what happened to the diff rather than what the code does. Once this merges the rename is history and the comment is permanent noise on a call site that needs no annotation. The rename itself is a fine change â€” the rationale just belongs in the commit message.

(The rename is complete and consistent â€” declaration at `:149` and the sole call site here both updated, and the method is `private` â€” so there's no half-applied-rename risk.)

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -128,6 +128,7 @@ public FlashLfqEngine(
 
             // IsoTracker settings
             bool isoTracker = false,
+            double acceptablePeakWidth = 0.4,
```
trishorts, line 131: Non-blocking, and I've already checked the thing that would have made it blocking: inserting a new optional parameter mid-list rather than appending silently re-binds any *positional* caller past this point.

I verified there aren't any. All 48 in-repo call sites pass the identification list positionally and everything after it by name, and both MetaMorpheus call sites (`PostSearchAnalysisTask.cs:574`, `PostGlycoSearchAnalysisTask.cs:527`) name every argument including the first. So nothing misbinds and the mzLibâ†’MM integration job will stay green.

Worth keeping in mind for the future though: appending would have been free here, and the risk is real for any out-of-repo positional caller. Grouping it with the other IsoTracker settings is defensible â€” just flagging that the alternative existed.

trishorts (changes requested):
Thanks for this — the noise problem in IsoTracker output is real and a configurable width filter is the right shape for it. One blocking issue, plus some smaller notes.

**The blocker:** the new setting can only tighten the filter, never loosen it. `QuantifyIsobaricPeaks` still constructs `new XICGroups(xicGroup)`, whose constructor calls `BuildSharedPeaks()` with no argument, so the hard-coded `windowLimit = 0.3` has already removed every narrower region before `AcceptablePeakWidth` is ever consulted. The effective threshold is `max(0.3, AcceptablePeakWidth)`, so any value at or below 0.3 — including 0, the natural way to opt out — silently does nothing. Threading the value through `XICGroups` into `BuildSharedPeaks` and dropping the duplicate engine-side check would make one threshold govern the filter and fix this. Details inline.

**Worth knowing:** the `peakWindow > 0.001` guard this replaces was dead code — everything reaching it was already at least 0.3 min wide. So the real behavioural change here is 0.3 to 0.4 min, not 0.001 to 0.4. That's a smaller change than it looks, but it still means regions in (0.3, 0.4] that master keeps are now dropped for every existing IsoTracker user, with no opt-in. That's the part I'd most like to see a test around.

**Things I checked so you don't have to:**
- Inserting `acceptablePeakWidth` mid-list in the convenience constructor does *not* break anything. All 48 in-repo call sites and both MetaMorpheus call sites (`PostSearchAnalysisTask.cs:574`, `PostGlycoSearchAnalysisTask.cs:527`) pass named arguments, so nothing misbinds.
- Not validating the new knob matches the house convention — none of the other 27 `FlashLfqParameters` settings are validated either, so I'm not asking for it here.
- The doc comment you added is more than the file's existing bar (only 2 of 28 properties carry one). Just needs its boundary wording corrected.
- The `SharePeakTrimming` to `ExtremaTrimming` rename is complete and consistent, and the method is private — no half-applied rename risk.

**Heads-up on a likely conflict:** #942 (IsoTracker performance improvement) also edits `XICGroups.cs` and was updated today. Since the fix above needs a change to the `XICGroups` constructor, it may be worth coordinating with @Alexander-Sol on which PR lands that signature change.

PR #883, opened 2025-05-13
Averagine Generalization

nbollis:
Created an abstract `AverageResidue` which contains all information that was previously stored statically in `DeconvolutionAlgorithm` 

`Averagine` Keeps the old behavior
`OxyriboAveragtide` Is for RNA molecules

These can be injected into the algorithm via the `DeconvoltuionParameters`


mzLib/MassSpectrometry/Deconvolution/AverageResidue/OxyriboAveragtide.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public override double[] GetAllTheoreticalMasses(int index) => AllMasses[index];
+
+    public override double[] GetAllTheoreticalIntensities(int index) => AllIntensities[index];
+
+    public override double GetDiffToMonoisotopic(int index) => DiffToMonoisotopic[index];
+    static OxyriboAveragtide()
+    {
+        // Magic numbers determined by counting atoms in the main 4 canonical RNA bases. 
+        double averageC = 9.5;
```
trishorts, line 27: these numbers are quite "low-res" compared to the same values in the protein averagine

trishorts (commented):
this all looks good. please comment on/correct the sig figs for the averagetide and i will approve.

mzLib/MassSpectrometry/Deconvolution/AverageResidue/OxyriboAveragtide.cs
nbollis: This is an extremely oversimplistic approach and should be further refined.

Alexander-Sol (dismissed):

Alexander-Sol (approved):

PR #885, opened 2025-05-21
Fix The bug on RunErrorChecking

RayMSMS:
In RunErrorChecking, if we find the same Apex peaks, the system will merge the peaks and their ID. In some cases, the two peptide IDs have similar mass but different base sequences (like PEPTIEK and KETIPEP). However, the IsoTracker will be crushed while trying to output it. In this PR, I edit the RevisedModifiedPeptides function to resolve the problem.

(1)New DetetionType: IsoTrack_MSMS, the MS2Id got from isoTracker. It will be used in ResolvedPeptide as a higher priority to conserve the peak.
(2)Peak merging handle: When peaks are merged in RunErrorCheck, we need to label the stored peak and try the peak as ambiguous. Their intensity will be 0 in the peptide output.
(3)Bug on MBRscore constructor: if all peak are ambiguous, there is a crush in the constructor, I make a try and set to handle it. If it happened, MBRscore will return null and stop the processing


mzLib/FlashLFQ/Peptide.cs
```diff
@@ -158,13 +158,34 @@ public void SetDetectionType(SpectraFileInfo fileInfo, DetectionType detectionTy
...
+                {
+                    RetentionTimes[peak.SpectraFileInfo] = peak.ApexRetentionTime;
+                    Intensities[peak.SpectraFileInfo] = 0;
+                    if (peak.DetectionType != DetectionType.IsoTrack_Ambiguous) // If the detectionType is not IsoTrack_amb but with more than one id. That extra id is added by runErrorChecking
+                    {
+                        DetectionTypes[peak.SpectraFileInfo] = DetectionType.MSMSAmbiguousPeakfinding;
+                    }
+                    else
+                    {
+                        DetectionTypes[peak.SpectraFileInfo] = peak.DetectionType;
```
Alexander-Sol, line 179: Shouldn't these always be either MSMSAmbiguous or IsoTrack_Ambiguous?

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -842,6 +844,7 @@ internal void RevisedModifiedPeptides()
                         var allSeq = isoPeptidePeaks
                             .Where(p => p != null)
                             .SelectMany(p => p.Identifications)
+                            .Where(p=>p.BaseSequence == originalPeptide.BaseSequence)// do not output the peptide with different base sequence in the peptide result
                             .Select(p => p.ModifiedSequence)
                             .Distinct()
```
Alexander-Sol, line 849: We still need to modify the peptides with the different base sequence by setting it to ambiguous/intensity to zero

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -831,8 +835,14 @@ private MbrScorer BuildMbrScorer(List<ChromatographicPeak> acceptorFileIdentifie
                 .ToList();
             double medianAcceptorLogIntensity = acceptorFileLogIntensities.Median();
             Normal logIntensityDistribution = new Normal(acceptorFileLogIntensities.Median(), acceptorFileLogIntensities.InterquartileRange() / 1.36);
-
-            return new MbrScorer(apexToAcceptorFilePeakDict, acceptorFileIdentifiedPeaks, ppmDistribution, logIntensityDistribution);
+            try // if the constructor fails, we don't want to crash the program
```
nbollis, line 838: Why would the constructor fail? This seems sketchy

nbollis (approved):

trishorts (approved):

PR #886, opened 2025-05-23
MassIndexingEngine

zhuoxinshi:
Created MassIndexingEngine class in PeakIndexing with peak finding and XIC construction with neutral masses


mzLib/MassSpectrometry/PeakIndexing/MassIndexingEngine.cs
```diff
@@ -0,0 +1,138 @@
...
+                    IndexedPeaks[roundedMass].Add(new IndexedMass(envelope, scanArray[scanIndex].RetentionTime, scanIndex, scanArray[scanIndex].MsnOrder));
+                }
+            }
+            if (IndexedPeaks == null || IndexedPeaks.Length == 0)
+                return false;
+            else
+                return true;
+        }
+
+        public IndexedMass GetIndexedPeak(double mass, int charge, int zeroBasedScanIndex, PpmTolerance ppmTolerance)
```
nbollis, line 48: This looks like a copy and paste from the base class. You can delete

nbollis (changes requested):
The indexed mass should be revised to be immutable. More care should be taken to reuse methods from parent indexing engine when applicable and avoid duplicated code.

mzLib/MassSpectrometry/PeakIndexing/IndexedMass.cs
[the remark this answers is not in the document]
Alexander-Sol: M is named that way because it can represent monoisotopic mass or m/z depending on the indexing engine. However it would be good to have a summary comment on the property that explains exactly that

mzLib/MassSpectrometry/PeakIndexing/MassIndexingEngine.cs
```diff
@@ -0,0 +1,138 @@
...
+            List<int> peakIndicesInBins = bins.Select(b => BinarySearchForIndexedPeak(b, zeroBasedScanIndex)).ToList();
+            return GetBestPeakFromBins(bins, mass, charge, zeroBasedScanIndex, peakIndicesInBins, ppmTolerance);
+        }
+
+        public static IndexedMass GetBestPeakFromBins(List<List<IndexedMass>> allBins, double mass, int charge, int zeroBasedScanIndex, IList<int> peakIndicesInBins, PpmTolerance ppmTolerance)
+        {
+            IndexedMass bestMass = null;
+            for (int i = 0; i < allBins.Count; i++)
+            {
+                var tempPeak = GetPeakFromBin(allBins[i], mass, charge, zeroBasedScanIndex, peakIndicesInBins[i], ppmTolerance);
```
Alexander-Sol, line 61: It looks like you're iterating over all bins here, which is inefficient. you should select the bins that are in tolerance

Alexander-Sol (changes requested):
Need more test coverage and more comments in MassIndexingEngine. Specifically, I would like to see comments that explain why so many functions need to be re-implemented.

nbollis (dismissed):

mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
```diff
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
```
Alexander-Sol, line 6: Do you need System.Runtime.InteropServices??

mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
```diff
@@ -121,7 +123,7 @@ public List<IIndexedPeak> GetXic(double mz, double retentionTime, Tolerance ppmT
         /// <param name="missedScansAllowed"> the number of successive missed scans allowed before the xic is terminated </param>
         /// <param name="maxPeakHalfWidth"> the maximum distance from the apex RT of the XIC to both start RT and end RT </param>
         /// <returns> A list of IIndexedPeak objects, ordered by retention time </returns>
-        public List<IIndexedPeak> GetXic(double mz, int zeroBasedStartIndex, Tolerance ppmTolerance, int missedScansAllowed, double maxPeakHalfWidth = double.MaxValue)
+        public List<IIndexedPeak> GetXic(double mz, int zeroBasedStartIndex, Tolerance ppmTolerance, int missedScansAllowed, double maxPeakHalfWidth = double.MaxValue, int? charge = null)
```
Alexander-Sol, line 126: I would like a summary comment for the new arguments, especially because if it is not null, you won't get any results at all

Alexander-Sol (changes requested):
I would like the peak indexing engine to throw an exception if charge != null and peak is not IndexedMass. Otherwise, this could result in failures in peak finding without any apparent reason.

 Also, amend all summary comments for methods with the new parameter that explain what it is (and emphasize that it should remain null most of the time)

mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
```diff
@@ -257,12 +258,17 @@ internal List<List<T>> GetBinsInRange(double mz, Tolerance ppmTolerance)
...
-        /// Returns the peak that is closest to the target mz
+        /// <param name="charge"> an optional parameter used only for IIndexedMass and massIndexingEngine; must be null for mz peak indexing </param>
+        /// Returns the peak that is closest to the target mz from one bin
         /// </summary>
-        internal static T GetPeakFromBin(List<T> bin, double mz, int zeroBasedScanIndex, int peakIndexInBin, Tolerance ppmTolerance)
+        internal static T GetPeakFromBin(List<T> bin, double mz, int zeroBasedScanIndex, int peakIndexInBin, Tolerance ppmTolerance, int? charge = null)
         {
             T? bestPeak = default(T);
             if (peakIndexInBin < 0 || peakIndexInBin >= bin.Count) return bestPeak;
+            if (charge != null && bin.FirstOrDefault() is not IndexedMass)
```
Alexander-Sol, line 268: If the bin is empty and charge is not equal to null, this will throw an exception. I think you should check that bin.FirstOrDefault() != null

Alexander-Sol (approved):

RayMSMS (approved):

PR #892, opened 2025-06-10
PpmToleranceWithNotch

zhuoxinshi:
Added a new type of Tolerance that works with notches for neutral mass peak finding; Changed all PpmTolerance to Tolerance in existing peak finding methods


mzLib/MzLibUtil/PpmToleranceWithNotch.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MzLibUtil
+{
+    public class PpmToleranceWithNotch : Tolerance
+    {
+        private readonly double[] AcceptableSortedMassShifts;
+        public const double NotchStep = 1.00335483810;
```
nbollis, line 12: Use Constants as defined in mzlib.

mzLib/MzLibUtil/PpmToleranceWithNotch.cs
```diff
@@ -0,0 +1,63 @@
...
+            : base(value)
+        {
+            var massShifts = new List<double>();
+            for (int i = 1; i <= protonNotches; i++)
+            {
+                massShifts.Add(NotchStep * i);
+                massShifts.Add(-NotchStep * i);
+            }
+            AcceptableSortedMassShifts = massShifts.OrderBy(p => p).ToArray();
+            NumNotches = protonNotches;
```
nbollis, line 25: Logic is wrong. The num notches, as written, will be 2 * proton notches.+ 1. where the 1 is from the true mass, no notches.

mzLib/MzLibUtil/PpmToleranceWithNotch.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace MzLibUtil
+{
+    public class PpmToleranceWithNotch : Tolerance
+    {
+        private readonly double[] AcceptableSortedMassShifts;
+        public const double NotchStep = 1.00335483810;
+        public PpmTolerance PpmTolerance { get; init; }
+
+        public PpmToleranceWithNotch(double value, int protonNotches)
```
nbollis, line 15: This means we will go +- protonNotches always. Is this the intended behavior? Do you want to define positive and negative notch steps?

mzLib/MzLibUtil/PpmToleranceWithNotch.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                massShifts.Add(NotchStep * i);
+                massShifts.Add(-NotchStep * i);
+            }
+            AcceptableSortedMassShifts = massShifts.OrderBy(p => p).ToArray();
+            NumNotches = protonNotches;
+            PpmTolerance = new PpmTolerance(value);
+        }
+
+        int NumNotches { get; init; }
```
nbollis, line 29: Is this property necessary? If so it should be made public. If not, it should be removed

mzLib/MzLibUtil/PpmToleranceWithNotch.cs
```diff
@@ -0,0 +1,63 @@
...
+    public class PpmToleranceWithNotch : Tolerance
+    {
+        private readonly double[] AcceptableSortedMassShifts;
+        public const double NotchStep = 1.00335483810;
+        public PpmTolerance PpmTolerance { get; init; }
+
+        public PpmToleranceWithNotch(double value, int protonNotches)
+            : base(value)
+        {
+            var massShifts = new List<double>();
```
nbollis, line 18: Don't forget to add 0 as an acceptable mass shift

mzLib/Test/TestRangeAndTolerances.cs
```diff
@@ -364,5 +365,22 @@ public void TolerancePPMGetRange()
 
             Assert.AreEqual("±1.0000 PPM", tol.ToString());
         }
+
+        [Test]
+        public void PpmToleranceWithNotch()
```
nbollis, line 370: You should add tests to make sure values that fall between the notches are excluded.

nbollis (changes requested):

PR #894, opened 2025-06-11
PeakTracing

zhuoxinshi:
Created ExtractedIonChromatogram as a generic XIC class that works with any type of peak objects implementing IIndexedPeak. Added an abstract class for XIC spline and implemented cubic/linear spline and the option of adding 0 intensity points before spline

### DIFF
--- a/mzLib/FlashLFQ/PeakIndexingEngine/IFlashLfqIndexingEngine.cs
+++ b/mzLib/FlashLFQ/PeakIndexingEngine/IFlashLfqIndexingEngine.cs
@@ -21,7 +21,7 @@ namespace FlashLFQ.Interfaces
         /// </summary>
         /// <param name="mz"> the m/z of the peak to be searched for </param>
         /// <param name="zeroBasedScanIndex"> the zero based index of the scan where the peak is to be found </param>
-        public IIndexedPeak GetIndexedPeak(double mz, int zeroBasedScanIndex, PpmTolerance tolerance);
+        public IIndexedPeak GetIndexedPeak(double mz, int zeroBasedScanIndex, Tolerance tolerance);
         /// <summary>
         /// A generic method of peak tracing across the retention time. Finds peaks with a given mz that occur on either side of a given
         /// retention time. Peak searching iterates backwards through the scans until the peak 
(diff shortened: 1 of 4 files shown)


mzLib/MassSpectrometry/PeakIndexing/ExtractedIonChromatogram.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace MassSpectrometry
+{
+    public class ExtractedIonChromatogram
+    {
+        public List<IIndexedPeak> Peaks { get; set; }
+
+        public double ApexRT => Peaks.OrderByDescending(p => p.Intensity).First().RetentionTime;
+        public int ApexCycle => Peaks.OrderByDescending(p => p.Intensity).First().ZeroBasedScanIndex;
+        public double AveragedM => AverageM();
```
Alexander-Sol, line 18: For ApexRT and ApexCycle, you should set those in the constructor so you don't have to order the peak list multiple times

mzLib/MassSpectrometry/PeakIndexing/ExtractedIonChromatogram.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading.Tasks;
+
+namespace MassSpectrometry
+{
+    public class ExtractedIonChromatogram
+    {
+        public List<IIndexedPeak> Peaks { get; set; }
+
+        public double ApexRT => Peaks.OrderByDescending(p => p.Intensity).First().RetentionTime;
+        public int ApexCycle => Peaks.OrderByDescending(p => p.Intensity).First().ZeroBasedScanIndex;
```
Alexander-Sol, line 17: I don't know what ApexCycle means. I think ApexScanIndex would be more descriptive

mzLib/MassSpectrometry/PeakIndexing/ExtractedIonChromatogram.cs
```diff
@@ -0,0 +1,49 @@
...
+        public double AveragedM => AverageM();
+        public (double, double)[] XYData { get; set; }
+        public double[] NormalizedPeakIntensities { get; set; }
+        public double StartRT => Peaks.Min(p => p.RetentionTime);
+        public double EndRT => Peaks.Max(p => p.RetentionTime);
+        public int StartCycle => Peaks.Min(p => p.ZeroBasedScanIndex);
+        public int EndCycle => Peaks.Max(p => p.ZeroBasedScanIndex);
```
Alexander-Sol, line 24: All of these can be set in the constructor

Alexander-Sol (changes requested):

Alexander-Sol (approved):

PR #895, opened 2025-06-17
add UniProt required Xml entry attributes

trishorts:
The UniProt XML database is required to have certain attributes in the <entry> tag that we currently don't write. Also, we don't report correct status for chain peptides with no begin. Consequently, the xml databases that we write cannot be read by ProSight in ProteomeDiscoverer. This PR fixes these two problems. 

- It was validated by Ryan that ProSight can read this new database.
- when searching in MM with original UniProt and the revised XML, I get exactly the same number of psms, peptides and proteins
- constructed a local nuget and tested MetaMorpheus read/write/write reduced; use written and use reduced. works.

Here's the full XSD for the UniProt XML, which hass all the requirements:
[https://ftp.uniprot.org/pub/databases/uniprot/current_release/knowledgebase/complete/uniprot.xsd](https://urldefense.com/v3/__https:/ftp.uniprot.org/pub/databases/uniprot/current_release/knowledgebase/complete/uniprot.xsd__;!!Mak6IKo!MRKPLTc4xoeULPH-ikftIfusnENBsX0yYnNFYW_NP6Qzao4IwxPp2hfYOoRF41OBI7fuU9HFpxSyAzV36fLydItATJY0074o$)

Here is the email from fellers explaining the problems. This PR fixes these two errors:.

image


mzLib/Test/DatabaseTests/TestProteomicsReadWrite.cs
```diff
@@ -88,6 +89,98 @@ public void Test_readUniProtXML_writeProteinXml()
...
+            }
+            // Clean up the output file after the test
+            if (File.Exists(outputPath))
+            {
+                File.Delete(outputPath);
+            }
+        }
+
+        [Test]
+        public void Test_readUniProtXML_featureBeginEndPosition()
```
nbollis, line 139: I believe this test will pass even if it was not written correctly.

nbollis (approved):

PR #905, opened 2025-07-07
Optimize Memory Usage in RAM

RayMSMS:
This update introduces several changes aimed at reducing memory consumption:

**Prune indexedEnginePeak**

- Generate targetMzs from the relevant IDs.
- Retain only the indexedPeak entries whose m/z values match the generated list.

**Switch Data Type of Intensity, mz and retention time from double to Float (64 bits -> 32 bits)**
- Reduces precision from ~16 to ~8 decimal digits.
- Precision loss is minimal and within acceptable limits (approximately 0.02 ppm).
- Unit tests have been updated to allow a wider tolerance range accordingly.
Cover


mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -2148,6 +2160,41 @@ private bool IDsEqual(List<Identification> idList1, List<Identification> idList2
...
+        /// peptides in the current dataset.
+        /// </summary>
+        /// <returns>A list of doubles representing the calculated m/z values of interest.</returns>
+        private List<double> GetTargetMz()
+        {
+            HashSet<double> interestedMass = new HashSet<double>();
+            foreach (var peptide in ModifiedSequenceToIsotopicDistribution)
+            {
+                var id = _allIdentifications.First(p => p.ModifiedSequence == peptide.Key);
+                List<double> massOfInterest = peptide.Value.Select(p => p.massShift + id.MonoisotopicMass).ToList();
```
Alexander-Sol, line 2175: should be plural. "massesOfInterest"

mzLib/FlashLFQ/IsoTracker/XIC.cs
```diff
@@ -100,16 +100,17 @@ private List<IIndexedPeak> PadPeaks()
 
             for (int i = 1; i < 6; i++) //add 5 peaks after the last peak
             {
-                paddedPeaks.Add(new IndexedMassSpectralPeak(0, 0, 0, lastPeak.RetentionTime + gap * i));
+                paddedPeaks.Add(new IndexedMassSpectralPeak(0, 0, 0, (float)(lastPeak.RetentionTime + gap * i)));
```
Alexander-Sol, line 103: unnecessary cast

Alexander-Sol (changes requested):
Left several comments. There are still several properties FlashLfqEngine and XIC that used doubles that should be converted to floats

trishorts (dismissed):
make sure patch code coverage is good

mzLib/FlashLFQ/IsoTracker/XIC.cs
```diff
@@ -174,11 +175,11 @@ public double AlignXICs(XIC referenceXIC, int resolution = 1000)
         /// <exception cref="ArgumentException"></exception>
         public void BuildSmoothedCubicSpline(int smoothDegree)
```
nbollis, line 176: Collection allocations here need to be worked on. On RT, we call ToArray then ToList later on. On the smoothed intensity, we get an array back from the method and then ToArray is called again. In another case, smoothed intensity comes back as an array, ToList is called, then ToArray is called to send it into another method

nbollis (changes requested):
Two issues stand out to me, unnecessary collection allocations, and unnecessary rounding. 

Every time you round, your introduce error. This error is greater than the precision of a float. 
There is no need to round 14.9999999 to 15.000 as they are functionally the same.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -2148,6 +2160,45 @@ private bool IDsEqual(List<Identification> idList1, List<Identification> idList2
...
+        // Helper method for float version
+        private List<float> GetAllPossibleMzs(HashSet<float> targetMasses)
+        {
+            HashSet<float> targetMzs = new();
+            foreach (var mass in targetMasses)
+            {
+                _chargeStates.ForEach(chargeState =>
+                {
+                    targetMzs.Add((float)mass.ToMz(chargeState));
+                });
```
Alexander-Sol, line 2195: You shouldn't need this cast

mzLib/MassSpectrometry/PeakIndexing/IndexedMassSpectralPeak.cs
```diff
@@ -6,16 +6,16 @@ namespace MassSpectrometry
...
         public int ZeroBasedScanIndex { get; init; }
-        public double Mz { get; init; }
-        public double M => Mz;
-        public double RetentionTime { get; init; }
-        public double Intensity { get; init; }
+        public float Mz { get; init; }
+        public float M => Mz;
+        public float RetentionTime { get; init; }
+        public float Intensity { get; init; }
         public IndexedMassSpectralPeak(double mz, double intensity, int zeroBasedScanIndex, double retentionTime)
```
Alexander-Sol, line 13: You should change this constructor to take in floats instead of doubles. You'll also need to modify IndexingEngine to do the cast on the constructor call.

Alternately, add a new constructor that takes in floats

mzLib/MassSpectrometry/PeakIndexing/PeakSpline/XicSpline.cs
```diff
@@ -44,10 +45,10 @@ public abstract class XicSpline
         /// </summary>
         public void SetXicSplineXYData(ExtractedIonChromatogram xic, bool cycle = false, double start = -1, double end = -1)
         {
-            double[] peakRts = null;
+            float[] peakRts = null;
             if (cycle)
             {
-                peakRts = xic.Peaks.Select(p => (double)p.ZeroBasedScanIndex).ToArray();
+                peakRts = xic.Peaks.Select(p =>(float)p.ZeroBasedScanIndex).ToArray();
             }
```
Alexander-Sol, line 52: This was bugged to begin with. It should take in retention times, not scan indexes

PR #913, opened 2025-07-21
Memory usage improving: generate the targetMassList for indexEngine pruning

RayMSMS:
In the previous PR, I generated **TargetMzs** to assist with IIndexedPeak pruning.

In this PR, I further optimize the process **by reducing the size of the TargetMzs list**, focusing only on the masses relevant to isoTracker.

Summary of changes:

1. The new individual function to create PeptideGroupsForIsoTracker: Group peptides by their base sequence and monoisotopic mass. And followed by any motifs filtering.
2. Iterate through each PeptideGroupsForIsoTracker, and use the associated isotopic distribution to generate the refined TargetMzs.

### DIFF
--- a/mzLib/FlashLFQ/FlashLfqEngine.cs
+++ b/mzLib/FlashLFQ/FlashLfqEngine.cs
@@ -52,6 +52,10 @@ namespace FlashLFQ
         /// </summary>
         internal bool IsoTrackerIsRunning { get; private set; }
         /// <summary>
+        /// The list of IsoPeptide Group with the same base sequence and monoisotopic mass. Only for IsoTracker analysis
+        /// </summary>
+        internal List<IGrouping<(string BaseSequence, double MonoisotopicMassGroup), Identification>> PeptideGroupsForIsoTracker { get; private set; } 
+        /// <summary>
         /// The dictionary of isobaric peaks for each modified sequence
         /// </summary>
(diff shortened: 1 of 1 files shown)


mzLib/FlashLFQ/IsoTracker/IsobaricPeptideGroup.cs
```diff
@@ -0,0 +1,37 @@
...
+            MonoisotopicMass = monoisotopicMass;
+            Identifications = identifications ?? new List<Identification>();
+        }
+        public override bool Equals(object obj)
+        {
+            if (obj is IsobaricPeptideGroup other)
+            {
+                // Compare BaseSequence and MonoisotopicMass (rounded to 4 decimal places for floating point safety)
+                return BaseSequence == other.BaseSequence &&
+                       Math.Round(Convert.ToDouble(MonoisotopicMass), 4) == Math.Round(Convert.ToDouble(other.MonoisotopicMass), 4);
```
Alexander-Sol, line 26: Why are you converting to double here?
Usually, the pattern for this type of comparison is
`Math.Abs(MonoisotopicMass - other.MonoisotopicMass) < 0.0001`
where you subtract and make sure the absolute difference is less than the desired tolerance

Alexander-Sol (dismissed):

Alexander-Sol (approved):

nbollis (approved):

PR #919, opened 2025-07-23
Modified MakeIdentifications to fix issue with IQuantifiableResultFiles

Alexander-Sol:
Previously, calling the MakeIdentifications adapter to create Identifications for FlashLFQ failed to distinguish between targets and decoys when making identifications. Whoops.

Now, identifications are labeled as targets or decoys. 

Additionally, information on the PSM score and QValue are included (for metamorpheus search results).


trishorts (approved):
sicker pr

PR #942, opened 2025-08-22
IsoTracker performance improvement

Alexander-Sol:
Optimized parallelization to increase the speed of IsoTracker. This lead to a 50% decrease in total runtime when running IsoTracker + MBR on 24 files.

When tested on 24 files, most of the performance increase comes from optimizations to the AddIsoPeaks method. However, when tested on a large number of files (370), performance increased by...


nbollis (dismissed):

trishorts (approved):
please check copilot comments

trishorts:
I tried to bring this up to master and stopped, deliberately — it needs you rather than a merge resolution.

The conflict is in `FlashLfqEngine.cs` and it is not adjacency. Master restructured the same IsoTracker quantification loop this PR restructures, adding the `RequireMultipleIdsInOneFiles` gate and reorganising the per-file XIC construction. This branch reorganised the same block for performance and added `Stopwatch` instrumentation around `BuildXIC`. Four conflict blocks, the largest 77 lines, and they overlap in intent rather than merely in position.

Resolving that means deciding how your restructure composes with master's, which is a design call on your optimisation. More to the point: **the speedup would need re-measuring anyway.** The loop this PR optimises is not the loop that is on master now, so whatever the original numbers were, they no longer describe the baseline. Merging it without re-measuring would ship a PR claiming a speedup nobody has verified still exists.

One thing that may make this cheaper. The PR has a separable half that does *not* conflict at all:

- `MassSpectrometry/PeakIndexing/IndexingEngine.cs` (+42) — the efficient XIC retrieval method
- `Test/FlashLFQ/IndexingEngineTests.cs` (+100) — its tests

That part is self-contained and mergeable on its own. Splitting it out would land the reusable piece now and leave the engine-loop restructure as a smaller PR that can be re-measured against current master on its own terms.

Say the word and I'll do the split, or resolve the loop conflict a particular way if you tell me which structure you want to keep.

trishorts:
Heads-up on a collision with #1191 (approved, not yet merged): it removes the four CS8625 warnings in `IndexingEngine.cs` by annotating the parameters that accept null, including `matchedPeaks`, which becomes `Dictionary<IIndexedPeak, ExtractedIonChromatogram?>?`.

The new `GetXicFromScanRange` here declares the old shape:

```csharp
int? charge = null, Dictionary<IIndexedPeak, ExtractedIonChromatogram> matchedPeaks = null)
```

`IndexingEngine.cs` is `#nullable enable`, so after #1191 merges that line re-adds a CS8625 and leaves the new method inconsistent with its two siblings. Worth matching `GetXic`/`GetXicByScanIndex` when you rebase — a null value in that dictionary marks a peak that was examined but did not form an XIC.

PR #944, opened 2025-08-26
XicSpline toString

zhuoxinshi:
added a toString method for XicSpline settings


nbollis (dismissed):

trishorts:
Merged master and pushed — this is mergeable again.

The only conflict was the using block: the test file moved from `mzLib/Test/TestXic.cs` to `mzLib/Test/MassSpectrometryTests/TestXic.cs`, and both sides had reordered `using MathNet.Numerics.Distributions;`. Kept master's ordering. Both halves of the PR survive intact — `XicSpline.ToString()` and `TestXicSplineToString`.

`TestXic` is 15/15 green locally.

acesnik (approved):
🤖 Approving. Splitting on `'\n'` and trimming in the test is the right call — `AppendLine` would otherwise make this platform-dependent.

One suggestion: `GetType().Name` is worth using because there are two subclasses, but the test only covers `XicCubicSpline`, and a hardcoded string would pass it identically. A second case on `XicLinearSpline` is what proves the type name differentiates. Two cosmetics if you're in there again: the `$` on `XicSplineSettings: ` has no placeholder, and there's a stray blank line at end of file.

PR #950, opened 2025-09-02
Modomics Modificaiton Parsing

nbollis:
Adds support for Modomics, an rna modificaiton database. 

This required a few things. 
1. Moving static base chemical formula declarations to the `Chemistry` project to allow Transcriptomics and Omics to reference it. 
2. Some bespoke parsing of the modomics format. 

They have numerous classes of modifications represented here and their chemical formulas are all encoded slightly differently. I have added support for ternimal caps and nucleoside modifications. Future work can be done to extend this to full nucleotide modifications and their other odd sub-classes. 

Known issue: Different methylations in their base loss product ions can produce slightly different masses if the mod dissociates or is retained. Likewise, modifications across the phosphate backbone have different fragmentation properties because they fall between residues instead of being neatly on one residue. Previous iterations of adding RNA mods to mzLib has done so through the rnaMods.txt file in which base losses or backbone mods are recorded explicitely. This information is not contained within the modomics csv or json file (both are needed to get full mzLib mod info) and the modifications are therefore loaded as our basic default modification instead of the specially derived modifications. Future work should
1. Parse specialized mod fragmentation properties at time of loading modomics mods. 
2. Construct the derived modification type so that the fragments at these cleavage sites are handled properly. 

That is out of scope for this PR as a tiny fraction of all total fragment ions will be off, the gain in the mod fidelity from having all these additional mods is worth this trade-off.


mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,620 @@
...
+            {
+                case "nucleoside":
+                    // MODOMICS nucleoside formulas describe the entire (neutral) nucleoside; the residue
+                    // already carries the canonical nucleoside scaffold, so removing it yields the shift.
+                    referenceFormulaToRemove = moietyDefinition.NucleosideChemicalFormula;
+                    break;
+                case "nucleotide" when isTerminalCap:
+                    // Generic-"N" cap formulas embed a base-less ribose for the capped residue; removing only
+                    // that ribose keeps the cap nucleoside and phosphate chain as the terminal mass shift.
+                    referenceFormulaToRemove = moietyDefinition.NucleosideChemicalFormula - moietyDefinition.BaseChemicalFormula - Formulas.WaterChemicalFormula;
```
trishorts: **Critical - every one of the 7 `5' Terminal Cap` mods carries a wrong mass shift (+14.99 to +18.01 Da).**

Two compounding causes, one broken computation:

1. **This line subtracts an extra `H2O`.** The quantity the residue already carries is the base-less ribosyl group, `Nucleoside - BondedBase` = `C5H9O4` (identical for all four bases). Subtracting a further `H2O` leaves `C5H7O3`, so the derived shift keeps two hydrogens and one oxygen the oligo already has.
2. **The published cap formulas are anions, consumed as neutral (line 311).** The `case "nucleotide"` arm just below (line 316) rejects real-base nucleotides *precisely because* MODOMICS publishes them as inconsistent phosphate anions - but the cap arm consumes `fullFormula` untouched. `Replace("+", "")` at line 222 does nothing here because the charge is not in the formula string. Confirmed from the source SMILES: `m7GpppN`/`m2,7GpppN`/`m2,2,7GpppN` are net -2, `GpppN`/`mpppN` net -3, `mpN` net -1, `mmpN` neutral.

The two errors partially cancel, which is why this survives casual inspection. End-to-end mass balance against mzLib's own model - `RNA("A")` is exactly adenosine (267.0968), so the reference frame is unambiguous:

```
derived neutral m7GpppA (C21H29N10O17P3) = 786.0925
RNA("A") + correct shift (C11H16N5O13P3) = 786.0925   OK
RNA("A") + loader shift  (C11H16N5O14P3) = 802.0874   +15.9949 (one oxygen)
```

All seven caps, loader vs. charge-neutralised-source minus `C5H9O4`:

```
cap           loader                  correct                  error      delta
m7GpppN       C11H16N5O14P3 534.9907  C11H16N5O13P3 518.9957  +15.9949   O
m2,7GpppN     C12H18N5O14P3 549.0063  C12H18N5O13P3 533.0114  +15.9949   O
m2,2,7GpppN   C13H20N5O14P3 563.0220  C13H20N5O13P3 547.0270  +15.9949   O
GpppN         C10H13N5O14P3 519.9672  C10H14N5O13P3 504.9801  +14.9871   H-1O
mpppN         CH4O10P3      268.9017  CH5O9P3       253.9146  +14.9871   H-1O
mpN           CH4O4P        110.9847  CH3O3P         93.9820  +17.0027   HO
mmpN          C2H7O4P       126.0082  C2H5O3P       107.9976  +18.0106   H2O
```

Caps are one of the two feature classes this PR claims to add, and no capped-oligo precursor built with these will match a real spectrum.

**Fix:** subtract only `Nucleoside - BondedBase` (`C5H9O4`), and neutralise the published formula by the charge implied by its SMILES before subtracting. The cap assertion in `ModomicsLoaderTests` currently locks in the implementation output rather than a mass-balance-derived value, so it needs updating alongside.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,620 @@
...
+                .ToList(),
+            ShortName = shortName,
+            Smile = GetStringOrNull(modEntry, "smile"),
+            MoietyType = moietyTypeByShortName.GetValueOrDefault(shortName, string.Empty),
+        };
+    }
+
+    private static IEnumerable<ModomicsConversionOutcome> ConvertDto(ModomicsDto dto)
+    {
+        var neutralFormulaText = dto.Formula.Replace("+", string.Empty, StringComparison.Ordinal);
```
trishorts, line 222: **Major - stripping `+` deletes the charge marker but keeps the proton it denotes, so `3-methylcytidine on C` loads 1.0078 Da heavy.**

MODOMICS is not self-consistent about what it publishes alongside a `+`, and this line assumes one convention for all of them:

| entry | published formula | published H count is | loader shift | correct |
|---|---|---|---|---|
| `m7G` | `C11H15N5O5+` | already neutral | `CH2` 14.0157 | correct |
| `m2,7G`, `m2,2,7G`, `m2,7Gm` | neutral | already neutral | `C2H4` / `C3H6` | correct |
| `m3C` | `C10H16N3O5+` | **the cation** | `CH3` 15.0235 | `CH2` 14.0157 |

Verified by loading the mods: `m3C` comes out as `CH3` / 15.0235. MODOMICS' own `mass_monoiso` for m3C (257.1012 = cytidine 243.0855 + 14.0157) says `CH2`, curated `RnaMods.txt` uses `C1H2` for methylations, and every other single methylation in this loader gives `CH2`. In a polymer the N3-methylcytidine ring cation is charge-balanced by a deprotonated phosphate, so the neutral residue shift is `CH2`.

The object is also internally inconsistent: its `ChemicalFormula` is `CH3` while the derived base-loss portion is `CH2`, stranding a bare `H` on the ribose. `MethylsHaveCorrectFormula` (ModomicsLoaderTests.cs:292) asserts the `CH3` value as a deliberate exception, but nothing justifies why this one methylation should differ.

**One caution on the fix:** reconciling against `mass_monoiso` and rejecting mismatches would be wrong here - MODOMICS' `mass_monoiso` for `m7G` is 299.123, i.e. guanosine + 16.0313, which is itself a data error (a methylation is +14.0157). A hard gate on that field would discard `m7G`, `m2,7G`, `m2,2,7G` and `m2,7Gm`, all of which load correctly today. The formula is the outlier for `m3C`; the mass is the outlier for `m7G`. Safer is to drop one hydrogen per positive charge when the SMILES shows a cationic centre and the formula's H count matches the cation.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,620 @@
...
+                continue;
+            }
+
+            ChemicalFormula referenceFormulaToRemove;
+            switch (dto.MoietyType)
+            {
+                case "nucleoside":
+                    // MODOMICS nucleoside formulas describe the entire (neutral) nucleoside; the residue
+                    // already carries the canonical nucleoside scaffold, so removing it yields the shift.
+                    referenceFormulaToRemove = moietyDefinition.NucleosideChemicalFormula;
```
trishorts, line 309: **Major - anionic formulas on `nucleoside`-typed rows load 2.0157 Da light.**

This branch assumes a `nucleoside`-typed row publishes the *neutral* whole nucleoside. Two rows break that: `Ar(p)` (2'-O-ribosyladenosine (phosphate)) and `Gr(p)` (2'-O-ribosylguanosine (phosphate)). Both carry a phosphate and both are published as **dianions** - their SMILES each show two `[O-]`. The `Replace("+", "")` guard at line 222 only handles cations, so they fall straight through to the neutral-nucleoside path.

This is exactly the protonation hazard the `case "nucleotide"` arm at line 316 refuses to guess at, but because these rows are labelled `nucleoside` the guard never fires.

Verified by loading:

```
short   target   loader shift   mass_monoiso implies   delta
Ar(p)   A          209.9929           212.0085       -2.0156
Gr(p)   G          209.9929           212.0086       -2.0157
```

The correct shift is `C5H9O7P` (212.0086) - a ribosyl phosphate joined at the 2'-O - not `C5H7O7P`. MODOMICS' `mass_monoiso` agrees (479.1053 - adenosine 267.0968 = 212.0086). Both are real tRNA modifications.

**Fix:** don't treat `Moiety type` alone as implying a neutral formula. Detecting a phosphorus-bearing formula on a `nucleoside`-typed row and applying the same caution as the `nucleotide` branch would catch both, as would neutralising by the SMILES-implied charge (the same helper the caps need).

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,620 @@
...
+        if (productIons is null)
+        {
+            return null;
+        }
+
+        // Every published ion becomes a diagnostic ion.
+        var ions = productIons.ToList();
+        var diagnosticIons = new Dictionary<DissociationType, List<double>>
+        {
+            { DissociationType.AnyActivationType, ions },
```
trishorts: **Major - diagnostic ions are stored as protonated m/z in a property documented to hold neutral mass.**

`InterpretProductIons` puts MODOMICS `product_ions` straight into `DiagnosticIons[AnyActivationType]`. `ModomicsDto.ProductIons`' own doc calls these "protonated fragment ions generated from the precursor [M+H]+" - i.e. m/z of cations. The consumer contract is the opposite; `PeptideWithSetModifications.cs:554`:

> the diagnostic ion is assumed to be annotated in the mod info as the *neutral mass* of the diagnostic ion, not the ionized species

and it passes the stored value straight in as the `Product`'s neutral mass, applying `.ToMz(1)` only to build the label.

132 of the 167 loaded mods carry diagnostic ions under this convention, e.g.:

```
m5Cm    5,2'-O-dimethylcytidine on C     -> [126.0667]
m2,7G   N2,7-dimethylguanosine on G      -> [180.0885]
```

The computed replacement at line 396 is ambiguous too: `canonicalBaseCationMass` is `(bondedBase + H2)`, which for 1-methyladenosine is 150.0780 - neither the neutral free base (`C6H7N5` = 149.0701) nor a correct cation m/z (150.0774; it adds a whole hydrogen atom rather than a proton, so it is off by the electron mass).

**Impact is latent, not active** - `OligoWithSetMods` doesn't generate `ProductType.D` from `DiagnosticIons` today, only the peptide path does, so nothing is currently mis-fragmenting. But the wrong unit is being baked into 132 shipped modifications and asserted by the new tests, so it becomes wrong the moment RNA diagnostic-ion support lands.

**Fix:** store the neutral monoisotopic mass and let the consumer add the proton, matching the documented contract; keep the m/z form for reporting only.

trishorts (commented):
Reviewed the MODOMICS loader payload. Every finding below was verified by building the branch and actually loading the mods (167 loaded / 28 terminal / 275 not-yet-representable), then cross-checking against the source JSON's own formulas, SMILES charge states, and `mass_monoiso` - not by reading the diff. Numbers quoted are loader output.

The architecture is good: the `NotYetRepresentable` reason taxonomy is the right call over silent drops, and the deliberate non-loads listed in the PR body all behave as described. The base-loss/backbone limitations flagged as out of scope in the description I've left alone.

Four blocking findings, all in `ModomicsLoader.cs`. The first is a hard mass error on all seven terminal caps; the other three are per-entry mass or unit errors.

One broader observation worth acting on separately: **14 of 139 loaded nucleoside mods disagree with MODOMICS' own `mass_monoiso` by more than 0.01 Da.** Three of those (`m3C`, `Ar(p)`, `Gr(p)`) are loader bugs - findings #2 and #3 below. The rest look like upstream data errors (`m7G` is +2.0157, `m6t6A` is -14.0156, the selenouridines differ by a Se isotope choice). Emitting that reconciliation as a load-time *report* alongside `NotYetRepresentableEntries`, rather than as a hard gate, would surface them without discarding entries the loader currently gets right.

trishorts:
many minor comments not posted

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,620 @@
...
+        {
+            // No portion matches, or two portions are isobaric at nominal precision: the ions do not
+            // measure a unique answer, so keep the plain representation rather than guess.
+            return null;
+        }
+
+        var baseLocalizedPortion = matchingPortions[0];
+        var isEntirelySugarLocalized = string.IsNullOrEmpty(baseLocalizedPortion.Formula);
+        return new DerivedBaseLoss(
+            isEntirelySugarLocalized ? BaseLossBehavior.Default : BaseLossBehavior.Modified,
```
trishorts, line 479: **Major - `BaseLossBehavior.Suppressed` is unreachable, so 2'-O-methyls load with base-loss semantics that contradict the curated file.**

`DeriveBaseLossModification` can only ever return `Default` or `Modified` — nothing in this loader emits `Suppressed`. But "nothing sits on the base" and "base loss is suppressed" are different facts, and the product ion only measures the first. `BaseModification`'s own class doc (`BaseModification.cs:19-21`) states the intended contract:

> 2-O methylation: BaseLossType = Suppressed, BaseLossModification = null
> 2-Om6 methylation: BaseLossType = Suppressed, BaseLossModification = CH3

and the curated ground truth agrees — `Resources/RnaMods.txt`:

```
ID   2'-O-Methyladenosine     CF C1 H2    BL Suppressed
ID   N6,2'-O-dimethyladenosine CF C2 H4   BL Suppressed:C1H2
```

The MODOMICS path produces the opposite for the identical chemistry, and the new test locks it in — `ModomicsLoaderTests.cs:228` asserts `BaseLossType == BaseLossBehavior.Default` for `2'-O-methyladenosine on A`. So `Mods.AllRnaModsList` now holds two entries with the same target, same `C1H2` formula, and contradictory fragmentation behaviour; whichever one a search picks changes the predicted ion series for the single most common RNA modification class.

The inference is the problem, not the measurement: an empty base-localized portion is consistent with *both* `Default` and `Suppressed`, and the ions cannot separate them. That is the same situation the `matchingPortions.Count != 1` branch just above already handles correctly by declining to guess.

**Fix:** don't map "empty portion" onto `Default`. Either detect ribose localization (the `2'-O-` prefix is explicit in the MODOMICS name) and emit `Suppressed`, or return `null` for the empty-portion case so the entry loads as a plain `Modification` and the curated `RnaMods.txt` entry stays authoritative. Both are better than asserting normal base loss at a residue where the curated file says it is suppressed.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Fixed. `m3C` now neutralizes to the CH2 residue shift instead of keeping the extra proton, and I added stronger coverage around the registered Modomics entry as follow-up validation. There is not a curated `RnaMods.txt` counterpart for `m3C`, so the extra consistency check uses the loader result plus the Modomics registry surface.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Fixed this by neutralizing the cap formulas before deriving the residue shift and by removing only the base-less ribose scaffold for the capped residue. I also updated the cap assertions so the expected formulas are mass-balanced.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Fixed by neutralizing the phosphorus-bearing `nucleoside` rows before deriving the residue shift, so `Ar(p)` and `Gr(p)` no longer load light. I added focused assertions for both entries.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Fixed. The loader now stores diagnostic ions as neutral masses and still uses the published ion m/z values only for base-loss inference, which matches the downstream `DiagnosticIons` contract. The focused Modomics loader tests are green after the update.

nbollis:
I addressed the specific actionable review findings above. I did not have any additional concrete 'minor comments' attached to this note to act on, but I can follow up if there are particular remaining nits you want called out separately.

nbollis:
Thanks — I addressed the four blocking loader issues from this review: cap mass handling, cationic/anionic formula neutralization, and diagnostic-ion unit storage. I left the broader reconciliation-report idea out of this pass and treated it as separate follow-up scope. Focused `ModomicsLoaderTests` are passing locally (13/13).

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Fixed with a conservative scope. The loader now emits `BaseLossBehavior.Suppressed` for explicitly confirmed `2'-O-...` Modomics entries, including mixed cases like `N6,2'-O-dimethyladenosine`, while keeping the measured base-localized portion when present. I left broader suppression inference out for now because the product ions alone do not prove it; this should likely be expanded as more cases are explicitly validated.

mzLib/Omics/Modifications/Mods.cs
```diff
@@ -28,7 +30,7 @@ static Mods()
             .ToDictionary(m => m.IdWithMotif);
 
         LoadAllRnaModifications();
-        AllRnaModsList = MetaMorpheusRnaModifications.ToList();
+        AllRnaModsList = MetaMorpheusRnaModifications.Concat(ModomicsRnaModifications).ToList();
```
trishorts, line 33: **Major — folding MODOMICS into `AllKnownMods` changes what a mass-only lookup returns for *protein* modifications.**

`AllRnaModsList` feeds `AllKnownMods`, which is the candidate set for `GlobalModificationLookup` (`GlobalModificationLookup.cs:23`). When `TryResolve` gets a mass with no target residue, `SelectBestCandidate` skips the residue filter, finds all survivors share one chemical formula, and returns **the shortest `OriginalId`** (`ModificationLookupBase.cs:414-416`).

The MODOMICS entry `inosine` is 7 characters and carries `H-1N-1O` — the deamidation formula. It is now shorter than every protein candidate.

I ran it on both refs. Same probe, `CanonicalModification.AtNTerminus(mass: 0.984015583, targetResidue: null)` through `GlobalModificationLookup.Instance.TryResolve`:

| | `AllKnownMods` | candidates within 1 mDa | resolves to |
|---|---|---|---|
| merge-base `a8250440` | 3278 | 13 | **`Asn->Asp on N`** |
| this branch | 3445 | 14 | **`inosine on A`** |

The new candidate list, sorted the way the tie-break sorts it: `inosine`(7), `Asn->Asp`(8), `Gln->Glu`(8), `Deamidated`(10), `Citrulline`(10), `Deamidation`(11), `Citrullination`(14). A +0.98402 Da shift on a peptide now annotates as an RNA nucleoside.

Two things that narrow this, so it is sized fairly:

- It needs the residue-less path. With a target residue the filter runs first and protein mods win.
- I checked methyl as a control and it does **not** flip — `+14.01565` resolves to `Methyl on E` on both refs. So this is the inosine/deamidation collision specifically, not a general widening.

**Suggested fix:** keep the MODOMICS set additive. `ModomicsRnaModifications` and `ModificationNamingConvention.Modomics` are both new surface and cost nothing; the question is whether line 33 should fold them into the pre-existing `AllRnaModsList`/`AllKnownMods` by default. If the merge is deliberate, a regression test pinning mass-only resolution of `0.98402` would at least make the choice explicit.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,692 @@
...
+        }
+
+        // Every published protonated ion becomes a neutral diagnostic mass.
+        var ions = productIons.Select(p => p.ToMass(1)).ToList();
+        var diagnosticIons = new Dictionary<DissociationType, List<double>>
+        {
+            { DissociationType.AnyActivationType, ions },
+        };
+
+        var measured = moietyType == "nucleoside"
```
trishorts, line 413: **Major — 11 of the 31 `2'-O-` entries never become a `BaseModification` at all, so the suppression rule never reaches them.**

`IsExplicitlyConfirmedSuppressedBaseLoss` is only consulted after `DeriveBaseLossModification` returns non-null. Two earlier exits skip it: `productIons is null` (line 369) and `measured is null` (line 388).

I loaded the resource and counted every mod whose `OriginalId` contains `2'-O-`:

```
total 2'-O- mods = 31 | Suppressed = 20 | plain Modification (not BaseModification) = 11
```

The 11 that miss it:

```
2'-O-ribosyladenosine (phosphate)                    ions=null
2'-O-ribosylguanosine (phosphate)                    ions=null
2'-O-methyluridine 5-oxyacetic acid methyl ester     ions=null
2'-O-methyl-5-hydroxymethylcytidine                  ions=null
5-carboxymethylaminomethyl-2'-O-methyluridine        ions=YES
5-carbamoylmethyl-2'-O-methyluridine                 ions=YES
1,2'-O-dimethylinosine                               ions=YES
2'-O-methylpseudouridine                             ions=YES
2-thio-2'-O-methyluridine                            ions=YES
2'-O-methylinosine                                   ions=YES
5-(isopentenylaminomethyl)-2'-O-methyluridine        ions=YES
```

Worth noting because it changes the fix: **8 of the 11 do have published product ions.** So "no product ions" is only part of it — those 8 reach `DeriveBaseLossModification` and it returns null, presumably because the ions match no unique portion or two indistinguishable ones. Guarding only the `productIons is null` path would still leave 8 of them wrong.

The result is that chemically identical 2'-O ribose modifications get two different fragmentation contracts depending on what MODOMICS happened to publish, and `2'-O-methylpseudouridine` and `2'-O-methylinosine` — ordinary, well-characterised residues — land on the wrong side.

**Suggested fix:** evaluate the name-based rule for every nucleoside entry independently of the ion-derived measurement, emitting `BaseLossBehavior.Suppressed` with a null `BaseLossModification` when no portion could be measured, and keep the ion-derived portion as a refinement rather than a precondition.

trishorts (commented):
Full review of the MODOMICS loader. This is a lot of careful work, and the derivation approach — measuring the base-localized portion from the published primary ion rather than assuming all-on-base — is genuinely nice; the `2'-O-methyladenosine` vs `3-methylcytidine` pair in the tests makes the case well.

Four things below. I ran every one of them rather than reading for them, and where I quote a number it came out of a probe on this branch, checked against the merge-base `a8250440` where the question was "is this a regression?".

The one I would most want a decision on is the first: a mass-only modification lookup with no target residue now returns the RNA mod `inosine` where master returned `Asn->Asp`, because `inosine` is the shortest `OriginalId` among the candidates sharing the deamidation formula. That reaches protein-side consumers, which I do not think is intended by "add MODOMICS RNA mods".

Two things I checked and am deliberately **not** reporting, so they do not get re-litigated later:

- The general claim that folding MODOMICS into `AllKnownMods` broadly changes protein-side resolution does **not** hold. I used `+14.01565` as a control and it resolves to `Methyl on E` on both refs. Only the inosine/deamidation formula collision actually flips.
- The static-constructor fragility around `LoadAllRnaModifications` is pre-existing, not introduced here — master already ran the same unguarded loads.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,692 @@
...
+    private static ProductIonInterpretation? InterpretProductIons(
+        string modomicsName, List<double>? productIons, string moietyType, ReferenceMoietyDefinition moiety, ChemicalFormula modificationFormula)
+    {
+        if (productIons is null)
+        {
+            return null;
+        }
+
+        // Every published protonated ion becomes a neutral diagnostic mass.
+        var ions = productIons.Select(p => p.ToMass(1)).ToList();
```
trishorts: **Question — the secondary diagnostic ions are 40-85 ppm off, which is wider than any product tolerance they would be matched at.**

Every published token goes through `p.ToMass(1)`, i.e. `nominal - 1.00728`, and `ModomicsDto.ProductIons:53` documents these as nominal integer m/z. Only the primary ion is replaced with a computed accurate mass (line 393). I can see from the comment at `ModomicsLoaderTests.cs:214` that this is deliberate, so I am raising it as a question rather than a defect.

The reason I would still raise it: `Modification.DiagnosticIons` is consumed as monoisotopic neutral masses and matched at ppm tolerance (`PeptideWithSetModifications.cs:547-548`), and the error is well outside that. Measured on this branch for `5-carboxymethyl-2-thiouridine` (`"187/169/141"`), deriving the expected secondaries from the *accurate* primary the loader already computes:

```
stored ions:  186.0099, 167.9927, 139.9927

primary  186.0099  (accurate, computed)
ion 2    167.9927  vs primary-H2O    = 167.9993   ->  -6.6 mDa   (-39 ppm)
ion 3    139.9927  vs primary-H2O-CO = 140.0044   -> -11.7 mDa   (-84 ppm)
```

Scope: 132 loaded mods carry diagnostic ions; **26 of them have more than one**, so 26 mods ship at least one ion that cannot match at a realistic tolerance.

For the record, the accurate values are derivable from what the loader already has — the comment at line 415 states the relationship, "every other ion is that cation minus a small neutral such as water or ammonia." Deriving them, or publishing only the primary and keeping the nominal remainder somewhere that does not claim sub-Dalton precision, would both work. Not blocking if you would rather land it and refine later — but the current state stores a nominal integer in a field typed and consumed as an exact mass, and nothing marks which entries are which.

mzLib/Test/Transcriptomics/ModomicsLoaderTests.cs
```diff
@@ -0,0 +1,335 @@
...
+            var report = ModomicsLoader.LoadModomics();
+
+            Assert.That(report.LoadedModifications, Is.Not.Empty);
+            Assert.That(report.LoadedCount, Is.EqualTo(report.LoadedModifications.Count));
+            Assert.That(report.LoadedModifications.All(m => !string.IsNullOrWhiteSpace(m.IdWithMotif)), Is.True);
+            Assert.That(report.LoadedModifications.All(m => m.IdWithMotif.Contains(" on ")), Is.True);
+        }
+
+        [Test]
+        public void LoadModomics_SkipsUnknownBases()
```
trishorts, line 26: **Minor — this assertion cannot fail as written.**

The loader only ever emits motifs it recognises; everything else is diverted to `NotYetRepresentableEntries`. Confirmed on this branch:

```
distinct motifs loaded : A, C, G, U
any "on X"             : False
loaded                 : 167
not representable      : 275
```

So `Any(m => m.IdWithMotif.Contains("on X"))` is false by construction, and would stay false even if the unknown-base handling regressed in most plausible ways — the failure mode this is guarding against would show up as an entry landing in the wrong *known* motif, or as a silent drop, neither of which this catches.

`LoadModomics_TracksNotYetRepresentableEntriesWithReasons` just below already does the real work by asserting on `Reason`. Pinning the count of `UnsupportedReferenceMoiety` entries, or asserting that a specific known-generic entry (`"xG"`, `"xX"`) appears in `NotYetRepresentableEntries` and not in `LoadedModifications`, would make this one earn its place.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Acknowledged — you're right that the secondary ions are stored as `nominal - 1.00728` neutral masses rather than true monoisotopic masses, and that the test comment documents this as deliberate. I'm treating this as a known limitation for this pass rather than a blocker, since the primary ion (the one actually used for base-loss measurement) carries the accurate computed mass and you flagged it non-blocking. Deriving the secondaries from the accurate primary (per the line-415 relationship) is a sensible follow-up — I'll revisit it after this PR lands. Nothing in the stored representation claims sub-Dalton precision beyond the primary; noted that 26 multi-ion mods ship at least one nominal secondary today.

mzLib/Omics/Modifications/Mods.cs
nbollis: Agreed — folding MODOMICS into `AllKnownMods` flipped a residue-less mass-only lookup of the deamidation mass to the RNA mod `inosine`. Adopted the additive option: `AllKnownMods` is now built from protein mods + curated `MetaMorpheusRnaModifications` only, while `ModomicsRnaModifications` remain fully addressable through `AllRnaModsList`, `AllKnownRnaModsDictionary`, and the `Modomics` naming convention. Added a regression test (`TryResolve_ResidueLessDeamidationMass_ResolvesToProteinModNotRnaInosine`) pinning a `0.984015583` mass shift with no target residue back to `Asn->Asp on N`. The `+14.01565` methyl control still resolves to `Methyl on E` as you verified.

mzLib/Test/Transcriptomics/ModomicsLoaderTests.cs
nbollis: Fair — the `Any("on X")` assertion was vacuous by construction. I reworked `LoadModomics_SkipsUnknownBases` to pin the actual failure mode: it asserts the generic-purine stand-in `xX` (Modomics id 98) appears in `NotYetRepresentableEntries` with `EmptyFormula` and the `X` reference moiety, and that neither it nor any `on X` IdWithMotif lands in `LoadedModifications`. This now fails if the unknown-base handling regresses into a wrong known motif or a silent drop.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
nbollis: Good catch on the 8-with-ions case — the null-ions guard alone would have missed them. `InterpretProductIons` now evaluates the `2'-O-` name rule first, independently of both the ion measurement and the presence of published ions: a name-marked nucleoside with no ions, or with ions that measure no unique portion, now emits `BaseLossBehavior.Suppressed` with a null `BaseLossModification`, while the measured base-localized portion is kept as a refinement when it resolves (e.g. `N6,2'-O-dimethyladenosine` stays Suppressed with its `C1H2` portion). `2'-O-methylpseudouridine`/`2'-O-methylinosine` now land on the Suppressed side like their chemically identical cousins. Verified against the focused `ModomicsLoaderTests` run (38 passing).

mzLib/Omics/Modifications/Mods.cs
trishorts: Verified on `55b4799a`. `AllKnownMods` is back to the merge-base content — I measure **3278**, the same number I reported for `a8250440` — `inosine` is no longer in it, and the MODOMICS set stays fully reachable through `AllKnownRnaModsDictionary` (3345 keys, `inosine on A` present) and `ModsByConvention[Modomics]`. The new regression test pins the residue-less deamidation lookup back to `Asn->Asp`. Resolving.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
trishorts: Verified. Loading the resource on this branch, all **31 of 31** `2'-O-` entries now come back as `BaseModification` with `BaseLossType = Suppressed` — previously 20 Suppressed / 11 plain. That includes the 8 that publish ions `DeriveBaseLossModification` can't resolve, which was the part a null-ions guard alone would have missed. Emitting `Suppressed` together with a measured partial formula is also exactly the `2-Om6` case documented on `BaseModification.cs:20`, and `ValidModification` is true for all 167 loaded mods. Resolving.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
trishorts: Agreed, and that's the right call — I raised it as a question rather than a defect precisely because the primary ion (the one actually used for base-loss measurement) carries the computed accurate mass. Resolving as a known limitation; worth an issue if the secondary ions ever get matched at ppm tolerance.

mzLib/Test/Transcriptomics/ModomicsLoaderTests.cs
trishorts: Verified: `xX` is the sole `ShortName == "xX"` entry, tracked with `Reason = EmptyFormula` and reference moiety `X`, and neither an `on X` `IdWithMotif` nor an `"unknown modification"` `OriginalId` reaches `LoadedModifications`. The assertion can now actually fail. Resolving.

mzLib/Omics/Modifications/IO/Modomics/ModomicsLoader.cs
```diff
@@ -0,0 +1,720 @@
...
+
+        if (measured is null)
+        {
+            // Caps, and nucleosides whose ions do not measure a unique base-localized portion: keep
+            // every ion verbatim with plain base-loss semantics.
+            return new ProductIonInterpretation(diagnosticIons, BaseLossType: null, BaseLossModification: null);
+        }
+
+        // The measured topology gives the primary ion its accurate neutral mass.
+        var primaryIonIndex2 = productIons.IndexOf(productIons.Max());
```
trishorts, line 425: **Minor, non-blocking — `primaryIonIndex2`.**

The rename is forced (CS0136 against the inner `primaryIonIndex` at line 395), but the numbered name is the symptom rather than the problem: lines 393–399 and 425–429 are now the same "give the primary ion its accurate neutral mass" block, and they have to be kept in sync by hand.

Lifting it into a local function would let both sites keep the original name and drop the duplication:

```csharp
static void ApplyAccuratePrimary(List<double> ions, List<double> published, double accurateMass)
{
    var index = published.IndexOf(published.Max());
    if (index >= 0) ions[index] = accurateMass;
}
```

Purely a readability call — the behaviour is right either way, and I confirmed both paths produce the expected masses.

trishorts (approved):
Approving. Round-3 verification on `55b4799a`, run against the branch rather than read off the diff:

| Finding | Was | Now |
|---|---|---|
| MODOMICS folded into `AllKnownMods` | 3445 candidates, `+0.98402` residue-less → `inosine on A` | 3278 — identical to merge-base `a8250440`; `inosine` out; regression test pins `Asn->Asp` |
| 11 of 31 `2'-O-` entries never became a `BaseModification` | 20 Suppressed / 11 plain | **31 / 31 Suppressed**, including the 8 with unmeasurable published ions |
| Secondary diagnostic ions 40–85 ppm off | — | accepted as a documented limitation (raised as a question, not a defect) |
| Vacuous `on X` assertion | could not fail | pins `xX` → `EmptyFormula` + moiety `X`, absent from `LoadedModifications` |

Also checked: `ValidModification` is true for all 167 loaded mods; `Suppressed` carrying a measured partial formula is the documented `2-Om6` case (`BaseModification.cs:20`), not a contradiction; MODOMICS mods stay name-addressable via `AllModsKnownDictionary` and `ModsByConvention[Modomics]`, so the additive fix narrows the mass-only candidate set without hiding anything.

`dotnet test` across Omics / Transcriptomics / Modification: **1417 passed, 0 failed**. CI green on all five checks.

One non-blocking nit inline. The known limitations you list in the PR body — backbone mods and base-loss fragmentation for the phosphate-spanning entries — read as reasonable follow-up scope to me, not merge blockers.

Alexander-Sol (approved):

PR #955, opened 2025-09-16
Ptm stoichiometry

pcruzparri:
New PR with changes from PR #797 and more. 

Description: 
PTM stoichiometry classes created for each structural level (protein group -> protein -> (base) peptide - modification). Probably the most important class to understand is the QuantifiedPeptide class for which an object represents a collection of post-translationally modified variants sharing the same base sequence. This is in contrast with other Peptide classes that represent a peptide with a given _full_ sequence. The `QuantifiedProtein` class, which stores the `QuantifiedPeptide` objects, handles peptide-to-protein indexing and obtaining of the modification stoichiometry for the protein. The `QuantifiedModification` class is primarily a data class, and the `QuantifiedProteinGroup` class is mostly a container for the different `QuantifiedProtein` members of the protein group. Lastly, the `PositionFrequencyAnalysis` class has a method to take a list of (full sequence, protein group list, intensity) tuples and create a collection of `QuantifiedProteinGroup` objects.


acesnik:
Hi there! @trishorts asked me to take a look at this one. It seems like a great pull request for getting modification ratios from results.

On stoichiometry, it's worth considering Jesper Olsen's work on stoichiometry measurements back in 2010 and what's been going on since then. Getting a true stoichiometry measure has required some tricks in the past to get the fully unoccupied quantity. I think the ratios might be useful, but I've struggled in the past when I've calculated such ratios to claim confidently that they're stoichiometries. You could consider using that original data to compare how ratios stack up to the measured stoichiometries if you're writing this up in a paper. https://www.science.org/doi/abs/10.1126/scisignal.2000475

I haven't looked at this logic in a long time, and it looks like it's been changed a bit, but do you think there is work to be done in `mzLib/Omics/BioPolymer/VariantApplication.cs`, or do you think there's logic to add to these new methods regarding sequence variations? Proteins with amino acid variations may have modifications listed within those data structures, https://github.com/acesnik/mzLib/blob/master/mzLib/Omics/BioPolymer/SequenceVariation.cs#L24. Should those be quantified and tested for these ratio calculations in the situations when amino acid variations are added?

pcruzparri:
@acesnik thank you for your thoughts and the paper you shared! I had not come across it and still need to get through it more closely, but it seems like a great reference for a paper. 

You are absolutely right that this calculation just outputs the residue-specific intensity ratios, and that is not the full picture of a site-occupancy calculation. I still need to compare the values obtained from this analysis to the residue-specific PSM count ratios we currently output in MetaMorpheus, but the goal is to output something quantitatively closer to a mod's stoichiometry. I'm thinking that the correction/scaling of intensities to account for peptide ionization efficiency would be a follow-up enhancement to this code (or maybe on the MetaMorpheus end?). There is still a little more exploring I'd like to do before deciding on which approaches we'd like to support on mzLib or MetaMorpheus for stoichiometry calculation, but at least whoever would like to implement their own normalization/scaling strategies can do so from the output being facilitated here in the meantime. 

As for sequence variations, variants in MetaMorpheus will get their own accessions, be treated as different proteins, and be placed accordingly into protein groups. Once the protein groups and variant sequences are extracted, this code will treat it as any other protein group. As for using this code directly from mzlib, sequence variants would need to be passed in as different proteins. In either case, the code here (specifically, the `SetUpQuanitficationFromFullSequences()` method) is written to only take in a full peptide sequence, the protein group names the full sequence belong to, and the peptide's intensity and then create new quantification objects. All mod information is parsed from the full sequence string and localized using the full sequence and the provided protein sequences (which would require passing each variant protein sequence separately).  

Please let me know if you have further thoughts/concerns I can address!

mzLib/MzLibUtil/PositionFrequencyAnalysis/PositionFrequencyAnalysis.cs
```diff
@@ -0,0 +1,68 @@
+﻿using Easy.Common.Extensions;
+using System.Collections.Generic;
+
+namespace MzLibUtil.PositionFrequencyAnalysis
+{
+    public class PositionFrequencyAnalysis
+    {
+        public Dictionary<string, QuantifiedProteinGroup> ProteinGroups { get; private set; }
+
+        //public Dictionary<string, (QuantifiedPeptide QuantifiedPeptide, string ProteinGroups)> Peptides { get; private set; }
```
Alexander-Sol: Remove commented out property

mzLib/MzLibUtil/PositionFrequencyAnalysis/QuantifiedProtein.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            foreach (var peptide in Peptides.Values)
+            {
+                // if peptide position in protein is unknown, set it using the protein sequence
+                if (peptide.OneBasedStartIndexInProtein == -1)
+                {
+                    peptide.OneBasedStartIndexInProtein = Sequence.IndexOf(peptide.BaseSequence) + 1;
+                }
+                // if peptide has no modifications, add to all its positions
+                if (!peptide.ModifiedAminoAcidPositions.IsNotNullOrEmpty())
```
Alexander-Sol: IsNullOrEmpty() would be slightly cleaner

mzLib/Test/TestMzLibUtil.cs
```diff
@@ -163,8 +169,192 @@ public void TestRemoveSpecialCharacters()
             string cleanSeq = seqWithHash.ToString();
             ClassExtensions.RemoveSpecialCharacters(ref cleanSeq, specialCharacter: "#");
             Assert.AreEqual("PEPTIDE", cleanSeq);
+        }
+
+        [Test]
+        public void TestQuantifiedModification()
+        {
+            var quantmod = new QuantifiedModification(idWithMotif: "TestMod: ModX on AAY", positionInPeptide: 1, positionInProtein: 2, intensity: 10);
```
Alexander-Sol: IdWithMotif just refers to the part after the colon. The full mod string is {Modification Type}: {Id with motif}

Alexander-Sol (changes requested):
Left a couple of review comments. Two things that jump out:
- Lack of summary comments. There are a lot of dictionaries in use, and I'm not always sure what the keys are just from looking at them. It would be nice to have summary comments that explain the dictionaries
- IdWithMotif doesn't include the modification type (the part before the colon).

mzLib/MzLibUtil/PositionFrequencyAnalysis/QuantifiedPeptide.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <summary>
+    /// A class to store information about a quantified peptides sharing the same base sequence.
+    /// </summary>
+    public class QuantifiedPeptide
+    {
+        public HashSet<string> FullSequences { get; set; }
+        public string BaseSequence { get; set; }
+        public QuantifiedProtein ParentProtein { get; set; }
+        public int OneBasedStartIndexInProtein { get; set; }
+        public Dictionary<int, Dictionary<string, QuantifiedModification>> ModifiedAminoAcidPositions { get; set; }
```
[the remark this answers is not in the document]
pcruzparri: The string stores the name of the modification, just for easier/faster access to the mod object. Added extra commenting in latest commit.

Alexander-Sol (dismissed):

mzLib/MzLibUtil/PositionFrequencyAnalysis/QuantifiedProteinGroup.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Represents a group of proteins for quantification purposes.
+    /// </summary>
+    public class QuantifiedProteinGroup
+    {
+        /// <summary>
+        /// The name of the protein group, typically a concatenation of protein accessions in the 
+        /// format "ProteinA;ProteinB", "ProteinA|ProteinB", or "ProteinA;ProteinB|ProteinC".
+        /// </summary>
+        public string Name { get; set; }
```
Alexander-Sol: Could you also include optional properties for the Gene Name and the Organism?

mzLib/MzLibUtil/ClassExtensions.cs
```diff
@@ -25,6 +25,9 @@ namespace MzLibUtil
 {
     public static class ClassExtensions
     {
+        public static readonly string ModificationPattern = @"-?\[(.+?)(?<!\[I+)\]";
```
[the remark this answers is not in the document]
pcruzparri: this has nothing to do with ambiguous mods, so pipe is irrelevant.

trishorts (dismissed):

mzLib/MassSpectrometry/ScanMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+/// <param name="NativeId">Vendor-native scan identifier string.</param>
+/// <param name="FullFilePath">Absolute or relative path to the originating spectra file.</param>
+/// <param name="PrecursorCharge">Charge state assigned to the deconvoluted precursor.</param>
+/// <param name="PrecursorMonoisotopicPeakMz">Monoisotopic m/z of the deconvoluted precursor.</param>
+/// <param name="PrecursorMass">Neutral monoisotopic mass of the precursor, derived from m/z and charge.</param>
+/// <param name="PrecursorIntensity">MS1 intensity of the precursor ion.</param>
+/// <param name="PrecursorEnvelopePeakCount">Number of peaks in the precursor isotopic envelope.</param>
+/// <param name="PrecursorFractionalIntensity">Fraction of precursor intensity relative to envelope total. -1 if unavailable.</param>
+/// <param name="OneOverK0">Inverse reduced ion mobility (1/K0) for TIMS data; null for non-IMS instruments.</param>
+public record ScanMetadata(
```
Alexander-Sol, line 28: Really like this

mzLib/Omics/BioPolymerGroup/BioPolymerGroupType.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// Protein group — occupancy is calculated at protein-level coordinates
+    /// using <see cref="ModificationOccupancyCalculator.CalculateProteinLevelOccupancy"/>.
+    /// </summary>
+    Protein,
+
+    /// <summary>
+    /// Peptide group — occupancy is calculated in peptide-local coordinates
+    /// using <see cref="ModificationOccupancyCalculator.CalculatePeptideLevelOccupancy"/>.
+    /// </summary>
+    Peptide,
```
nbollis: Why are peptide and oligo separate?

Alexander-Sol (dismissed):

mzLib/Omics/BioPolymerGroup/SampleGroupResult.cs
```diff
@@ -0,0 +1,144 @@
+using MassSpectrometry;
+
+namespace Omics.BioPolymerGroup;
+
+/// <summary>
+/// Bundles quantification and modification occupancy data for a single sample group
+/// (Condition × BiologicalReplicate). Each group contributes 2 columns (SpectralCount + CountOccupancy)
+/// or 4 columns (+Intensity + IntensityOccupancy) when intensity data is available.
+/// </summary>
+public sealed class SampleGroupResult
```
nbollis, line 10: Should this implement ISampleInfo?

trishorts: i think that would be a good idea

mzLib/Omics/BioPolymerGroup/BioPolymerGroupType.cs
pcruzparri: Fixed. When first written, I separated the two in case i wanted to have specific methods/options checking for different localization names, though the implementation didn't really address it. I updated the enum to just have two types, namely "Parent"(addressing BioPolymer objects) and "DigestionProduct" (addressing BioPolymerWithSetMods objects).

nbollis (approved):

trishorts (approved):

PR #956, opened 2025-09-17
Whats wrong with variant application and xml

trishorts:
Overview of Major Changes
1.	SequenceVariantDescription.cs renamed VariantCallFormat.cs – this was done for accuracy. All of the parameters of the old class actually map onto VCF file format. This class was not meant to contain generic data about sequence variants. So renaming improves expectations.
2.	VariantCallFormat.cs extended in updated. All of the information about reading VCF files was done several years ago and unsurprisingly there have been updates to the information provided. I examined what is presently reported in VCF files and enabled the newly named method to be able to retrieve all of that information accurately. This revised class should now be able to handle most anything that we throw at it. Depth and Zygosity are key parameters. We use depth to filter which variants get applied. And Zygosity tells us if we have a complete or partial change when applying the variant.
3.	SnpEffAnnotation revisions. Similar story to VCF. Revisions were mostly to better handle string splitting. SnpEff appears out of favor. I don’t expect lots of use. We are keeping this for backwards compatibility. The large block of commented code is for historical reference. It was and is not used. But, it does explain some things. Comments explaining the comments are added.
4.	SequenceVariation revisions. One stop to handle variants from GPTMD, UniProt, VCF. There are a couple of versions because information passed can be different. Sometimes there is only a begin position and sometimes begin and end. We convert everything to the latter. Sequence variants have a modification dictionary because sequence variants can have modifications. In the case where we have a point mutation, it might seem like overkill. But we could have two different mods at the same point. In other cases where we have multi-amino acid change, the modification can be somewhere along the string. Keep a list of mods and positions helps. More later in VariantApplication. Description sometimes comes from VCF and sometimes it comes from UniProt. There is some difference in handling of this parameter. AreValid was a key method as not everything sent in is actually valid. Go figure. Split multi-sample VCF metadata was super confusing to me. But, it turns out that not every line in a VCF describes a single observation. In fact you could have a 100 different samples on the same vcf line. Each would have their own read depth and zygosity. The code here handles the variety of samples thrown at it. 
5.	VariantApplication – It was not straightforward deciding how to combine UniProt variants and VCF variants because VCF has depths and zygosity and uniprot does not. Also, counts were bugged. There was no way to specify how many variants per isoform and how many total isoforms. This had to be redone. And many old unit tests were effect by this change. A key in generating combinations of isoforms was making sure that there were no variant collisions. This happens routinely when you have two variants attempting to change the same sequence or sequence that has totally disappeared once a big truncation is applied. This is new code. There are a couple key aspects of applying variants. One, the sequence length can change, which changes all modification coordinates. Two, and modifications on a sequence variant need to get promoted to the variant protein modification dictionary so they are in the correct spot for digestion or not, etc. Then, the applied sequence variant needs to be removed from the list of sequence variants so its not hanging around like a sausage. Formerly, one only got variants from this method. Now it returns the consensus too. Still working to resolve best handling of this.
6.	PeptideWithSetModifications. There were some big changes in the IntersectsAndIdentifiesVariation method. This method is replete with edge cases and was formerly bugged. Revised so all the old unit tests pass and better handling of edge cases. Many more comments. 
7.	DecoyProteinGenerator. Manual addition of nics write decoys method. This can be  complicated with variants. When reversing a sequence one also needs to adjust all the coordinates of the sequence variants and their modifications. An extra complication is when the variant is at the start and we don’t reverse the start. 
8.	RNAdecoygenerator see decoyproteingenerator.
9.	Proteindbloader has new parameters for LoadProteinXML that indicate the desired number of variants per isoform, required minimum sequencing depth and total combinatorial number of isoforms desired. Method converts and variants discovered as modifications in GPTMD to actual sequence variants. All constructors are temporarily retained that convert maxHeterozygousVariants to maxVariantIsoforms etc. 
10.	ProteinDbWriter.  Fasta change b/c write decoys changes. 

Core Additions (SequenceVariation)
•	New semantic / convenience properties:
•	SearchableAnnotation (VCF line preferred over Description).
•	ReferenceAllele / AlternateAllele passthrough to VariantCallFormat.
•	Classification flags: IsPointSubstitution, IsMultiResidueSubstitution, IsInsertion, IsDeletion, IsStopGain, IsLikelyFrameshift.
•	LegacyVariantDescription (Obsolete shim).
•	Genotype expansion:
•	SplitPerGenotype(...) creates per‑sample variants with depth filtering, heterozygous handling, and optional reference emission (suppresses invalid no‑ops).
•	Combination utilities:
•	CombineEquivalent(...) merges equivalent variants (merges modifications, aggregates descriptions).
•	Modification management:
•	TryAddModification / AddModifications with positional validation against edited span, stop gains, deletions.
•	Validation of OneBasedModifications during construction (throws on invalid positions).
•	Enhanced validation logic:
•	AreValid() now rejects “no‑op” variants (OriginalSequence == VariantSequence and no variant‑specific modifications).
•	Frameshift heuristic (length change without insertion/deletion classification and not a stop gain).
•	Equality / hashing:
•	Description intentionally excluded from Equals / GetHashCode.
•	Modification dictionaries compared as position + multiset of modification identifiers.
•	SimpleString() retained; spans use Begin-End form.
•	Robust constructor overload accepting parsed VariantCallFormat or raw VCF string.
Core Additions (VariantCallFormat)
•	Lightweight VCF parser with:
•	REF / ALT allele capture.
•	ANN (SnpEff) extraction via SnpEffAnnotation (assumed existing).
•	FORMAT token parsing into per‑sample dictionaries.
•	GT parsing (handles '/', '|' separators; preserves '.' as missing).
•	AD parsing with validation + numeric extraction (TryParseAD).
•	AlleleIndex resolution against ANN’s allele.
•	Zygosity classification per sample (Unknown / Homozygous / Heterozygous).
•	Legacy boolean maps (Homozygous / Heterozygous) retained.
•	Truncated line tolerance (IsTruncated).
•	Validation helpers for restricted vocabularies.
Core Changes / Refactors (VariantApplication)
•	GetVariantBioPolymers:
•	Early exits clarified; over‑strict validation fallback (uses raw variants if all fail AreValid()).
•	Uses new ApplyAllVariantCombinations pipeline.
•	ApplyAllVariantCombinations:
•	Genotype-aware expansion via SplitPerGenotype before combination enumeration.
•	Collapses equivalent variants first (CombineEquivalent).
•	Filters no‑ops aggressively.
•	Deterministic combination generation; guards for invalid / overlapping variants (ValidCombination).
•	Improved safety / resilience:
•	Widespread try/catch wrappers prevent a single malformed variant from collapsing processing.
•	SafeAreValid helper.
•	ApplySingleVariant:
•	Rebuilds SequenceVariation instance for applied result (keeping original span semantics).
•	Adjusts truncation products, modifications, and subsequent variant coordinates.
•	Handles premature stops (sequence truncation) and length recalculations (mass & length attribute normalization via reflection).
•	Index adjustment helpers:
•	AdjustSequenceVariationIndices: careful overlap & same VCF record handling.
•	AdjustTruncationProductIndices: stop-gain aware length clipping.
•	AdjustModificationIndices: preserves or shifts mods, avoids out-of-range issues.
•	Name/accession formatting:
•	GetVariantName aggregates de‑duplicated descriptors (VCF preferred).
•	CombineDescriptions / CombineSimpleStrings centralize formatting.
•	Nucleotide substitution mod conversion:
•	ConvertNucleotideSubstitutionModificationsToSequenceVariants converts “nucleotide substitution” modifications into SequenceVariation entries, pruning the original modifications.
•	Data sanitation:
•	SanitizeVariantData(...) drops invalid variants, prunes impossible modification coordinates, reconciles applied variant list, yields human-readable notes.
Behavioral / Validation Shifts (Potential Breaking Changes)
•	“No‑op” variants (no sequence change and no variant-specific modifications) now invalid (throw in constructor or AreValid() returns false). Existing workflows that stored reference-only variants must adapt (e.g., attach a modification or skip creation).
•	Equality semantics changed (Description ignored, modifications included). Hash-based collections may re-cluster.
•	Modification coordinate validation is stricter (invalid positions now throw on construction rather than being silently kept).
•	SplitPerGenotype filters out heterozygous reference “ref copy” when it would be a no‑op (as asserted in tests).
•	Frameshift heuristic might classify certain previous substitutions differently (downstream logic relying on old classification may need review).
New Test Coverage (SequenceVariationNewPropertiesTests)
•	Verifies SearchableAnnotation precedence (VCF over description).
•	Checks allele passthrough, classification predicates, stop-gain vs frameshift disambiguation.
•	Ensures constructor rejects no‑ops unless variant-specific modifications exist.
•	Validates modification position enforcement & deletion + modification rejection.
•	Genotype splitting logic (only alt entries produced when ref copy invalid).
•	CombineEquivalent merges descriptions and modifications.
•	Equality semantics ignoring Description.
•	Convenience constructor end coordinate logic.
•	SimpleString formatting rules.
•	LegacyVariantDescription continuity.
Reliability / Robustness Improvements
•	Defensive cloning of modification lists when combining or splitting variants.
•	Frequency-based multiset comparison for modifications (stable deterministic hashing).
•	Reflection-based safe mass/length updates tolerant of missing or ambiguous residues.
•	Comprehensive guards to prevent index drift after frameshifts / stops.
•	SanitizeVariantData provides a pre-write data hygiene pipeline.
Performance Considerations
•	Combination generation uses iterative index method (avoids recursion allocations).
•	Early dismissal of invalid or depth-insufficient genotypes.
•	OrderBy stages localized (only where deterministic ordering needed).
•	HashCode struct used for consolidated hashing.
Suggested Upgrade Notes for Consumers
1.	Audit any creation of “reference-only” variants; adapt to new validation or skip creating them.
2.	Revisit equality-based dictionaries/sets: Description no longer differentiates variants.
3.	Ensure downstream tools expecting original modification leniency handle earlier exceptions.
4.	If using pre-existing variant combination logic, validate new genotype expansion doesn’t inflate search space unexpectedly; tune maxSequenceVariantsPerIsoform / maxSequenceVariantIsoforms.
5.	Review reporting paths that consumed previous variant name/accession formatting (they may now be longer with multiple descriptors).


zhuoxinshi (dismissed):

acesnik (commented):
I'm a little lost. Is the goal is to make the variant application amenable to UniProt's variant annotations? I see you've extended the SnpEff description to include the whole VCF lines, which makes me think you're also doing some stuff with RNAs? Are you trying to take in new VCF sources?

trishorts:
> I'm a little lost. Is the goal is to make the variant application amenable to UniProt's variant annotations? I see you've extended the SnpEff description to include the whole VCF lines, which makes me think you're also doing some stuff with RNAs? Are you trying to take in new VCF sources?

yes and no. we are doing stuff w/ RNA (we have a whole new RNA-MS that is about to come online). While the creation of RNA databases is also included in this PR (including nucleotide substitutions and modifications), that was not the origin of VCF. I think VCF goes back to spritz, where we wanted to filter the application of certain amino acid substitutions based on read depth of the associated nucleotide variant. Also, we are capturing the result of the variation from VCF (substitution, termination....). So, even though we don't have plans really to do much in this area I wanted to preserve and update the old behavior. Therefore, I looked around to see what is going on in VCF world now and made sure that we could read any kind of VCF data that was sent at us. 

So, no variants that come inside UniProt come w/ vcf data. They just have some description, like a dbsnp reference. So, we have to be able to incorporate those variants using the same code that is used for variants discovered using something like spritz. The VCF line becomes the string description AND it also is decomposed to populate the VCF object which has depth etc.

mzLib/Omics/BioPolymer/VariantCallFormat.cs
```diff
@@ -0,0 +1,331 @@
...
+    /// <code>
+    /// var vcf = new VariantCallFormat(vcLine);
+    /// foreach (var (sampleId, gt) in vcf.Genotypes)
+    /// {
+    ///     var z = vcf.ZygosityBySample[sampleId];
+    ///     var ad = vcf.AlleleDepths[sampleId];
+    /// }
+    /// </code>
+    /// </summary>
+    public class VariantCallFormat
```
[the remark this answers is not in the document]
trishorts: yes. VCF is a better variable name

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,167 +1,857 @@
...
+                                 string? variantCallFormatDataString = null,
+                                 Dictionary<int, List<Modification>>? oneBasedModifications = null)
         {
             OneBasedBeginPosition = oneBasedBeginPosition;
             OneBasedEndPosition = oneBasedEndPosition;
             OriginalSequence = originalSequence ?? "";
             VariantSequence = variantSequence ?? "";
-            Description = new SequenceVariantDescription(description);
+            Description = description;
+            VariantCallFormatData = variantCallFormatDataString is null ? null : new VariantCallFormat(variantCallFormatDataString);
```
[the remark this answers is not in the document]
trishorts: yes. we only use the vcf as text line as description when there is VCF data. in uniprot there is not vcf data. just reference to dbsnp or toerh.

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
```diff
@@ -206,7 +206,7 @@ public static Dictionary<string, int> WriteXmlDatabase(Dictionary<string, HashSe
                     {
                         writer.WriteStartElement("feature");
                         writer.WriteAttributeString("type", "sequence variant");
-                        writer.WriteAttributeString("description", hm.Description.ToString());
+                        writer.WriteAttributeString("description", hm.VariantCallFormatData?.ToString() ?? hm.Description);
```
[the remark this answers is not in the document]
trishorts: we keep it for something like this:
image

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,167 +1,857 @@
...
+            OriginalSequence != VariantSequence;
+
+        /// <summary>
+        /// True if substitution length >1 but same length (multi-nucleotide / multi-amino-acid).
+        /// </summary>
+        public bool IsMultiResidueSubstitution =>
+            OriginalSequence?.Length > 1 &&
+            VariantSequence?.Length == OriginalSequence.Length &&
+            OriginalSequence != VariantSequence &&
+            !IsPointSubstitution;
```
nbollis: !IsPointSubstitution is a redundant check as the length check already does this.

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,167 +1,857 @@
...
+        /// True if a deletion (variant empty).
+        /// </summary>
+        public bool IsDeletion =>
+            string.IsNullOrEmpty(VariantSequence) &&
+            !string.IsNullOrEmpty(OriginalSequence);
+
+        /// <summary>
+        /// True if variant introduces a stop (* at end).
+        /// </summary>
+        public bool IsStopGain => VariantSequence?.EndsWith("*", StringComparison.Ordinal) == true;
```
nbollis: if bool == true?

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,167 +1,857 @@
...
+
+        /// <summary>
+        /// Heuristic frameshift flag: length difference not equal & not simple stop gain only.
+        /// (Refine if you have explicit annotation elsewhere.)
+        /// </summary>
+        public bool IsLikelyFrameshift =>
```
nbollis: What do you mean by frame shift? How can we evaluate a frame shift from only protein sequences?

mzLib/Omics/BioPolymer/VariantApplication.cs
nbollis: The fact that every single function is entirely rewritten is very concerning. Was all of this necessary to create the desired changes?

mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -351,14 +442,15 @@ public static List<Protein> LoadProteinFasta(string proteinDbLocation, bool gene
...
-            return Merge(toReturn).ToList();
+            var toRetrun = generateTargets ? targets.Concat(decoys).ToList() : decoys;
+            return MergeProteins(toRetrun).ToList();
         }
 
         /// <summary>
         /// Merge proteins that have the same accession, sequence, and contaminant designation.
         /// </summary>
-        public static IEnumerable<Protein> Merge(IEnumerable<Protein> mergeThese)
+        // inside MergeProteins(IEnumerable<Protein> mergeThese)
```
nbollis: ??

mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -430,15 +526,119 @@ public static IEnumerable<Protein> Merge(IEnumerable<Protein> mergeThese)
...
-                    );
+                );
             }
         }
+        /// <summary>
+        /// Finds groups of proteins that share the same accession and base sequence.
+        /// Intended to identify cases where an applied-variant entry appears twice
+        /// (e.g., one parsed from XML and another created via variant expansion).
+        /// </summary>
+        internal static IEnumerable<IGrouping<(string accession, string baseSequence), Protein>> FindDuplicateGroupsByAccessionAndBaseSequence(
```
nbollis: Why create a second set of merging duplicate functions when we already have a merge function? 

Is this functionality essential to the proper loading of sequence variants? If so, RNA will need it as well.

mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -323,7 +407,14 @@ public static List<Protein> LoadProteinFasta(string proteinDbLocation, bool gene
                         }
                         else
                         {
-                            targets.Add(protein);
+                            if (protein.IsDecoy)
```
nbollis: Bad if else statements. We just checked if it was a decoy six lines up

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
```diff
@@ -41,632 +44,1045 @@ public static Dictionary<string, int> WriteXmlDatabase(Dictionary<string, HashSe
...
-                foreach (var nucleicAcid in nonVariantRna)
+                WriteEndDocument(writer);
+            }
+
+            return newModResEntries;
+        }
+
+        /// <summary>
+        /// Writes a FASTA file for a list of proteins.
+        /// </summary>
```
nbollis: Why is fasta writing in your files changed?

mzLib/UsefulProteomicsDatabases/ProteinXmlEntry.cs
```diff
@@ -172,19 +157,7 @@ private void ParseEntryAttributes(XmlReader xml)
             DatabaseVersionEntryTag = xml.GetAttribute("version");
             XmlnsEntryTag = xml.GetAttribute("xmlns");
         }
-        /// <summary>
-        /// Parses some attributes of a &lt;sequence&gt; XML element and assigns their values to the corresponding properties of the ProteinXmlEntry.
```
nbollis: Comment was deleted again :(

mzLib/UsefulProteomicsDatabases/ProteinXmlEntry.cs
```diff
@@ -268,20 +219,13 @@ private static UniProtSequenceAttributes.FragmentType ParseFragmentType(string f
             return UniProtSequenceAttributes.FragmentType.unspecified;
         }
 
-        // Helper method to compute the monoisotopic mass of a sequence.
-        /// <summary>
-        /// Computes the monoisotopic mass of the given sequence.
-        /// Returns 0 if the sequence is empty.
-        /// </summary>
```
nbollis: :(

nbollis (changes requested):
Some issues left over from copilot: 
1. Rewrote entire sections of code that should not have been affected by changing how variants operate. 
2. Deleted a bunch of comments unnecessarily. 
3. Commented about what change in the code occurred due to your copilot prompt, but does not add any understanding to what is happening and why in the function. 


Max variant isoforms being 1 to get the canonical out feels bad. The sequence we get out is not a variant isoform.

mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -645,223 +645,194 @@ public bool IncludesSpliceSite(SpliceSite site)
...
-
-        /// <summary>
-        /// Checks if sequence variant and peptide intersect, also checks if the seuqence variatn can be identified whether they intersect
-        /// or not (ie if the variant causes a cleavage site generating the peptide). Returns a tuple with item 1 being a bool value
-        /// representing if the varaint intersects the peptide and item 2 beign abool that represents if the variatn is identified.
-        /// </summary>
-        /// <param name="pep"></param>
-        /// <param name="appliedVariation"></param>
-        /// <returns></returns>
         public (bool intersects, bool identifies) IntersectsAndIdentifiesVariation(SequenceVariation appliedVariation)
```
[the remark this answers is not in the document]
trishorts: reverted

mzLib/Omics/BioPolymer/SnpEffAnnotation.cs
```diff
@@ -1,72 +1,40 @@
...
-        public string HGVSNotationProteinLevel { get; }
+        public string HGVSNotationDnaLevel { get; } = string.Empty;
+        public string HGVSNotationProteinLevel { get; } = string.Empty;
         public int OneBasedTranscriptCDNAPosition { get; }
         public int TranscriptCDNALength { get; }
         public int OneBasedCodingDomainSequencePosition { get; }
         public int CodingDomainSequenceLengthIncludingStopCodon { get; }
         public int OneBasedProteinPosition { get; }
         public int ProteinLength { get; }
-
```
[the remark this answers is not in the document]
trishorts: all comments back in snpeff

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,167 +1,857 @@
...
+
+        #endregion
+
+        #region Convenience / Interval Logic
+
+        /// <summary>Simple concatenated representation (Original + Begin(+/-End) + Variant).</summary>
         public string SimpleString()
         {
-            return OriginalSequence + OneBasedBeginPosition.ToString() + VariantSequence;
+            if (OneBasedBeginPosition == OneBasedEndPosition || (OriginalSequence?.Length ?? 0) <= 1)
```
Alexander-Sol: Why not just make this the ToString() method. Is there another ToString method that I missed?

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,167 +1,857 @@
...
+                return $"{(OriginalSequence ?? string.Empty)}{OneBasedBeginPosition}{(VariantSequence ?? string.Empty)}";
+            }
+            return $"{(OriginalSequence ?? string.Empty)}{OneBasedBeginPosition}-{OneBasedEndPosition}{(VariantSequence ?? string.Empty)}";
         }
 
+        internal bool Intersects(SequenceVariation segment) =>
+            segment.OneBasedEndPosition >= OneBasedBeginPosition && segment.OneBasedBeginPosition <= OneBasedEndPosition;
+
+        internal bool Intersects(TruncationProduct segment) =>
+            segment.OneBasedEndPosition >= OneBasedBeginPosition && segment.OneBasedBeginPosition <= OneBasedEndPosition;
```
Alexander-Sol: These Intersects methods will only return true if the segment is a subset of this SequenceVariation, but not if they merely overlap, e.g.,
`segment.OneBasedEndPosition >= OneBasedBeginPosition && segment.OneBasedEndPosition  <= OneBasedEndPosition && segment.OneBasedBeginPosition  OneBasedEndPosition`

Is this the intended behavior?

Alexander-Sol (changes requested):
The changes made are far reaching and extensive, which makes this PR difficult to review. 

If possible, please reduce the size of this PR, eliminating unnecessary changes and all changes not directly related to sequence variants. If this PR can be split up into multiple, smaller PRs, that would be preferred.

mzLib/Omics/BioPolymer/VariantApplication.cs
```diff
@@ -121,188 +212,388 @@ public static List<TBioPolymerType> ApplyVariants<TBioPolymerType>(TBioPolymerTy
...
+                adjustedProteolysisProducts,
+                adjustedModifications,
+                individual);
+
+            // Normalize UniProt sequence attributes (length + mass)
+            try
+            {
+                var seq = created?.BaseSequence;
+                if (!string.IsNullOrEmpty(seq))
+                {
```
Alexander-Sol: Inside this try statement, you're using reflection to access properties of your variant. This is concerning, as reflection should rarely be used in production code. It's slow and finicky.

Can you refactor this to avoid using reflection?

mzLib/UsefulProteomicsDatabases/ProteinDbWriter.cs
```diff
@@ -41,632 +44,1045 @@ public static Dictionary<string, int> WriteXmlDatabase(Dictionary<string, HashSe
...
+                    writer.WriteStartElement("location");
+                    writer.WriteStartElement(seqVar == null ? "position" : "subposition");
+                    writer.WriteAttributeString(seqVar == null ? "position" : "subposition",
+                        positionModKvp.Key.ToString(CultureInfo.InvariantCulture));
+                    writer.WriteEndElement(); // position/subposition
+                    writer.WriteEndElement(); // location
+                    writer.WriteEndElement(); // feature or subfeature
+                }
+            }
+        }
```
Alexander-Sol: Large chunks of this file were re-written. Are all rewrites necessary?

trishorts:
Closing — the original variant-application monolith has been decomposed: Parts 1–4 are merged and the remainder lands via the #968 → #969 → #970 → #972 stack. Conflicting as a unit.

PR #958, opened 2025-10-10
out matchedPeaks in GetAllXics

zhuoxinshi:


trishorts:
Merged master and pushed, with one change I'd flag for you.

**The conflict** was mechanical: master moved the test file to `Test/MassSpectrometryTests/` and swept `Assert.That` to `NUnit.Framework.Assert.That`, so every line this PR touched conflicted. I took master's file and re-applied the `out matchedPeaks` calls on top.

**The change worth a look:** I made the out-parameter version an *overload* rather than a changed signature.

`GetAllXics` has 12 call sites. Inserting `out matchedPeaks` ahead of the optional `cutPeakDiscriminationFactor` breaks every one of them, and three landed on master after this branch was written, so they were not updated here:

- `Test/MassSpectrometryTests/Deconvolution/TestXicVsMassTraceComparison.cs:82` and `:104`
- `Test/MassSpectrometryTests/Deconvolution/JurkatMs1FeatureGeneratorViaXic.cs:98`

The build would have failed on those. With an overload, no existing caller changes at all, and the four-argument version just delegates with `out _`. Both stay `virtual`.

I also strengthened the assertion slightly: it now checks each peak maps back to the XIC that actually contains it, not only that the counts agree. If you'd rather have the breaking signature, say so and I'll put it back — but then those three call sites need updating in this PR.

`TestXic` is 17/17 green locally.

mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
```diff
@@ -196,9 +196,20 @@ public List<IIndexedPeak> GetXicByScanIndex(double m, int zeroBasedStartIndex, T
...
+            return GetAllXics(peakFindingTolerance, maxMissedScanAllowed, maxRTRange, numPeakThreshold, out _, cutPeakDiscriminationFactor);
+        }
+
+        /// <summary>
+        /// As above, additionally reporting which XIC each indexed peak was traced into. Callers that only want the
+        /// XICs use the overload without it; this one exists for callers that need to get from a peak back to the
+        /// chromatogram it belongs to, which the returned list alone cannot answer without a second pass.
+        /// </summary>
+        /// <param name="matchedPeaks"> every peak that ended up in an XIC, mapped to that XIC </param>
+        public virtual List<ExtractedIonChromatogram> GetAllXics(Tolerance peakFindingTolerance, int maxMissedScanAllowed, double maxRTRange, int numPeakThreshold, out Dictionary<IIndexedPeak, ExtractedIonChromatogram> matchedPeaks, double? cutPeakDiscriminationFactor = null)
```
acesnik, line 209: Both this and the four-argument overload at 198 are `virtual`, and 198 delegates here. A subclass that overrides only one of them gets silently bypassed through the other path. Dropping `virtual` from the delegating overload and keeping it here, on the one that does the work, costs nothing now and removes the trap. Nothing overrides `GetAllXics` today, so this is about the next subclass rather than a current bug.

mzLib/Test/MassSpectrometryTests/TestXic.cs
```diff
@@ -142,18 +142,20 @@ public static void TestGetAllXics()
...
             }
+            NUnit.Framework.Assert.That(matchedPeaks.Count, Is.EqualTo(xics.Sum(xic => xic.Peaks.Count)));
+            NUnit.Framework.Assert.That(xics.All(xic => xic.Peaks.All(peak => matchedPeaks[peak] == xic)));
 
             //Test with massIndexingEngine
             var deconParameters = new ClassicDeconvolutionParameters(1, 20, 4, 3);
             var allMasses = Deconvoluter.Deconvolute(testScans[0].MassSpectrum, deconParameters);
             var massIndexingEngine = MassIndexingEngine.InitializeMassIndexingEngine(testScans, deconParameters);
-            var massXics = massIndexingEngine.GetAllXics(new PpmTolerance(20), 2, 2, 3);
+            var massXics = massIndexingEngine.GetAllXics(new PpmTolerance(20), 2, 2, 3, out matchedPeaks);
```
acesnik, line 158: This and the three below it (179, 184, 195) moved to the new overload but never assert their `out` value — only the mapping at 145 is checked. Leaving these four on the four-argument overload would shrink the diff and make the one real assertion easier to spot.

acesnik (approved):
🤖 Approving. No overload ambiguity, since an `out` argument cannot be omitted, and `GetAllXics` has no overrides anywhere in the library today, so the delegation is safe as it stands.

One note that isn't in the diff: nothing in the library calls the new out parameter yet. Worth linking the PR that needs it, so it doesn't sit as unused public surface.

trishorts:
Heads-up on a collision with #1191 (approved, not yet merged), which annotates this same dictionary as `Dictionary<IIndexedPeak, ExtractedIonChromatogram?>`.

This PR promotes it to public API as a non-nullable `out Dictionary<IIndexedPeak, ExtractedIonChromatogram>` documented as *"every peak that ended up in an XIC, mapped to that XIC"*. That isn't what it holds. `GetAllXics` also does `matchedPeaks.Add(peak, null)` for a peak that was examined but fell below `numPeakThreshold` — the line is unchanged by this PR — so the dictionary is a *considered-peaks* set with an optional payload, and `matchedPeaks[peak]` can hand a caller `null`.

Two consequences if this merges as-is:

- The `out` parameter's value type should be `ExtractedIonChromatogram?`, and the doc should say what a null value means, or callers will dereference it.
- `Assert.That(matchedPeaks.Count, Is.EqualTo(xics.Sum(xic => xic.Peaks.Count)))` only holds on data where every peak forms an XIC. It passes on the fake scans here; it is not a property of the method.

Simplest path is to let #1191 land first and rebase — it makes the nullability explicit and the test assertion's real precondition visible.

acesnik:
This is looks like it's ready to rebase

PR #969, opened 2025-10-25
Update Handling of Sequence Variants (Part 6): SequenceVariant AreValid

trishorts:
the existing method for evaluating if a sequence variant is valid just isn't up to snuf. This PR adds improved bool for AreValid that considers all changes at the sequence and modification level of a variant


acesnik:
🤖 The strengthened `AreValid` is worth having — master still has the thin version (`SequenceVariation.cs:314`, just `OneBasedBeginPosition > 0 && OneBasedEndPosition >= OneBasedBeginPosition`), so the "isn't up to snuff" diagnosis is accurate and the replacement genuinely adds the no-op check and the modification-position validation.

But I do not think this should be reviewed line by line in its current form, and there is one interaction that needs measuring before it lands at all.

## The interaction, which post-dates this branch

`5ec37ea9` — "A variant database was not one decoy per target, each a reversal of it (#1154)" — added this to `DecoyProteinGenerator.cs:354`:

```csharp
return decoyVariations.Where(v => v.AreValid()).ToList();
```

That call site has **0 occurrences at this branch's merge-base and 0 on this branch**, and 1 on master. So a rebase drops a much stricter `AreValid` into a filter that did not exist when this was written. A decoy variation that changes nothing — same original and variant sequence, no modifications — is currently valid and would become invalid, and therefore silently dropped from the decoy set.

The decoy set drives q-values, and #1154 existed precisely to make the decoy-per-target count correct. Dropping some again by a different route would undo it in a way no test on either branch is positioned to catch. Worth measuring the decoy count before and after on a real variant database, rather than reasoning about whether no-op decoy variations occur.

## Why re-cutting beats rebasing

Master already has the other half of this PR. `SequenceVariation.cs:193` and `:199` on master are already:

```csharp
public string Description { get; }
public VariantCallFormat? VariantCallFormatDataString { get; }
```

So the `Description` retype and the VCF split — a large share of the 13-file diff — have landed by another route. What is left that is genuinely new is `AreValid` and `GetInvalidModificationPositions`. Re-cut against master, this is a small focused PR about a validity predicate, which is a far better thing to review than +2435/−507.

The mechanical state supports that too. Every one of the 13 files has also changed on master since the merge-base, and three cases have no clean resolution:

- `Test/TestPeptideWithSetMods.cs` — **deleted on master** by `c120f7a1` (#1057, test class grouping). This PR adds +160/−50 to a file that no longer exists, so those tests have to be re-homed by hand.
- `truncationsExpected.tsv` and `RnaDecoyGenerator.cs` — already **byte-identical** to master, so they contribute nothing.
- `TestVariantProtein.cs` and `ProteinXmlEntry.cs` — both sides independently rewrote them (+1323/−134 vs +1693/−169, and +259/−84 vs +292/−110).

## Two smaller things, for whenever it is re-cut

- The description is one sentence for a change that alters a validity predicate consumed by variant application and decoy generation. Given the decoy interaction above, it wants a paragraph on what becomes invalid that previously was not, and what that does downstream.
- This diff adds `// Always construct a VariantCallFormat so tests relying on non-null VCF objects pass.` Justifying a production invariant by what tests need is backwards — if the object is always constructed then the property does not need to be `VariantCallFormat?`, and if it can be null the tests should handle that. Worth resolving in one direction.

Not approving, but not because of the code: the predicate looks right to me and the docs are better than what they replace (+180 doc lines against −41). It is that reviewing this diff would mostly be reviewing changes master already has, and the one thing that most needs review — the decoy filter — cannot be seen from this branch at all.

acesnik:
🤖 Closing this one in favour of #1222, which carries its substance forward.

The strengthened `AreValid` was the piece of this series that was genuinely new — master still had the position-only predicate — and #1222 is that logic re-cut against current master, co-authored to you since it is your work. It also answers the question this branch could not: `AreValid` is consumed by `DecoyProteinGenerator`'s filter and by `GetVariantBioPolymers`' branch selection, and measuring all four variant fixtures before and after showed eight no-op decoy variations dropped in `VariantModsGPTMD.xml` and no change to any protein count. Your predicate is unchanged in it; the diagnosis that the old one "isn't up to snuff" was right.

Closing rather than rebasing because the rest of the diff had gone stale underneath it rather than because anything was wrong with it: the `Description` / `VariantCallFormatDataString` split has since landed on master by another route, `Test/TestPeptideWithSetMods.cs` was deleted by #1057, and both `TestVariantProtein.cs` and `ProteinXmlEntry.cs` were independently rewritten on master. Reviewing it would have been mostly reviewing changes master already had.

Please reopen if I have got that wrong — in particular if there is something in the 40 commits here that #1222 does not carry and should. The `GetConsensusAndVariantBioPolymers` API is deliberately left out of #1222 rather than dropped, and is worth its own PR once `maxVariantsPerIsoform` actually does something.

Thanks for the work — the `AreValid` rules are a real improvement and they are now on a branch that can land.

PR #970, opened 2025-10-25
Update Handling of Sequence Variants (Part 7): Point LoadProteinXML to new VariantApplicationMethod

trishorts:
current LoadProteinXML has two int variables: maxVariantsForCombinatorics =4 and minAlleleDepth = 1. VariantApplication now has three consensusPlusVariantIsoforms = 1, minAlleleDepth = 0 and maxVariantsPerIsoform = 0. This PR changes LoadProteinXML to use all three variables. It also calls the new GetConsensusAndVariantBioPolymers method in VariantApplication

this may break some tests in MM where variants are expected but none are called for. The former LoadXML called for 4 with a min depth of 1. defaults now are zero.


acesnik:
🤖 Reviewed the delta against #969 rather than the whole diff, since this branch is #969 plus four commits (39 insertions, 14 deletions across two files). Everything in my comment on #969 applies here too, because this contains it.

Three things in the delta, and the first means this cannot land as written.

## `maxVariantsPerIsoform` is never read

The description says "This PR changes LoadProteinXML to use all three variables." It uses two. The third is threaded through both public signatures and then dropped:

```
VariantApplication.cs:23    declared on GetConsensusAndVariantBioPolymers
VariantApplication.cs:51    legacy wrapper passes maxVariantsPerIsoform: 0
ProteinDbLoader.cs:84       declared on LoadProteinXML
ProteinDbLoader.cs:160      passed through
```

That is every occurrence. The body of `GetConsensusAndVariantBioPolymers` calls `ApplyAllVariantCombinations(protein, variations, consensusPlusVariantIsoforms)` at `:33` and `ApplyVariants(..., maxAllowedVariantsForCombinitorics: consensusPlusVariantIsoforms, minAlleleDepth)` at `:37`, and neither `ApplyVariants` (`:126`) nor `ApplyAllVariantCombinations` (`:519`) accepts a per-isoform cap at all.

The two doc comments disagree about this, which is worth noting because one of them is the honest one. `VariantApplication.cs:21` says it "is reserved for per-isoform combinatoric caps" — reserved, i.e. not implemented. `ProteinDbLoader.cs:73` describes it as "an explicit per-isoform combinatoric cap to control explosion" — i.e. as working. A public parameter that silently does nothing is worse than no parameter here specifically, because the thing a caller would set it for is protection against combinatorial explosion, and they would get none while believing they had it.

Either implement it or take it out of both signatures until it exists.

## The behaviour change is a reduction, not a rename

`consensusPlusVariantIsoforms` is passed straight through as `maxAllowedVariantsForCombinatorics` — the same underlying knob the old `maxHeterozygousVariants` fed. So the rename is only a rename; what actually changes is the **default, from 4 to 1**, while `minAlleleDepth` goes 1 to 0 in the opposite direction.

So a default `LoadProteinXML` caller goes from "up to 4 heterozygous variants, minimum allele depth 1" to "up to 1, minimum depth 0". That is the MetaMorpheus breakage you anticipate, and it is worth stating as the headline rather than as a closing caveat: it changes variant output for every MetaMorpheus user who loads a variant XML without naming these parameters, not just tests that "expected variant proteins when they don't explicitly call for them". Same shape as #1186 — it needs to be a deliberate, measured decision with the MetaMorpheus side coordinated, not a default that moves as a side effect of a rename.

## The parameter insertion is a silent-rebinding hazard

`maxVariantsPerIsoform` goes in at position 11, before `addTruncations`. Callers that passed `addTruncations` positionally fail to compile, since `bool` will not bind to `int` — that is the lucky case. A caller that stopped earlier, `LoadProteinXML(..., maxThreads, 4, 1)`, still compiles and now means `consensusPlusVariantIsoforms: 4, minAlleleDepth: 1` instead of `maxHeterozygousVariants: 4, minAlleleDepth: 1`. Same numbers, and the meaning of the first one has moved.

Named-argument callers break loudly on the rename, which is fine. It is the positional ones that change quietly. Appending the new parameter after `decoyIdentifier`, or keeping the old overload as a delegating wrapper the way `GetVariantBioPolymers` does at `:48-52`, would avoid it.

## Minor

`ProteinDbLoader.cs` — the new summary block has a stray `/// /// ` on its second line.

acesnik:
🤖 Closing alongside #969, which this was stacked on, in favour of #1222.

#1222 deliberately does not include this part — pointing `LoadProteinXML` at the new API is a separate decision from the validity predicate, and there were two things worth settling first rather than losing:

- `maxVariantsPerIsoform` is threaded through `LoadProteinXML` and `GetConsensusAndVariantBioPolymers` and never read. Neither `ApplyVariants` nor `ApplyAllVariantCombinations` takes a per-isoform cap, so a caller setting it for protection against combinatorial explosion gets none.
- The default moves from `maxHeterozygousVariants = 4` to `consensusPlusVariantIsoforms = 1` while `minAlleleDepth` moves 1 → 0. Since `consensusPlusVariantIsoforms` is passed straight through as `maxAllowedVariantsForCombinatorics`, the rename is only a rename and the real change is the default — which alters variant output for every caller who does not name these parameters.

Both are worth a PR of their own against current master, where the diff would be small enough to argue about properly. Please reopen if you would rather carry this branch forward instead — the observations are in the thread either way, so nothing is lost by closing.

Thanks for the work.

PR #974, opened 2025-10-27
Handling of applied variants with and without modifications

trishorts:
Few things here:
1. Simple expansion of some unit tests to better reveal current behavior for the treatment of sequence variants that are supported by VCF data. 
2. Homozygous variants with depth greater than specified results in only the production of the variant protein or RNA. you do not get the consensus protein or RNA in that case. This was a surprise.
3. List of sequence variants in proteins and RNAs wasn't properly adjust when variants get applied. Desired behavior is to remove applied variants from the sequence variants list when they get applied.
4. This PR helps display proper handling of modifications on variants.


trishorts:
Closing — stalled mod-on-variant branch (Nov 2025). Its real bugfixes (XML-writer-wrote-consensus-only, variant-list pruning on apply, homozygous-depth behavior, mods-on-variants) and test assets are captured in #1083 for folding into the active variant stack. Branch `variantModificationTestExpansion` retained for cherry-picking.

PR #978, opened 2025-11-11
Added handling of Compensation Voltages for data with FAIMS

acarrMagnet:
I need to work with data that is more than likely generated with FAIMS. It would be very convenient if there was some way to filter MsDataScans by compensation voltages in my favorite MS processing package. This pull request adds a compensation voltage property to MsDataScan, adds code to read value from .raw file trailer to Thermo Raw File Reader, and adds a unit test with a .raw file generated using FAIMS.


nbollis (approved):
10/10. Would review again

PR #984, opened 2025-12-01
isobaric quant engine and results

trishorts:


trishorts:
Closing as a deliberate architecture decision, not as housekeeping — worth stating plainly so the record is useful later.

This PR proposes a **separate `mzLib/IsobaricQuant/` project** — `IsobaricQuantEngine` (735 lines), `IsobaricQuantResults`, `TmtFileInfo`, `TmtPlexAnnotation` — as a dedicated isobaric pipeline alongside the general one. #996 asks for the opposite: a single engine that does not care whether the intensities came from label-free peak areas or isobaric reporter ions, unified by `ISpectralMatch.Intensities` (length 1 for LFQ, one entry per channel for TMT).

That unified path is what shipped, in #1024, and it is what has been built on since:

- `mzLib/Quantification/` with `QuantificationEngine`, `QuantMatrix<T>`, and the normalization / roll-up / collapse strategy trio
- `ReferenceChannelNormalization`, which is isobaric-aware and reads `IsobaricQuantSampleInfo.IsReferenceChannel`
- `Test/Quantification/TmtSpikeInTests.cs` in CI, and a real-data TMT harness in `Development/QuantificationDevelopment/`
- #1193, which makes the engine write its results onto `IBioPolymerGroup` and implements `QuantificationWriter`
- #1194, which adds `SampleCollapseStrategy`

Keeping a second engine would mean two normalization stacks, two roll-up stacks, and two sets of output writers to keep in step — and MetaMorpheus would have to choose between them at the call site rather than choosing a strategy set.

For the record: the last substantive commit here is 2025-12-02; the recent timestamp is a base sync, not new work.

`TmtFileInfo` and `TmtPlexAnnotation` are not lost — the same two types are alive in MetaMorpheus PR #2598 (`tmtExperimentBuilder`), which reads a `TmtDesign.txt`. The open work there is to have them produce mzLib `IExperimentalDesign` / `IsobaricQuantSampleInfo` objects rather than paralleling them, which needs a reference-channel flag and a reporter m/z. Worth also looking at the SDRF-Proteomics support that landed in #1138 and #1174 — SDRF already encodes channel-to-sample mapping and may be the better carrier than a bespoke format.

PR #985, opened 2025-12-03
Quant interfaces

Alexander-Sol:


mzLib/FlashLFQ/Interfaces/IQuantifiablePeptide.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    /// <summary>
+    /// Defines the interface for peptides that can be quantified
+    /// </summary>
+    public interface IQuantifiablePeptide
+    {
+        /// <summary>
+        /// The peptide sequence
+        /// </summary>
+        string Sequence { get; }
```
nbollis: We should probably use BaseSequence or FullSequence so it maps better with our existing objects.

mzLib/FlashLFQ/ProteinQuantification/GenericProteinQuantificationEngine.cs
nbollis: Love the dependency injection and inversion of control here

nbollis (commented):
What about the interface that is already in readers? How does this play into this?

trishorts:
Leaving this one open, unlike its siblings — flagging why, since #1000, #1001 and #984 were just closed as superseded by #1024 and this one is not.

None of the work here is on master:

| File | On master? |
|---|---|
| `FlashLFQ/ProteinQuantification/MedianPolishProteinQuantificationStrategy.cs` (330 lines) | no |
| `FlashLFQ/ProteinQuantification/Top3ProteinQuantificationStrategy.cs` | no |
| `FlashLFQ/ProteinQuantification/GenericProteinQuantificationEngine.cs` | no |
| `FlashLFQ/Interfaces/IQuantifiable{Peptide,ProteinGroup,SpectraFile}.cs` | no |

Master's FlashLFQ still has only the Bayesian `ProteinQuantificationEngine`.

This is the label-free half of #996 — the branch in the workflow diagram where FlashLFQ computes and then populates a length-1 `double[]` per PSM so that the shared engine can take over. That half has not been started anywhere else: FlashLFQ still owns its own `FlashLfqResults` roll-up and normalization, and nothing populates `ISpectralMatch.Intensities` for LFQ.

So this is the only existing attempt at it, and closing it would discard the `IQuantifiable*` seam and two protein-quant strategies that master does not otherwise have.

It will need a rebase (last commit 2025-12-03, and master has since moved to .NET 10), and the interfaces will want reconciling against `Omics/IHasSampleIntensities` from #1193 — but that is follow-up work on a live idea, not a superseded one.

Issue #996, opened 2025-12-17
Add Quantification project to mzLib, refactor MetaMorpheus and FlashLFQ to support


Alexander-Sol:
Here's a workflow for how quantification will work in the new regime

<img width="10345" height="6593" alt="Image" src="https://github.com/user-attachments/assets/b2d05ede-4cd2-408e-9851-cc47f93d4475" />

If anyone wants to edit this diagram, they can [do so at this link](https://www.mermaidchart.com/app/projects/c1d5cccb-22e9-43a4-b90b-38e0af51a814/diagrams/163d0b35-a13e-4169-8426-40d1f4bd94b1/share/invite/eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJkb2N1bWVudElEIjoiMTYzZDBiMzUtYTEzZS00MTY5LTg0MjYtNDBkMWY0YmQ5NGIxIiwiYWNjZXNzIjoiRWRpdCIsImlhdCI6MTc2NTk1NDkyN30.SAGMCJ9kZdGq_EmAStT7nETYxxfyBuahGatTD5Fi7Ok)

PR #993, opened 2025-12-12
Add Interface IBioPolymerGroup interface and BioPolymerGroup class to Omics

trishorts:
This pull request introduces a new class for spectral match handling with fragment-level sequence coverage calculation, and makes minor adjustments to null handling and property ordering in the `SpectraFileInfo` class. The most significant change is the addition of the `BaseSpectralMatch` class, which provides a reusable and extensible implementation for storing and comparing spectral matches, as well as calculating sequence coverage from fragment ions.

### New spectral match implementation

* Added a new `BaseSpectralMatch` class in `mzLib/Omics/BaseSpectralMatch.cs` that implements both `ISpectralMatch` and `IHasSequenceCoverageFromFragments`. This class provides a robust structure for storing spectral match information, calculating fragment-level sequence coverage, and supporting equality and comparison operations.

### Improvements and fixes in `SpectraFileInfo`

* Removed unnecessary null-coalescing (`?? string.Empty`) for `FullFilePathWithExtension` and `Condition` in the `SpectraFileInfo` constructor, and adjusted the order of property assignments for clarity.
* Cleaned up the `CompareTo` method in `SpectraFileInfo` by removing an outdated comment, improving code readability.

PR #992, opened 2025-12-11
ISampleInfo interface added to MassSpectrometry to enable generic key for both LFQ and IsoBaricQuant

trishorts:
This pull request introduces a new abstraction for sample metadata in quantification workflows by adding the `ISampleInfo` interface and updating sample classes to implement it. It also provides a new class for isobaric quantification samples and enhances the existing `SpectraFileInfo` class with interface conformance, improved equality, and comparison logic. These changes establish a unified structure for handling sample information across different quantification strategies.

**New sample abstraction and interface:**

* Introduced the `ISampleInfo` interface in `ISampleInfo.cs`, defining common properties (file path, condition, biological/technical replicate, fraction) and requiring equality and comparison implementations for consistent sample identification and grouping.

**Isobaric quantification sample support:**

* Added the `IsobaricQuantSampleInfo` class implementing `ISampleInfo`, representing samples in isobaric (TMT/iTRAQ) quantification workflows. This class includes isobaric-specific properties (channel label, reporter ion m/z, plex ID, reference channel flag), and implements custom equality, hashing, and comparison based on file path and channel label.

**Enhancements to label-free sample metadata:**

* Updated `SpectraFileInfo` to implement `ISampleInfo`, adding interface methods and improving equality, hashing, and sorting. Now, sample identity is determined by all key properties (file path, condition, biological replicate, fraction, technical replicate), providing robust sample distinction and comparison.
* Improved documentation and property descriptions in `SpectraFileInfo` for clarity and consistency with the new interface.

PR #1000, opened 2025-12-18
Quant Project (WIP)

Alexander-Sol:
This PR builds off of Shortreeds PR #993 . First step in addressing Issue #996


mzLib/MassSpectrometry/ExperimentalDesign/IExperimentalDesign.cs
nbollis: It may be worth moving some of the static Experimental Design methods out of metamorpheus and into mzlib as either static funcctions or interface members.

mzLib/Quantification/QuantificationWriter.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Omics.BioPolymerGroup;
+
+namespace Quantification
+{
+    /// <summary>
+    /// This will need to be implemented or changed or relocated later. Right now it's just a placeholder.
+    /// </summary>
+    public static class QuantificationWriter
```
nbollis: If writing require unique functions, you could possibly do this as an interface and an abstract factory. That way the engine constructor calls a writer factor, passes in quant parameters, and receives either the TMT writer or the LFQ writer.

trishorts:
Closing as superseded, with the one piece that never landed harvested into #1194.

The architecture this PR introduced reached master by a different route — principally #1024 "Quant project TMT Implementation" (merged 2026-03-03), together with #992 (`ISampleInfo`), #993 (`IBioPolymerGroup`), and #1028 (lightweight spectral match).

Comparing this branch's head (`3be08af`) against master file-by-file, matching by basename so that files master **moved** are not counted as missing: **32 of the 35 files have landed.** Several were relocated during the reorganization — `Omics/ISpectralMatch.cs` → `Omics/SpectralMatch/ISpectralMatch.cs`, `Test/Omics/*Tests.cs` → `Test/Omics/{BioPolymerGroupTests,SpectralMatchTests,SampleInfo}/` — which is most of why the diff still reads as +7665. Master has since gone further, gaining `SumCollapse`, `MeanCollapse`, `GlobalMedianNormalization`, `ReferenceChannelNormalization`, `MedianRollUp`, `Omics/BioPolymerGroup/SampleGroupResult.cs`, and a CI-resident `Test/Quantification/TmtSpikeInTests.cs`.

The three files that were still only here:

| File | Disposition |
|---|---|
| `Quantification/Strategies/Collapse/SampleCollapseStrategy.cs` | ported in #1194 |
| `Test/Quantification/CollapseStrategyTests.cs` | rewritten as `SampleCollapseStrategyTests` in #1194 |
| `Test/Quantification/QuantMatrixTests.cs` | ported in #1194 |

`SampleCollapseStrategy` was worth keeping: `SumCollapse` and `MeanCollapse` both group by `Condition + '_' + BiologicalReplicate`, so they collapse fraction *and* technical replicate at once, and there is no way to sum fractions while keeping technical replicates. The parameterized version subsumes both as special cases and composes.

Two things were changed on the way in. `CreateCollapsedSampleInfo` returned a `SpectraFileInfo` for every input, which discarded the channel label, plex, reporter m/z and reference-channel flag of an `IsobaricQuantSampleInfo` — since collapse runs between peptide and protein roll-up, that would have silently disabled `ReferenceChannelNormalization` downstream. And `QuantMatrixTests` could no longer compile as written, because `BaseSpectralMatch` has since become abstract; it is rebuilt against `MockSpectralMatch` and extended to the argument-validation paths.

Thanks @Alexander-Sol — the shape of `QuantificationEngine`, `QuantMatrix<T>` and the three strategy interfaces is all still there on master, essentially as designed here.

Resolving this branch's conflicts would mean re-doing several months of master by hand to recover one 178-line file, so closing rather than rebasing.

PR #1001, opened 2025-12-18
Simple quant writer2

trishorts:


trishorts:
Closing as superseded by #1193, which finishes the job this PR started.

The `QuantificationWriter` here was the first real implementation — `RawQuantification.tsv`, max-channel detection, deterministic ordering by file then scan. #1193 implements all three writer methods (raw, peptide, protein group), returns the path written, and disambiguates repeated sample labels.

**One thing from this PR was better and has been folded into #1193**: naming the raw file's value columns by case rather than uniformly. This PR emitted a single `Intensity` column when there is one value per match and `Reporter_n` columns when there are several; #1193 had `Intensity_n` for both. That distinction is now in #1193, with `RawHeader` extracted so the naming is directly testable.

The `// Placeholder naming - will be updated with actual TMT channel names` note is worth recording as a deliberate non-goal rather than a TODO. A spectral match carries no channel labels — only `Intensities`, which is positionally aligned with the experimental design's `ISampleInfo` array. Naming the columns "126", "127N" … needs the design at the writer, and would be wrong for a run whose files do not all share one plex. The named-channel columns belong in the peptide and protein files, which do have the design, and #1193 labels those by sample.

Everything else here has landed: `BioPolymerGroup`/`IBioPolymerGroup` via #993, the `ExperimentalDesign` types via #992, `BaseSpectralMatch` and `IHasSequenceCoverageFromFragments` (now under `Omics/SpectralMatch/`), and the `Quantification` project itself via #1024. `RollUpStrategyBase` exists in neither this branch's final state nor master.

Note also that `ISpectralMatch.QuantValues` has since been renamed `Intensities`, so this branch would need that rename before it could build against master.

PR #1027, opened 2026-02-25
Dia

trishorts:
> **Project of origin:** none — standalone


acesnik (commented):
🤖 Read this for structure rather than line by line, and I want to be upfront that a line review is not what I am offering — 30,255 lines of C# is past the point where one reviewer's read means much, and this has sat 187 days without anyone attempting it. What follows is the set of decisions I think have to be settled *before* a line review is worth anyone's time, and where the seams are if you want to split it. That is the same thing that unstuck #2598.

First, a number worth correcting because it is discouraging people: the PR header says +108,669, and about 90,000 of that is two mzML fixtures — `Test/Dia/DiaTestData/Fig2HeLa-0-5h_MHRM_R01_T0_Scans0to500.mzML` alone is 6 MB, and GitHub reports both as `0/0` because the patches are omitted. The actual code is **~18.5k production, 5.6k tests, and 7.2k under `Development/`**. That is large but not unreviewable, and it reads very differently to someone deciding whether to pick it up.

## Three decisions that gate a line review

**ILGPU lands in every mzLib consumer's dependency graph.** `MassSpectrometry.csproj` gains an unconditional `<PackageReference Include="ILGPU" Version="1.5.3" />` — no `Condition`, no `PrivateAssets` — in the most central assembly in the library. MetaMorpheus, Spritz and FlashLFQ all take mzLib as a package, so all three acquire ILGPU transitively. `GpuDeviceDetector` is careful at runtime (reflection over `TryLoadAssembly("ILGPU")`, and its own remarks say macOS is `IsGpuAvailable = false` because CUDA is unavailable and OpenCL deprecated), so nothing *breaks* — but on Apple Silicon we would be shipping a dependency the code itself states can never be used, into a library that has just finished the work to be architecture-neutral.

Related and concrete rather than philosophical: `mzLib.nuspec` mentions ILGPU **zero times**. The nuspec hand-assembles its dependency groups, so as it stands the published package would contain an assembly referencing ILGPU while declaring a dependency set that omits it. Whichever way the GPU question goes, that has to be reconciled.

**Which classifier is the classifier?** Between `Dia/Scoring`, `Dia/FDR` and `Dia/Classifiers` there are three implementations behind `IDiaClassifier` — `DiaGradientBoostedClassifier` (747), `DiaNeuralNetClassifier` (585), `DiaLinearDiscriminant` (693) — plus `DiaFdrEngine` (948) and `DiaFdrEngineLegacy` alongside it. The interface says the plurality is deliberate, and having three while you pick one is exactly right during development. Shipping three is a different claim: each is a scoring regime whose FDR someone has to trust, and `Legacy` in a filename that has never been on master is a strong hint that the experimental record is being carried rather than the conclusion. Also worth saying: this would be a second ML stack in mzLib alongside ML.NET FastTree, and TorchSharp/libtorch already in the nuspec makes a third.

**Do the two DIA subsystems coexist?** Master already has `EngineLayer/DIA/` from #2546 and #2573 — library-free, pseudo-MS2 from precursor-fragment and XIC grouping, then an ordinary database search. This is library-based: query a spectral library, score against it, own FDR. Both are legitimate and most tools ship one, some ship both. But it is a product decision rather than a review finding, and it determines whether `mzLib/MassSpectrometry/Dia` should share the peak-indexing and XIC machinery that already exists (`PeakIndexing/ExtractedIonChromatogram`, `PeakSpline/`) or stay separate. Answering it after 18.5k lines of review is answering it too late.

## Where the seams are

Roughly in dependency order, each of which stands alone and is reviewable in an afternoon:

1. **`Readers/SpectralLibrary/DiaNNSpecLibReader.cs` + `DiaNNBinaryStructs.cs`** — a DIA-NN library reader, two new files, no dependency on any of the rest. This one could go today.
2. **`Dia/DataStructures`, `Dia/Interfaces`, `Dia/Extraction`, `Dia/Ms1`** — the scan-index and fragment-extraction layer, ~1.7k, and the natural place to settle the question of sharing with `PeakIndexing`.
3. **`Dia/Scoring`** minus the FDR engines — the four scorers plus `DiaFeatureExtractor`, ~4.2k of largely pure functions over extracted signal, which is the easiest part of this to test hard.
4. **`Dia/Calibration`** — ~2.7k, iRT and the bootstrap calibrator.
5. **`Dia/FDR` + `Dia/Classifiers`** — after the classifier decision, one implementation.
6. **The engine** — `DiaSearchEngine`, `DiaLibraryQueryGenerator`, `DiaPeakGroupDetector`, `Orchestration`, which is the part that needs everything above.
7. **`Dia/Gpu`** — gated entirely on the ILGPU decision, and cleanly separable because the CPU extractor already exists beside it.
8. **`Development/Dia`** — 7.2k lines and a `Development.csproj` that now becomes an `Exe` with a `StartupObject` and a `ProjectReference` to `PredictionClients`. Benchmarks are worth having; whether they belong in the shipped repo or on a branch is worth deciding rather than defaulting.

One thing that must not merge whichever way the rest goes: the nuspec version is set to `9.9.932`, which I take to be a local pin for testing against MetaMorpheus, the same way `9.9.1186` was used on the KZ work. Beyond that the nuspec diff is pure reindentation.

Happy to review (1) properly this week if that is useful, and to take the scorers in (3) after — those two are where I can say something worth having without the architecture questions being settled first.

PR #1037, opened 2026-03-19
Treat Entrapment like Contaminants

nbollis:
Added new properties to Protein, RNA, Biopolymer, and BioPolymerGroup which tracks if the protein is an entrapment protein. 

This is done in a similar manner to contaminants in that the database reader will set the IsEntrapment of our biopolymer based upon a bool parameters in the reader. 

This is done in a similar manner to decoys in that the database reader will prepend a string to the accession if 1) it is not already there and 2) it is an entrapment databases based upon a bool parameter to the reader.


Alexander-Sol (commented):
This all looks fine, but the idea of having code designed to specifically handle entrapment sequences makes me very, very nervous. The whole idea behind entrapment is to make sure that the algorithm is blind to the labels in a way that isn't true for decoy sequences. I worry that having access to a property that denotes an entrapment sequence could lead to contamination/issues down the line. 

The DB loading stuff makes sense, but idk if it should be a public property for biopolymers

Alexander-Sol (dismissed):

zhuoxinshi (approved):

Alexander-Sol (approved):

PR #1039, opened 2026-03-26
Readers: FDRBench files

nbollis:
FDRBench is a tool for false discovery rate (FDR) control evaluation in proteomics. It provides two main functions: (1) build entrapment databases using randomly shuffled target sequences or using sequences from foreign species, and (2) estimate the false discovery proportion (FDP) using the lower bound, combined, and paired methods.

https://github.com/Noble-Lab/FDRBench/tree/main

This PR adds reader support for two of their files so we can run entrapment evaluations from C#


acesnik (commented):
🤖 This reads FDRBench's peptide and protein result files so an entrapment evaluation can be driven from C#. It was speculative when you opened it in March; it is not any more, and that is the main thing I want to say. trishorts has a fifteen-PR stack in flight (#1249, #1250, #1253–#1265) that builds entrapment databases, with both of FDRBench's arms — shuffled targets and foreign species — and a QC report. This PR is the other half of that loop.

The code is in good shape. The detection ordering is the subtle part and it is right: `_fdrbench_peptide.tsv` also ends with MsFragger's `peptide.tsv`, so checking FDRBench first is load-bearing — and the two `TestCaseData` entries pin it, so a future reorder fails a test rather than silently reading an FDRBench file as MsFragger. `InvariantCulture` on both `CsvConfiguration`s is right for a numeric TSV. I also went looking for a regression where the diff appears to delete the header-sniffing block that separates `Tsv_FlashDeconv` from MsFragger; it is still there, relocated below the new checks. No regression.

**Two things worth acting on.**

It is `CONFLICTING` and last touched 2026-04-10, so a rebase comes before a real review — I read it at its head rather than merged.

And the interoperability question, which is the reason I would settle this before it lands rather than after. Your fixture's `protein` column reads `AAAPAPEEEMDECEQALAAEPK_target`, so FDRBench discriminates entrapment from target by a **suffix on the protein field**. The stack's `EntrapmentAccession.Format` emits `RANDOM_<target>_f<n>` — a prefix plus a fold suffix — and `ProteinDbLoader` classifies entrapment by testing whether an accession *contains* `"Random"`. Those are different conventions, so a MetaMorpheus search against a stack-generated database produces accessions FDRBench cannot classify without a translation step, and this PR's stated purpose is exactly that composition. Whichever side moves, it is cheaper to decide now than after both halves have shipped and users have databases in both dialects.

The stronger version of the same point: **#1260 is currently designing report columns for an unnamed consumer**, and its open question is whether a consumer can reconstruct peptide-level *r* from what the report emits. FDRBench's three estimators — lower bound, combined, paired — are that consumer, with published definitions. Pointing #1260 at them would replace a guess with a specification, and that seems like the most valuable thing this PR can contribute beyond the readers themselves.

Smaller: the new `TestCaseData` paths use backslash literals, so they will not resolve off Windows. Consistent with the rest of the file, and mzLib's `Test` project is Windows-only anyway, so nothing in CI would notice either way.

Not verified: I did not run anything — `Test/Test.csproj` is `net*.0-windows` via `mzPlot` and cannot execute here. FDRBench's suffix convention is inferred from your fixture, not from its documentation, so if it is configurable that weakens the interoperability point and I would rather hear so.

PR #1040, opened 2026-03-30
protease dictionary embedded resource new default behavior

trishorts:
Summary

    This PR migrates ProteaseDictionary and RnaseDictionary to an embedded-resource-first architecture. All default
    digestion agent definitions are compiled into the assembly and serve as the immutable baseline. Custom agents are
    additive-only — they supplement the defaults but can never replace them.

    Embedded Resource Rules

    1. Immutable baseline. Embedded definitions (proteases.tsv, protease_mods.txt, rnases.tsv) are the authoritative
    source of truth. They are loaded once by the static constructor and cannot be overridden by any custom file.
    2. Additive-only custom loading. LoadAndMergeCustomProteases and LoadAndMergeCustomRnases add new entries to the
    dictionary. If a custom file contains a name that already exists in the embedded set, it is skipped — the embedded
    definition always wins.
    3. Collision reporting, not crashing. Name collisions are reported via CustomDigestionAgentLoadResult.Skipped rather
     than throwing, so callers can warn users without aborting.
    4. Duplicate names within a file are malformed. A single custom file containing two entries with the same name
    throws MzLibException. This is a data error the caller should fix.
    5. Atomic loading. All files are parsed and validated before any entry is committed to the dictionary. If any file
    in a batch is missing or malformed, the dictionary remains completely unchanged.
    6. Private setter. The Dictionary property setter is private — external code cannot swap the entire dictionary
    instance.
    7. No default custom file. mzLib does not ship a default custom file. Downstream consumers (MetaMorpheus,
    ProteaseGuru, etc.) supply their own paths.
    8. Embedded mod caching. Embedded protease modifications (protease_mods.txt) are parsed once via Lazy<T> and cached
    for the process lifetime. LoadEmbeddedProteaseMods() returns a defensive copy each call.
    9. Null safety. Passing null as the paths parameter throws ArgumentNullException. Passing null for proteaseMods
    automatically resolves to the embedded modifications.
    10. Result immutability. CustomDigestionAgentLoadResult.Added and .Skipped are IReadOnlyList<string>, backed by
    ReadOnlyCollection — callers cannot mutate the reported results.

This pull request introduces a new result type for loading custom digestion agents and improves the reliability and clarity of digestion-related tests. The main changes include the addition of a record to report which custom agents were added or skipped, refactoring tests to ensure proper cleanup of custom proteases, and minor corrections to test assertions for RNA digestion.

**Custom Digestion Agent Loading:**

* Added a new record `CustomDigestionAgentLoadResult` in `Omics.Digestion` to encapsulate the results of loading custom proteases or RNases, providing read-only lists of added and skipped agent names.

**Test Reliability and Cleanup:**

* Refactored protease-related tests in `TestSeqCoverage.cs` to use `try/finally` blocks, ensuring that custom proteases are always removed from `ProteaseDictionary.Dictionary` after each test, preventing side effects between tests. [[1]](diffhunk://#diff-af9ce3e78979fc4fac30c0477a2d7e06c0c9ca01ddddfed866cfdc23b32aa1b8R40-R41) [[2]](diffhunk://#diff-af9ce3e78979fc4fac30c0477a2d7e06c0c9ca01ddddfed866cfdc23b32aa1b8R50-R64) [[3]](diffhunk://#diff-af9ce3e78979fc4fac30c0477a2d7e06c0c9ca01ddddfed866cfdc23b32aa1b8R75-R160)
* Updated the `ReadCustomFile` test to use the new `LoadAndMergeCustomProteases` method, verify which proteases were actually added, and clean up all added entries in a `finally` block.

**Test Assertion Corrections:**

* Fixed conditional logic in RNA digestion modification tests to use `if` instead of `else if`, ensuring all relevant assertions are checked for higher values of `MaxMods`. [[1]](diffhunk://#diff-720603470a9e5e5c004864a41e17e582aa71d06b512451cc8e52657968e4050bL1085-R1086) [[2]](diffhunk://#diff-720603470a9e5e5c004864a41e17e582aa71d06b512451cc8e52657968e4050bL1154-R1147) [[3]](diffhunk://#diff-720603470a9e5e5c004864a41e17e582aa71d06b512451cc8e52657968e4050bL1163-R1156)

**Test Initialization Simplification:**

* Removed the unnecessary `[OneTimeSetUp]` method from `TestDigestion.cs` that reset the RNase dictionary, as it is no longer needed.


nbollis:
Blocking (must address before merge)

---

mzLib/Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
Lines 65-69: Public API LoadProteaseDictionary(string path, List<Modification>) was removed. This is a breaking change for downstream consumers. Either restore it as a thin wrapper that parses without mutating global state, or mark [Obsolete] with migration guidance. Do not delete without deprecation period.
Lines 320-325: ResetToDefaults(...) was removed. This is still referenced in embedded resource comments (proteases.tsv:27) and may be used by consumers. Restore or obsolete, don't delete silently.

mzLib/Transcriptomics/Digestion/RnaseDictionary.cs
Lines 39-45: Public API LoadRnaseDictionary(string path) was removed. Same concern as above.
Lines 144-147: ResetToDefaults() was removed. Still referenced in rnases.tsv:24.

mzLib/Test/TestProteinDigestion.cs
Lines 139-167 (CNBrProteinDigestion): The test comment claims custom proteases "must have names that do NOT collide with the embedded resource," but then loads DoubleProtease.tsv which contains CNBr, CNBr_old, CNBr_N. These names collide with embedded entries and would be skipped under new semantics. The test passes because it falls back to embedded definitions, not because it exercises custom file loading. Either rename the custom entries to truly unique names, or rewrite the test to assert the skip behavior explicitly.
---

Non-blocking (suggestions)

---

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
Line 27: Comment references ResetToDefaults() which no longer exists. Update or remove.

mzLib/Transcriptomics/Digestion/rnases.tsv
Line 24: Comment references ResetToDefaults() which no longer exists. Update or remove.

mzLib/Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
Lines 12-52: The class docstring repeatedly describes the baseline as "immutable" and claims embedded entries "cannot be overridden," but the live Dictionary remains mutable by contents (add/remove/overwrite via indexer). Consider toning down to "protected baseline / private setter" to avoid overselling what the code actually enforces.

mzLib/Test/TestSeqCoverage.cs
Lines 116-157 (ReadCustomFile): This test was rewritten to use global dictionary mutation (LoadAndMergeCustomProteases) instead of standalone parsing. This weakens isolation—consider preserving at least one standalone parser test to verify file parsing correctness independently of merge semantics.
mzLib/Test/TestProteaseDictionaryEmbeddedMods.cs
General: Many new tests are well-structured, but the file has grown significantly. Consider splitting atomicity tests into a separate fixture to keep focused regression suites.

General
Scope: The PR bundles a focused behavior change (additive-only custom agents, atomic merge) with extensive comment rewrites and opportunistic cleanup. For future PRs, separate behavior changes from documentation/style churn to ease review.

---

Summary for maintainers

The core behavior change is sound (embedded baseline should be authoritative; custom files should merge atomically with skip-on-collision semantics). However, the PR removes public APIs without deprecation and contains at least one test (CNBrProteinDigestion) whose assertions no longer match its documentation under the new rules. Address the blocking items before merge.

trishorts:
Thanks for the thorough review — addressed all blocking items and most of the non-blocking suggestions.

## Blocking

- **Restored deprecated public APIs.** Added `[Obsolete]` thin wrappers for `LoadProteaseDictionary(path, mods)` and `ResetToDefaults(mods)` on `ProteaseDictionary`, and `LoadRnaseDictionary(path)` and `ResetToDefaults()` on `RnaseDictionary`. The parser wrappers do not mutate global state (matches pre-PR behavior); `ResetToDefaults` rebuilds the dictionary from the embedded resource. Both messages point downstream consumers at `LoadAndMergeCustomProteases` / `LoadAndMergeCustomRnases` and note migration guidance.
- **Fixed `CNBrProteinDigestion`.** Took the rename path you suggested: the three colliding rows in `DoubleProtease.tsv` (`CNBr`, `CNBr_old`, `CNBr_N`) are now `CNBr_custom`, `CNBr_old_custom`, `CNBr_N_custom`. Updated the test's dictionary lookups to match and added `Assert.That(result.Added, Contains.Item(...))` for each so the test now actually verifies the merge accepted the custom entries instead of silently falling back to the embedded definitions. The `Test1`/`Test2`/`Test3` rows used by `TestSeqCoverage.ReadCustomFile` are untouched, so that test is unaffected.

## Non-blocking — addressed

- **`proteases.tsv` / `rnases.tsv` comment blocks.** Removed the dead `ResetToDefaults()` references, rewrote the merge-rules section to reflect skip-on-collision semantics, and removed the stale "or override built-in definitions" wording on the intro lines. Also fixed a misleading example in `proteases.tsv` that used the colliding name `trypsin` — replaced it with `MyTrypsinNoProline` and added a note that a custom entry named `trypsin|P` would be skipped.
- **Class docstrings.** Toned down "immutable-baseline" → "protected-baseline" and removed "cannot be overridden by any custom file" / "additive only — they can never replace an embedded definition" in both `ProteaseDictionary` and `RnaseDictionary`. Added an explicit note that the `Dictionary` property has a private setter but the dictionary's contents remain mutable via the indexer, which matches what the code actually enforces.

## Non-blocking — deferred to a follow-up

- **Standalone parser test in `TestSeqCoverage`.** Agree the merge-coupled test weakens isolation. Holding off because the cleanest fix requires either making `ParseProteaseLines` `internal` + adding `InternalsVisibleTo("Test")` to `Proteomics.csproj`, or introducing a new public parse helper. Both feel like architectural choices that deserve their own PR rather than being bundled here.
- **Splitting `TestProteaseDictionaryEmbeddedMods.cs`.** Agree the file has grown. Pure reorganization with no behavior change — keeping it separate to honor the scope feedback below.

## Scope feedback

Point taken on separating behavior changes from doc/comment churn in future PRs.

## Other notes worth surfacing

- The new `LoadAndMergeCustomProteases` returns `CustomDigestionAgentLoadResult` rather than the old `List<string>`. That's an intentional part of the behavior change, but it is technically a breaking signature change for that method on top of the deletions you flagged. Mentioning it here in case a follow-up `[Obsolete]` overload returning `List<string>` would be useful.
- The deprecated `LoadProteaseDictionary` wrapper inherits the new "throw if a referenced mod is missing" behavior from `CreateProtease`, where the pre-PR version silently produced a mod-less protease in that case. I think throwing is the correct behavior, but it's a subtle change downstream consumers may notice through the wrapper.
-

trishorts:
## CI failure explanation (expected breaking change)

The build is failing due to an **expected breaking change** in MetaMorpheus’ dependency/API around the protease and RNase dictionaries.

In the failing job log:

- Workflow run/job: https://github.com/smith-chem-wisc/mzLib/actions/runs/25063295344/job/73423303650?pr=1040
- Ref (commit): `8e5e4fdc4bd2e9b2b532e6a0021ceb11971a63c3`

### What failed

Compilation fails in `MetaMorpheus/MetaMorpheus/EngineLayer/GlobalVariables.cs` with:

- `CS0200: Property or indexer 'ProteaseDictionary.Dictionary' cannot be assigned to -- it is read only`
- `CS0200: Property or indexer 'RnaseDictionary.Dictionary' cannot be assigned to -- it is read only`

This means MetaMorpheus currently tries to **replace** the entire dictionary object:

```csharp
ProteaseDictionary.Dictionary = ...;
RnaseDictionary.Dictionary = ...;

nbollis (dismissed):

pcruzparri (dismissed):

trishorts:
@copilot resolve the merge conflicts in this pull request

PR #1043, opened 2026-04-09
Generic Deconvolution Scoring

trishorts:
# Generic Deconvolution Scorer and Target-Decoy FDR Infrastructure

## Overview

This PR adds a generic post-hoc scorer and target-decoy FDR framework for isotopic
envelope deconvolution. The scorer operates on any `IsotopicEnvelope` regardless of
which algorithm produced it — Classic, IsoDec, or FLASHDeconv — and produces a
comparable quality score in [0, 1].

The motivating problem is that mzLib's three deconvolution algorithms place
incomparable values in `IsotopicEnvelope.Score`:

- **Classic** uses an empirical heuristic with no upper bound, designed for peptide data
- **IsoDec** places a cosine similarity in [0, 1]
- **FLASHDeconv** places a logistic-regression Qscore in [0, 1]

A caller cannot apply a single threshold across methods. This PR fixes that by
recomputing a method-agnostic quality score from the envelope's peak list and the
Averagine model.

---

## New Files

All scorer files live in a new `MassSpectrometry/Deconvolution/Scoring/` folder.

### `EnvelopeScoreFeatures.cs`

A `readonly struct` carrying four computed feature values for a single envelope.
It is a pure data carrier with no scoring logic.

| Field | What it measures |
|---|---|
| `AveragineCosineSimilarity` | Cosine similarity between the observed isotope intensity distribution and the theoretical Averagine distribution at the envelope's mass |
| `AvgPpmError` | Mean absolute ppm error of observed peaks vs their theoretical m/z positions, anchored at the apex peak |
| `PeakCompleteness` | Fraction of expected Averagine isotope peaks (>1% of theoretical max intensity) that were actually observed |
| `IntensityRatioConsistency` | Uniformity of the per-peak observed/theoretical intensity ratios, expressed as `1 / (1 + CV²)` where CV is the coefficient of variation of those ratios |

The fourth feature captures something the cosine alone cannot: a real isotope envelope
is the Averagine pattern scaled by a single abundance factor, so `observed[n] /
theoretical[n]` should be approximately the same constant across all isotope positions
(low CV → value near 1.0). A noise envelope has erratic ratios with no physical
coherence (high CV → value near 0.0). The constructor's fourth parameter defaults to
`0.0` for backward compatibility.

### `DeconvolutionScorer.cs`

A `static` class with four methods:

- `ComputeFeatures(IsotopicEnvelope, AverageResidue)` — computes all four features from
  the envelope's peak list and the Averagine model. Does not access the raw spectrum.
- `ComputeScore(EnvelopeScoreFeatures)` — combines the four features into a logistic
  score in [0, 1].
- `ScoreEnvelope(IsotopicEnvelope, AverageResidue)` — convenience method.
- `ScoreEnvelopes(IEnumerable<IsotopicEnvelope>, AverageResidue)` — lazy enumeration
  returning `(envelope, score)` pairs. Does not mutate `envelope.Score`.

**Score formula:**

```
linear = -3.4494 × cosine
       +  1.6341 × ppmError
       + -4.7795 × completeness
       + -2.1883 × ratioConsistency
       + -4.3142

score  = 1 / (1 + exp(linear))
```

**Weights are calibrated** — see Validation section below.

### `DeconvolutionDecoyGenerator.cs`

An `internal static` class that produces decoy `DeconvolutionParameters` instances
for target-decoy FDR estimation. Decoys are generated by cloning the target parameters
and setting `DecoyIsotopeDistance = 0.9444 Da` — a physically impossible isotope
spacing taken directly from OpenMS FLASHDeconv (`noise_iso_delta_`).

Also adds `BuildDecoySpectrum(MzSpectrum, double)` for Classic deconvolution, which
infers isotope spacing from observed peak-to-peak distances rather than from a
parameter. See Known Limitations.

### `DeconvolutionQValueCalculator.cs`

A `static` class implementing standard target-decoy q-values (Käll et al. 2008):

```
q(t) = #{decoys with score ≥ t} / #{targets with score ≥ t}
```

enforced to be monotonically non-decreasing as the score threshold decreases. Provides:

- `AssignQValues(targetScores, decoyScores)` — returns q-values in input order.
- `AssignQValues(targetPairs, decoyPairs)` — returns `(Envelope, Score, QValue)` triples.
- `GetFdrThreshold(targetScores, decoyScores, maxFdr)` — minimum score achieving a given FDR.

---

### `DeconvolutionParameters.cs`

Two new properties with backward-compatible defaults:

- `DecoyIsotopeDistance` — isotope spacing used during deconvolution, defaults to
  `Constants.C13MinusC12`. Set to `0.9444` by `DeconvolutionDecoyGenerator`.
- `IsDecoyRun` — informational flag, defaults to `false`.

Also adds `internal virtual ShallowClone()` via `MemberwiseClone()`.

### `IsoDecDeconvolutionParameters.cs`

- `ToIsoSettings()` now uses `DecoyIsotopeDistance` for `mass_diff_c` instead of the
  hard-coded `MassDiffC` property, so the shifted spacing reaches the C DLL.
- Overrides `ShallowClone()` to null the `_isoSettings` cache on the clone, forcing
  `ToIsoSettings()` to rebuild with the updated `DecoyIsotopeDistance`. Without this,
  the cached struct silently retains the original value and the decoy shift never
  reaches the DLL.

### `Deconvoluter.cs`

Adds `DeconvoluteWithDecoys(MzSpectrum, DeconvolutionParameters, MzRange?)` and an
`MsDataScan` overload. Returns `(List<IsotopicEnvelope> Targets, List<IsotopicEnvelope> Decoys)`.

### `SpectralSimilarity.cs`

Adds `CosineOfAlignedVectors(ReadOnlySpan<double>, ReadOnlySpan<double>)`.

### `Ms2ScanWithSpecificMass.cs` (MetaMorpheus)

Adds one optional constructor parameter and one read-only property:

```csharp
public double PrecursorDeconvolutionScore { get; }
// constructor parameter: double precursorDeconvolutionScore = -1.0
```

Default `-1.0` = not computed. All existing call sites unaffected.

---

### `TestDeconvolutionScorerUnit.cs`

12 unit tests (Groups A–C): feature correctness on synthetic envelopes, score bounds
and ordering, `ScoreEnvelopes` API immutability.

### `TestDeconvolutionScorerIntegration.cs`

14 integration tests (Groups D–G): real Classic output on synthetic spectra, decoy
pass, q-value end-to-end, edge cases.

### `TestShallowClone.cs`

15 unit tests covering `ShallowClone` on Classic (base) and IsoDec (override), plus
end-to-end `MakeDecoyParameters`. The critical test primes the `_isoSettings` cache,
clones, sets `DecoyIsotopeDistance = 0.9444`, calls `ToIsoSettings()`, and asserts
`mass_diff_c == 0.9444`.

### `TestDeconvolutionScorerValidation.cs`

Three validation tests tagged `[Category("DeconvolutionValidation")]` — not run in CI.
See Validation section below.

---

### Experiment design

For each algorithm, `DeconvoluteWithDecoys` was run on 2,000 MS1 scans. Target
envelopes are labelled T (real isotope spacing); decoy envelopes are labelled D
(shifted isotope spacing, 0.9444 Da). Features are computed post-hoc from the peak
list. A logistic regression is fit on the labelled set using
`fit_scorer_weights.py`.

### Results

**Classic bottom-up** (A549, Velos Orbitrap, z=1–12, n=146,899 T / 146,899 D):
AUC = **0.50**. All feature means identical between T and D to 4 decimal places.
Classic's `Peak2satisfiesRatio` filter pre-enforces Averagine shape on every recruited
envelope — including decoys from a shifted spectrum — so no post-hoc feature can
distinguish them.

**Classic top-down** (Yeast, z=1–60, n=same structure):
AUC = **0.50**. Same result. The pre-filtering problem is fundamental to Classic and
independent of charge range.

**IsoDec top-down** (Yeast `05-26-17_B7A_yeast_td_fract8_rep2.mzML`, n=381,098 T /
198,445 D):
AUC = **1.00**. All four features strongly separate targets from decoys:

| Feature | Target mean | Decoy mean | Delta |
|---|---|---|---|
| Cosine | 0.871 | 0.064 | +0.807 |
| PpmError | 0.919 | 22.046 | −21.127 |
| Completeness | 0.949 | 0.223 | +0.726 |
| RatioConsistency | 0.700 | 0.229 | +0.471 |
| GenericScore | 0.981 | 0.010 | +0.971 |

IsoDec passes `DecoyIsotopeDistance` directly to the C DLL as `mass_diff_c`. The decoy
pass therefore finds isotope series at the wrong spacing, producing candidate masses at
genuinely different positions. The post-hoc scorer correctly identifies them as noise.

The logistic weights in `DeconvolutionScorer.cs` are calibrated from this experiment.

### Target score distribution (IsoDec top-down, targets only)

Inspecting the target distribution without the decoy comparison:

| Threshold | Count | Fraction |
|---|---|---|
| GenericScore < 0.5 | 5,405 | 1.4% |
| GenericScore < 0.7 | 7,153 | 1.9% |
| GenericScore < 0.9 | 11,448 | 3.0% |
| GenericScore ≥ 0.9 | 369,650 | 97.0% |

97% of target envelopes score above 0.9. There is a small but real low-scoring tail
(~3%) representing genuinely poor deconvolution events within real data.

### Implication for q-value usage

The IsoDec decoy distribution does not overlap the target distribution — the two are
perfectly separated. This means q-values computed from this decoy pool will be near
zero for all targets and a q-value threshold is not a meaningful filter. The underlying
cause is that shifted-isotope decoys for IsoDec represent an extreme worst case (22 ppm
error, 22% completeness) rather than a realistic model of borderline false positives.

**The recommended use of the scorer is as a continuous quality score, not a binary FDR
filter.** Specifically:

- Pass `PrecursorDeconvolutionScore` to `Ms2ScanWithSpecificMass` and let PEP learn its
  weight relative to other PSM features. PEP will naturally downweight PSMs whose
  precursor scored in the low tail.
- Use the score as a loose pre-filter (e.g. discard envelopes below 0.3) to reduce the
  number of MS2 scans queried, without committing to a calibrated FDR threshold.
- Reserve `DeconvolutionQValueCalculator` for future use once a more realistic null
  distribution is available — one that models borderline deconvolution events rather
  than obvious noise.

---

### Scoring individual envelopes

```csharp
var model    = new Averagine();
var envelope = Deconvoluter.Deconvolute(scan, parameters).First();
double score = DeconvolutionScorer.ScoreEnvelope(envelope, model);
// score in [0, 1]; higher = better quality
```

### Appending to a PSM for PEP

```csharp
var scan = new Ms2ScanWithSpecificMass(
    mzLibScan, precursorMz, charge, filePath, commonParams,
    precursorDeconvolutionScore: DeconvolutionScorer.ScoreEnvelope(precursorEnvelope, model));
// scan.PrecursorDeconvolutionScore available as a PEP feature
// Defaults to -1.0 at all existing call sites
```

### Target-decoy q-value pipeline (for future use)

```csharp
// Accumulate across all files before computing q-values
var allTargets = new List<IsotopicEnvelope>();
var allDecoys  = new List<IsotopicEnvelope>();

foreach (var file in rawFiles)
    foreach (var scan in MsDataFileReader.GetDataFile(file).LoadAllStaticData().GetMS1Scans())
    {
        var (t, d) = Deconvoluter.DeconvoluteWithDecoys(scan, parameters);
        allTargets.AddRange(t);
        allDecoys.AddRange(d);
    }

var results = DeconvolutionQValueCalculator.AssignQValues(
    DeconvolutionScorer.ScoreEnvelopes(allTargets, model).ToList(),
    DeconvolutionScorer.ScoreEnvelopes(allDecoys,  model).ToList());
```

---

## Known Limitations

**Classic decoys are not discriminating.** AUC = 0.50 for both bottom-up and top-down
Classic, for the same structural reason: `Peak2satisfiesRatio` pre-enforces Averagine
shape on every recruited envelope, including those found on a shifted-spectrum decoy
pass. The scorer still produces a valid quality score for Classic output — it just
cannot be calibrated via this decoy strategy.

**Shifted-isotope decoys model extreme noise, not borderline events.** The IsoDec decoy
envelopes have 22 ppm error and 22% completeness — these are obviously wrong, not
borderline. A realistic null distribution for FDR estimation should model the boundary
between good and bad deconvolution, which requires a different decoy construction
strategy. This is deferred to future work.

**No spectrum access.** All features are computed from the peak list alone. SNR-type
features requiring the raw spectrum are not available under the deconvolute-and-discard
assumption.

---

## Calibration Path

The current weights were calibrated on IsoDec top-down yeast data. Future recalibration
opportunities in priority order:

1. **Cross-algorithm consensus** — once FLASHDeconv is merged, use agreement between
   Classic, IsoDec, and FLASHDeconv within 5 ppm as positive labels and algorithm-
   specific envelopes not confirmed by the others as negative labels. This models
   borderline deconvolution quality rather than obvious noise.
2. **FLASHDeconv decoys** — FLASHDeconv uses `DecoyIsotopeDistance` natively in Step 2
   candidate mass search, expected to produce a more realistic null distribution than
   IsoDec's perfect separation.
3. **PEP feedback** — once `PrecursorDeconvolutionScore` is included in the PEP feature
   vector, the model's learned weight for this feature provides indirect validation of
   whether the score adds information beyond existing PSM features.


mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
nbollis, line 1: It seems like the only time these new parameters are used is when we clone the parameters. 

I think we can get rid of IsDecoyRun and then change DecoySpacing to ExpectedIsotopeSpacing.

mzLib/MassSpectrometry/Deconvolution/Deconvoluter.cs
```diff
@@ -46,6 +47,65 @@ public static IEnumerable<IsotopicEnvelope> Deconvolute(MzSpectrum spectrum,
...
+                DeconvolutionParameters parameters,
+                MzRange rangeToGetPeaksFrom = null)
+        {
+            var targets = Deconvolute(spectrum, parameters, rangeToGetPeaksFrom).ToList();
+
+            DeconvolutionParameters decoyParams =
+                DeconvolutionDecoyGenerator.MakeDecoyParameters(parameters);
+            var decoys = Deconvolute(spectrum, decoyParams, rangeToGetPeaksFrom).ToList();
+
+            return (targets, decoys);
```
nbollis, line 71: This appears to create a new decon parameters, with the decoy specific fields set then call decovolute again. 

Lets change this to parameters.ToDecoyParameters()

mzLib/MassSpectrometry/Deconvolution/Scoring/DeconvolutionDecoyGenerator.cs
nbollis, line 1: Can get rid of this class if we use ToDecoyParameters

mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
```diff
@@ -60,7 +60,10 @@ internal IsoSettings ToIsoSettings()
             min_score_diff = MinScoreDiff,
             minareacovered = MinAreaCovered,
             isolength = IsoLength,
-            mass_diff_c = MassDiffC,
+            // Use DecoyIsotopeDistance rather than MassDiffC so that a decoy pass
+            // (DecoyIsotopeDistance = 0.9444 Da) is correctly forwarded to the DLL.
+            // For normal runs DecoyIsotopeDistance == Constants.C13MinusC12 == MassDiffC.
+            mass_diff_c = DecoyIsotopeDistance,
```
nbollis: This feels odd in that we are using the `DecoyIsotopeDistance` to set the target distance as well. Do we need the decoy isotope distance to be a settable parameter? 

One alternative would be to remove the property and use two static/const fields, target distance and decoy distance. Then the derived parameters `ToDeconParams` method can either use these constants or set their own for special cases.

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
nbollis: Lets add a public abstract DeconvolutionParameters ToDecoyParameters(). This will allow each of the decon algorithms to set what is needed for them to operate in decoy mode. 

Can do it with caching like this to avoid repeated object construction in the derived classes. 
```
private DeconvolutionParameters? _decoyParams = null
public override DeconvolutionParameters ToDecoyParams() => _decoyParams ?? new(....)
```

mzLib/MassSpectrometry/Deconvolution/Scoring/DeconvolutionDecoyGenerator.cs
trishorts: deleted

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
trishorts: i think i did this

mzLib/MassSpectrometry/Deconvolution/Deconvoluter.cs
trishorts: see above. decoy generator gone and we have a virtual todecoyparams in params

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
```diff
@@ -13,16 +12,35 @@ public abstract class DeconvolutionParameters
...
+        /// The returned object is lazily constructed and cached — calling this multiple
+        /// times returns the same instance.
+        /// Subclasses that do not support decoy deconvolution may leave this default
+        /// implementation, which returns <c>null</c>.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="DeconvolutionParameters"/> configured for decoy spacing,
+        /// or <c>null</c> if this parameter type does not support decoy deconvolution.
+        /// </returns>
+        public virtual DeconvolutionParameters? ToDecoyParameters() => null;
```
nbollis: Abstract without default to guarantee override on subsequent decon params that we make

trishorts: done now

mzLib/MassSpectrometry/Deconvolution/Scoring/EnvelopeScoreFeatures.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// The computed feature vector for a single <see cref="IsotopicEnvelope"/>,
+    /// produced by <see cref="DeconvolutionScorer.ComputeFeatures"/> and consumed
+    /// by <see cref="DeconvolutionScorer.ComputeScore"/>.
+    ///
+    /// All four features are computable from the envelope's peak list and an
+    /// <see cref="AverageResidue"/> model alone — no raw spectrum access is required.
+    /// This allows the scorer to operate after the spectrum has been discarded
+    /// (deconvolute-and-discard callers).
+    /// </summary>
+    public readonly struct EnvelopeScoreFeatures
```
Alexander-Sol, line 21: This is pretty big for a struct ( 64 bytes ). What is the advantage of using a struct over a class? Can we use a Record instead?

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -404,6 +404,43 @@ public static double[] NormalizeSpectrumSum(double[] spectrum)
...
+        /// against a theoretical Averagine distribution after the two vectors have already been
+        /// aligned to the same isotope index grid.
+        /// </summary>
+        /// <param name="a">First intensity vector (e.g. observed isotope intensities).</param>
+        /// <param name="b">Second intensity vector (e.g. theoretical Averagine intensities).</param>
+        /// <returns>
+        /// Cosine similarity in [0, 1] for non-negative input vectors.
+        /// Returns 0 if either vector is all-zero or if the lengths differ.
+        /// </returns>
+        public static double CosineOfAlignedVectors(ReadOnlySpan<double> a, ReadOnlySpan<double> b)
```
Alexander-Sol, line 422: This looks fine, but I think we have similar functionality in SpectralSimilarity

mzLib/MassSpectrometry/Deconvolution/Deconvoluter.cs
trishorts: This was already addressed -- DeconvoluteWithDecoys now calls parameters.ToDecoyParameters() directly at Deconvoluter.cs:61; DeconvolutionDecoyGenerator.MakeDecoyParameters path gone.

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
trishorts: This was already addressed -- IsDecoyRun removed and DecoySpacing renamed to ExpectedIsotopeSpacing; verified at DeconvolutionParameters.cs:19.

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
trishorts: This was already addressed -- ToDecoyParameters is now abstract at DeconvolutionParameters.cs:43 (was virtual with => null default); ExampleNewDeconvolutionParametersTemplate.cs:17 overrides with => null.

mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
trishorts: This was already addressed -- DecoyIsotopeDistance property removed; ToIsoSettings now sets mass_diff_c = ExpectedIsotopeSpacing (IsoDecDeconvolutionParameters.cs:69) where ExpectedIsotopeSpacing is a base-class property set per-instance by ToDecoyParameters().

mzLib/MassSpectrometry/Deconvolution/Scoring/DeconvolutionDecoyGenerator.cs
trishorts: This was already addressed -- DeconvolutionDecoyGenerator.cs file deleted; Deconvoluter uses parameters.ToDecoyParameters() directly.

mzLib/MassSpectrometry/Deconvolution/Scoring/EnvelopeScoreFeatures.cs
trishorts: Going to leave this as-is. Keeping readonly struct: actual size is 32 bytes (4 doubles), not 64 as the comment estimated. ComputeFeatures runs once per envelope on the batch-scoring hot path (thousands per scan); a class/record-class would heap-allocate per envelope. The 32-byte copy cost on the single ComputeFeatures->ComputeScore boundary is a worthwhile tradeoff for avoiding GC pressure at scale. Added a <remarks> doc block on the struct documenting the choice.

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
trishorts: Going to leave this as-is. Not a duplicate: SpectralSimilarity.CosineSimilarity() is an instance method that handles mz-tolerance peak matching internally; CosineOfAlignedVectors is a static helper that assumes alignment is already done. DeconvolutionScorer.ComputeFeatures does its own isotope-index alignment to Averagine positions, so it needs the no-matching variant. A future DIA method will also use this static helper with pre-matched feature vectors -- alignment is the caller's responsibility, the helper just computes the metric.

Alexander-Sol (approved):

nbollis (approved):

PR #1045, opened 2026-04-09
Real OpenMS FLASHDeconv using exe from provided installation path

trishorts:
> **Project of origin:** `MetaFlashDecon`

# Add `RealFLASHDeconvolution` — wrapper for the official FLASHDeconv executable (OpenMS)

## Summary

This PR adds `RealFLASHDeconvolution` as a new first-class deconvolution method in mzLib. It wraps the official **FLASHDeconv executable from OpenMS** behind the standard `DeconvolutionParameters` / `DeconvolutionAlgorithm` / `Deconvoluter` interface, so MetaMorpheus can use it transparently alongside Classic, IsoDec, and the reverse-engineered `FLASHDeconvolution`.

The design follows the `IsoDecAlgorithm` pattern — an external binary invoked at runtime, with results converted to `IsotopicEnvelope` objects. IsoDec uses a DLL via P/Invoke; `RealFLASHDeconvolution` uses `FLASHDeconv.exe` via `System.Diagnostics.Process`.

---

## Motivation

mzLib already contains a reverse-engineered pure-C# implementation of FLASHDeconv (`FLASHDeconvolutionAlgorithm`). This PR gives users access to the **official validated binary** from OpenMS without leaving the mzLib/MetaMorpheus pipeline, enabling direct comparison and providing a gold-standard reference implementation.

---

## Files Added

| File | Location | Description |
|---|---|---|
| `RealFLASHDeconvolutionParameters.cs` | `MassSpectrometry/Deconvolution/Parameters/` | User-configurable parameters; maps to FLASHDeconv CLI flags |
| `RealFLASHDeconvolutionAlgorithm.cs` | `MassSpectrometry/Deconvolution/Algorithms/` | Writes temp mzML, runs `FLASHDeconv.exe`, parses TSV output |
| `TestRealFLASHDeconvolutionInfrastructure.cs` | `Test/Deconvolution/` | Infrastructure + real-exe tests (11 tests, all passing) |

## Files Modified

| File | Change |
|---|---|
| `DeconvolutionType.cs` | Added `RealFLASHDeconvolution` enum value |
| `Deconvoluter.cs` | Added `RealFLASHDeconvolution` case to factory switch |

---

## Architecture

```
DeconvolutionType.RealFLASHDeconvolution
        │
        ▼
Deconvoluter.CreateAlgorithm()
        │
        ▼
RealFLASHDeconvolutionAlgorithm
        │
        ├─ Per-scan:   Deconvolute(MzSpectrum, MzRange)
        │              → writes temp mzML → runs FLASHDeconv.exe → parses _ms1.tsv
        │
        └─ Whole-file: DeconvoluteFile(mzmlPath, parameters)   [preferred]
                       → passes full mzML directly → parses _ms1.tsv
                       → returns Dictionary<int scanNum, List<IsotopicEnvelope>>
```

**`DeconvoluteFile` is the preferred mode.** FLASHDeconv's multi-scan feature tracing only works on complete files. The per-scan `Deconvolute()` path exists for compatibility with the standard `Deconvoluter` interface but will produce sparse results on simple spectra.

---

### Minimal
```csharp
var p = new RealFLASHDeconvolutionParameters(
    flashDeconvExePath: @"C:\Program Files\OpenMS-3.0.0\bin\FLASHDeconv.exe");

// Whole-file (preferred)
var byScan = RealFLASHDeconvolutionAlgorithm.DeconvoluteFile(mzmlPath, p);

// Per-scan via standard interface
var envelopes = Deconvoluter.Deconvolute(scan, p);
```

### Exe auto-detection

The algorithm searches these locations automatically if `FLASHDeconvExePath` is null:

- `C:\Program Files\OpenMS-3.x.x\bin\FLASHDeconv.exe` (several versions)
- `/usr/bin/FLASHDeconv`, `/usr/local/bin/FLASHDeconv`
- System `PATH`

---

## FLASHDeconv CLI flags (verified against OpenMS 3.0.0-pre-HEAD-2023-06-17)

| Parameter | Flag |
|---|---|
| Input mzML | `-in` |
| Feature TSV (required) | `-out` |
| Per-spectrum TSV (MS1, MS2) | `-out_spec ms1.tsv ms2.tsv` |
| Tolerance | `-Algorithm:tol` |
| Charge range | `-Algorithm:min_charge` / `-Algorithm:max_charge` |
| Mass range | `-Algorithm:min_mass` / `-Algorithm:max_mass` |
| Cosine threshold | `-Algorithm:min_isotope_cosine` |

---

### mzML writer
The algorithm contains a self-contained mzML writer. Several requirements were found empirically against the OpenMS 3.0 pre-release parser:

- **No UTF-8 BOM** — `new UTF8Encoding(encoderShouldEmitUTF8Identifier: false)`. OpenMS rejects BOM silently by crashing with a misleading error.
- `<run>` element must have `defaultInstrumentConfigurationRef="IC1"` — in addition to the `<spectrum>` element.
- `softwareList`, `instrumentConfigurationList` (id=`IC1`), and `dataProcessingList` (id=`dp`) are all required.

### `-out_spec` requires two paths for mixed MS1/MS2 files
Passing a single path to `-out_spec` when the input contains both MS1 and MS2 causes FLASHDeconv to write MS2 output to that path and produce no MS1 output. Two paths must always be passed; we read only the MS1 one.

### Trailing tab in TSV header
FLASHDeconv writes a trailing `\t` after the last column name in the header row. The TSV parser strips empty strings from the split header with `.Where(h => !string.IsNullOrEmpty(h))` to prevent every data row from being rejected by a length mismatch.

### Single-peak envelopes
The standard `-out_spec` TSV provides per-envelope summary rows, not individual isotope peak m/z values. Each returned `IsotopicEnvelope` therefore contains a single peak at the representative m/z (midpoint of `RepresentativeMzStart` and `RepresentativeMzEnd`). For full isotope peak lists, use `-out_topFD` with the existing `MsAlign` reader.

### Score mapping
`Qscore` is used as the envelope score when present; `MassSNR` is used as a fallback; `IsotopeCosine` is the final fallback.

---

## Tests

All 11 tests pass. Tests are in `TestRealFLASHDeconvolutionInfrastructure.cs`.

**Infrastructure tests (no exe required):**
- Parameters class defaults and explicit values
- `DeconvolutionType.RealFLASHDeconvolution` enum value exists
- Empty spectrum short-circuits without calling exe
- Bad exe path throws `FileNotFoundException`

**Real-exe tests (require FLASHDeconv at known path or `FLASHDECONV_EXE` env var):**
- `RealFLASH_RealExe_WholeFile_SmallYeast_ReturnsEnvelopes` — `DeconvoluteFile` on `SmallCalibratibleYeast.mzml` returns envelopes
- `RealFLASH_RealExe_WholeFile_SmallYeast_EnvelopeCountMatchesReference` — count matches reference run (3149 ± 5)
- `RealFLASH_RealExe_WholeFile_SmallYeast_EnvelopeFieldsValid` — all field validity checks
- `RealFLASH_RealExe_WholeFile_SmallYeast_MassesInRange` — all masses within requested bounds

---

## Required User Setup

1. Download and install **OpenMS** from [openms.de](https://www.openms.de)
2. Set `RealFLASHDeconvolutionParameters.FLASHDeconvExePath` to the `FLASHDeconv` binary, **or** add the OpenMS `bin/` directory to `PATH`

For MetaMorpheus integration, `GlobalSettings.FLASHDeconvExecutablePath` should be added and wired to a Browse button in the Settings UI (deferred to a follow-up PR).

---

## Known Limitations

- **Speed:** Each `Deconvolute()` call launches a new process and writes/reads temp files. For whole-file work, use `DeconvoluteFile()` directly.
- **Single-peak envelopes:** Use `RetainMsAlignOutput = true` + the `MsAlign` reader for full isotope peak lists.
- **MS2 not wired:** Only the `_ms1.tsv` file is parsed. MS2 support is a straightforward extension.
- **Scan number extraction:** `SmallCalibratibleYeast.mzml` (and similar files) produce `ScanNum = 0` for all rows because FLASHDeconv cannot parse the native scan ID format. Data is still returned correctly, just all keyed to scan 0.

---

## References

- Jeong et al. (2020) *Cell Systems* 10(2):213–218.e6. DOI: 10.1016/j.cels.2020.01.003
- OpenMS source: `src/openms/source/ANALYSIS/TOPDOWN/SpectralDeconvolution.cpp`


mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
nbollis, line 1: Cool with using local resources and enabling this decon in our system. There should be a public facing method allowing someone to add a path to the registry it checks. That way Metamorpheus users can set their local paths. 

Question: why do this instead of binding up a version?

Alexander-Sol (dismissed):

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
trishorts: claude convinced me not to and I didn't really want to put up a fight. Here's the critical text:

# Bundling FLASHDeconv.exe with mzLib / MetaMorpheus

## The Problem

Unlike IsoDec's `isodeclib.dll` (which is a self-contained ~2 MB DLL), `FLASHDeconv.exe` is part of OpenMS — a full C++ application that links against the OpenMS shared library (`OpenMS.dll` on Windows, ~50–100 MB) plus its own dependencies (HDF5, Boost, zlib, etc.). You can't just copy the `.exe` — it will crash immediately looking for `OpenMS.dll` and a dozen other files.

There are three realistic approaches, from simplest to most complete.

---

## Option 1 — Copy the whole OpenMS `bin/` folder (simplest, works today)

After installing OpenMS, go to `C:\Program Files\OpenMS-3.x.x\` and look at the structure:

```
OpenMS-3.x.x/
  bin/
    FLASHDeconv.exe     ← the tool
    OpenMS.dll          ← the main library (~80 MB)
    Qt6Core.dll         ← Qt runtime
    Qt6Gui.dll
    ... ~30-50 other DLLs
  share/
    OpenMS/             ← data files (CV terms, etc.) — also required
```

**FLASHDeconv.exe needs `OPENMS_DATA_PATH`** set to the `share/OpenMS/` directory or it fails to start. This is the key fact that makes simple bundling tricky.

The minimal set you actually need to copy into mzLib (from a real OpenMS install) is:

```
MassSpectrometry/
  FLASHDeconv.exe
  OpenMS.dll
  + all DLLs that FLASHDeconv.exe depends on
  share/
    OpenMS/       ← critical data files
```

**How to find the exact DLL list:** On a machine with OpenMS installed, open a Developer Command Prompt and run:

```cmd
dumpbin /dependents "C:\Program Files\OpenMS-3.x.x\bin\FLASHDeconv.exe"
```

Then for each DLL it lists, check if it's in the OpenMS `bin/` folder (those need to be copied) vs. a system DLL like `kernel32.dll` (those don't).

---

## Option 2 — Ship a self-contained zip via GitHub LFS or a NuGet package (recommended for a real repo)

This is how serious tools handle large native binaries.

### Step 1 — Create a zip of the minimal runtime

On a machine with OpenMS installed:

```powershell
# Run this after installing OpenMS 3.x
$openmsDir = "C:\Program Files\OpenMS-3.x.x"
$outDir = "flashdeconv_bundle_win64"

New-Item $outDir -ItemType Directory

# Copy the exe and all DLLs from bin/
Copy-Item "$openmsDir\bin\FLASHDeconv.exe" $outDir
Get-ChildItem "$openmsDir\bin\*.dll" | Copy-Item -Destination $outDir

# Copy the required share/OpenMS data folder
Copy-Item "$openmsDir\share\OpenMS" "$outDir\share\OpenMS" -Recurse
```

Then zip it: `Compress-Archive $outDir flashdeconv_win64.zip`

### Step 2 — Store it in GitHub LFS

```bash
git lfs track "*.zip"
git lfs track "FLASHDeconv.exe"
git lfs track "OpenMS.dll"
git add .gitattributes
git add MassSpectrometry/FLASHDeconv/
git commit -m "Add FLASHDeconv binary bundle"
```

### Step 3 — Set `OPENMS_DATA_PATH` in the algorithm

This is the critical piece. Update `RealFLASHDeconvolutionAlgorithm.RunFLASHDeconv()` to set the environment variable before launching the process:

```csharp
// In RunFLASHDeconv(), add to ProcessStartInfo:
proc.StartInfo.EnvironmentVariables["OPENMS_DATA_PATH"] =
    Path.Combine(AppContext.BaseDirectory, "share", "OpenMS");
```

And in the `.csproj`, copy the whole bundle to output:

```xml
<ItemGroup>
  <!-- The exe -->
  <None Update="FLASHDeconv\FLASHDeconv.exe">
    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
  </None>
  <!-- All DLLs in the bundle folder -->
  <None Update="FLASHDeconv\*.dll">
    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
  </None>
  <!-- The share/OpenMS data directory -->
  <None Update="FLASHDeconv\share\OpenMS\**\*">
    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    <Link>share\OpenMS\%(RecursiveDir)%(Filename)%(Extension)</Link>
  </None>
</ItemGroup>
```

And update `ResolveExePath()` to look in the bundle folder first:

```csharp
private static string ResolveExePath(string? explicitPath)
{
    if (!string.IsNullOrWhiteSpace(explicitPath) && File.Exists(explicitPath))
        return explicitPath!;

    // Look next to the running assembly (bundled)
    string assemblyDir = AppContext.BaseDirectory;
    string[] bundleCandidates = {
        Path.Combine(assemblyDir, "FLASHDeconv.exe"),               // flat layout
        Path.Combine(assemblyDir, "FLASHDeconv", "FLASHDeconv.exe"), // subfolder layout
    };
    foreach (string candidate in bundleCandidates)
        if (File.Exists(candidate)) return candidate;

    // Fall back to PATH search
    // ... (existing PATH search code)
}
```

---

## Option 3 — Just document it (pragmatic for a research tool)

For a research codebase like MetaMorpheus that's used by proteomics labs, the most common pattern is:

1. Put instructions in the README: "Install OpenMS 3.x from openms.de"
2. The algorithm searches common install paths automatically
3. The GUI has a "Browse..." button to set the path

Update `ResolveExePath()` with common install locations:

```csharp
string[] wellKnownPaths = {
    // Windows
    @"C:\Program Files\OpenMS-3.5.0\bin\FLASHDeconv.exe",
    @"C:\Program Files\OpenMS-3.4.0\bin\FLASHDeconv.exe",
    @"C:\Program Files\OpenMS-3.0.0\bin\FLASHDeconv.exe",
    // Linux package install
    @"/usr/bin/FLASHDeconv",
    @"/usr/local/bin/FLASHDeconv",
    // macOS
    @"/Applications/OpenMS-3.5.0/bin/FLASHDeconv",
};
foreach (string p in wellKnownPaths)
    if (File.Exists(p)) return p;
```

---

## Recommendation

**Start with Option 3** right now — it costs 10 lines of code and works for any lab that already has OpenMS installed. Then if you want zero-setup for users, do Option 2 (zip + LFS) as a follow-up. Option 1 is fine for personal use but creates a maintenance burden when OpenMS updates.

**The one line you must not forget regardless of option** is setting `OPENMS_DATA_PATH` in `ProcessStartInfo.EnvironmentVariables` before launching the process:

```csharp
proc.StartInfo.EnvironmentVariables["OPENMS_DATA_PATH"] =
    Path.Combine(Path.GetDirectoryName(exePath)!, "..", "share", "OpenMS");
```

Without it, `FLASHDeconv.exe` will fail to start even if all the DLLs are present, because it can't find its CV term files.

nbollis: Cool. Makes sense. 

That bottom bit needs to be in a static constructor for FlashDeconvAlgorithm. 

We also need a public facing method to add a new path to the dictionary of paths that it checks. This will enable MetaMorpheus to register a new path to it.

nbollis (changes requested):

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
trishorts: On the registry method — I think the user-facing piece you're after is already there, just under a different name. The single config knob is
  `RealFLASHDeconvolutionParameters.FLASHDeconvExePath` (a `string?`); whatever a consumer sets there goes through the resolver as the explicit path:

  ```
  ResolveExePath(explicitPath):
    1. explicitPath set + file exists  → return it
    2. explicitPath set + missing       → throw FileNotFoundException
    3. walk DefaultWellKnownPaths       → first match
    4. walk PATH env var                → first match
    5. throw FileNotFoundException
  ```

  For MetaMorpheus, the chain is: settings UI → `GlobalSettings.FLASHDeconvExecutablePath` (`GlobalSettings.toml`) → `params.FLASHDeconvExePath` → mzLib uses it. The well-known list never
   runs in that path because step (1) returns. Same shape for any non-MM consumer with their own settings mechanism:

  | Consumer | "User has unusual install" path |
  |---|---|
  | MetaMorpheus | UI → `GlobalSettings.FLASHDeconvExecutablePath` → params |
  | CLI tool | `--flashdeconv-exe` / env var → params |
  | Library | host's config / DI → params |
  | Script | hardcoded or argparse → params |

  The hardcoded `DefaultWellKnownPaths` list is just convenience for "user installed OpenMS in the default location and didn't bother configuring anything" — not where end-users should
  configure paths.

  A public `static IList<string> AdditionalSearchPaths` registry would only save plumbing in the niche case of a consumer with *no* settings system at all that still wants auto-discovery
  extended. Cost: static mutable state in a library (multiple consumers in one process can't disagree), a second config surface competing with `FLASHDeconvExePath` (now two right answers;
   docs have to disambiguate), and no new capability — anything the registry could express, the explicit-path knob already does.

  Happy to add it later if a real consumer hits a case where the explicit-path approach is awkward, but I'd rather not introduce mutable static state speculatively.

RayMSMS (dismissed):

trishorts:
Please compare the diff to ensure that you are meeting all of the contractual obligations.

# mzLib Spectral Deconvolution — Contractual Obligations

Distilled from the current wiki page:
https://github.com/smith-chem-wisc/mzLib/wiki/Spectral-Deconvolution-(Decharging-and-Deisotoping)

Last edited: Nov 25, 2025 (2 revisions)

---

## 1. Supported Algorithms

The wiki formally documents **two** supported algorithms:

| Algorithm | Class | Parameters Class |
|---|---|---|
| Classic Deconvolution | `ClassicDeconvolutionAlgorithm` | `ClassicDeconvolutionParameters` |
| IsoDec | `IsoDecAlgorithm` | `IsoDecDeconvolutionParameters` |

**Obligation:** Any new algorithm must be listed in this table with a description of what it is best for and its key features.

---

## 2. Architecture Contracts

### 2.1 `Deconvoluter` (Static Context Class)

Required public API:

```csharp
public static IEnumerable<IsotopicEnvelope> Deconvolute(
    MzSpectrum spectrum,
    DeconvolutionParameters deconvolutionParameters,
    MzRange rangeToGetPeaksFrom = null);

public static IEnumerable<IsotopicEnvelope> Deconvolute(
    MsDataScan scan,
    DeconvolutionParameters deconvolutionParameters,
    MzRange rangeToGetPeaksFrom = null);
```

**Obligation:** Both overloads must exist and be public. The factory pattern (`CreateAlgorithm`) must remain private.

---

### 2.2 `DeconvolutionParameters` (Abstract Base)

Required properties as documented:

```csharp
public abstract DeconvolutionType DeconvolutionType { get; protected set; }
public int MinAssumedChargeState { get; set; }
public int MaxAssumedChargeState { get; set; }
public Polarity Polarity { get; set; }
public AverageResidue AverageResidueModel { get; set; }
```

**Obligation:** All five of these properties must be present on the base class. The wiki shows `AverageResidueModel` as part of the base — any addition to the base class (e.g. `ExpectedIsotopeSpacing`, `ToDecoyParameters()`) is an extension of this contract, not a violation, but should be documented.

---

### 2.3 `DeconvolutionAlgorithm` (Abstract Base)

Required structure:

```csharp
public abstract class DeconvolutionAlgorithm
{
    protected readonly AverageResidue AverageResidueModel;
    protected readonly DeconvolutionParameters DeconvolutionParameters;

    internal abstract IEnumerable<IsotopicEnvelope> Deconvolute(
        MzSpectrum spectrum,
        MzRange range);
}
```

**Obligation:** `Deconvolute` must be `internal abstract`. The two protected fields must be present.

---

### 2.4 `IsotopicEnvelope` (Result Type)

Required public interface:

```csharp
public readonly List<(double mz, double intensity)> Peaks;
public double MonoisotopicMass { get; private set; }
internal double MostAbundantObservedIsotopicMass { get; private set; }
public readonly int Charge;
public readonly double TotalIntensity;
public readonly int PrecursorId;
public double Score { get; private set; }
```

**Obligation:** All of these members must exist. `Score` must be settable (privately). `MostAbundantObservedIsotopicMass` is internal only.

---

### 2.5 `DeconvolutionType` Enum

Documented values:

```csharp
public enum DeconvolutionType
{
    ClassicDeconvolution,
    IsoDecDeconvolution,
    ExampleNewDeconvolutionTemplate
}
```

**Obligation:** These three values must exist. New algorithms require a new enum value and a corresponding case in `CreateAlgorithm()`.

---

## 3. Averagine Models

Documented models:

| Model | Biomolecule | Formula |
|---|---|---|
| `Averagine` | Proteins/Peptides | C4.9384 H7.7583 N1.3577 O1.4773 S0.0417 |
| `OxyriboAveragine` | RNA/Nucleotides | C9.75 H12.25 N3.75 O7 P |

**Obligation:** Both models must remain available. New models (e.g. `DecoyAveragine`) should be added to this table in a wiki update.

---

## 4. `ClassicDeconvolutionParameters` Contract

Required constructor and parameters:

```csharp
new ClassicDeconvolutionParameters(
    minCharge: int,
    maxCharge: int,
    deconPpm: double,
    intensityRatio: double,
    polarity: Polarity = Polarity.Positive,
    averageResidueModel: new Averagine()  // optional
);
```

Required properties:
- `DeconvolutionTolerancePpm`
- `IntensityRatioLimit`

**Obligation:** These constructor parameters and property names must remain stable. The wiki example shows `deconPpm` and `intensityRatio` as the positional parameter names.

---

## 5. `IsoDecDeconvolutionParameters` Contract

Required user-accessible parameters (all must remain):

| Parameter | Type | Default | Description |
|---|---|---|---|
| `PhaseRes` | `int` | 8 | Precision of encoding matrix |
| `CssThreshold` | `float` | 0.7 | Min cosine similarity score |
| `MatchTolerance` | `float` | 5 | Peak matching tolerance (ppm) |
| `MaxShift` | `int` | 3 | Max allowed isotopic shift |
| `MzWindow` | `float[]` | [-1.05, 2.05] | m/z window |
| `KnockdownRounds` | `int` | 5 | Iterative refinement rounds |
| `MinAreaCovered` | `float` | 0.20 | Min isotopic pattern coverage |
| `DataThreshold` | `float` | 0.05 | Relative intensity threshold |
| `ReportMultipleMonoisos` | `bool` | true | Report multiple monoisotopic peaks |

**Obligation:** All nine parameters must remain with their documented names, types, and defaults.

---

## 6. Factory Pattern Contract

`CreateAlgorithm()` must map each `DeconvolutionType` to its algorithm:

```csharp
private static DeconvolutionAlgorithm CreateAlgorithm(DeconvolutionParameters parameters)
{
    return parameters.DeconvolutionType switch
    {
        DeconvolutionType.ClassicDeconvolution => new ClassicDeconvolutionAlgorithm(parameters),
        DeconvolutionType.IsoDecDeconvolution => new IsoDecAlgorithm(parameters),
        _ => throw new MzLibException("DeconvolutionType not yet supported")
    };
}
```

**Obligation:** Every new `DeconvolutionType` value must have a corresponding case here, or the `_` fallback will throw at runtime.

---

## 7. Extensibility Contract (New Algorithm Checklist)

The wiki explicitly defines the checklist for adding a new algorithm:

- [ ] Add `DeconvolutionType` enum value
- [ ] Create parameter class inheriting from `DeconvolutionParameters`
- [ ] Implement algorithm class inheriting from `DeconvolutionAlgorithm`
- [ ] Override `Deconvolute()` with algorithm logic
- [ ] Update `CreateAlgorithm()` factory in `Deconvoluter`
- [ ] Add unit tests for algorithm
- [ ] Document parameters and use cases
- [ ] **Update the wiki with algorithm details**

**Obligation:** All eight items must be completed for any new algorithm. The wiki update is explicitly listed as a requirement.

---

## 8. Parameter Classes Contract

Each parameter class must:

- Inherit from `DeconvolutionParameters`
- Set `DeconvolutionType` to the matching enum value
- Call `base(minCharge, maxCharge, polarity, averageResidueModel)` in the constructor
- Expose algorithm-specific parameters as public properties

---

## 9. Deconvolution Flow Contract

The documented steps are:

1. Input validation — check spectrum type and range
2. Algorithm selection — factory creates appropriate algorithm from parameters
3. Charge state scanning — test charge states within specified range
4. Pattern matching — compare observed peaks to theoretical isotopic distributions
5. Scoring — evaluate quality of each isotopic envelope match
6. Result generation — return scored isotopic envelopes

**Obligation:** `NeutralMassSpectrum` must be short-circuited before algorithm creation, returning peaks directly as single-peak envelopes.

---

## 10. Integration Contracts

### Dependencies (must remain)
```
MassSpectrometry
  ↓ MzLibUtil
  ↓ Chemistry (ChemicalFormula in AverageResidue)
```

### External Libraries (must remain available)
- **IsoDec**: `isodeclib.dll`, `isogenmass.dll`
- **MathNet.Numerics**: Statistical calculations
- **Intel MKL**: `libmmd.dll`, `svml_dispmd.dll`

### Downstream consumers (must not break)
- **Proteomics**: Precursor mass determination from MS1
- **FlashLFQ**: Quantification of isotopic envelopes
- **MetaMorpheus**: Precursor refinement
- **mzLib applications**: Any charge state determination use case

---

## 11. `IsotopicEnvelope` Filtering Contract

The wiki documents these as supported filtering patterns — they must continue to work:

```csharp
// By score
envelopes.Where(e => e.Score > 100)

// By charge
envelopes.Where(e => e.Charge == 2)

// By mass range
envelopes.Where(e => e.MonoisotopicMass >= 1000 && e.MonoisotopicMass <= 5000)

// By intensity
envelopes.OrderByDescending(e => e.TotalIntensity).First()
```

**Obligation:** `Score`, `Charge`, `MonoisotopicMass`, and `TotalIntensity` must remain public readable properties/fields.

---

## 12. Range Handling Contract

```csharp
// Must support explicit m/z range
var range = new MzRange(500, 1500);
Deconvoluter.Deconvolute(spectrum, deconParams, range);

// Must support isolation window
Deconvoluter.Deconvolute(scan, deconParams, scan.IsolationRange);
```

**Obligation:** The third `rangeToGetPeaksFrom` parameter must default to `null` (full spectrum) and accept any `MzRange`.

---

## Summary — What the Wiki Does NOT Yet Document

These are gaps in the current wiki relative to the work in progress. Each represents an obligation to update the wiki before or alongside the PR merge:

| Feature | Status |
|---|---|
| `FLASHDeconvolution` algorithm | Not documented |
| `RealFLASHDeconvolution` algorithm | Not documented |
| `DecoyAveragine` model | Not documented |
| `ExpectedIsotopeSpacing` on base class | Not documented |
| `ToDecoyParameters()` abstract method | Not documented |
| `DeconvoluteWithDecoys()` on Deconvoluter | Not documented |
| `DeconvolutionScorer` generic scorer | Not documented |
| Target-decoy FDR workflow | Not documented |

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
```diff
@@ -0,0 +1,476 @@
...
+        /// </summary>
+        internal override IEnumerable<IsotopicEnvelope> Deconvolute(MzSpectrum spectrum, MzRange range)
+        {
+            var p = DeconvolutionParameters as RealFLASHDeconvolutionParameters
+                ?? throw new MzLibException("Deconvolution params and algorithm do not match");
+
+            range ??= spectrum.Range;
+            if (spectrum.Size == 0) return Enumerable.Empty<IsotopicEnvelope>();
+
+            string exePath = ResolveExePath(p.FLASHDeconvExePath);
```
nbollis: Still resolving EXE path at decon call :(

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
```diff
@@ -0,0 +1,476 @@
...
+        internal static Dictionary<int, List<IsotopicEnvelope>> DeconvoluteFile(
+            string mzmlPath,
+            RealFLASHDeconvolutionParameters p,
+            FLASHDeconvRunner? runner)
+        {
+            if (!File.Exists(mzmlPath))
+                throw new FileNotFoundException($"mzML not found: {mzmlPath}", mzmlPath);
+
+            FLASHDeconvRunner effectiveRunner = runner ?? RunFLASHDeconvDefault;
+            string exePath = ResolveExePath(p.FLASHDeconvExePath);
```
nbollis: Same here :(

nbollis (changes requested):
A step in the right direction but missed the mark. The intended revision was to allow us to statically register an EXE path and then the parameters does not need to store the path and the algorithm does not need to resolve at each decon call

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
```diff
@@ -0,0 +1,415 @@
...
+            double bpMz = n > 0 ? spectrum.XArray[Array.IndexOf(spectrum.YArray, bpInt)] : 0;
+
+            string polTerm = polarity == Polarity.Negative
+                ? @"<cvParam cvRef=""MS"" accession=""MS:1000129"" name=""negative scan""/>"
+                : @"<cvParam cvRef=""MS"" accession=""MS:1000130"" name=""positive scan""/>";
+
+            string mzB64 = B64(spectrum.XArray);
+            string intB64 = B64(spectrum.YArray);
+
+            var sb = new StringBuilder();
```
nbollis, line 325: Use the mzml write instead of doing it custom from scratch

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
```diff
@@ -0,0 +1,415 @@
...
+            string stderr = stderrTask.Result;
+            if (proc.ExitCode != 0)
+                throw new MzLibException(
+                    $"FLASHDeconv exited with code {proc.ExitCode}.\n" +
+                    $"Args: {args}\nSTDOUT: {stdout}\nSTDERR: {stderr}");
+        }
+
+        // ── TSV parser ────────────────────────────────────────────────────────
+
+        internal static Dictionary<int, List<IsotopicEnvelope>> ParseSpecTsvByScan(
```
nbollis, line 252: Use readers instead of custom parsing if possible

mzLib/MassSpectrometry/Deconvolution/Algorithms/FlashDeconvExePathRegistry.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// exists, then caches the result so subsequent <see cref="Resolve(string?)"/>
+        /// calls with the same path return immediately.
+        /// </summary>
+        /// <exception cref="ArgumentException">path is null/whitespace.</exception>
+        /// <exception cref="FileNotFoundException">path does not exist on disk.</exception>
+        public static void Register(string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+                throw new ArgumentException("Path must be a non-empty string.", nameof(path));
+            if (!File.Exists(path))
```
nbollis, line 80: Validation here is weak. It only checks that the single path exists and could approve me registering a .psmtsv as a flashdecon exe

nbollis (changes requested):
Cool with registration. A few small details to look at

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
trishorts: ● Concrete blocker discovered. I started writing the swap, then hit this:

  - MzmlMethods lives in the Readers project.
  - RealFLASHDeconvolutionAlgorithm lives in the MassSpectrometry project.
  - Readers/Readers.csproj:20 already references MassSpectrometry.csproj.

  So MassSpectrometry cannot reference Readers — that's a circular project reference, which MSBuild rejects. To actually
   use MzmlMethods from here, one of:

  1. Move RealFLASHDeconvolutionAlgorithm (and friends) out of MassSpectrometry into Readers — but that breaks the
  Deconvoluter factory + DeconvolutionType enum that lives in MassSpectrometry and dispatches to it.
  2. Move MzmlMethods (1100+ lines) out of Readers into MassSpectrometry — touches every existing mzML write path in the
   codebase.
  3. Extract the mzML writer into a new shared lower-level project both can reference.

  All three are real restructures, all are well beyond this PR's scope ("Real OpenMS FLASHDeconv using exe from provided
   installation path").

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
trishorts: Context: ParseSpecTsvByScan (live file line 252, ~55 lines) parses FLASHDeconv's TSV output into IsotopicEnvelope
  records. "Readers" almost certainly refers to mzLib's Readers project — but the same project-graph blocker from
  fix_019 applies: Readers depends on MassSpectrometry, so MassSpectrometry can't reference Readers back.

  The alternative interpretation: @nbollis might mean "use a generic CSV/TSV library." CsvHelper is already in
  MassSpectrometry.csproj:14 as a package reference, so that's available without restructuring.

  The catch: the parser is more than "read a TSV." FLASHDeconv's output format has fingerprints mzLib's generic readers
  couldn't know about:
  - Trailing tab in header → empty columns (currently stripped)
  - Column-name aliases (MonoisotopicMass OR MonoMass; Qscore/Score; MassSNR/SNR)
  - Optional columns (RepresentativeMzEnd, Qscore, MassSNR) with degraded fallbacks
  - Score-priority fallback chain (Qscore → SNR → IsotopeCosine)
  - Polarity sign-flip on charge
  - Mass-range filtering
  - Construction of IsotopicEnvelope with computed repMz = (mzStart + mzEnd) / 2

  CsvHelper could absorb the row-split (~5 lines saved). Every other piece of domain logic stays. There's no generic
  FLASHDeconv reader to slot in.

  My recommendation: disagree with the same architectural rationale as fix_019, plus: "the parser isn't a generic TSV
  reader, it's FLASHDeconv-specific domain logic with format-quirk handling that no library reader would know about.
  CsvHelper would absorb the row-split (~5 lines) but the rest of the logic is irreducible." Alternative is disagree
  with a "happy to bring in CsvHelper for the row-split if you want, but the win is small" hedge.

mzLib/MassSpectrometry/Deconvolution/Algorithms/FlashDeconvExePathRegistry.cs
trishorts: Added a filename-shape guard: the basename must start with "FLASHDeconv"
  (case-insensitive). Catches the .psmtsv mistake at registration time
  with a clear ArgumentException naming the bad filename. Permissive
  enough to accept version-suffixed binaries (FLASHDeconv-3.5.exe) and
  the Linux extension-less name.

  Test updates:
  - New: RegisterWrongFilename_ThrowsArgumentException pins the new check.
  - CreateFakeExeFile helper renamed its output from realflash_fake_<guid>.exe
    to FLASHDeconv_fake_<guid>.exe so existing Register-based tests still
    pass the validation.
  - RegisterMissingPath_ThrowsFileNotFound: renamed its nonexistent path
    to start with "FLASHDeconv" so it still exercises the File.Exists
    branch, not the new filename-shape branch.

trishorts:
PR #1045 adds DeconvolutionType.RealFLASHDeconvolution to the mzLib enum (confirmed: the enum now contains
  ClassicDeconvolution, ExampleNewDeconvolutionTemplate, IsoDecDeconvolution, RealFLASHDeconvolution). MetaMorpheus's
  switch handles the first three but not the fourth → the default: fires → every test that constructs a
  DeconHostViewModel (directly, or transitively via DeconExplorationTabViewModel, or via a static [OneTimeSetUp] in
  ChimeraGroupViewModelTests) crashes. The 56 TypeInitializationException failures are a knock-on from a single
  static-init crash in ChimeraGroupViewModelTests.

  What you need to fix in mzLib

  Nothing. This isn't a mzLib bug.

  - mzLib's PR #1045 correctly added a new public enum value.
  - mzLib's own dispatch (Deconvoluter.CreateAlgorithm) is exhaustive.
  - MetaMorpheus's DeconHostViewModel switch is intentionally non-exhaustive with a fail-loud default — the comment on
  line 117 says so. The whole point is to force the MetaMorpheus maintainer to make a conscious decision when a new
  deconvolution type lands.

  What MetaMorpheus needs (your fix is here): add a case DeconvolutionType.RealFLASHDeconvolution: arm to
  DeconHostViewModel.cs:44-119. Two reasonable shapes:     1. All 114 failure stacks trace through DeconHostViewModel..ctor (PowerShell match across failures.json). It really is
   one root cause, not a shared cluster hiding a second one. After you add the case arm, those 114 tests are the ones
  that will turn green.
  2. MetaMorpheus has zero references to FlashDeconvExePathRegistry, RealFLASHDeconvolution, or
  FLASHDeconvExecutablePath anywhere in the tree. That means:
    - MM never calls Register() today, so the filename-shape validation we tightened earlier this session can't bite it.
    - No Deconvoluter.Deconvolute(...) call path in MM would dispatch to the new algorithm post-fix.
    - There's no second wave of failures waiting to surface once you patch the switch.

  So one switch arm in MetaMorpheus (case DeconvolutionType.RealFLASHDeconvolution: continue;) closes the entire
  integration gap. The mzLib side of PR #1045 has nothing waiting to be repaired.

  The only forward-looking note: when MM actually wants to expose RealFLASHDeconvolution to users, they'll need both (a)
   a real GUI case arm and (b) a startup hook to call
  FlashDeconvExePathRegistry.Resolve(GlobalSettings.FLASHDeconvExecutablePath) once and stash the result on params. But
  that's a feature-add, not a bug fix — not needed to green the tests.

mzLib/MassSpectrometry/Deconvolution/Algorithms/RealFLASHDeconvolutionAlgorithm.cs
trishorts: The mzLib Deconvoluter factory's contract is per-scan: every DeconvolutionAlgorithm subclass must implement
  Deconvolute(MzSpectrum, MzRange) — i.e., take an in-memory spectrum and return envelopes. ClassicDeconvolution and
  IsoDecDeconvolution both do this purely in memory.

  RealFLASHDeconvolutionAlgorithm wraps an external executable (FLASHDeconv.exe). That exe only reads mzML from disk —
  there's no in-memory API, no stdin streaming. So when a caller invokes Deconvoluter.Deconvolute(spectrum,
  RealFLASHDeconvolutionParameters, range), the algorithm has an in-memory MzSpectrum that the exe can't consume
  directly. WriteSingleScanMzml materializes that one spectrum to a temporary mzML file so we can shell out, then we
  delete the temp file in the finally.

  The class doc already flags the per-scan path as inferior:

  ▎ Per-scan: writes a temp single-scan mzML. No multi-scan feature tracing; results may be sparse.
  ▎ Whole-file: passes the full mzML directly. Enables multi-scan feature tracing; strongly preferred.

  DeconvoluteFile(mzmlPath, params) is a separate static entry point that bypasses the factory and hands FLASHDeconv a
  complete mzML the caller already has on disk — no writer needed.

  Is it absolutely necessary?

  Only if you want the per-scan factory path to work.

PR #1046, opened 2026-04-11
 Unify IRetentionTimePredictor: add batch method and implement on Koina RetentionTimeModel

trishorts:
# Unify IRetentionTimePredictor: add batch method and implement on Koina RetentionTimeModel

## Summary

- Adds a default `PredictRetentionTimes` batch method to `IRetentionTimePredictor` so callers can predict retention times for multiple peptides in a single call
- Makes `RetentionTimeModel` (the Koina abstract base class) implement `IRetentionTimePredictor`, bringing Prosit2019iRT and Prosit2020iRTTMT under the same interface as ChronologerRetentionTimePredictor
- Adds a project reference from PredictionClients to Chromatography to enable this cross-project interface implementation

## Motivation

MetaMorpheus currently has two separate `IRetentionTimePredictor` interfaces (one in mzLib, one in EngineLayer) and uses an adapter to bridge Koina models to the MetaMorpheus-side interface. This PR consolidates to a single interface in mzLib that both local predictors (Chronologer) and remote predictors (Koina/Prosit) satisfy directly. MetaMorpheus can then consume `IRetentionTimePredictor` without an adapter layer, which is required for wiring Chronologer into the PEP pipeline as the default RT predictor.

### `IRetentionTimePredictor.cs` (Chromatography)

Added a default interface method `PredictRetentionTimes(IEnumerable<IRetentionPredictable>)` that returns `Dictionary<string, double?>` keyed by `peptide.FullSequence`. The default implementation loops over `PredictRetentionTime` per peptide. Koina overrides this with its own batch implementation. Chronologer inherits the default.

### `RetentionTimeModel.cs` (PredictionClients/Koina)

`RetentionTimeModel` now declares `: IRetentionTimePredictor` and implements all interface members:

- **`PredictorName`** maps to the existing `ModelName` property
- **`SeparationType`** returns `SeparationType.HPLC` (all current Koina RT models are HPLC)
- **`PredictRetentionTime` (single)** delegates to the batch method
- **`PredictRetentionTimes` (batch)** overrides the interface default. It looks up results from the `Predictions` collection (populated by `RunInferenceAsync`), converting UNIMOD-format keys back to mzLib format for matching. If inference hasn't been run yet, it triggers `RunInferenceAsync` automatically.
- **`GetFormattedSequence`** returns `peptide.FullSequence`

### `PredictionClients.csproj`

Added `<ProjectReference>` to `Chromatography.csproj`. No circular dependency exists (Chromatography does not reference PredictionClients).

### `RetentionTimePredictorInterfaceTests.cs` (NEW — Test project)

10 NUnit 4 tests covering the interface contract:

- **Default batch method (5 tests):** Uses a mock predictor to verify the default `PredictRetentionTimes` implementation handles empty input, null results, mixed results, and keys by `FullSequence` (not `BaseSequence`)
- **Chronologer (2 tests):** Confirms `ChronologerRetentionTimePredictor` satisfies the interface and the default batch method produces real predictions via TorchSharp inference
- **Koina (3 tests):** Confirms `Prosit2019iRT` and `Prosit2020iRTTMT` satisfy the interface with correct `PredictorName` and `SeparationType` values; one live-API test validates batch override end-to-end

## What did NOT change

- **`ChronologerRetentionTimePredictor`** — already implements the interface; inherits the default batch method with no code changes
- **`Prosit2019iRT` / `Prosit2020iRTTMT`** — inherit all interface members from `RetentionTimeModel` with no code changes
- **No existing method signatures were modified** — this is purely additive

## Test plan

- [x] All new `RetentionTimePredictorInterfaceTests` pass (mock, Chronologer, and Koina interface tests)
- [x] Koina live-API batch test passes
- [x] All existing `ChronologerPredictorTests` pass (no regressions)
- [x] All existing Koina tests pass (`Prosit2019iRTTests`, `Prosit2020iRTTMTTests`)
- [x] Full `dotnet build` with zero errors and zero new warnings
- [x] Full test suite passes with no regressions


.github/workflows/dotnet.yml
nbollis, line 1: This will disable any test labeled integration from our github actions. Is this intended? 

If it is intended, we should rename it to MakesWebRequests or something like that. In our github actions, we use integration to mean sticking mzlib into metamorpheus.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
```diff
@@ -1,26 +1,70 @@
...
+    /// Duplicate <c>FullSequence</c> entries result in redundant prediction work in the
+    /// default implementation (the last result silently overwrites earlier ones in the
+    /// output dictionary).
     /// </summary>
-    /// <returns>Predicted retention time in predictor-specific units, or null if prediction not possible</returns>
-    double? PredictRetentionTime(IRetentionPredictable peptide, out RetentionTimeFailureReason? failureReason);
+    /// <exception cref="ArgumentNullException">
+    /// Thrown when <paramref name="peptides"/> is null.
+    /// </exception>
+    IReadOnlyDictionary<string, double?> PredictRetentionTimes(
```
nbollis: I am a bit worried about stability here. There are several different failure modes and we would not know what is happening if our peptides in the returned dictionary are null. 

One approach would be to return a nullable failure reason with the peptide.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
```diff
@@ -1,26 +1,70 @@
...
+        IEnumerable<IRetentionPredictable> peptides)
```
nbollis: Dictionary return is likely overkill. I think IEnumerable would work better here for upstream applications. 

However,this may have implications with Koina, so would be interested in Peter's thoughts.

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
```diff
@@ -261,6 +264,152 @@ protected virtual List<PeptideRTPrediction> ResponseToPredictions(
...
+        /// ensures consistent behavior between single and batch call sites.
+        /// </summary>
+        public double? PredictRetentionTime(IRetentionPredictable peptide,
+            out RetentionTimeFailureReason? failureReason)
+        {
+            failureReason = null;
+
+            if (peptide is null)
+            {
+                failureReason = RetentionTimeFailureReason.PredictionError;
```
nbollis: Bad failure reason. peptide was null, this is not a prediction error

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
```diff
@@ -261,6 +264,152 @@ protected virtual List<PeptideRTPrediction> ResponseToPredictions(
...
+
+                var predictions = Predict(inputs);
+
+                foreach (var prediction in predictions)
+                {
+                    results[prediction.FullSequence] = prediction.PredictedRetentionTime;
+                }
+
+                // Fill null for any inputs missing from predictions (defensive)
+                foreach (var peptide in peptideList)
```
pcruzparri: This and the previous foreach loop are inefficient, and null filling already built into the predictions. Once you get the predictions, you can generate your dictionary with just `var results = predictions.ToDictionary(pred => pred.FullSequence, pred.PredictedRetentionTime)`

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
pcruzparri: +1 to this note. There are two design choices for error handling. Maybe an IRetentionPrediction interface analogous to the IRetentionPredictable could merge these. 

Currently, chronologer addresses this by predicting a peptide at a time and returning an out parameter with the failure reason, which always maps to the input predictable. The expectation for Chronologer (correct me if I'm wrong, Nic) is that the caller will do failure catching as predictions are returned. 

Koina, instead, returns a prediction record, with the original sequence queried, the model validated sequence, the retention time, and any priority warning/error associated to that sequence (usually from validation). That way, those are still available after batch predictions.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
pcruzparri: I think Koina-wise it is technically fine, since this interface seems to wrap the current Koina implementation. It does seem redundant, but I mind it less on the interface definition here and more on the interface implementation on Koina, where the redundancy is more present. 

See comments on the RetentionTimeModel file.

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
```diff
@@ -261,6 +264,152 @@ protected virtual List<PeptideRTPrediction> ResponseToPredictions(
...
+        /// All Koina RT models target HPLC separation by default.
+        /// </summary>
+        public virtual SeparationType SeparationType => SeparationType.HPLC;
+
+        /// <summary>
+        /// Implements IRetentionTimePredictor.PredictRetentionTime by delegating to the
+        /// batch method with a list-of-one. This avoids duplicate validation logic and
+        /// ensures consistent behavior between single and batch call sites.
+        /// </summary>
+        public double? PredictRetentionTime(IRetentionPredictable peptide,
```
pcruzparri: I think the implementations of the PredictRetentionTime(s) methods rewrites the Koina logic. These two methods should simply call Koina's Predict() method rather than validating here _and_ having the model validate again internally. You could/should wrap the call in PredictRetentionTimes() to Koina's Predict() method in a try/catch for network error catching. Alternatively, you could update Koina's Predict method to have an out parameter for request prediction issues (currently not added).

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
```diff
@@ -261,6 +264,152 @@ protected virtual List<PeptideRTPrediction> ResponseToPredictions(
...
+
+            return new ReadOnlyDictionary<string, double?>(results);
+        }
+
+        /// <summary>
+        /// Returns the Koina-format (API-ready) sequence for a peptide using the same
+        /// internal sequence conversion logic as the batch prediction pipeline.
+        /// Returns null with a failure reason if conversion is not possible.
+        /// </summary>
+        public string? GetFormattedSequence(IRetentionPredictable peptide,
```
pcruzparri, line 289: This one as well should just wrap Koina's TryCleanSequence(), and report the failure reason from that method.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
trishorts: Quick clarification needed -- @pcruzparri +1'd the fix_020 stability concern and proposed a concrete shape: an IRetentionPrediction interface analogous to IRetentionPredictable, holding original sequence, model-validated sequence, RT, and warning/error -- effectively lifting Koina's PeptideRTPrediction record to interface form. This is a substantial interface-contract change that would ripple through every caller and likely doubles this PR's scope. Asking whether the preferred path is (a) land it in this PR, (b) follow-up PR with coordinated caller updates across mzLib + MetaMorpheus, or (c) keep the existing dictionary shape and address stability differently. Combining the response with fix_020.

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
trishorts: This was already addressed -- GetFormattedSequence (RetentionTimeModel.cs:403-434) already wraps TryCleanSequence and maps outcomes to RetentionTimeFailureReason: successful clean -> return apiSequence; cleaned-null -> IncompatibleModifications; exception -> PredictionError; empty input -> EmptySequence. The third out param of TryCleanSequence is a WarningException? (free-text), not enum-mappable; currently discarded as out _..

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
trishorts: Going to leave this as-is. Both loops have specific engineering rationale that a bare ToDictionary(...) would lose. Loop 1 normalizes NaN/Infinity to null so callers doing arithmetic on the dictionary values don't silently corrupt (added in this PR's review pass). Loop 2 is a defensive fill ensuring every input FullSequence is keyed in the output even if the predictor omits failed entries; we deliberately kept it after evaluating an engine finding that flagged the same defensive code. The current shape isn't an efficiency problem at the batch sizes Koina handles, and collapsing it would force re-adding both behaviors.

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
trishorts: This was already addressed -- Null peptide path now returns RetentionTimeFailureReason.EmptySequence (not PredictionError). See RetentionTimeModel.cs:295-305. Same EmptySequence reason is also returned for null/empty BaseSequence/FullSequence..

mzLib/PredictionClients/Koina/AbstractClasses/RetentionTimeModel.cs
trishorts: Going to leave this as-is. Premise does not match the live code. PredictRetentionTimes already calls Koina's Predict() (line 370) wrapped in try/catch (lines 364-393). The pre-call work (null/empty FullSequence filter, DistinctBy dedup) is input sanity to avoid redundant API calls and to honor the documented contract; it does not re-implement Koina's internal model/sequence validation. Post-call NaN/Infinity normalization is a contract requirement from a separate review fix (fix_008) so the documented 'null means prediction was not possible' invariant holds end-to-end. Modifying Koina's Predict() signature with an out parameter is shared by several Koina model types and is out-of-scope for this PR.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
trishorts: Quick clarification needed -- @nbollis suggested IEnumerable instead of dictionary, deferring to @pcruzparri on Koina implications. Dictionary keyed by FullSequence was chosen because Koina returns predictions in unordered batches; positional IEnumerable would force re-sorting. Asking @nbollis and @pcruzparri to confirm intended upstream usage pattern before reshaping the interface.

@pcruzparri  says the interface itself is fine and redirects to their comments on RetentionTimeModel.cs. Those implementation-side comments were addressed in fix_022 (already-addressed: null peptide now returns EmptySequence) and fix_023 (disagree: NaN normalization + defensive fill loops kept, with rationale). No code change required at this site..

.github/workflows/dotnet.yml
trishorts: Renaming the [Category("Integration")] label only on this PR's new tests would change CI routing for the 4+ pre-existing live-Koina test fixtures (Prosit2019iRT, Prosit2020iRTTMT, Prosit2020IntensityHCD, PFly2024FineTuned), pulling them into the fast build job and triggering real HTTP calls there. Renaming all of them is out-of-scope for this PR. Leaving the label as Integration for now; will revisit ASAP as a follow-up that touches all live-API fixtures together with a coordinated yml change.

nbollis: I understand the intention, it's just that integration is a bad keyword for this as we already use it to describe the CI/CD workflow of loading mzlib into Metamorpheus and running Metamorpheus tetst

nbollis (approved):
I don't like how the koina test category is called integration. But I won't hold this up

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
```diff
@@ -31,16 +35,66 @@
...
     /// </summary>
     SeparationType SeparationType { get; }
 
+    /// <summary>
+    /// Predicts retention time for a single peptide.
+    /// Returns null when prediction is not possible (invalid sequence, unsupported
+    /// modifications, model error, etc.). The <paramref name="failureReason"/> out
+    /// parameter is set when null is returned.
+    /// </summary>
+    double? PredictRetentionTime(IRetentionPredictable peptide,
```
nbollis: Why add this new method? In your previous PR, you changed the name to PredictRetentionTimeEquivalents.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
```diff
@@ -31,16 +35,66 @@
...
+    /// Duplicate <c>FullSequence</c> entries result in redundant prediction work in the
+    /// default implementation (the last result silently overwrites earlier ones in the
+    /// output dictionary).
+    ///
+    /// Null elements in <paramref name="peptides"/> are skipped.
+    /// </summary>
+    /// <exception cref="ArgumentNullException">
+    /// Thrown when <paramref name="peptides"/> is null.
+    /// </exception>
+    IReadOnlyDictionary<string, double?> PredictRetentionTimes(
```
nbollis: Same as other comment

nbollis (changes requested):

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
trishorts: Fair question — there's enough naming overlap here that I want to lay out exactly what each method does and why both pairs exist.

  The interface has **two pairs**, where each pair is a singular + a batch (plural-with-trailing-s) form. So four methods, but only two distinct concepts:

  | | Singular | Batch |
  |---|---|---|
  | **"Plain RT" pair** | `PredictRetentionTime` | `PredictRetentionTimes` |
  | **"Equivalent" pair** | `PredictRetentionTimeEquivalent` | `PredictRetentionTimeEquivalents` |

  The trailing `s` on the plurals is the batch form, not a different concept. So `PredictRetentionTime` ↔ `PredictRetentionTimes` is one pair; `PredictRetentionTimeEquivalent` ↔
  `PredictRetentionTimeEquivalents` is the other. That's the structural reason there are four method names rather than two.

  ### Why two *pairs* and not just one

  The two pairs differ in their **batch return shape**, not just their name:

  - **`PredictRetentionTimes`** returns `IReadOnlyDictionary<string, double?>` keyed by `FullSequence`. Simple shape: lookup-by-sequence, null value indicates "no prediction available."
  **No per-peptide failure reason** is preserved in the batch result. This is the **older, simpler shape** kept for backward compatibility with the existing concrete predictors
  (`ChronologerRetentionTimePredictor`, `SSRCalc3RetentionTimePredictor`, `CZERetentionTimePredictor`) — all of which already return dictionaries via this method.

  - **`PredictRetentionTimeEquivalents`** returns `IReadOnlyList<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)>`. Richer shape: each
  tuple carries the original peptide reference plus a per-peptide failure reason when the prediction couldn't be made. This addresses the diagnostic-loss concern (`fix_017` from this
  morning's review pass — same kind of "the dictionary form discards too much information for failed entries") by exposing the failure-reason channel that the singular method already had
  via `out`.

  So the **Equivalent pair isn't a duplicate** — it's the newer richer-diagnostic API, intentionally added alongside the older simpler one.

  ### Backward compat

  Renaming or removing `PredictRetentionTime`/`PredictRetentionTimes` would break the three concrete predictors that already implement it (`Chronologer`, `SSRCalc3`, `CZE`) and any
  downstream consumer that depends on the dictionary shape. That's the "previous PR's choice" you remember — that PR landed the `PredictRetentionTime[s]` shape. This PR adds the
  `Equivalent[s]` shape on top of it, rather than replacing it.

  ### When to use which

  - **`PredictRetentionTime[s]`** (plain): the simple "predict and give me a number per peptide" path. Use when you don't need per-peptide failure diagnostics; null in the dictionary
  value is enough.
  - **`PredictRetentionTimeEquivalent[s]`**: when you want to know *why* a particular peptide failed (mod incompatibility vs. model error vs. invalid sequence vs. ...) without falling
  back to per-peptide singular calls.

  Happy to add a doc note on the interface itself spelling this out if it would help future readers — let me know.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
trishorts: see above

pcruzparri (approved):

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
nbollis: No. I do not buy it. Why would we need two entrance points that are identical? If the caller does not need the per peptide failure information they can ignore it. This seems to be reverting the work of the previous PR. The intention from the previous PR seems to be clear, remove PreditcRetentionTime by making it obsolete.

.github/workflows/dotnet.yml
nbollis: Change the CI flag to be Koina instead of Integration.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
trishorts: Fair point — accepted. Pushed <5b58ee5e61e0e6cbf50127c8d4c3a90363dea3af>:

- PredictRetentionTime and PredictRetentionTimes are now [Obsolete] on IRetentionTimePredictor (and on the Koina RetentionTimeModel), with messages pointing callers at the PredictRetentionTimeEquivalent[s] pair.
  - In RetentionTimeModel, I inverted ownership so the Equivalent[s] pair holds the real Koina HTTP-batch logic (dedup, NaN normalization, Predictions clear, failure-reason fallback). PredictRetentionTime[s] are now thin obsolete shims that delegate to it. This matches the pattern already in place on the RetentionTimePredictor abstract  base, where PredictRetentionTime has been an [Obsolete] shim around PredictRetentionTimeEquivalent since the prior PR.
  - The interface's default PredictRetentionTimes was also redirected to call PredictRetentionTimeEquivalent so it doesn't self-trigger CS0618.
 - Existing tests still cover the legacy surface behind a file-level #pragma warning disable CS0618 to lock in  backward-compatible behavior until the methods are removed.

nbollis: Why add a new method to mark it obsolete. The claim that we need a new obsolete interface method to make an existing contract valid does not make sense to em. Am I missing something? 

Best approach would likely be to remove the obsoletes entirely, could be convinced otherwise.

.github/workflows/dotnet.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - name: Add coverlet collector (Test)
       run: cd mzLib && dotnet add Test/Test.csproj package coverlet.collector -v 6.0.2
     - name: Test
-      run: cd mzLib && dotnet test --no-build --verbosity normal --collect:"XPlat Code Coverage" /p:CoverletOutputFormat=cobertura ./Test/Test.csproj
```
nbollis, line 28: Change the CI flag to be Koina instead of Integration.

.github/workflows/dotnet.yml
```diff
@@ -25,14 +25,14 @@ jobs:
...
+      run: cd mzLib && dotnet test --no-build --verbosity normal --filter "Category!=Koina" --collect:"XPlat Code Coverage" /p:CoverletOutputFormat=cobertura ./Test/Test.csproj
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v4
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         verbose: true
         files: mzLib/Test*/TestResults/*/coverage.cobertura.xml
-  integration:
+  koina:
```
nbollis: Very close. You properly changed the name of the koina testing tag but we should keep this chunk of tests called integration as it is the one that clones metamorpheus

nbollis (approved):

zhuoxinshi (approved):

Alexander-Sol (approved):

PR #1048, opened 2026-04-14
Retention Time Prediction Infrastructure: Batch Support, Factory, and Generalized Naming

trishorts:
# Retention Time Prediction Infrastructure: Parallel Batch Support, Factory, and Generalized Naming

## Overview

This PR extends the `IRetentionTimePredictor` interface introduced by Nic. The original draft added four capabilities; code review surfaced a number of correctness, safety, and API-shape issues that have now been addressed. The final PR ships:

1. **Generalized naming** — `PredictRetentionTime` → `PredictRetentionTimeEquivalent`. The predicted value is predictor-specific: a time (CZE), an iRT (Chronologer), or a hydrophobicity score (SSRCalc3).
2. **Parallel batch prediction** — `PredictRetentionTimeEquivalents` returns a materialized list using range-partitioned parallel execution, with per-subclass opt-in for concurrency safety.
3. **Streaming batch prediction** — `StreamRetentionTimeEquivalents` exposes the same producer as a live, single-use feed.
4. **Factory** — predictors are created by `RetentionTimePredictorFactory.Create(PredictorType)`, a dedicated static class (no longer a method on the interface).
5. **Deterministic disposal** — concrete predictors with native resources (Chronologer's TorchSharp model) are reliably cleaned up via an `IDisposable` contract on the base class.

---

## Files Changed

| File | Change Type | Description |
|------|-------------|-------------|
| `IRetentionTimePredictor.cs` | Modified | Renamed single-item method; added batch and streaming signatures; **removed** the nested `PredictorType` enum and static `Create` (moved to dedicated factory) |
| `RetentionTimePredictorFactory.cs` | **New** | Static factory + top-level `PredictorType` enum. Keeps the interface free of transitive TorchSharp / native dependencies |
| `RetentionTimePredictor.cs` | Modified | Base class now implements `IDisposable`; added parallel `ProduceResults` producer with exception capture, single-use guard, disposable `BlockingCollection`, serial fast-path, validated `maxThreads`, ceiling-division partitioning, `IsConcurrentPredictionSafe` opt-in, and `CanonicalAminoAcids` as a `HashSet<char>` |
| `ChronologerRetentionTimePredictor.cs` | Modified | `Dispose()` changed from hiding (`new`) to `override`; explicit `maxThreads: 1` override for the batch method as a documented hook for future tensor-batching; default `IsConcurrentPredictionSafe => false` |
| `SSRCalc3RetentionTimePredictor.cs` | Modified | Opts in to `IsConcurrentPredictionSafe => true` (pure function over readonly state); `CanonicalAminoAcids` call site updated to `HashSet` |
| `CZERetentionTimePredictor.cs` | Modified | Opts in to `IsConcurrentPredictionSafe => true` |
| `ChronologerPredictorTests.cs` | Modified | Method-rename call-site updates; removed unused `using Readers;` and `using System.Threading;` |
| `RetentionTimeEquivalentBatchTests.cs` | **New** | Comprehensive batch/streaming test suite |
| `RetentionTimePredictionIntegrationTests.cs` | Modified | Long-peptide integration test tightened to the universal null-result/failure-reason contract |

---

### `IRetentionTimePredictor.cs`

- Renamed `PredictRetentionTime` → `PredictRetentionTimeEquivalent`.
- Added batch and streaming signatures (both accept `int maxThreads = 1`):
  ```csharp
  IReadOnlyList<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)>
      PredictRetentionTimeEquivalents(IEnumerable<IRetentionPredictable> peptides, int maxThreads = 1);

  IEnumerable<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)>
      StreamRetentionTimeEquivalents(IEnumerable<IRetentionPredictable> peptides, int maxThreads = 1);
  ```
- **Removed** the nested `PredictorType` enum and static `Create` factory. Having the factory reference every concrete predictor on the interface forced all consumers of `IRetentionTimePredictor` to transitively link TorchSharp and its native binaries. Moving it to a dedicated class restores open/closed and trims the dep footprint for callers that only need a non-Chronologer predictor.

### `RetentionTimePredictorFactory.cs` (new)

```csharp
public enum PredictorType { SSRCalc3, Chronologer, CZE }

public static class RetentionTimePredictorFactory
{
    public static IRetentionTimePredictor Create(PredictorType type) { ... }
}
```
- `PredictorType` is now a top-level enum.
- XML docs on each enum value call out the native-dep footprint (TorchSharp on Chronologer) so callers can make informed choices.

### `RetentionTimePredictor.cs`

- **`IDisposable` on the base class.** Subclasses override `Dispose`; the base provides a no-op. This replaces the prior pattern where `ChronologerRetentionTimePredictor` *hid* `Dispose` with `new`, which leaked the TorchSharp model when callers held a base-typed reference.
- **`ProduceResults` hardening:**
  - Public `maxThreads` is validated at the API boundary; `< 1` throws `ArgumentOutOfRangeException`.
  - The `BlockingCollection<T>` is wrapped in `using` so its internal `SemaphoreSlim` is released immediately rather than deferred to the finalizer queue.
  - Background `Task.Run` body is wrapped in `try/catch/finally`; exceptions are captured into a `ConcurrentQueue<Exception>` and rethrown as an `AggregateException` after the consumer drains. `CompleteAdding()` always runs in `finally`, eliminating the prior deadlock mode where a thrown producer left the consumer blocked forever.
  - A single-use guard (closure-boxed `started` flag) on `StreamRetentionTimeEquivalents` throws `InvalidOperationException` on re-enumeration; `peptides.ToList()` is hoisted to the caller's thread so non-rewindable sources are materialized exactly once.
  - Serial fast-path: when `effectiveThreads <= 1`, results are yielded directly without `BlockingCollection` or `Task.Run` — eliminating synchronization overhead for the default single-threaded case.
  - Ceiling division: `rangeSize = (count + effectiveThreads - 1) / effectiveThreads`, so 10 peptides × 7 threads yields ≤ 7 partitions (not 10).
- **`protected virtual bool IsConcurrentPredictionSafe => false`** — safe-by-default gate. Subclasses opt in explicitly. `ProduceResults` computes `effectiveThreads = IsConcurrentPredictionSafe ? maxThreads : 1`, so a subclass with mutable state cannot silently race even if a caller passes `maxThreads > 1`.
- **`CanonicalAminoAcids`** changed from `char[]` to `HashSet<char>`. The previous `Array.IndexOf` was O(n × 20) per peptide — roughly 600M comparisons on a 100k-peptide, 15-residue-average workload.

### `ChronologerRetentionTimePredictor.cs`

- `Dispose` now uses `override`, not `new` — deterministic native-model release regardless of the callsite's reference type.
- Explicit override of `PredictRetentionTimeEquivalents` that forces `maxThreads: 1`. Kept as a local declaration of intent and a clean hook point for future true tensor-batching; redundant with `IsConcurrentPredictionSafe = false` but useful as a code-level anchor.

### `SSRCalc3RetentionTimePredictor.cs` / `CZERetentionTimePredictor.cs`

- Both opt in with `protected override bool IsConcurrentPredictionSafe => true` after audit — each is a pure function over readonly state.

---

## Behaviour & Contracts

- **Prediction results unchanged.** All three predictors run through the same validation and scoring.
- **Result order** from `PredictRetentionTimeEquivalents` matches input order: the parallel producer preserves index-aligned placement. Callers can index by position.
- **`maxThreads` contract:** `< 1` throws `ArgumentOutOfRangeException` synchronously. `maxThreads > 1` is honored only when the subclass declares `IsConcurrentPredictionSafe = true`.
- **Exception propagation:** producer exceptions now surface as `AggregateException` to the caller after the consumer drains (previously: deadlock or silent loss).
- **Streaming single-use:** `StreamRetentionTimeEquivalents` enumerates exactly once; a second `GetEnumerator()` throws `InvalidOperationException`.
- **Failure reason contract:** `result == null ⇔ failureReason != null`. Which reason a predictor returns (too long, too short, unhandled AA, unhandled mod, …) is per-predictor; cross-predictor tests only verify the universal contract.

---

## Breaking Changes

| Change | Impact | Migration |
|--------|--------|-----------|
| `IRetentionTimePredictor.Create` removed; `PredictorType` promoted to top-level | Any caller using `IRetentionTimePredictor.Create(IRetentionTimePredictor.PredictorType.X)` | `RetentionTimePredictorFactory.Create(PredictorType.X)` |
| `PredictRetentionTime` → `PredictRetentionTimeEquivalent` | All single-item call sites | Rename |
| `CanonicalAminoAcids` type: `char[]` → `HashSet<char>` | External subclasses reading the field as `char[]` | Use `Contains` (and treat as a set) |
| Base class now declares `IDisposable` | Existing subclasses hiding `Dispose` with `new` | Change to `override` |

---

## Usage Examples

```csharp
// Factory creation
using var predictor = RetentionTimePredictorFactory.Create(PredictorType.SSRCalc3);

// Single prediction
var value = predictor.PredictRetentionTimeEquivalent(peptide, out var reason);

// Materialized parallel batch (SSRCalc3 / CZE: concurrent; Chronologer: serial under the hood)
var results = predictor.PredictRetentionTimeEquivalents(peptides, maxThreads: 8);

// Streaming (single-use; re-enumeration throws)
foreach (var (value, peptide, reason) in predictor.StreamRetentionTimeEquivalents(peptides, maxThreads: 8))
{
    // process as each completes
}
```

---

## Test Coverage

`RetentionTimeEquivalentBatchTests.cs`:

| Region | Tests |
|--------|-------|
| Basic batch functionality | Count matches input; all valid peptides succeed; all input peptides present |
| Single vs. batch consistency | All three predictors; deterministic across repeated calls; tests key by peptide reference (not fragile `BaseSequence`) |
| Multi-threaded batch | `maxThreads > 1` matches `maxThreads = 1`; 500 distinct 7-mer peptides with value assertions at partition boundaries (indices `0, 1, 49, 50, 99, 100, 249, 250, 499`) |
| Streaming API | Correctness, `maxThreads` honored, single-use enforcement (second `GetEnumerator` throws) |
| Mixed valid/invalid | Multiset comparison for duplicate sequences; empty/invalid sequences produce accurate reasons; Chronologer-specific `SequenceTooLong` is a standalone test |
| Edge cases | Empty input (no deadlock); single-item batch; `maxThreads < 1` throws `ArgumentOutOfRangeException`; `BlockingCollection` disposal verified indirectly via repeated-call stress |
| Factory & disposal | All `PredictorType` values functional; invalid enum throws; factory-returned predictors are `using`-safe |

`RetentionTimePredictionIntegrationTests.cs`:

- Long-peptide cross-predictor test now enforces only the universal `null ⇔ reason` contract, not a Chronologer-specific `SequenceTooLong`.


mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
```diff
@@ -56,6 +56,24 @@ protected RetentionTimePredictor(SequenceConversionHandlingMode sequenceHandling
     /// </summary>
     protected abstract double? PredictCore(IRetentionPredictable peptide, string? formattedSequence = null);
 
+    /// <summary>
+    /// Default batch implementation — loops over singles. Predictors that support true
+    /// batched inference (e.g. Chronologer) should override this method.
+    /// </summary>
+    // TODO: Chronologer should override this with a true batched tensor call
+    public virtual IReadOnlyList<RetentionTimeEquivalentResult> PredictRetentionTimeEquivalents(IEnumerable<IRetentionPredictable> peptides)
```
nbollis: IEnumerable return. No need to materialize to a list inside this method.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
```diff
@@ -28,7 +28,7 @@ protected RetentionTimePredictor(SequenceConversionHandlingMode sequenceHandling
         SequenceHandlingMode = sequenceHandlingMode;
     }
 
-    public double? PredictRetentionTime(IRetentionPredictable peptide, out RetentionTimeFailureReason? failureReason)
+    public double? PredictRetentionTimeEquivalent(IRetentionPredictable peptide, out RetentionTimeFailureReason? failureReason)
```
nbollis, line 37: Why rename?

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
```diff
@@ -16,11 +42,17 @@ public interface IRetentionTimePredictor
...
     /// <summary>
-    /// Predicts retention time for a given peptide.
+    /// Predicts a retention time equivalent for a given peptide.
+    /// The value may represent a time, iRT, or hydrophobicity depending on the predictor.
     /// Returns null if prediction cannot be made.
     /// </summary>
-    /// <returns>Predicted retention time in predictor-specific units, or null if prediction not possible</returns>
-    double? PredictRetentionTime(IRetentionPredictable peptide, out RetentionTimeFailureReason? failureReason);
+    /// <returns>Predicted value in predictor-specific units, or null if prediction not possible</returns>
+    double? PredictRetentionTimeEquivalent(IRetentionPredictable peptide, out RetentionTimeFailureReason? failureReason);
```
nbollis: Why rename this method?

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
```diff
@@ -89,3 +107,17 @@ protected virtual bool ValidateBasicConstraints(IRetentionPredictable peptide, o
         return true;
     }
 }
+/// <summary>
+/// The result of a retention time equivalent prediction for a single peptide.
+/// The predicted value may represent a time, iRT, or hydrophobicity depending on the predictor used.
+/// </summary>
+public readonly record struct RetentionTimeEquivalentResult(
```
nbollis: What is the purpose of this? Seems like we could just return the existing objects instead of creating something new.

mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
nbollis: Do we need to add override here? The parent interface has it as virtual. Will calling dispose on an interface casted chronologer predictor call the empty one in the interface or this one?

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: not necessary but as it says in the overview, sometimes we return hydrophobicity, sometimes retnetion time, sometimes iRT. So, i was attempting to make it generic.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
trishorts: see above. not weded to the idea. maybe better to leave it alone.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: only way i could think to get  original input, retention time and failure reason all in one.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
nbollis: Makes sense. We also return electrophoretic migration time in there. 

The issue is that I have a few downstream applications that already use this function from Mzlib and will break upon updating.

trishorts: we could leave the original and mark it as outdated? depracated? something like that. no need to eliminate it.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
nbollis: How about returning IEnumerable<(double, peptide, RetenionTimeFailureReason?)> from the batch method? Not weded to this, just wary of all that object construction in higher-throughput applications

trishorts: # C# Batch Analysis Return Type Comparison

## Overview

This document compares two approaches for returning results in a C#
batch analysis context:

1.  `IEnumerable<(double, string, enum?)>`
2.  `readonly record struct`

------------------------------------------------------------------------

## Option 1: IEnumerable\<(double, string, enum?)\>

### Semantics

Represents a **lazy sequence** of tuples. Results may be generated on
demand.

### Strengths

-   **Lazy Evaluation / Streaming**
    -   Supports `yield return`
    -   Enables early consumption of results
    -   Efficient for large datasets
-   **Low Allocation**
    -   Tuples are value types
    -   Avoids full materialization
-   **Composable**
    -   Works well with LINQ

### Weaknesses

-   **Threading Risks**
    -   Not thread-safe
    -   Deferred execution can lead to unpredictable behavior
    -   Execution occurs on the consumer thread
-   **Hidden Costs**
    -   May perform expensive operations during enumeration
-   **Poor Readability**
    -   Tuple fields lack semantic clarity
    -   Harder to maintain and extend

------------------------------------------------------------------------

## Option 2: readonly record struct

### Example

``` csharp
public readonly record struct Result(
    double Score,
    string Name,
    MyEnum? Type
);
```

### Semantics

Represents a **concrete, immutable data structure**.

### Strengths

-   **Clarity**
    -   Named fields improve readability
-   **Thread Safety**
    -   Immutable and safe for parallel use
-   **Versionability**
    -   Easier to extend and evolve
-   **Debugging**
    -   Better tooling and visibility

### Weaknesses

-   **More Boilerplate**
    -   Requires explicit type definition
-   **Copy Cost**
    -   Value-type copying (minimal for small structs)

------------------------------------------------------------------------

## Key Insight

These options address different concerns:

  Concern           Best Choice
  ----------------- ---------------
  Streaming         IEnumerable
  Thread Safety     record struct
  Readability       record struct
  Lazy Execution    IEnumerable
  Maintainability   record struct

------------------------------------------------------------------------

## Recommended Approach

Use both together:

``` csharp
public readonly record struct Result(
    double Score,
    string Name,
    MyEnum? Type
);

public IEnumerable<Result> Analyze(...)
```

### Why?

-   Maintains streaming capability
-   Provides strong typing
-   Ensures safer threading behavior
-   Improves API clarity

------------------------------------------------------------------------

## When to Use Tuples

Use tuples only when:

-   Scope is local
-   Usage is short-lived
-   Semantics are obvious
-   Not part of public API

------------------------------------------------------------------------

## Additional Consideration

For parallel batch processing, consider alternatives:

-   `List<Result>`
-   `IReadOnlyList<Result>`
-   `IAsyncEnumerable<Result>`

These may better align with push-based or asynchronous workflows.

------------------------------------------------------------------------

## Conclusion

-   Tuples are lightweight but fragile
-   Record structs are explicit and maintainable

**Recommendation:** Use `readonly record struct` for data
representation, optionally combined with `IEnumerable<T>` for streaming.

mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
trishorts: i changed it back

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: # IEnumerable`<T>`{=html} vs IReadOnlyList`<T>`{=html} in C

## Overview

This document compares two common return types in C#:

-   `IEnumerable<T>`
-   `IReadOnlyList<T>`

The choice between them defines how consumers interact with your data
and what guarantees your API provides.

------------------------------------------------------------------------

## Core Difference

### IEnumerable`<T>`{=html}

> You can **iterate** over the data.

### IReadOnlyList`<T>`{=html}

> You can **index, count, and repeatedly access** the data safely.

------------------------------------------------------------------------

## IEnumerable`<T>`{=html}

### Guarantees

-   Supports `foreach`
-   Minimal contract

### Does NOT Guarantee

-   Count
-   Indexing
-   Repeatability
-   Performance characteristics

------------------------------------------------------------------------

### Strengths

-   **Flexible**
    -   Works with lazy evaluation and LINQ
-   **Streaming**
    -   Can produce results incrementally
-   **Low overhead**
    -   Avoids full materialization

------------------------------------------------------------------------

### Weaknesses

-   **Multiple Enumeration Risk**

``` csharp
var count = results.Count();
var first = results.First();
```

May recompute results or cause performance issues.

-   **Hidden Cost**
    -   Could trigger expensive computation during iteration
-   **Unclear Shape**
    -   Caller does not know size, cost, or structure

------------------------------------------------------------------------

## IReadOnlyList`<T>`{=html}

### Guarantees

-   `Count` is available
-   Indexing (`[i]`)
-   Stable ordering
-   Safe repeated access

------------------------------------------------------------------------

### Strengths

-   **Clear Contract**
    -   Fully materialized and finite
-   **Performance Predictability**
    -   O(1) indexing and count
-   **Safer Consumption**
    -   No accidental recomputation

------------------------------------------------------------------------

### Weaknesses

-   **Requires Materialization**
    -   Must compute all results upfront
-   **Less Flexible**
    -   Not suitable for streaming pipelines

------------------------------------------------------------------------

## Decision Guide

  Question                             Use
  ------------------------------------ ---------------------------
  Need streaming / lazy execution?     IEnumerable`<T>`{=html}
  Need predictable, stable results?    IReadOnlyList`<T>`{=html}
  Need indexing or Count?              IReadOnlyList`<T>`{=html}
  Multi-threaded usage?                IReadOnlyList`<T>`{=html}
  Expensive or stateful computation?   IReadOnlyList`<T>`{=html}

------------------------------------------------------------------------

## Recommendation for Batch Analysis

For batch and potentially threaded analysis:

**Use `IReadOnlyList<T>`**

### Reasons

-   Safer in multi-threaded scenarios
-   Avoids repeated execution
-   Clear and predictable API
-   Matches "compute once, consume many" workflows

------------------------------------------------------------------------

## When to Use IEnumerable`<T>`{=html}

Use it when:

-   You need true streaming
-   Data is very large
-   You want lazy pipelines
-   You clearly document single-pass behavior

------------------------------------------------------------------------

## Best Practice Pattern

``` csharp
public IReadOnlyList<Result> Analyze(...)
```

-   Perform computation internally
-   Materialize results
-   Return as `IReadOnlyList<T>`

------------------------------------------------------------------------

## Key Insight

-   `IEnumerable<T>` → represents **computation**
-   `IReadOnlyList<T>` → represents **data**

------------------------------------------------------------------------

## Common Mistake

``` csharp
return results.ToList(); // typed as IEnumerable<T>
```

This hides useful guarantees from the caller.

------------------------------------------------------------------------

## Conclusion

-   `IEnumerable<T>` is powerful but can be ambiguous and error-prone
-   `IReadOnlyList<T>` is explicit, safe, and predictable

**Recommendation:** Prefer `IReadOnlyList<T>` for returning results,
especially in batch or threaded contexts.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: added an obsolete     public double? PredictRetentionTime(IRetentionPredictable peptide, out RetentionTimeFailureReason? failureReason)
        => PredictRetentionTimeEquivalent(peptide, out failureReason);

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: we are gonna use the tuple for now

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
```diff
@@ -51,11 +86,187 @@ protected RetentionTimePredictor(SequenceConversionHandlingMode sequenceHandling
...
+    /// <param name="maxThreads">Degree of parallelism; must be at least 1.</param>
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// Thrown when <paramref name="maxThreads"/> is less than 1.
+    /// </exception>
+    /// <exception cref="AggregateException">
+    /// Thrown after enumeration completes if the background producer encountered one or more
+    /// unhandled exceptions. Per-peptide prediction errors are reported via the
+    /// <see cref="RetentionTimeFailureReason"/> element of each tuple, not thrown.
+    /// </exception>
+    public virtual IEnumerable<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)> StreamRetentionTimeEquivalents(
```
trishorts: ## `StreamRetentionTimeEquivalents` — iterator semantics create a silent single-use contract

The current implementation uses `yield return` (an iterator method) to expose the parallel producer as an `IEnumerable<T>`. This is convenient, but it has a subtle contract that isn't visible at the call site.

### The problem

Every call to `GetEnumerator()` on the returned enumerable **re-runs the method body from scratch** — that's how C# iterator methods work. Each enumeration creates a fresh `BlockingCollection`, spawns a new `Task.Run`, and calls `peptides.ToList()` again. This produces three failure modes:

**1. Non-rewindable sources are consumed on the first pass and empty (or wrong) on the second**
```csharp
IEnumerable<IRetentionPredictable> oneShot = ReadPeptidesFromStream(file);
var stream = predictor.StreamRetentionTimeEquivalents(oneShot, maxThreads: 8);

foreach (var r in stream) { ... }   // consumes oneShot
foreach (var r in stream) { ... }   // oneShot is empty → 0 results, no error
```
Common triggers: LINQ chains over a `StreamReader`, DB cursors, any `IEnumerable` backed by a generator.

**2. Rewindable sources silently double CPU and allocations**
```csharp
var peptides = GetPeptides().ToList();
var stream = predictor.StreamRetentionTimeEquivalents(peptides);

var count = stream.Count();                    // enumeration 1: runs full pipeline
var results = stream.ToList();                 // enumeration 2: runs it again
```
Works correctly, but at 2× cost. `count + ToList` and other innocuous double-enumeration patterns (`.Any()` then `foreach`, Linq operators that probe count, etc.) all double the work.

**3. The caller has no way to know any of this is happening**
`IEnumerable<T>` is the universal "you can foreach this as many times as you want" contract. There's no signal in the type, the method name, or the doc comment that this particular enumerable is single-use. Reviewers and future maintainers will write double-enumeration code in good faith.

### Why `IAsyncEnumerable<T>` doesn't fix it

`async` iterator methods re-execute on every `GetAsyncEnumerator()` call for the same reason. The problem is iterator-method semantics, not sync-vs-async.

### Options worth discussing

| Option | Behavior | Tradeoffs |
|---|---|---|
| **A. Fail-fast single-use guard** | Hoist `peptides.ToList()` to the caller's thread; wrap the iterator in an outer method that throws `InvalidOperationException` on re-enumeration (closure-boxed `started` flag) | Surfaces the contract loudly; tiny behavior break for callers that currently iterate twice |
| **B. Return `IReadOnlyList<T>`** | Drop streaming; materialize eagerly like the batch method | Simplest; loses the "emit as they complete" benefit for very large batches |
| **C. Replay caching** | First enumeration streams + caches; subsequent enumerations replay | Caller-friendly; retains memory until GC'd; most code |
| **D. Minimal: hoist `ToList()` + document** | No runtime guard, just make the source-consumption behavior deterministic and document "single-use" in XML | Cheap; trusts the caller |
| **E. `IAsyncEnumerable<T>` + single-use guard** | Bigger API change, same guard required | Only worth it if we also want per-item backpressure; the guard is still needed |

My leaning is **A** — the hard failure is better than silent wrong results, and the fix is small. But I want input from anyone who has a streaming consumer in mind that would iterate twice by design.

mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
```diff
@@ -36,10 +36,21 @@ public ChronologerRetentionTimePredictor(
...
     }
 
+    /// <inheritdoc />
+    /// <remarks>
+    /// Chronologer's TorchSharp model is guarded by a single lock in <see cref="PredictCore"/>,
+    /// so multi-threaded dispatch would only serialize on that lock and add partition/aggregation
+    /// overhead. This override forces sequential execution until true tensor-batched inference
+    /// is implemented.
+    /// </remarks>
+    public override IReadOnlyList<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)> PredictRetentionTimeEquivalents(
```
nbollis: Why does this override exist? The effect is that chronologer will only ever run single threaded.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
```diff
@@ -23,12 +29,41 @@ public abstract class RetentionTimePredictor : IRetentionTimePredictor
...
+    /// reentrant (no shared mutable fields, no non-reentrant native handles, no locks that
+    /// would merely re-serialize threads defeating the parallel path).
+    /// </para>
+    /// <para>
+    /// This gate affects batch parallelism only. Single-call predictions via
+    /// <see cref="PredictRetentionTimeEquivalent"/> are unaffected: callers always control
+    /// thread use at that entry point.
+    /// </para>
+    /// </remarks>
+    protected virtual bool IsConcurrentPredictionSafe => false;
```
nbollis: What is the purpose of this if all implemented predictors are thread safe currently? Feels unnecessary.

trishorts: that may or may not all be threadsafe. I don't actually know because I did not actually look at each one. and I guess i don't know what will be implemented in the future. this just forces the predictor to state that it is threadsafe, otherwise we assume it is not. a safeguard that is probabaly unlikely now. i'm not sure that makes it bad.

mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
trishorts: This override forces sequential execution until true tensor-batched inference is implemented. I am simply not prepared to think about what chronologer does and doesn't do at this very moment. but it will absolutely be the next thing that I think about because I do want to default to chronologer in metamorpheus.

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
```diff
@@ -51,11 +86,166 @@ protected RetentionTimePredictor(SequenceConversionHandlingMode sequenceHandling
...
+    /// </exception>
+    /// <exception cref="AggregateException">
+    /// Thrown if the background producer encountered one or more unhandled exceptions while
+    /// enumerating or partitioning the input. Per-peptide prediction errors are reported via the
+    /// <see cref="RetentionTimeFailureReason"/> element of each tuple, not thrown.
+    /// </exception>
+    public virtual IReadOnlyList<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)> PredictRetentionTimeEquivalents(
+        IEnumerable<IRetentionPredictable> peptides, int maxThreads = 1)
+    {
+        if (maxThreads < 1)
```
Alexander-Sol, line 202: Usually if max threads is < 1, we either set it to one or set it to (MaxThreads - 1) or  1.

Alexander-Sol (dismissed):

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: changed to maxThreads = 1 if maxThreads < 1

mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
nbollis: I spend time making the same instance of chronologer callable from multiple threads and it handles the gating. Therefore you should be able to use the method in the parent class. 

This also means that all predictors are thread safe.

trishorts: this override is now deleted

Alexander-Sol (dismissed):

mzLib/Chromatography/RetentionTimePrediction/RetentionTimePredictor.cs
trishorts: i removed this

Alexander-Sol (approved):

pcruzparri (approved):

PR #1054, opened 2026-04-27
Generic Deconvolution Scoring for `IsotopicEnvelope`

trishorts:
## Summary

mzLib's deconvolution algorithms place values on different scales in `IsotopicEnvelope.Score`:

- **Classic** uses an empirical heuristic with no upper bound, designed for peptide data.
- **IsoDec** uses a cosine similarity in `[0, 1]`.

A downstream caller — most importantly MetaMorpheus — cannot apply a single threshold or ranking across methods because the scores live on different scales. This PR adds a method-agnostic quality score in `[0, 1]` that is recomputed from the envelope's peak list and an Averagine model. Because the score is derived from the envelope rather than produced by the algorithm, the same envelope produces the same score regardless of which algorithm yielded it.

The design is **purely additive**:

- Existing `IsotopicEnvelope.Score` semantics are unchanged. Every algorithm continues to populate it the same way it always has, so existing call sites that order, threshold, or aggregate on `Score` are unaffected.
- A new optional `GenericScore` property carries the method-agnostic value when callers want it.
- A small ergonomic layer makes the new score easy to consume from MetaMorpheus and any other downstream caller without modifying those projects.

---

### Two scores, side-by-side

```csharp
public class IsotopicEnvelope
{
    // Unchanged: algorithm-specific. Set at construction time. Different scales per algorithm.
    public double Score { get; private set; }

    // Nullable, [0, 1] when set, comparable across algorithms.
    public double? GenericScore { get; private set; }
    public void SetGenericScore(double genericScore);

    // Read-only conveniences:
    public bool HasGenericScore => GenericScore.HasValue;
    public double GenericOrFallbackScore => GenericScore ?? Score;
}
```

`GenericScore` is null until the scorer or one of the convenience entry points populates it. `HasGenericScore` reads more cleanly at call sites than `GenericScore.HasValue`. `GenericOrFallbackScore` exists to ease migration: a call site that currently sorts or thresholds on `Score` can switch to `GenericOrFallbackScore` without changing wiring, and start working correctly the moment the generic score is populated upstream.

### Why post-hoc rather than per-algorithm?

The scorer operates on the envelope's peak list and an `AverageResidue` model. It does not access the raw `MzSpectrum` and does not consult the algorithm that produced the envelope. This is what makes scores comparable across algorithms **by construction** — the same envelope produces the same score regardless of provenance.

The cost is that features which require the original spectrum (e.g. signal-to-noise) are not available. The scorer assumes deconvolute-and-discard.

---

## Feature vector

`EnvelopeScoreFeatures` carries four values:

| Field | What it measures | Computed from |
|---|---|---|
| `AveragineCosineSimilarity` | Cosine similarity between the observed isotope intensity vector and the theoretical Averagine distribution at the envelope's mass. Range `[0, 1]`. | `SpectralSimilarity.CosineOfAlignedVectors` over an aligned `(observed[k], theoretical[k])` vector pair, where observed peaks are matched to theoretical isotope positions within a 10 ppm window. |
| `AvgPpmError` | Mean absolute ppm error of observed peaks vs the nearest theoretical isotope position, anchored at the apex (most-intense) peak. | For each peak: `n = round((mz - apexMz) / isotopeStep)`, `theorMz = apexMz + n × isotopeStep`, error = `|mz - theorMz| / theorMz × 1e6`. |
| `PeakCompleteness` | Fraction of theoretical isotope positions above 1% of theoretical max intensity that have a matching observed peak. Range `[0, 1]`. | Of theoretical isotopes with `theor[k] > 0.01 × max(theor)`, count how many have `observed[k] > 0` after the 10 ppm alignment. |
| `IntensityRatioConsistency` | Uniformity of observed/theoretical intensity ratios across matched isotope positions, expressed as `1 / (1 + CV²)` where CV is the coefficient of variation. Range `[0, 1]`. | A real isotope envelope is the Averagine pattern scaled by a single abundance factor, so `observed[k] / theor[k]` should be approximately constant (low CV → near 1.0). Noise has erratic ratios (high CV → near 0.0). |

The fourth feature captures something cosine alone cannot: cosine is invariant to scale per-peak, so a "noisy but Averagine-shaped" vector and a real envelope can score similarly. Ratio consistency penalizes the former.

---

## Score formula

```
linear = -3.4494 × cosine
       +  1.6341 × ppmError
       + -4.7795 × completeness
       + -2.1883 × ratioConsistency
       + -4.3142

score  = 1 / (1 + exp(linear))
```

Weights are calibrated against IsoDec top-down yeast data and verified locally to satisfy the unit-test bounds:

- `(0.95, 1.5, 0.95, 0.95)` → score > 0.5 ✓
- `(0.20, 25.0, 0.20, 0.10)` → score < 0.5 ✓
- Range check across a 3×3×3×3 grid of inputs → all in `[0, 1]` ✓

The weights are flagged as **provisional** in a class-level doc comment with a recommendation to recalibrate against labelled training data.

---

### Scoring infrastructure

```csharp
namespace MassSpectrometry
{
    public readonly struct EnvelopeScoreFeatures { ... }   // four-feature data carrier

    public static class DeconvolutionScorer
    {
        public static EnvelopeScoreFeatures ComputeFeatures(IsotopicEnvelope, AverageResidue);
        public static double ComputeScore(EnvelopeScoreFeatures);
        public static double ScoreEnvelope(IsotopicEnvelope, AverageResidue);
        public static IEnumerable<(IsotopicEnvelope, double)> ScoreEnvelopes(
            IEnumerable<IsotopicEnvelope>, AverageResidue);
    }
}
```

`ScoreEnvelopes` is lazy. It performs eager argument-null validation (sequence, model) so the `ArgumentNullException` surfaces from the call site rather than from the first `MoveNext`. Null elements within the sequence are silently skipped — they are not an error, just nothing to score. `ScoreEnvelopes` does not mutate `envelope.Score`.

### `Deconvoluter` overloads

The existing `Deconvolute` overloads are unchanged. Two new overloads compute the generic score for every yielded envelope:

```csharp
// Existing — unchanged:
IEnumerable<IsotopicEnvelope> Deconvoluter.Deconvolute(MsDataScan, DeconvolutionParameters, MzRange?);
IEnumerable<IsotopicEnvelope> Deconvoluter.Deconvolute(MzSpectrum, DeconvolutionParameters, MzRange?);

// New — always-on generic scoring:
IEnumerable<IsotopicEnvelope> Deconvoluter.DeconvoluteWithGenericScoring(MsDataScan, DeconvolutionParameters, MzRange?);
IEnumerable<IsotopicEnvelope> Deconvoluter.DeconvoluteWithGenericScoring(MzSpectrum, DeconvolutionParameters, MzRange?);
```

Each yielded envelope has `GenericScore` set before yielding. The algorithm-specific `Score` is unchanged.

### `IsotopicEnvelopeExtensions` (ergonomic layer)

```csharp
namespace MassSpectrometry
{
    public static class IsotopicEnvelopeExtensions
    {
        public static double GetOrComputeGenericScore(this IsotopicEnvelope, AverageResidue);
        public static double GetOrComputeGenericScore(this IsotopicEnvelope, DeconvolutionParameters);
    }
}
```

`GetOrComputeGenericScore` is **idempotent**: if `GenericScore` is already set (by a prior `DeconvoluteWithGenericScoring` call, by manual `SetGenericScore`, or by a previous extension call), the cached value is returned without recomputation. Otherwise the score is computed via `DeconvolutionScorer.ScoreEnvelope`, cached on the envelope via `SetGenericScore`, and returned. The two overloads exist so callers that already hold a `DeconvolutionParameters` object don't have to reach into it for the model.

---

## Three access patterns

These three patterns let downstream callers pick the lowest-friction option for their context.

**Always-on** — every yielded envelope has `GenericScore` set:
```csharp
foreach (var env in Deconvoluter.DeconvoluteWithGenericScoring(spectrum, parameters))
{
    double s = env.Score;                  // algorithm-specific
    double g = env.GenericScore.Value;     // generic [0, 1], comparable across algorithms
}
```

**Opt-in, manual** — score individual envelopes after deconvolution:
```csharp
var env = Deconvoluter.Deconvolute(spectrum, parameters).First();
double g = DeconvolutionScorer.ScoreEnvelope(env, parameters.AverageResidueModel);
env.SetGenericScore(g);                                  // optional cache
```

**One-line ergonomic** — for downstream callers (e.g. MetaMorpheus) that already hold an envelope:
```csharp
foreach (IsotopicEnvelope env in scan.GetIsolatedMassesAndCharges(
    spectrum, commonParams.PrecursorDeconvolutionParameters))
{
    double genericScore = env.GetOrComputeGenericScore(
        commonParams.PrecursorDeconvolutionParameters);
    // existing env.Score logic, plus genericScore for PEP features, thresholding, etc.
}
```

---

### New files

| File | Purpose |
|---|---|
| `MassSpectrometry/Deconvolution/Scoring/EnvelopeScoreFeatures.cs` | `public readonly struct` carrying the four computed feature values. Pure data carrier — no scoring logic. |
| `MassSpectrometry/Deconvolution/Scoring/DeconvolutionScorer.cs` | `public static class` with `ComputeFeatures`, `ComputeScore`, `ScoreEnvelope`, `ScoreEnvelopes`. Reuses `SpectralSimilarity.CosineOfAlignedVectors` for cosine. |
| `MassSpectrometry/Deconvolution/Scoring/IsotopicEnvelopeExtensions.cs` | `public static class` with `GetOrComputeGenericScore` extension methods. Lives in `MassSpectrometry` namespace so a single `using MassSpectrometry;` brings it in. |

### Modified files

| File | Change |
|---|---|
| `MassSpectrometry/MzSpectra/IsotopicEnvelope.cs` | Added nullable `GenericScore` property, `SetGenericScore(double)` setter, `HasGenericScore` flag, and `GenericOrFallbackScore` migration convenience. Constructors and existing members untouched. |
| `MassSpectrometry/Deconvolution/Deconvoluter.cs` | Added `DeconvoluteWithGenericScoring` overloads (`MzSpectrum` and `MsDataScan`) for the always-on path. Existing `Deconvolute` overloads unchanged. |

### New tests

| File | Coverage |
|---|---|
| `Test/Deconvolution/TestDeconvolutionScorerUnit.cs` | Feature correctness, score bounds and ordering, `ScoreEnvelopes` API. |
| `Test/Deconvolution/TestDeconvolutionScorerNegativeCharge.cs` | `+z` / `-z` parity for cosine, ppm error, completeness, ratio consistency, and final score. |
| `Test/Deconvolution/TestDeconvolutionScorerIndependentValidation.cs` | Physical validation against angiotensin II and insulin B-chain envelopes built from formula-derived (not Averagine-derived) intensities. Guards against circular self-validation. |
| `Test/Deconvolution/TestScoreEnvelopesNullHandling.cs` | Null sequence and null model throw eagerly; null elements within the sequence are skipped; empty sequence returns empty. |
| `Test/Deconvolution/TestIsotopicEnvelopeExtensions.cs` | `GetOrComputeGenericScore` cache-miss / cache-hit / null-guard contracts; `HasGenericScore` round-trip; `GenericOrFallbackScore` fallback and override. |

---

## Test coverage

All tests pass. Approximately 35 tests across the five files.

**Group A — feature correctness on synthetic envelopes**
- A1 perfect envelope cosine ≥ 0.99
- A2 perfect envelope ppm error < 0.01
- A3 perfect envelope completeness == 1.0
- A4 half-peaks envelope completeness ∈ `[0.3, 0.7]`
- A5 50-ppm-shifted envelope cosine < 0.5 (peaks fall outside the matching window → observed vector is all zeros)
- A6 single zero-intensity peak does not throw, returns finite features

**Group B — score bounds and ordering**
- B1 high-quality features → score > 0.5
- B2 low-quality features → score < 0.5
- B3 score in `[0, 1]` across input grid (3×3×3×3)
- B4 better features give higher score

**Group C — `ScoreEnvelopes` API**
- C1 returns one pair per input envelope
- C2 does not mutate `envelope.Score`
- C3 empty input returns empty

**Negative-charge parity**
- Cosine, ppm error, completeness, ratio consistency, and final score all match between `+z` and `-z` for the same mass within 0.01
- All five negative charges parametrized (`-1, -2, -3, -5, -10`) produce finite features
- Synthetic peaks ascend in m/z for all negative charges (regression coverage: previously failed when isotope step was negative)

**Null-handling**
- Null sequence → `ArgumentNullException("envelopes")` thrown eagerly
- Null model → `ArgumentNullException("model")` thrown eagerly
- Null elements within the sequence are silently skipped
- Empty sequence returns empty

**Independent physical validation** (guard against circular self-validation)
- Angiotensin II at z=2 with formula-derived isotope intensities → cosine > 0.85
- Insulin B-chain at z=4 with formula-derived intensities → cosine > 0.85, completeness > 0.5
- Off-by-one shifted angiotensin envelope → cosine drops below 0.85
- Multiple mass ranges (800, 2000, 5000 Da) at various charges → all features finite

**`IsotopicEnvelopeExtensions`**
- Cache-miss path computes via `DeconvolutionScorer.ScoreEnvelope` and stashes via `SetGenericScore`; result is in `[0, 1]` and equals the cached value
- Cache-hit path returns the cached value verbatim, without recomputation (verified by pre-setting a value the scorer would not produce)
- Null envelope on either overload throws `ArgumentNullException("envelope")`
- Null model on the `(envelope, AverageResidue)` overload throws `ArgumentNullException("model")`
- Null parameters on the `(envelope, DeconvolutionParameters)` overload throws `ArgumentNullException("parameters")`
- `HasGenericScore` round-trips
- `GenericOrFallbackScore` returns `Score` when generic is unset, returns `GenericScore` when set, never mutates `Score`

---

## Migration

Existing call sites need no changes — `IsotopicEnvelope.Score` keeps its old meaning, every `Deconvolute` overload keeps its old behavior, and every algorithm continues to populate `Score` the way it always did.

A call site that currently writes:

```csharp
foreach (var env in Deconvoluter.Deconvolute(spectrum, parameters))
{
    if (env.Score > someThreshold) { ... }
}
```

can opt in to comparable scoring by switching to either:

```csharp
foreach (var env in Deconvoluter.DeconvoluteWithGenericScoring(spectrum, parameters))
{
    if (env.GenericScore.Value > someThreshold) { ... }
}
```

…or, for the per-envelope ergonomic path:

```csharp
foreach (var env in Deconvoluter.Deconvolute(spectrum, parameters))
{
    if (env.GetOrComputeGenericScore(parameters) > someThreshold) { ... }
}
```

`GenericOrFallbackScore` is provided for callers that want to flip a threshold expression to the generic score *if it has been computed* but degrade to the algorithm-specific value otherwise. It is a migration aid — new code should compute the generic score explicitly rather than rely on the fallback.

---

## Known limitations

**Provisional logistic weights.** The weights `(-3.4494, 1.6341, -4.7795, -2.1883, intercept -4.3142)` were calibrated against IsoDec top-down yeast targets vs. shifted-isotope decoys. They produce sensible scores in the test suite but have not been independently validated against labelled data in this repo. The class-level doc comment in `DeconvolutionScorer.cs` flags this and recommends recalibration when labelled data becomes available.

**No spectrum access.** The scorer operates only on the envelope's peak list, not on the raw `MzSpectrum`. SNR-type features that require the original spectrum are not available under the deconvolute-and-discard assumption.

**Classic decoys are not discriminating.** Relevant only for any future target-decoy work: Classic's `Peak2satisfiesRatio` filter pre-enforces Averagine shape on every recruited envelope, including those from a shifted-spectrum decoy pass, so post-hoc features cannot distinguish targets from decoys for Classic. The scorer still produces a valid continuous quality score for Classic output; it just cannot be calibrated via the shifted-isotope decoy strategy.

**FLASHDeconv is not in scope.** It is not in the current release of mzLib and is excluded from this PR.

**Target-decoy and q-value infrastructure is deferred.** A target-decoy framework using shifted-isotope decoys was prototyped in a longer design doc but is not included here. For IsoDec on top-down data, shifted-isotope decoys produce extreme noise rather than borderline cases, so q-values from this strategy don't add information beyond the continuous score. The continuous score is the recommended primary use of the scorer until a more realistic null distribution is available.

---

## Recommended follow-ups

These belong in subsequent PRs, not in this one.

1. **MetaMorpheus integration.** With the extension method in place, the downstream MetaMorpheus PR is small and obvious — typically one added line per call site. No mzLib changes required.
2. **Recalibrate logistic weights** on labelled data once available. The current weights are explicit about being provisional.
3. **Cross-algorithm consistency test.** Assert that the same envelope passed through `DeconvolutionScorer.ScoreEnvelope` produces the same score regardless of which algorithm produced it. This is true by construction (the scorer ignores the algorithm and only sees the peak list), but a test would lock the property in.
4. **Validate on real-data files** in the repo (`Averaged_221110_CytoOnly.mzML`, `Averaged_221110_HGHOnly.mzML`, `Averaged_221110_UbiqOnly.mzML`) — score real envelopes from each algorithm and confirm the score distributions look reasonable.
5. **Target-decoy and q-value infrastructure.** Worth revisiting once a more realistic null distribution strategy is identified (e.g. cross-algorithm consensus rather than shifted-isotope).

---

## Compile and test status

- `MassSpectrometry` project: builds clean.
- `Test` project: builds clean.
- All previously-existing unit tests: pass.
- All new unit tests across the five new test files: pass.
- No changes outside mzLib.


mzLib/MassSpectrometry/Deconvolution/Deconvoluter.cs
nbollis, line 1: Major red flag. The whole point of this decon structure was to have a single entrance point and all behavioral changes are handled by dependency injection. Consider putting a scorer in the parameters or as an optional parameter on the main call or a Boolean 'UseGenericScore' in the parameters and then deconvoluter sends the results from the algorithm to be scored in the deconvolute method

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
nbollis, line 1: Do we want two score fields? Several things use IsoEnvelope.score within MM. 

Is it better to have one score fields and the caller of deconvolute determines which score is calculated (default per algorithm or generic)?

nbollis (changes requested):
Looks pretty good except for the public entrance points. We can discuss if you'd like

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
```diff
@@ -67,6 +82,37 @@ public IsotopicEnvelope(int id, List<(double mz, double intensity)> peaks, doubl
             MostAbundantObservedIsotopicMass = peaks.MaxBy(p => p.intensity).mz * Math.Abs(chargestate);
         }
 
+        /// <summary>
+        /// Sets the optional generic deconvolution score for this envelope. The generic score is
+        /// produced by <see cref="DeconvolutionScorer"/> and is comparable across deconvolution
+        /// algorithms. Calling this method does not change <see cref="Score"/>.
+        /// </summary>
+        /// <param name="genericScore">Generic deconvolution score, expected in [0, 1].</param>
```
Alexander-Sol, line 90: Why not just make it a public Set?

Alexander-Sol (dismissed):

mzLib/MassSpectrometry/Deconvolution/Deconvoluter.cs
trishorts: ▎ Good call — restoring the single-entrance-point design. Took the bool UseGenericScore route from your three options:
  ▎
  ▎ - Added UseGenericScore (default false) to DeconvolutionParameters.
  ▎ - Folded the post-pass scoring into Deconvolute(...) itself, gated by the flag. The implementation is in a separate iterator helper (AddGenericScoring) so the NeutralMassSpectrum
  ▎ early-return stays eager rather than getting deferred to first MoveNext.
  ▎ - Removed both DeconvoluteWithGenericScoring overloads outright (no [Obsolete] shim) — the architectural intent is "all behavioral variation through the parameters object," and a
  ▎ transitional shim would just leave the violation in the codebase. Updated the three D-group integration tests + one doc cref to use the new path.
  ▎
  ▎ Caller migration: Deconvoluter.DeconvoluteWithGenericScoring(spec, p) → p.UseGenericScore = true; Deconvoluter.Deconvolute(spec, p). Both Score and GenericScore are still populated
  ▎ when the flag is on (algorithm sets Score, post-pass sets GenericScore).
  ▎
  ▎ Let me know if you'd prefer the proper-DI path (option c — IDeconvolutionScorer reference in parameters) instead — happy to layer that on as a follow-up if/when a second scorer enters
  ▎  the picture.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
trishorts: Good question. Going to keep the two fields — here's the rationale:

The two scores answer different questions, not the same question computed differently:
 - Score is algorithm-specific (Classic's intensity-weighted, IsoDec's own metric). Existing MM consumers including the ones you flagged — depend on that semantic contract.
 - GenericScore is the cross-algorithm comparable score this PR adds.

If I collapsed to a single Score whose content depends on UseGenericScore, any MM call site that reads envelope.Score would silently get a different score shape whenever UseGenericScore is set upstream — a footgun where the field's semantic meaning depends on caller context that the reader can't see.

The migration convenience for "generic if available, otherwise fall back to the algorithm score" is IsotopicEnvelope.GenericOrFallbackScore — added in this PR specifically for that pattern.

Open to revisiting if you have a specific MM call site in mind where the two-field design is causing pain.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
trishorts: Fair question — opted for the explicit method for two small reasons rather than any strong principle:

- The method's parameter is non-nullable double, so callers can't accidentally clear the score via envelope.GenericScore = null. The property type stays double? to express "not yet
computed".
- Routing all mutations through one method leaves a single place to add a range/invariant check later (e.g. [0, 1]) once the provisional weights stabilize and we want to enforce
bounds, without changing the API.

Neither is load-bearing today, so I'm not strongly attached. Happy to flip to { get; set; } if you'd prefer it — just let me know and it's a 2-line change.

nbollis (approved):

zhuoxinshi (approved):

PR #1056, opened 2026-04-27
Spectrum-Aware Generic Deconvolution Scoring

trishorts:
## Summary

Adds an opt-in spectrum-aware path to `DeconvolutionScorer` that produces a richer feature vector and a richer score when the caller still has the original `MzSpectrum`. The existing envelope-only path is preserved verbatim and remains the canonical "comparable across algorithms" score. `Deconvoluter.DeconvoluteWithGenericScoring(MzSpectrum, …)` is upgraded to use the new path automatically — its public signature does not change, and any caller already using it picks up the richer score for free.

Two scoring philosophies now coexist:

- **Envelope-only.** Cosine, ppm error, completeness, intensity-ratio consistency. Computable from the envelope's peak list and an `AverageResidue` model alone. Comparable across deconvolution algorithms by construction. Use when the spectrum is no longer available (the deconvolute-and-discard MetaMorpheus call site) or when scoring envelopes from disparate sources on the same scale.
- **Envelope + spectrum.** The four envelope-only features plus local SNR and a competing-peak ratio. Requires access to the original `MzSpectrum`. More discriminating, but not strictly cross-algorithm comparable because preprocessing differences (peak picking, smoothing, threshold choice) can leak into spectrum-derived features.

The two scores answer different questions and **should not be compared on the same threshold**. A 0.5 cutoff in one system does not correspond to 0.5 in the other; consumers should pick separate thresholds — or recalibrate the logistic — for each scoring path.

---

## Motivation

The envelope-only scorer that ships in mzLib today is a clean post-hoc scorer that computes four features purely from an `IsotopicEnvelope`'s peak list. That algorithm-blind property is valuable — it is what makes the score directly comparable across Classic, IsoDec, and (eventually) FLASHDeconv — but it costs the scorer some real discrimination power. Specifically, the envelope-only scorer cannot detect:

1. **Local SNR.** A small-but-clean envelope and a small-but-noisy envelope are indistinguishable in the envelope-only feature space.
2. **Competing peaks at theoretical positions.** An envelope extracted from the shoulder of a much larger feature looks fine in isolation.
3. *Spectral context for the apex* (how the envelope's apex compares against the broader spectrum's intensity distribution) — out of scope here, see "Deferred features" below.
4. *Unmatched intensity in the envelope's m/z footprint* (whether the envelope captures only a small fraction of intensity in its m/z range, typical of chimerics) — also out of scope here.

This PR closes (1) and (2). Items (3) and (4) are deferred — each requires its own validation and weight calibration, and bundling them together would dilute focus.

---

### `EnvelopeScoreFeatures` — additive struct expansion

Two new readonly fields and a `HasSpectrumFeatures` predicate:

```csharp
public readonly double LocalSignalToNoise;     // NaN under the envelope-only path
public readonly double CompetingPeakRatio;     // NaN under the envelope-only path
public bool HasSpectrumFeatures => !double.IsNaN(LocalSignalToNoise);
```

A new six-arg constructor accepts both spectrum-aware fields. The existing four-arg constructor's signature is **unchanged** (including its `intensityRatioConsistency = 0.0` default) and now chains to the six-arg constructor with `double.NaN` sentinels for the spectrum-aware fields. NaN as the "not measured" sentinel is intentional: it propagates safely through arithmetic, so any caller that consumes these fields without checking `HasSpectrumFeatures` gets NaN out of their score, surfacing the bug loudly rather than hiding it.

`ToString()` is extended to append `SNR` and `CompetingPeakRatio` when present, and is unchanged for envelope-only structs.

### `DeconvolutionScorer` — new overloads, new scoring entry point

| New public surface | Purpose |
|---|---|
| `ComputeFeatures(IsotopicEnvelope, AverageResidue, MzSpectrum)` | Spectrum-aware feature computation. Reuses the existing two-arg `ComputeFeatures` verbatim for the four envelope-only features, then computes `LocalSignalToNoise` and `CompetingPeakRatio` against the spectrum. |
| `ComputeScoreWithSpectrumContext(EnvelopeScoreFeatures)` | Logistic over all six features. Throws `ArgumentException` if called on a feature struct without spectrum features (i.e. `HasSpectrumFeatures == false`) — no silent NaN propagation. |
| `ScoreEnvelope(IsotopicEnvelope, AverageResidue, MzSpectrum)` | Convenience wrapper: `ComputeFeatures` then `ComputeScoreWithSpectrumContext`. Mirrors the shape of the existing two-arg `ScoreEnvelope`. |

The original `ComputeFeatures(env, model)`, `ComputeScore(features)`, `ScoreEnvelope(env, model)`, and `ScoreEnvelopes(envelopes, model)` are byte-for-byte unchanged in behavior. The diff against master on `DeconvolutionScorer.cs` is purely additive: zero existing lines removed or modified.

#### Local SNR — `ComputeLocalSnr`

For each envelope peak at m/z = `mz_p`:

1. Extract the spectrum window `[mz_p − SnrWindowDa, mz_p + SnrWindowDa]` via `MzSpectrum.ExtractIndices`.
2. From the peaks in that window, exclude any whose m/z is within `MatchTolerancePpm` (10 ppm) of any envelope peak. The remaining peaks form the noise sample.
3. `SNR_p = envelope peak intensity / median(noise sample intensities)`. If the noise sample is empty after exclusion, or its median is non-positive, `SNR_p` is treated as missing and is not contributed to the aggregate.

Aggregate: **median** of the `SNR_p` values across all envelope peaks (mean is not used — a few high-SNR peaks shouldn't mask a generally-noisy envelope). Returns `0.0` if no `SNR_p` could be computed.

#### Competing-peak ratio — `ComputeCompetingPeakRatio`

For each theoretical isotope position above the 1% intensity floor that has a matched observed peak (10 ppm tolerance, same alignment as the envelope-only `ComputeFeatures` path):

1. Extract the spectrum window `[matchedMz − CompetingPeakWindowDa, matchedMz + CompetingPeakWindowDa]`.
2. Find the apex of that window **excluding envelope peaks** — i.e. the tallest peak the algorithm did *not* claim as part of this envelope. Neighboring isotopes are not "competition"; only non-envelope peaks count.
3. The position is "non-competing" if `matchedIntensity ≥ apexInWindow × (1 − ε)`. Otherwise it counts as competing.

Returns `non-competing / matched`, in `[0, 1]`. Returns `0.0` if no positions were matched, consistent with the existing scorer's handling of degenerate inputs.

#### SNR squashing

```csharp
private static double SquashSnr(double snr)
{
    if (double.IsNaN(snr) || snr <= 0.0) return 0.0;
    const double SnrScale = 6.0;
    return 1.0 - Math.Exp(-snr / SnrScale);
}
```

Maps unbounded raw SNR into `[0, 1]` for the logistic. Scale chosen so SNR ≈ 5 (typical good envelope) maps to ~0.6 and SNR ≈ 20 maps to ~0.96.

#### Provisional weights

```csharp
private const double CoefficientSnr             =  2.0;
private const double CoefficientCompetingRatio  =  1.5;
private const double InterceptSpectrumAware     = -2.5;
```

Sign convention matches the existing envelope-only weights: positive coefficient ⇒ feature increases P(true). The intercept is shifted negative relative to the envelope-only `Intercept` because the SNR squashing function and the additional `CompetingPeakRatio` term raise the typical linear score; without a more negative intercept, every envelope would score above 0.5 even at low cosine and completeness.

These weights are **not fit against labelled data**. They are chosen to satisfy the unit-test bounds (high-quality features → score > 0.5; low-quality features → score < 0.5; all scores in [0, 1]) and to honour the physical meaning (higher SNR = better, more local-apex matches = better). The class-level doc comment flags both the existing envelope-only weights and the new spectrum-aware weights as provisional.

The first four logistic weights (`CoefficientCosine`, `CoefficientPpmError`, `CoefficientCompleteness`, `CoefficientRatioConsistency`) are reused verbatim from the envelope-only `ComputeScore`.

### `Deconvoluter` — promote the always-on path to spectrum-aware

`DeconvoluteWithGenericScoring(MzSpectrum, …)` now passes the spectrum into the scorer:

```diff
  if (envelope.GenericScore == null)
  {
-     double score = DeconvolutionScorer.ScoreEnvelope(envelope, model);
+     // We have the spectrum here — use the spectrum-aware score.
+     double score = DeconvolutionScorer.ScoreEnvelope(envelope, model, spectrum);
      envelope.SetGenericScore(score);
  }
```

The public signature of both overloads is unchanged. The `MsDataScan` overload delegates to the `MzSpectrum` overload via `scan.MassSpectrum`, so it gets the upgrade for free. The `<remarks>` block on the `MzSpectrum` overload is updated to mention SNR / competing-peak ratio and to clarify that the `IsotopicEnvelopeExtensions.GetOrComputeGenericScore` extension still uses the envelope-only feature set (because by the time MetaMorpheus calls it, the spectrum is gone).

### `IsotopicEnvelopeExtensions` — explicitly unchanged

`GetOrComputeGenericScore` keeps the envelope-only path. By design: at MetaMorpheus's call site the spectrum has already been discarded; switching the extension to the spectrum-aware path would force callers either to retain spectra they don't want to retain, or to silently fall back to the envelope-only path with a different set of weights than the rest of the codebase. The clean answer is: callers who want spectrum-aware scoring use `DeconvoluteWithGenericScoring` upfront, where the spectrum is still in hand.

---

## Out of scope (deferred)

- **Apex spectral percentile.** Whether the envelope's apex is a top-quartile peak in the spectrum or near the noise floor.
- **Unmatched-intensity ratio.** What fraction of intensity in the envelope's m/z footprint the envelope claimed (low values flag chimerics).
- **Recalibration of the existing logistic weights.** The new `ComputeScoreWithSpectrumContext` uses a new intercept and two new coefficients; the original `ComputeScore` weights are not touched.
- **MetaMorpheus integration.** No changes outside mzLib.
- **FLASHDeconv.** Not in the current release of mzLib.

A `// TODO` comment is placed near the new feature methods marking where the two deferred features would naturally slot in.

---

## How to interpret a spectrum-aware score relative to an envelope-only score

The two scores are **not directly comparable** and **should not share a threshold**.

The envelope-only score asks: *given just the peak list and an Averagine model, how Averagine-shaped is this envelope?* It is comparable across deconvolution algorithms by construction — the scorer never sees the spectrum or the algorithm.

The spectrum-aware score asks: *given the full spectral context, is this envelope cleanly above the local noise floor and not extracted from a shoulder?* It is more discriminating than the envelope-only score for a single algorithm's output, but reflects preprocessing differences across algorithms and is therefore not strictly cross-algorithm comparable.

A reasonable consumer pattern:

- **Cross-algorithm comparison or downstream filtering after deconvolute-and-discard:** envelope-only score, with whatever threshold has been calibrated against your data.
- **Quality ranking of envelopes within a single algorithm's output, with the spectrum still in hand:** spectrum-aware score, with a separate threshold (likely lower than the envelope-only equivalent for the same precision target — the spectrum-aware score is more pessimistic about chimerics and shoulders).

---

## Files changed

| File | Status | Lines (before → after) | Notes |
|---|---|---|---|
| `MassSpectrometry/Deconvolution/Scoring/EnvelopeScoreFeatures.cs` | modified | 84 → 136 | Two new readonly fields, `HasSpectrumFeatures`, six-arg constructor; existing four-arg constructor signature preserved (delegates to six-arg with NaN sentinels). `ToString` extended for spectrum-aware fields only when present. |
| `MassSpectrometry/Deconvolution/Scoring/DeconvolutionScorer.cs` | modified | 312 → 636 | Purely additive. Five new private constants (`SnrWindowDa`, `CompetingPeakWindowDa`, `CoefficientSnr`, `CoefficientCompetingRatio`, `InterceptSpectrumAware`), three new public methods (`ComputeFeatures` overload, `ComputeScoreWithSpectrumContext`, `ScoreEnvelope` overload), six new private helpers (`ComputeLocalSnr`, `ComputeCompetingPeakRatio`, `SquashSnr`, `SpectrumWindowIndices`, `IsEnvelopePeakMz`, `Median`). Zero existing lines removed or modified. |
| `MassSpectrometry/Deconvolution/Deconvoluter.cs` | modified | 137 → 140 | One functional line in the `MzSpectrum` overload of `DeconvoluteWithGenericScoring`. `<remarks>` block updated on that overload. The `MsDataScan` overload is unchanged — it delegates and picks up the upgrade for free. |
| `Test/Deconvolution/TestDeconvolutionScorerSpectrumAware.cs` | new | 0 → 446 | 15 tests. |

---

## Tests

15 new tests in `TestDeconvolutionScorerSpectrumAware`, organized by feature. They share the `DeconvolutionTestHelpers` static class (`Model`, `TestMass`, `TestCharge`, `BuildPerfectEnvelope`) with the existing scorer fixtures via `using static` — no parallel helpers. A single local helper, `BuildSpectrum`, constructs an `MzSpectrum` from a list of `(mz, intensity)` peaks.

| Test | Asserts |
|---|---|
| `ComputeFeatures_SpectrumOverload_DefaultsForNullSpectrumThrows` | Null envelope, model, and spectrum each throw `ArgumentNullException` with the right `ParamName`. |
| `ComputeFeatures_SpectrumOverload_PreservesEnvelopeOnlyFeatureValues` | The four envelope-only fields produced via the spectrum-aware overload equal those from the envelope-only `ComputeFeatures` (within 1e-12). The spectrum-aware overload does not perturb the envelope-only features. |
| `ComputeFeatures_SpectrumOverload_HasSpectrumFeaturesIsTrue` | After spectrum-aware call, `HasSpectrumFeatures == true` and both spectrum-aware fields are finite. |
| `ComputeFeatures_NoSpectrum_HasSpectrumFeaturesIsFalse` | After envelope-only call, `HasSpectrumFeatures == false` and both spectrum-aware fields are NaN. |
| `LocalSnr_CleanSpectrum_HighSnr` | Perfect envelope embedded in a spectrum where every non-envelope peak is well below the envelope's mean intensity → SNR > 50. |
| `LocalSnr_NoisySpectrum_LowSnr` | Same envelope, padded with non-envelope noise at intensity ≈ envelope mean intensity → SNR < 5. |
| `CompetingPeakRatio_NoCompetingPeaks_RatioIsOne` | Spectrum contains only the envelope's peaks → ratio == 1.0. (At charge=1 for clean window geometry.) |
| `CompetingPeakRatio_CompetingPeakAtEveryPosition_RatioIsZero` | Taller competing peak 0.1 Da from each envelope peak → ratio == 0.0. |
| `CompetingPeakRatio_CompetingAtHalfThePositions_RatioNearHalf` | Taller competing peak only at every other envelope peak → ratio in [0.4, 0.6]. |
| `ComputeScoreWithSpectrumContext_RequiresSpectrumFeatures` | Four-arg `EnvelopeScoreFeatures` (no spectrum features) → `ArgumentException`. No silent NaN. |
| `ComputeScoreWithSpectrumContext_HighQuality_AbovePointFive` | Six-arg features with cosine 0.95, ppm 1.5, comp 0.95, ratio 0.95, snr 20, competing 1.0 → score > 0.5. |
| `ComputeScoreWithSpectrumContext_LowQuality_BelowPointFive` | Six-arg features with cosine 0.2, ppm 25, comp 0.2, ratio 0.1, snr 0.5, competing 0.0 → score < 0.5. |
| `ComputeScoreWithSpectrumContext_BoundedZeroOne` | 3×3×3 grid across cosine/comp/ratio with snr in {0, 5, 50} and competing in {0, 0.5, 1.0}: every score in [0, 1]. |
| `ScoreEnvelope_WithSpectrum_ReturnsSpectrumAwareScore` | `ScoreEnvelope(env, model, spec) == ComputeScoreWithSpectrumContext(ComputeFeatures(env, model, spec))` within 1e-12. |
| `DeconvoluteWithGenericScoring_PopulatesScoreFromSpectrumAwarePath` | Two-part wire-through proof: (a) every yielded envelope's `GenericScore` matches a direct spectrum-aware `ScoreEnvelope(env, model, cleanSpec)` call within 1e-12; (b) the spectrum-aware scorer produces a different score against a synthetic noisy spectrum than against the clean spectrum. Together these prove `DeconvoluteWithGenericScoring` is on the spectrum-aware path. |

### Why charge=1 for the competing-peak-ratio tests

`CompetingPeakWindowDa = 0.5` Da is wider than the isotope spacing at charge ≥ 3 (0.334 Da and below). At higher charges, an envelope's neighboring isotopes fall inside each other's competing-peak windows, which would confound assertions about "non-competing" positions if envelope peaks were not excluded from the apex search. With the envelope-exclusion logic in place, the tests still pass at any charge, but the geometry is cleanest at charge = 1 (envelope peaks ~1 Da apart, each window contains only its anchor envelope peak). The three competing-peak-ratio tests use `charge: 1` for that reason; the SNR and integration tests use the default `BuildPerfectEnvelope()` charge.

### Why the integration test does not deconvolute a noisy spectrum

The original design of the "stronger" wire-through proof was to deconvolute the same envelope from two spectra differing only in noise floor and assert that the resulting `GenericScore` values differ. This runs into harmonics: noise peaks placed at offsets like ±0.7 and ±1.3 Da from envelope peaks land near integer multiples of the z=10 isotope step (0.7 / 0.1003 ≈ 7, 1.3 / 0.1003 ≈ 13), which Classic can mistake for legitimate isotope positions of a charge-10 envelope. The yielded envelope sets differ between clean and noisy spectra, confounding direct comparison.

The committed test instead deconvolutes only the clean spectrum, then asserts:

1. **Spectrum is wired through:** the yielded envelope's `GenericScore` matches `DeconvolutionScorer.ScoreEnvelope(env, model, cleanSpec)` within 1e-12. (If the deconvoluter had silently used the envelope-only path, this equality would not hold.)
2. **Scorer is sensitive to spectral context:** `ScoreEnvelope(env, model, cleanSpec)` differs from `ScoreEnvelope(env, model, noisySpec)`, where `noisySpec` is constructed synthetically and never fed to Classic.

Together these establish that `DeconvoluteWithGenericScoring` is calling the spectrum-aware scorer with the actual spectrum it was given — without putting noise through Classic.

---

## Verification checklist

- [x] `EnvelopeScoreFeatures.cs` has six fields and two constructors. The four-arg constructor still compiles, preserves its `intensityRatioConsistency = 0.0` default, and produces a struct where `HasSpectrumFeatures == false`.
- [x] `DeconvolutionScorer.cs` exposes the new public methods: `ComputeFeatures(IsotopicEnvelope, AverageResidue, MzSpectrum)`, `ComputeScoreWithSpectrumContext(EnvelopeScoreFeatures)`, `ScoreEnvelope(IsotopicEnvelope, AverageResidue, MzSpectrum)`.
- [x] The original `ComputeFeatures(envelope, model)`, `ComputeScore(features)`, `ScoreEnvelope(envelope, model)`, and `ScoreEnvelopes(envelopes, model)` methods are byte-for-byte unchanged in behavior. The diff against master is purely additive on `DeconvolutionScorer.cs`. Existing tests in `TestDeconvolutionScorerUnit.cs`, `TestDeconvolutionScorerNegativeCharge.cs`, `TestDeconvolutionScorerIndependentValidation.cs`, and `TestScoreEnvelopesNullHandling.cs` pass without modification.
- [x] All 15 new tests in `TestDeconvolutionScorerSpectrumAware.cs` pass.
- [x] `Deconvoluter.DeconvoluteWithGenericScoring(MzSpectrum, ...)` uses the spectrum-aware path. Public signature unchanged. The `MsDataScan` overload delegates and picks up the upgrade for free.
- [x] `IsotopicEnvelopeExtensions.GetOrComputeGenericScore` is unmodified and continues to use the envelope-only path.
- [x] No file in `MassSpectrometry/Deconvolution/Algorithms/` was changed.
- [x] No file outside mzLib was changed.

---

## Self-checks

- [x] **Reuse, not duplicate.** The new `ComputeFeatures(env, model, spectrum)` calls `ComputeFeatures(env, model)` to obtain the four envelope-only features and forwards them verbatim. The cosine assignment occurs at exactly one site in `DeconvolutionScorer.cs` (inside the original envelope-only `ComputeFeatures`). The two paths cannot drift.
- [x] **Window arithmetic.** `MzSpectrum.ExtractIndices(min, max)` is inclusive on both ends; the empty-range sentinel `(1, 0)` is handled explicitly in both `ComputeLocalSnr` and `ComputeCompetingPeakRatio`.
- [x] **Empty-window handling in SNR.** When the SNR window contains only envelope peaks (noise sample empty after exclusion), the peak is skipped (not contributed to the aggregate). If every envelope peak's window is empty, `ComputeLocalSnr` returns `0.0` rather than `NaN` or `Infinity`.
- [x] **Idempotent deconvolution path.** `DeconvoluteWithGenericScoring` retains the `if (envelope.GenericScore == null)` guard. Envelopes that arrive already-scored are not re-scored.
- [x] **Tolerance vs. window units.** `MatchTolerancePpm = 10.0` is ppm and is converted to absolute Da at every call site. `SnrWindowDa = 2.0` and `CompetingPeakWindowDa = 0.5` are absolute Da and are passed directly to `ExtractIndices`. No mixing.
- [x] **NaN propagation.** `ComputeScoreWithSpectrumContext` throws `ArgumentException` (with `ParamName = "features"`) when called on a struct with `HasSpectrumFeatures == false`. The four-arg `EnvelopeScoreFeatures` constructor sets the spectrum-aware fields to `NaN`, not `0.0`, so the predicate works regardless of what the caller passed.
- [x] **Provisional-weight comment.** The class-level XML doc on `DeconvolutionScorer` and a dedicated `// ── Spectrum-aware logistic weights (PROVISIONAL) ──` comment block flag the new weights as provisional and recommend recalibration against labelled data, in the same tone as the existing comment for the envelope-only weights.

---

## Notable design choices and deviations from the planning prompt

These are noted explicitly so reviewers can flag any they disagree with.

**Logistic sign convention.** The existing `ComputeScore` uses `1 / (1 + exp(-linear))` with positive coefficients for "good" features and `Intercept = +4.3142`. The original planning prompt's `ComputeScoreWithSpectrumContext` snippet flipped this convention (`1 / (1 + exp(+linear))` with negative weights). The committed implementation preserves the existing class convention so the spectrum-aware path is internally consistent with the envelope-only path; future calibration work touches one set of constants, not two with opposite signs. The numerical scores are unchanged. `CoefficientSnr` and `CoefficientCompetingRatio` are positive in the committed version where the prompt's snippet had them negative; `InterceptSpectrumAware = -2.5` matches the prompt.

**`EnvelopeScoreFeatures` four-arg constructor signature.** The existing four-arg constructor takes `intensityRatioConsistency` with a default value of `0.0` (predates the ratio-consistency feature). The committed version preserves that signature, including the default, so that any caller that pre-dates this PR continues to compile. The new spectrum-aware fields default to `NaN` via the four-arg → six-arg delegation chain. Net: zero source-level breakage for existing callers.

**Competing-peak apex search excludes envelope peaks.** The prompt described the feature as catching "a taller peak that the algorithm chose not to claim", but the algorithmic detail of excluding envelope peaks from the apex search was not explicit. Excluding them is necessary at any charge ≥ 3 to avoid neighboring isotopes being flagged as their own competition, and is symmetric with how SNR excludes envelope peaks from the noise sample.

**Integration test does not deconvolute a noisy spectrum.** See "Why the integration test does not deconvolute a noisy spectrum" above. The spirit of the prompt's "stronger test" was preserved (prove the spectrum is being wired into the scorer), but the mechanism is rotated to avoid harmonic confounders inside Classic.

---

## Recommended next steps

1. **Recalibrate the spectrum-aware logistic weights** against labelled top-down data (the same training corpus used for the envelope-only weights, plus held-out spectra so SNR and competing-peak distributions can be estimated against ground truth).
2. **Add the deferred apex-percentile and unmatched-intensity-ratio features** in a follow-up PR. Each adds one feature to `EnvelopeScoreFeatures`, one helper to `DeconvolutionScorer`, and one weight to `ComputeScoreWithSpectrumContext`. The class-level TODO marks where they slot in.
3. **Cross-algorithm spectrum-aware sensitivity study.** The spectrum-aware score is not strictly comparable across algorithms in theory; quantifying how much it actually drifts (e.g. score distributions for the same spectrum under Classic vs IsoDec) would tell consumers whether they need separate thresholds per algorithm or can share a single one.
4. **MetaMorpheus integration.** Out of scope here, but with this PR landed, MetaMorpheus task code can switch from `Deconvoluter.Deconvolute` + `IsotopicEnvelopeExtensions.GetOrComputeGenericScore` (envelope-only) to `Deconvoluter.DeconvoluteWithGenericScoring` (spectrum-aware) at the precursor-deconvolution call sites where the spectrum is still in hand. No mzLib changes required to make that switch.


mzLib/MassSpectrometry/Deconvolution/Scoring/DeconvolutionScorer.cs
```diff
@@ -316,5 +443,194 @@ private static double ComputeAvgPpmError(
...
+        ///   1. Extract the window [mz_p − <see cref="SnrWindowDa"/>, mz_p + <see cref="SnrWindowDa"/>] from the spectrum.
+        ///   2. From the spectrum peaks in that window, exclude any peak whose m/z is within
+        ///      <see cref="MatchTolerancePpm"/> of any envelope peak (i.e. the envelope's own peaks).
+        ///   3. SNR_p = envelope peak intensity / median(remaining window intensities).
+        ///      If the remaining window is empty or its median is non-positive, SNR_p is treated
+        ///      as missing and is not contributed to the aggregate median.
+        /// Returns the median of the SNR_p values across all envelope peaks, or 0.0 if no SNR_p
+        /// could be computed (degenerate spectrum / envelope).
+        /// </summary>
+        private static double ComputeLocalSnr(IsotopicEnvelope envelope, MzSpectrum spectrum)
```
nbollis, line 459: This feels weird. Signal to noise here is defined as the median snr for each individual isotopic peak? Is this a normal way to do the calculation?

nbollis (approved):

mzLib/MassSpectrometry/Deconvolution/Scoring/DeconvolutionScorer.cs
pcruzparri: I think the "noise" here should be computed from either the positive values only or from the magnitude of all noise peaks. The latter is more robust, since it keeps the noise level and prevents potentially removing too many values for the median calculation. The negative noise seems like an issue here since it increases chances of not getting a median floor >=0 for the snr proxy when it can actually be obtained. Thoughts?

pcruzparri (approved):
see reply to Nic's comment. pointed out a stability concern, but won't hold this.

PR #1059, opened 2026-05-07
Spectral Library and Similarity Quality of Life 

nbollis:
No true functional change, all quality of life. 

## `LibrarySpectrum` Enhancements
1. IEquatable to Library Spectrum
2. Can now run spectral similarity directly from `LibrarySpectrum`

## `SpectralSimilarity` Enhancements
1. Added a ton of overloads for LibrarySpectrum inside of Spectral Similarity
2. Added an enum for the different types of similarity metrics
3. Added methods to call for different and/or many similarity metrics
4. Added overloads for library spectra to call those similarity metric methods.


mzLib/Omics/SpectralLibrary/LibraryExtensions.cs
```diff
@@ -0,0 +1,681 @@
...
+namespace Omics.SpectralLibrary
+{
+    public static class LibraryExtensions
+    {
+        /// <summary>
+        /// Computes spectral similarity between two library spectra.
+        /// </summary>
+        public static SpectralSimilarity ComputeSpectralSimilarity(
+            this LibrarySpectrum experimental,
+            LibrarySpectrum theoretical,
```
trishorts, line 15: **Minor** - New similarity extensions hardcode the 300 m/z cutoff with no opt-out

**Problem:**
Every new `ComputeSpectralSimilarity` overload calls the `SpectralSimilarity` constructor without passing `filterOutBelowThisMz`, so it defaults to 300, and the extension signatures expose `scheme`, `toleranceInPpm`, and `allPeaks` but not `filterOutBelowThisMz`. For `LibrarySpectrum` inputs (peptide fragment ions), low-mass ions such as b1/y1/y2 commonly fall below 300 m/z and are silently discarded with no way to override via the new API.

**Impact:**
Callers of the new "run similarity directly from LibrarySpectrum" API get silently truncated peak lists and potentially skewed similarity scores, with no parameter to disable the filter.

**Suggested fix:**
Add an optional `filterOutBelowThisMz` parameter to the extension methods (consistent with the underlying constructors), or document the fixed cutoff explicitly.

**Affected test scenarios:**
- LibrarySpectrum_ComputeSpectralSimilarity_PeaksBelowCutoff - library spectra with fragment ions below 300 m/z - peaks retained when cutoff overridden
- LibrarySpectrum_ComputeSpectralSimilarity_DefaultCutoff - documents that sub-300 m/z peaks are dropped by default

mzLib/Test/FileReadingTests/SpectralLibraryTests/TestLibrarySpectrum.cs
```diff
@@ -64,5 +69,1011 @@ public static void CrosslinkPsmFromTsvToLibrarySpectrumTest()
...
+
+        [Test]
+        public static void LibrarySpectrum_ComputeSpectralSimilarity_MzSpectrumToLibrary()
+        {
+            Product a = new Product(ProductType.b, FragmentationTerminus.N, 1, 1, 1, 0);
+            Product b = new Product(ProductType.b, FragmentationTerminus.N, 2, 2, 1, 0);
+            var peaks = new List<MatchedFragmentIon>
+            {
+                new MatchedFragmentIon(a, 400.0, 1000.0, 1),
+                new MatchedFragmentIon(b, 500.0, 2000.0, 1)
```
trishorts, line 311: **Major** - Directional overloads only tested with symmetric data; swapped args undetectable

**Problem:**
The ~30 new overload tests (`*ToLibrary`, `*ToMzSpectrum`, `*ToMsDataScan`, `*ToArrays`) all build the experimental and theoretical operands from identical peak data (e.g. `LibrarySpectrum_ComputeSpectralSimilarity_LibraryToMzSpectrum`, `MsDataScanToLibrary`, `ArraysToMsDataScan`). Because `CosineSimilarity` is symmetric and the inputs are identical, a wiring bug that swapped the experimental and theoretical arguments inside any extension method would still produce `1.0` and pass. The experimental/theoretical roles are not actually symmetric in `GetIntensityPairs` (only unpaired *experimental* peaks are appended when `allPeaks` is true; `KullbackLeiblerDivergence_P_Q` is direction-dependent), so this is a real, untested risk for an overload-heavy PR.

**Impact:**
A transposed-argument defect in any of the new extension overloads — the single most likely failure mode for this PR — would not be caught by the test suite.

**Suggested fix:**
For each directional shape, add at least one test using asymmetric inputs (different experimental vs theoretical peak sets) and assert against an expected value computed via the verified explicit constructor, so that an argument swap changes the result.

**Affected test scenarios:**
- LibrarySpectrum_ComputeSpectralSimilarity_LibraryToMzSpectrum_Asymmetric - experimental has an unpaired peak the theoretical lacks - result differs from the swapped orientation
- MsDataScan_ComputeSpectralSimilarity_Arrays_Asymmetric - KullbackLeiblerDivergence_P_Q on asymmetric inputs - matches explicit-constructor value, not the transposed value

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
```diff
@@ -216,7 +216,64 @@ public static double[] NormalizeSpectrumSum(double[] spectrum)
 
         #endregion normalization
 
-        #region similarityMethods
+        #region SimilarityMethods
+
```
trishorts, line 220: **Minor** - Revert cosmetic-only change (Rename)

**Problem:**
The `#region` label was renamed from `similarityMethods` to `SimilarityMethods` (capitalization only).

**Impact:**
`#region` directives are ignored by the compiler, so renaming the label has zero effect on compiled IL or runtime behaviour. The author explicitly does not want stylistic churn mixed with substantive changes - it makes the PR harder to review.

**Suggested fix:**
Revert these lines to match the base branch. No functional impact - this is purely undoing noise.

**Revert patch:**
```diff
--- a/mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
+++ b/mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
@@ -216,7 +216,7 @@ namespace MassSpectrometry.MzSpectra
 
         #endregion normalization
 
-        #region SimilarityMethods
+        #region similarityMethods
 
         public enum SimilarityMeasures
         {
```

**Affected test scenarios:**
- _No tests needed - revert restores behaviour already covered by existing tests._

mzLib/Omics/SpectralLibrary/LibrarySpectrum.cs
```diff
@@ -1,4 +1,3 @@
-﻿using System.Globalization;
 using System.Text;
 using Chemistry;
```
trishorts, line 3: **Minor** - Revert cosmetic-only change (Reordering)

**Problem:**
The unused `using System.Globalization;` directive (and the leading BOM) was deleted from the top of the file.

**Impact:**
The file still compiles without it, so the directive was unused; removing an unused using directive produces identical compiled IL. The author explicitly does not want stylistic churn mixed with substantive changes - it makes the PR harder to review.

**Suggested fix:**
Revert these lines to match the base branch. No functional impact - this is purely undoing noise.

**Revert patch:**
```diff
--- a/mzLib/Omics/SpectralLibrary/LibrarySpectrum.cs
+++ b/mzLib/Omics/SpectralLibrary/LibrarySpectrum.cs
@@ -1,3 +1,4 @@
+using System.Globalization;
 using System.Text;
 using Chemistry;
 using Easy.Common.Extensions;
```

**Affected test scenarios:**
- _No tests needed - revert restores behaviour already covered by existing tests._

mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
```diff
@@ -1,17 +1,12 @@
 ﻿using Chemistry;
 using MzLibUtil;
```
trishorts, line 2: **Minor** - Revert cosmetic-only change (Reordering)

**Problem:**
Five unused/redundant using directives (`System`, `System.Collections.Generic`, `System.IO`, `System.Linq`, `ThermoFisher.CommonCore.Data.Business`) were deleted from the top of the file.

**Impact:**
The file still compiles without them, so they were redundant; removing unused using directives produces identical compiled IL and does not touch the substantive constructor change below. The author explicitly does not want stylistic churn mixed with substantive changes - it makes the PR harder to review.

**Suggested fix:**
Revert these lines to match the base branch. No functional impact - this is purely undoing noise.

**Revert patch:**
```diff
--- a/mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
+++ b/mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
@@ -1,11 +1,16 @@
 using Chemistry;
 using MzLibUtil;
 using Omics.Fragmentation;
+using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
+using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using Easy.Common.Extensions;
 using Omics.Fragmentation.Peptide;
 using Omics.SpectrumMatch;
+using ThermoFisher.CommonCore.Data.Business;
 
 namespace Readers.SpectralLibrary
```

**Affected test scenarios:**
- _No tests needed - revert restores behaviour already covered by existing tests._

trishorts (changes requested):
Automated code review - 8 finding(s): 5 posted inline, 3 Major findings recorded in this summary (not anchored to diff lines).

## Findings not anchored to diff lines

The following 3 **Major** findings concern `mzLib/Omics/SpectralLibrary/LibrarySpectrum.cs` (lines 99/101). These lines fall outside the PR's diff hunks, so they could not be posted as inline comments - they are recorded here instead.

---

### Major - `IEquatable` `Equals` lacks `GetType` check, breaks symmetry with `CrosslinkLibrarySpectrum` (LibrarySpectrum.cs:99)

`Equals(LibrarySpectrum? other)` performs no runtime type check, whereas `Equals(object? obj)` does (`obj.GetType() != GetType()`). `LibrarySpectrum` is not sealed and has a derived type `CrosslinkLibrarySpectrum`. A base `LibrarySpectrum` compared against a `CrosslinkLibrarySpectrum` whose inherited fields coincide returns `true` from `baseSpec.Equals((LibrarySpectrum)crosslink)` but `false` from `crosslink.Equals((object)baseSpec)`. Also, since `CrosslinkLibrarySpectrum` does not override equality, two distinct crosslink spectra that differ only in `BetaPeptideSpectrum` are reported equal.

**Impact:** Asymmetric equality violates the `Equals` contract and corrupts `HashSet`/`Dictionary`/`Distinct`/`SequenceEqual` behavior. Crosslink library spectra can be wrongly de-duplicated or matched.

**Suggested fix:** Add a `GetType() == other.GetType()` guard at the top of the strongly-typed `Equals(LibrarySpectrum?)`, and have `CrosslinkLibrarySpectrum` override equality to include its beta-peptide state.

---

### Major - Equality/hashing depend on `MzSpectrum` base value-equality that may not exist (LibrarySpectrum.cs:101)

`Equals(LibrarySpectrum?)` calls `base.Equals(other)` and `GetHashCode()` calls `base.GetHashCode()`. If `MzSpectrum` does not override `Equals`/`GetHashCode`, `base.Equals` is reference equality and `base.GetHashCode` is an identity hash - `Equals` then returns `false` for two distinct value-equal spectra, and equal objects produce different hash codes.

**Impact:** If `MzSpectrum` does not provide value equality, the headline feature (`IEquatable<LibrarySpectrum>`) is broken and the new equality/hash tests fail.

**Suggested fix:** Verify `MzSpectrum` overrides `Equals`/`GetHashCode` by value. If it does not, drop the `base` calls and compare the relevant spectrum state directly (e.g. `XArray`/`YArray` element-wise).

---

### Major - `Equals`/`GetHashCode` rely on `MatchedFragmentIon` equality documented as broken (LibrarySpectrum.cs:101)

`Equals` uses `MatchedFragmentIons.SequenceEqual(other.MatchedFragmentIons)` and `GetHashCode` does `hash.Add(ion)` per `MatchedFragmentIon`. Both depend on `MatchedFragmentIon.Equals`/`GetHashCode`. The test file itself documents this dependency as unreliable: `CrosslinkPsmFromTsvToLibrarySpectrumTest` contains the comment "This test would be better if MatchedIon.equals method worked, but it breaks because the mz comparison is implemented incorrectly."

**Impact:** `LibrarySpectrum` equality and hashing may produce wrong results (false negatives or false positives), undermining the new `IEquatable` contract.

**Suggested fix:** Either fix `MatchedFragmentIon.Equals`/`GetHashCode` first, or compare ions in `LibrarySpectrum.Equals` by explicit fields (Mz, Intensity, Charge, annotation) rather than relying on `MatchedFragmentIon.Equals`.

mzLib/Test/FileReadingTests/SpectralLibraryTests/TestLibrarySpectrum.cs
nbollis: Addressed. I added asymmetric spectral-similarity tests for the library-spectrum wrappers so argument-order mistakes are now detectable. The new cases compare the extension-method result against an explicit `SpectralSimilarity` construction for asymmetric inputs.

mzLib/Omics/SpectralLibrary/LibraryExtensions.cs
nbollis: Addressed. I threaded an optional `filterOutBelowThisMz` parameter through the `LibrarySpectrum` spectral-similarity extension overloads, defaulting to the existing `300` value so current callers keep existing behavior, while allowing callers to opt out or override the cutoff.

mzLib/MassSpectrometry/MzSpectra/SpectralSimilarity.cs
nbollis: Addressed. I reverted the `#region SimilarityMethods` rename back to `#region similarityMethods` to avoid unrelated cosmetic churn in this PR.

mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
nbollis: Addressed. I restored the removed usings in `SpectralLibrary.cs` to avoid unrelated formatting/cleanup churn in this PR.

mzLib/Omics/SpectralLibrary/LibrarySpectrum.cs
nbollis: Addressed. I restored `using System.Globalization;` so this PR only carries behavioral changes related to the review items and does not include unrelated cosmetic removals.

nbollis:
Replying to the three unanchored findings in the review summary:

1. Addressed. I added a runtime type check in `LibrarySpectrum.Equals(LibrarySpectrum? other)` so a base `LibrarySpectrum` no longer compares equal to a `CrosslinkLibrarySpectrum`. I also added `CrosslinkLibrarySpectrum` equality/hash overrides so the beta-spectrum state participates in equality semantics.

2. I reviewed this one and did not make a change here. `MzSpectrum` already uses value-based equality for the X/Y arrays, and its object-level equality path includes a runtime type check. Based on the current implementation, I don’t see an additional fix needed in this PR.

3. I reviewed this as well and did not make a code change here. The current `MatchedFragmentIon` implementation already has explicit `Equals`/`GetHashCode` logic that compares product, charge, and rounded m/z/intensity values. The older concern in the tests appears to be stale relative to the current code.

trishorts (approved):

pcruzparri (approved):

PR #1060, opened 2026-05-08
Add MS:1000516 charge array support to MsDataScan + mzML I/O

pcarvalho75:
## What

Adds optional per-peak charge state support to `MsDataScan`, with corresponding read/write paths in the mzML I/O. The charge array is encoded as a third `<binaryDataArray>` per spectrum using PSI-MS controlled-vocabulary term [`MS:1000516` "charge array"](https://www.ebi.ac.uk/ols4/ontologies/ms/classes?obo_id=MS:1000516).

```xml
<binaryDataArray encodedLength="...">
  <cvParam cvRef="MS" accession="MS:1000521" name="32-bit float" value=""/>
  <cvParam cvRef="MS" accession="MS:1000576" name="no compression" value=""/>
  <cvParam cvRef="MS" accession="MS:1000516" name="charge array" value=""/>
  <binary>{base64}</binary>
</binaryDataArray>
```

## Why

mzML 1.0+ defines the charge array as a **first-class spec feature**, and tools like ProteoWizard's `msconvert`, Skyline, MZmine, mzR (Bioconductor), and pymzML already consume it when present. mzLib's writer didn't expose the API, forcing downstream callers (e.g. deisotopers wanting to publish per-peak charge state) to use side-car files.

The original code even acknowledges this gap:

```csharp
// ONLY WRITING M/Z AND INTENSITY DATA, NOT THE CHARGE! (but can add charge info later)
// CHARGE (and other stuff) CAN BE IMPORTANT IN ML APPLICATIONS!!!!!
```

This PR is the "later" the comment promised.

The motivating use case: [YADA 3.1](http://patternlabforproteomics.org/yada3/) (Yet Another Deconvolution Algorithm) deisotopes proteomic spectra and wants to publish per-peak charge state inline. Today it falls back to a side-car `.envelopes.tsv` file because mzLib's API doesn't accept charges. With this PR, YADA's deisotoped output becomes a self-describing mzML that any mzML-aware downstream tool can consume natively.

## Backwards compatibility

This change is **strictly additive**:

- Constructor parameter `chargeArray` defaults to `null`, so existing callers see no behavior change.
- mzML files without the charge array parse identically (existing `mzArray`/`intensityArray` handling is untouched).
- Conformant readers that don't recognize `MS:1000516` ignore the extra array per the mzML spec — so files emitted with charges remain compatible with any mzML consumer.

## Files changed

| File | Change |
|---|---|
| `mzLib/MassSpectrometry/MsDataScan.cs` | Adds nullable `int[] ChargeArray` property + optional constructor parameter (default `null`) |
| `mzLib/Readers/MzML/MzmlMethods.cs` | When `ChargeArray != null`, writes a third `<binaryDataArray>` with cvParam `MS:1000516` "charge array", 32-bit float, no compression. Updates the `count` attribute and array sizing logic accordingly. |
| `mzLib/Readers/MzML/Mzml.cs` | Recognizes `MS:1000516` cvParam, decodes the array (rounds the 32-bit float values back to `int`), populates `MsDataScan.ChargeArray`. |

~100 LOC total.

## Implementation notes

- **Encoding**: 32-bit float per PSI-MS convention (charges fit comfortably in 2^23). 64-bit would double file size for no precision benefit on integer values.
- **Compression**: no compression by default — charge arrays are small enough (1 int per peak ≈ ~25% of intensity-array size) that the zlib overhead isn't worth it. Could be made configurable in a follow-up if anyone needs it.
- **Array-list sizing**: the existing 2-or-5 array allocation became `2 + (3 if NoiseData) + (1 if ChargeArray)`, computed once per spectrum.
- **Reader robustness**: if the charge array length disagrees with the m/z array length, downstream code can detect and handle (`scan.ChargeArray.Length != scan.MassSpectrum.Size`). The reader doesn't enforce this — same policy as for noise arrays.

## Build status

Builds clean against current `master` (0 errors, 0 new CS warnings) on .NET 8 and .NET 10.

## Suggested test (would happily add to a follow-up)

```csharp
[Test]
public void ChargeArrayRoundTrip()
{
    var spectrum = new MzSpectrum(
        new[] { 100.0, 200.0, 300.0 },
        new[] { 1000.0, 2000.0, 3000.0 },
        shouldCopy: false);
    var charges = new[] { 2, 3, 1 };

    var scan = new MsDataScan(
        spectrum, oneBasedScanNumber: 1, msnOrder: 1,
        isCentroid: true, polarity: Polarity.Positive,
        retentionTime: 1.0,
        scanWindowRange: new MzRange(50, 1000),
        scanFilter: "FTMS + p NSI Full ms",
        mzAnalyzer: MZAnalyzerType.Orbitrap,
        totalIonCurrent: 6000, injectionTime: null,
        noiseData: null, nativeId: "scan=1",
        chargeArray: charges);

    var sourceFile = new SourceFile(
        nativeIdFormat: "scan number only nativeID format",
        massSpectrometerFileFormat: "mzML format",
        checkSum: null, fileChecksumType: "SHA-1",
        uri: new Uri("file:///test"),
        id: "test", fileName: "test.mzML");

    var dataFile = new GenericMsDataFile(new[] { scan }, sourceFile);
    string outPath = Path.Combine(TestContext.CurrentContext.TestDirectory, "charge_roundtrip.mzML");
    MzmlMethods.CreateAndWriteMyMzmlWithCalibratedSpectra(dataFile, outPath, writeIndexed: true);

    var reread = (Mzml)Mzml.LoadAllStaticData(outPath);
    var readBack = reread.GetOneBasedScan(1);
    CollectionAssert.AreEqual(charges, readBack.ChargeArray);
}
```

(`GenericMsDataFile` is just any concrete `MsDataFile` subclass — happy to use whatever pattern you prefer for the test project.)

Happy to iterate on naming, parameter placement, encoding choice, or test approach. Thanks for the great library!

### DIFF
--- a/mzLib/MassSpectrometry/MsDataScan.cs
+++ b/mzLib/MassSpectrometry/MsDataScan.cs
@@ -25,30 +25,31 @@ namespace MassSpectrometry
 {
     public class MsDataScan
     {
-        public MsDataScan(MzSpectrum massSpectrum, 
-            int oneBasedScanNumber, 
-            int msnOrder, 
-            bool isCentroid, 
-            Polarity polarity, 
-            double retentionTime, 
-            MzRange scanWindowRange, 
-            string scanFilter, 
+        public MsDataScan(MzSpectrum massSpectrum,
+            int oneBasedScanNumber,
+            int msnOrder,
+            bool isCentroid,
+            Polarity polarity,
+            double retentionTime,
+            MzRange scanWindowRange,
+            string scanFilter,
             MZAnalyzerType mzAnalyzer,
-            double totalIonCurrent, 
-            double? injectionTime, 
-            double[,] noiseData, 
-            string nativeId, 
-            double? selectedIonMz = null, 
-            int? selectedIonChargeStateGuess = null, 
-            double? selectedIonIntensity = null, 
+            double totalIonCurrent,
+            double? injectionTime,
+            double[,] noiseData,
+            string nativeId,
+            double? selectedIonMz = null,
+            int? selectedIonChargeStateGuess = null,
+            double? selectedIonIntensity = null,
             double? isolationMZ = null,
-            double? isolationWidth = null, 
-            DissociationType? dissociationType = null, 
-            int? oneBasedPrecursorScanNumber = null, 
-            double? selectedIonMonoisotopicGuessMz = null, 
+            double? isolationWidth = null,
+            DissociationType? dissociationType = null,
+            int? oneBasedPrecursorScanNumber = null,
+            double? selectedIonMonoisotopicGuessMz = null,
             string hcdEnergy = null,
-            string scanDescription = null, 
-            double? compensationVoltage = null)
+            string scanDescription = null,
+            double? compensationVoltage = null,
+            int[] chargeArray = null)
         {
             OneBasedScanNumber = oneBasedScanNumber;
             MsnOrder = msnOrder;
@@ -72,9 +73,10 @@ namespace MassSpectrometry
             SelectedIonMonoisotopicGuessMz = selectedIonMonoisotopicGuessMz;
             HcdEnergy = hcdEnergy;
             ScanDescription = scanDescription;
-            CompensationVoltage = compensationVoltage; 
+            CompensationVoltage = compensationVoltage;
+            ChargeArray = chargeArray;
 
-            // Ensure the charge of the selected ion matches the polarity of the scan 
+            // Ensure the charge of the selected ion matches the polarity of the scan
             SelectedIonChargeStateGuess = Polarity switch
             {
                 Polarity.Negative when selectedIonChargeStateGuess is > 0 => -selectedIonChargeStateGuess,
@@ -82,6 +84,17 @@ namespace MassSpectrometry
                 _ => selectedIonChargeStateGuess
             };
         }
+
+        /// <summary>
+        /// Per-peak charge state, parallel to <see cref="MassSpectrum"/>.XArray.
+        /// Length must equal MassSpectrum.Size when set; null means "no charge data captured".
+        /// Encoded in mzML as a third &lt;binaryDataArray&gt; with cvParam
+        /// MS:1000516 ("charge array"), 32-bit float, zlib-compressed.
+        ///
+        /// Only set by deisotopers (e.g. YADA's annotate mode); vendor-converted mzML
+        /// rarely includes it. Readers that don't understand MS:1000516 silently ignore it.
+        /// </summary>
+        public int[] ChargeArray { get; protected set; }
 
         /// <summary>
         /// The mass spectrum associated with the scan
--- a/mzLib/MassSpectrometry/MsDataScan.cs
+++ b/mzLib/MassSpectrometry/MsDataScan.cs
@@ -89,7 +89,9 @@ namespace MassSpectrometry
         /// Per-peak charge state, parallel to <see cref="MassSpectrum"/>.XArray.
         /// Length must equal MassSpectrum.Size when set; null means "no charge data captured".
         /// Encoded in mzML as a third &lt;binaryDataArray&gt; with cvParam
-        /// MS:1000516 ("charge array"), 32-bit float, zlib-compressed.
+        /// MS:1000516 ("charge array"), 32-bit float, no compression
+        /// (charge arrays are small enough that the zlib overhead isn't worth it; could
+        /// be made configurable in a follow-up if a downstream consumer needs it).
         ///
         /// Only set by deisotopers (e.g. YADA's annotate mode); vendor-converted mzML
         /// rarely includes it. Readers that don't understand MS:1000516 silently ignore it.
--- a/mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
+++ b/mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
@@ -1775,15 +1775,21 @@ namespace Test.FileReadingTests.SpectraFileReading
             //   1. construct an MsDataScan with a per-peak chargeArray,
             //   2. write it as mzML,
             //   3. re-read it,
-            //   4. assert the round-tripped ChargeArray matches the original element-for-element.
+            //   4. assert ChargeArray, m/z, AND intensity all round-trip element-for-element.
             //
             // Exercises both the writer branch in MzmlMethods (third <binaryDataArray> with
             // accession="MS:1000516") and the reader branch in Mzml.cs (cvParam recognition
             // + 32-bit-float decode rounded back to int[]).
-            var spectrum = new MzSpectrum(
-                new[] { 100.0, 200.0, 300.0 },
-                new[] { 1000.0, 2000.0, 3000.0 },
-                shouldCopy: false);
+            //
+            // m/z and intensity assertions cover the most likely class of regression in this
+            // PR: a bug in the writer's per-spectrum array-list sizing (count attribute,
+            // allocation index, off-by-one when slotting the third <binaryDataArray>) could
+            // leave the new charge array intact while corrupting or mis-indexing the m/z or
+            // intensity arrays. Without these assertions a "passing" green test could ship
+            // a corrupted m/z/intensity round-trip.
+            var mzs = new[] { 100.0, 200.0, 300.0 };
+            var intensities = new[] { 1000.0, 2000.0, 3000.0 };
+            var spectrum = new MzSpectrum(mzs, intensities, shouldCopy: false);
             var charges = new[] { 2, 3, 1 };
 
             var scan = new MsDataScan(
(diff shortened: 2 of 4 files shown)


nbollis (changes requested):
Nice addition to the codebase. That test would work out perfectly.

We use GenericMsDataFile as our concrete testing placeholder (primarily for writing a custom mzML to test a feature) exactly as you implemented. 

The integration test GitHub action is failing due to our changes, and not anything that you produced. 

If you added that simple regression test, I would be happy to approve.

pcarvalho75:
Thanks @nbollis! Pushed the requested test in [b68c595b](https://github.com/smith-chem-wisc/mzLib/pull/1060/commits/b68c595b).

Added two cases to `mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs`:

1. **`ChargeArrayRoundTrip`** — the round-trip you suggested. Constructs an `MsDataScan` with a per-peak `chargeArray = { 2, 3, 1 }`, writes it via `MzmlMethods.CreateAndWriteMyMzmlWithCalibratedSpectra`, re-reads it via `Mzml.LoadAllStaticData`, and asserts element-for-element equality. Exercises both the writer branch (third `<binaryDataArray>` with `MS:1000516`) and the reader branch (cvParam recognition + 32-bit-float decode rounded back to `int[]`).
2. **`ChargeArrayDefaultsToNullWhenAbsent`** — pins the strictly-additive behavior. A scan written without a `chargeArray` round-trips with `ChargeArray == null`, so a future change can't quietly start emitting an empty array.

Both pass locally on net8.0:
```
Passed!  - Failed: 0, Passed: 2, Skipped: 0, Total: 2
```

Should also bring patch coverage on `MzmlMethods.cs` and `Mzml.cs` close to 100% (codecov was flagging those branches as uncovered).

pcarvalho75:
Heads-up: I just force-pushed to fix the commit author on the original commit (it had gone out with a placeholder identity). **No code changes** — the diff is identical, both commits now properly attributed. Sorry for the notification noise.

nbollis (approved):

nbollis (dismissed):
10/10 would review again😊

mzLib/MassSpectrometry/MsDataScan.cs
```diff
@@ -25,30 +25,31 @@ namespace MassSpectrometry
...
+            double? isolationWidth = null,
+            DissociationType? dissociationType = null,
+            int? oneBasedPrecursorScanNumber = null,
+            double? selectedIonMonoisotopicGuessMz = null,
             string hcdEnergy = null,
-            string scanDescription = null, 
-            double? compensationVoltage = null)
+            string scanDescription = null,
+            double? compensationVoltage = null,
+            int[] chargeArray = null)
```
trishorts, line 52: **Problem:**
`GetOneBasedScanFromDynamicConnection` is the second reader path in `Mzml.cs` and processes the same `<binaryDataArray>` elements as the static reader, but its `CVPARAM` switch only handles `MS:1000514` (m/z) and `MS:1000515` (intensity) by setting `readingMzs` / `readingIntensities`. There is no branch for `MS:1000516`. The `BINARY` case then short-circuits with `if (!readingMzs && !readingIntensities) break;`, so charge-array bytes are never decoded. The constructed `MsDataScan` at the end of the method also never passes `chargeArray`. The static path (`GetMsDataOneBasedScanFromConnection`) was updated, but this one was missed.

**Impact:**
Any caller that uses dynamic loading (`InitiateDynamicConnection` + `GetOneBasedScanFromDynamicConnection`) on an mzML file written by this PR's writer will get scans with `ChargeArray == null`, silently discarding the per-peak charge information. This breaks the PR's stated read-side contract ("Recognizes `MS:1000516` cvParam, decodes the array, populates `MsDataScan.ChargeArray`") for half of the reader API and creates an asymmetry between the two load paths that is invisible to callers.

**Suggested fix:**
Add a `case "MS:1000516":` to the CVParam switch that sets a `readingCharges` flag, decode the binary data when that flag is set (rounding 32-bit floats back to `int` the same way the static path does), accumulate into an `int[] chargeArray`, and pass it as the `chargeArray:` named argument to the `new MsDataScan(...)` call at the end of the SPECTRUM end-tag branch.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
```diff
@@ -1768,6 +1768,122 @@ private MzSpectrum CreateMS2spectrum(IEnumerable<Fragment> fragments, int v1, in
             return new MzSpectrum(allMassesArray, allIntensitiessArray, false);
         }
 
+        [Test]
+        public static void ChargeArrayRoundTrip()
+        {
+            // Regression test for the MS:1000516 "charge array" round-trip:
```
trishorts, line 1774: **Problem:**
The new XML doc comment on `ChargeArray` states the array is "Encoded in mzML as a third <binaryDataArray> with cvParam MS:1000516 ("charge array"), 32-bit float, zlib-compressed." The PR description and the writer change explicitly use cvParam `MS:1000576` ("no compression") and call out that no compression is applied ("zlib overhead isn't worth it"). The doc comment therefore documents a contract that contradicts what the code does.

**Impact:**
A future maintainer reading the public property's XML doc will believe charge-array bytes are zlib-compressed and may build a downstream tool, debugging session, or test fixture around that assumption. The "comments are signal" rule applies — this comment was added by this PR and conflicts with the implementation it describes.

**Suggested fix:**
Update the doc comment to match the implementation: 32-bit float, **no compression** (and optionally note that compression was deferred and could be added later, mirroring the PR description).


**Problem:**
`ChargeArrayRoundTrip` asserts only `readBack.ChargeArray` equals the originally supplied `charges`. The PR scope explicitly changes the per-spectrum `count` attribute and the array-list sizing logic in the writer (`2 + (3 if NoiseData) + (1 if ChargeArray)`). A bug in that sizing — wrong `count`, wrong allocation index, off-by-one when slotting the third `<binaryDataArray>` — could corrupt or mis-index the m/z or intensity arrays without affecting the charge array, and the test would still pass.

**Impact:**
The single most likely class of regression introduced by this PR (writer sizing/indexing bugs around the new third array) is uncovered by the only test that targets it. A green test would let a corrupted m/z/intensity round-trip ship.

**Suggested fix:**
After the read-back, also assert `readBack.MassSpectrum.XArray` and `YArray` equal the originally supplied `mz` and `intensities` element-for-element (with `CollectionAssert.AreEqual`).

**Problem:**
The new sizing formula in scope is `2 + (3 if NoiseData) + (1 if ChargeArray)`, giving four distinct branches (2, 3, 5, 6 arrays per spectrum). `ChargeArrayRoundTrip` exercises only the `ChargeArray, no NoiseData` (3-array) branch; `ChargeArrayDefaultsToNullWhenAbsent` exercises only the baseline 2-array branch. The most complex branch — both `NoiseData` and `ChargeArray` set, producing six binary data arrays per spectrum — is untested. That branch is exactly where indexing/count bugs are most likely.

**Impact:**
A bug in the combined `NoiseData + ChargeArray` writer branch (wrong `count`, wrong array order, NoiseData and ChargeArray bytes swapped) would ship undetected. Downstream tools that consume both arrays would silently get wrong data.

**Suggested fix:**
Add a third round-trip test that constructs an `MsDataScan` with both a non-null `noiseData` (e.g. a small `double[3, N]`) and a non-null `chargeArray`, then asserts that all of: m/z array, intensity array, NoiseData, and ChargeArray round-trip correctly.

trishorts (commented):
nice PR. please consider a revision dealing with the highlighted problems

mzLib/MassSpectrometry/MsDataScan.cs
pcarvalho75: Fixed in feb8c55e.

Mirrored the static-reader path inside `GetOneBasedScanFromDynamicConnection`:

- New `readingCharges` flag and `chargeArray` local alongside the existing `readingMzs` / `readingIntensities`
- New `case "MS:1000516":` in the CVPARAM switch
- BINARY branch now decodes when any of the three flags is set; charge bytes get rounded back to `int[]` exactly the way the static path does
- The constructed `MsDataScan` at the end of the SPECTRUM end-tag branch now receives `chargeArray: chargeArray`

`InitiateDynamicConnection` callers will now see `MsDataScan.ChargeArray` populated for files that carry the third `<binaryDataArray>`, matching the static loader's behavior.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
pcarvalho75: All three sub-issues in this comment fixed in feb8c55e.

**1. Doc comment (zlib-compressed → no compression).** Updated the XML doc on `MsDataScan.ChargeArray` to say `32-bit float, no compression`, with a brief note that compression was deferred and could be added later — mirrors the PR description.

**2. `ChargeArrayRoundTrip` tightened.** After the read-back the test now also asserts:
```csharp
CollectionAssert.AreEqual(mzs, readBack.MassSpectrum.XArray);
CollectionAssert.AreEqual(intensities, readBack.MassSpectrum.YArray);
```
A writer sizing/indexing bug that corrupted m/z or intensity while leaving the charge array intact will now fail this test instead of silently shipping.

**3. New `ChargeArrayRoundTripWithNoiseData`.** Constructs an `MsDataScan` with both `noiseData = double[3, 2]` (mass / noise / baseline rows) and `chargeArray = { 1, 2, 3, 4 }`, exercising the 6-array writer branch. Asserts:
- m/z, intensity, and charge all round-trip element-for-element (NoiseData itself isn't read back today — there's a `// TODO: read this` against it — so we don't assert on it directly)
- The on-disk XML's spectrum-scoped `<binaryDataArrayList>` declares `count="6"` (regex over the first match in the file, so chromatogram arrays at file scope don't confuse the count)
- An `accession="MS:1000516"` cvParam is present

All three charge-array tests pass:
```
Passed ChargeArrayDefaultsToNullWhenAbsent
Passed ChargeArrayRoundTrip
Passed ChargeArrayRoundTripWithNoiseData
```
Full `TestMzML` (43 tests) and dynamic-connection (55 tests) suites still green.

nbollis (approved):

mzLib/MassSpectrometry/MsDataScan.cs
trishorts: new issue just showed up after this update (which was very nice btw).   **Problem:**
In `GetOneBasedScanFromDynamicConnection`, the newly-added `chargeArray` is read from the mzML in the file's original order and then passed straight into the `MsDataScan` constructor at the end of the SPECTRUM case. But before that constructor call, the existing code (a) optionally removes zero-intensity peaks via `Array.Sort(intensities, mzs)` + `SubArray` + `Array.Sort(mzs, intensities)`, (b) optionally runs `WindowModeHelper.Run(ref intensities, ref mzs, ...)` for peak filtering, and (c) unconditionally runs `Array.Sort(mzs, intensities)` to enforce m/z ordering. None of these touch `chargeArray`, so a charge value that originally belonged to peak *i* in the source mzML now corresponds to a different peak (or to a peak that has been filtered away) once the spectrum reaches the caller.

**Impact:**
Charges silently misalign with their peaks whenever the source mzML has any peaks with intensity below `0.01`, any out-of-order m/z, or `filterParams` is non-null. Callers can't detect this via the length-mismatch policy the PR documents - lengths can still match while values point at the wrong peaks. Self-roundtrip tests with pre-sorted, non-zero-intensity data won't expose the bug, so it can ship unnoticed and corrupt the very downstream consumers (YADA, mzR, pymzML) the PR is meant to serve.

**Suggested fix:**
Treat `chargeArray` as a third parallel array. Whenever `mzs`/`intensities` are reordered or trimmed, apply the same permutation/mask to `chargeArray`. Concretely: before the zero-intensity-removal `Array.Sort(intensities, mzs)`, build an index array and sort it instead so the resulting permutation can be applied to all three arrays; similarly carry `chargeArray` through `WindowModeHelper` (or perform an indexed filter first) and the final `Array.Sort`. The same sync gap exists in `GetMsDataOneBasedScanFromConnection` for the static path and should be fixed there too.

**Affected test scenarios:**
- ChargeArrayRoundTrip - mzML with unsorted m/z input - expect ChargeArray values still aligned to sorted m/z peaks
- ChargeArrayRoundTrip - mzML with at least one intensity < 0.01 - expect ChargeArray length matches MassSpectrum.Size and remaining charges align to surviving peaks
- ChargeArrayRoundTrip - LoadAllStaticData with `FilteringParams` that prunes peaks - expect ChargeArray.Length == MassSpectrum.Size and per-peak alignment preserved
-

trishorts (approved):

nbollis:
If you are interested, we can set up a call and discuss any component with of our codebase with you. 

It may be worthwhile for you to add yada to our deconvolution structure in the mass spectrometry project

pcarvalho75:
Thanks @nbollis, really appreciate the offer! We're in the final stretch of YADA 3.1 right now — once that ships I'd love to look at integrating into mzLib's deconvolution structure. Will reach out then to set up a call.

PR #1064, opened 2026-05-12
FromFile deconvolution algorithm

trishorts:
# Add FromFile deconvolution algorithm

## Summary

**What this is for.** After you run a deconvolution algorithm on MS1 spectra (or load one from disk), you get a list of features — each a (mass, charge, RT-window) triple representing a peptide or proteoform observed during a chromatographic peak. Before any of them can be searched against a protein database, you have to figure out which MS2 scans the instrument fragmented to interrogate them. That's a join: every MS2 scan has a retention time and an isolation m/z; every feature has an RT window and (once you divide by charge) an m/z; you match them up. Today every mzLib consumer reimplements that join — MetaMorpheus inside its `Ms2ScanWithSpecificMass` construction loop, ProteoformExplorer in its own pairing layer, ad-hoc scripts by hand. This PR adds the canonical version as a new `DeconvolutionAlgorithm` that plugs into the existing `Deconvoluter` factory, with no MetaMorpheus dependencies.

**What's in the PR.** A new deconvolution algorithm — `FromFileDeconvolutionAlgorithm` — and its parameters class `FromFileDeconvolutionParameters`, both living under `Readers/Deconvolution/`. The parameters take a feature-file path; the file format is auto-detected (FlashDeconv / TopFD `.ms1.feature`, Dinosaur `.feature.tsv`) via the existing `Readers.FileReader` + `SupportedFileType` machinery. The algorithm filters the loaded features by m/z range, RT-window overlap, and the parameters' `[MinAssumedChargeState, MaxAssumedChargeState]` and yields synthetic `IsotopicEnvelope` objects for the matches.

**The MM-side picture.** MetaMorpheus's existing precursor-deconvolution loop in `TaskLayer/MetaMorpheusTask.cs` calls `MsDataScan.GetIsolatedMassesAndCharges(precursorScan, deconParams)` per MS2 scan and builds an `EngineLayer.Util.Precursor` from each returned `IsotopicEnvelope`. That loop **does not change**. To switch from "deconvolute on the fly" to "load from file," MM only swaps the `DeconvolutionParameters` it constructs — `ClassicDeconvolutionParameters` → `new FromFileDeconvolutionParameters(filePath, minCharge, maxCharge)`. No new method calls, no new types in MM, no coordination PR.

```
external decon output         mzLib (this PR)                                       MetaMorpheus (no change required)
─────────────────────         ──────────────────────────────────────                ─────────────────────────────────
.ms1.feature (FlashDeconv / TopFD)
.feature.tsv (Dinosaur)
        │
        ▼
new FromFileDeconvolutionParameters(filePath, minCharge, maxCharge)   ← auto-detects format
        │
        ▼
foreach (ms2Scan) {
    var envelopes =
        ms2Scan.GetIsolatedMassesAndCharges(precursorMs1Scan, fromFileParams);     ← single call
    foreach (var env in envelopes) {
        var precursor = new Precursor(env, ...);                                    ← MM existing
        var scan = new Ms2ScanWithSpecificMass(ms2Scan, ...);                       ← MM existing
        // ... search ...
    }
}
```

**How the polymorphism slot works.** `DeconvolutionParameters` gains a virtual `CreateAlgorithm()` hook returning `null` by default. `FromFileDeconvolutionParameters` overrides it to return its own `FromFileDeconvolutionAlgorithm` instance. `Deconvoluter.CreateAlgorithm` calls `parameters.CreateAlgorithm()` first; if non-null, that wins. Otherwise it falls through to the existing enum-based switch for the in-project algorithms (`Classic`, `IsoDec`, `ExampleNew`). This lets `FromFile` live in `Readers` (which needs to drive the file-format machinery) without MassSpectrometry-side code needing to reference Readers types. `DeconvolutionAlgorithm.Deconvolute` is promoted from `internal abstract` to `protected internal abstract` so subclasses in other assemblies can override; existing in-project overrides updated to match.

**What's deliberately not in the PR.** No curation in the algorithm. No cross-charge consensus, no off-by-one correction, no harmonic collapse, no quality filtering, no MS2 fragment deconvolution, no isobaric-tag reporter extraction. Those concerns belong to the producer upstream (which already does them for FlashDeconv / TopFD output) or to the search engine downstream. The algorithm is a pure m/z + RT + charge filter — small enough that a reviewer can read it (~30 lines of body) and convince themselves it's correct without trusting any abstractions.

**Efficiency.** Features are stored in `FromFileDeconvolutionParameters` sorted by m/z, with a parallel `double[]` of m/z keys. The algorithm binary-searches for the first feature whose m/z is at or above the requested lower bound, iterates forward, and `yield break`s the moment m/z exceeds the upper bound. Per-MS2 query cost is O(log N + k) — typically log₂(10k) ≈ 14 comparisons plus 1–3 yielded envelopes vs the old O(N) linear scan.

**Behavior change in `MsDataScan.GetIsolatedMassesAndCharges`.** The `MzSpectrum` overload of `GetIsolatedMassesAndCharges` no longer auto-upgrades a plain `MzRange` to an `MzRtRange` when given FromFile params — it now lets the throw surface from `Deconvoluter.Deconvolute(MzSpectrum, ...)`. Callers wanting from-file decon must use the `MsDataScan` overload (which routes through `Deconvoluter.Deconvolute(MsDataScan, ...)` — that overload still auto-upgrades using the precursor scan's `RetentionTime`). This is the Nic-requested contract change; it makes the explicit "FromFile needs an RT anchor" requirement consistent between the two entry points.

**One small physical change to call out.** `ISingleChargeMs1Feature` moves from `Readers` to `MassSpectrometry` (via `git mv`, history preserved). `Deconvoluter`'s `Deconvolute` overloads inspect `parameters.DeconvolutionType` for the FromFile branch, and the interface they consume needs to be reachable from `MassSpectrometry`. No external consumer references the interface today (grep-checked on MetaMorpheus, ProteaseGuru, ProteoformExplorer), so the move is invisible outside this PR; five files inside `Readers` get a one-line `using MassSpectrometry;` added.

**A note on envelope fidelity.** For input coming from external readers, the returned envelope is built via the existing 3-arg `new IsotopicEnvelope(monoisotopicMass, intensity, charge)` constructor. `Peaks` is a single synthetic entry; `Score` is `double.MaxValue`. `NumberOfIsotopes` from the input interface is not surfaced because per-peak m/z and intensity are unknown for external output. Producers handing real envelopes (e.g. an in-house whole-file deconvolution) can construct the parameters via the internal `IEnumerable<ISingleChargeMs1Feature>` constructor instead and pass envelopes with real peak lists — the contract works lossless in that case.

## File-level changes

| File | Change |
|---|---|
| `MzLibUtil/MzRange.cs` | New `MzRtRange : MzRange` class — adds RT bounds, `Contains(mz, rt)`, `CompareTo(mz, rt)`, mean / width helpers |
| `MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs` | New `virtual CreateAlgorithm()` returning null. Polymorphism slot for out-of-project algorithms |
| `MassSpectrometry/Deconvolution/Algorithms/DeconvolutionAlgorithm.cs` | `Deconvolute` promoted from `internal abstract` → `protected internal abstract` |
| `MassSpectrometry/Deconvolution/Algorithms/{Classic,IsoDec,ExampleNew}*.cs` | Override accessor changed to match base (`internal override` → `protected internal override`) — no behavior change |
| `MassSpectrometry/Deconvolution/Deconvoluter.cs` | Factory tries `parameters.CreateAlgorithm()` first; falls back to enum switch. `Deconvolute(MsDataScan, ...)` auto-upgrades `MzRange` → `MzRtRange` for FromFile params using `scan.RetentionTime`. `Deconvolute(MzSpectrum, ...)` throws when given FromFile + plain `MzRange` |
| `MassSpectrometry/Deconvolution/DeconvolutionType.cs` | New `FromFile` enum entry |
| `MassSpectrometry/MsDataScan.cs` | `GetIsolatedMassesAndCharges` no longer auto-upgrades; extracts the `8.5` magic number into a const |
| `MassSpectrometry/ISingleChargeMs1Feature.cs` | Moved from `Readers/BaseClasses/` via `git mv` |
| `Readers/Deconvolution/FromFileDeconvolutionParameters.cs` | **New.** Public ctor takes file path; loads via `FileReader.ReadResultFile`. Internal ctor takes pre-loaded features for unit tests |
| `Readers/Deconvolution/FromFileDeconvolutionAlgorithm.cs` | **New.** Filters features by m/z + RT + charge, binary-searches the m/z lower bound |
| `Readers/{IMs1FeatureFile, DinosaurFeature, Ms1Feature, DinosaurTsvFile, Ms1FeatureFile}.cs` | One-line `using MassSpectrometry;` added (paired with the interface move) |
| `Test/TestMzRtRange.cs` | **New.** 15 tests covering MzRtRange ctors, properties, Contains, CompareTo, ToString |
| `Test/MassSpectrometryTests/TestFromFileDeconvolution.cs` | **New** (was `TestPrecursorPairing.cs`). 25 tests covering direct algorithm, MsDataScan integration, end-to-end fixture, error paths, and a test-only `DeconvolutionParameters` stub locking the "FromFile params must override CreateAlgorithm" contract |

## Test plan

- [x] New tests: 40 (15 in `TestMzRtRange.cs` + 25 in `TestFromFileDeconvolution.cs`). All pass.
- [x] Full `Test.csproj` suite: **3436/3436 pass**, 5 benchmarks skipped, 2 min 33 s on .NET 8.
- [x] Coverage on PR-touched files: **every class ≥ 90% line coverage** (most at 100%). Measured via `dotnet-coverage collect ... -f cobertura`.
- [ ] Reviewer to confirm: the move of `ISingleChargeMs1Feature` to `MassSpectrometry` is acceptable (no external consumer references it today — grep on MetaMorpheus / ProteaseGuru / ProteoformExplorer).
- [ ] Reviewer to confirm: the `DeconvolutionAlgorithm.Deconvolute` accessibility change from `internal abstract` to `protected internal abstract` is acceptable.

## Review-fix audit (Nic's comments on this PR)

| Comment | Status |
|---|---|
| "Good." (×2) — file-level positive | No action |
| "Could possibly improve efficiency by storing the features in an ordered collection" (FromFileDeconvolutionAlgorithm.cs line 47) | **Done** — features sorted by m/z in the parameters; algorithm binary-searches the lower bound and breaks on upper bound |
| "Better approach would be to take in a file path on construction and load the data inside of here. This may necessitate moving FromFileDeconParams to the readers project. This would allow you to take advantage of our Readers structure such as the DetectFileType method." (FromFileDeconvolutionParameters.cs file-level) | **Done** — both params and algorithm moved to `Readers/Deconvolution/`. Public ctor takes a string filePath, dispatches through `FileReader.ReadResultFile` which auto-detects the format via `SupportedFileType.GetResultFileType` |
| "This one should throw just like inside of Deconvoluter" (MsDataScan.cs line 192) | **Done** — `MsDataScan.GetIsolatedMassesAndCharges` no longer auto-upgrades for FromFile params. The `MzSpectrum` overload now throws via Deconvoluter, matching the canonical path. Locked by a new test (`MzSpectrumOverloadOfGimc_FromFileWithoutMzRtRange_Throws`) |
| Top-level summary review: "FromFileDecon parameters and algorithm will need to be moved to the readers project. The parameters will then take in the file path and load up the objects that are needed." | **Done** — addressed by the moves above. The polymorphic `DeconvolutionParameters.CreateAlgorithm()` hook is what lets the algorithm live outside MassSpectrometry while still plugging into `Deconvoluter`'s factory |


mzLib/MassSpectrometry/Deconvolution/Deconvoluter.cs
nbollis, line 1: Good.

mzLib/Readers/Deconvolution/FromFileDeconvolutionAlgorithm.cs
nbollis, line 1: Good.

mzLib/Readers/Deconvolution/FromFileDeconvolutionAlgorithm.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            var fromFileParams = DeconvolutionParameters as FromFileDeconvolutionParameters
+                ?? throw new MzLibException(
+                    "FromFileDeconvolutionAlgorithm requires FromFileDeconvolutionParameters");
+
+            var rangeWithRt = range as MzRtRange
+                ?? throw new MzLibException(
+                    "FromFileDeconvolutionAlgorithm requires an MzRtRange (m/z range + RT bounds)");
+
+            foreach (var feat in fromFileParams.Features)
```
nbollis: Could possibly improve efficiency by storing the features in an ordered collection within the FromFileParameters.

mzLib/MassSpectrometry/Deconvolution/Parameters/FromFileDeconvolutionParameters.cs
nbollis, line 1: Great foundation. A better approach would be to take in a file path on construction and load the data inside of here. 

This may necessitate moving FromFileDeconParams to the readers project. 
This would allow you to take advantage of our Readers structure such as the DetectFileType method.

mzLib/MassSpectrometry/MsDataScan.cs
```diff
@@ -184,11 +184,17 @@ public byte[] Get64BitNoiseDataBaseline()
...
-                ? new List<IsotopicEnvelope>()
-                : Deconvoluter.Deconvolute(precursorSpectrum, deconParameters,
-                    new MzRange(IsolationRange.Minimum - 8.5, IsolationRange.Maximum + 8.5))
-                    .Where(b => b.Peaks.Any(cc => isolationRange.Contains(cc.mz)));
+            const double isolationPadding = 8.5;
+
+            if (IsolationRange is null)
+                return new List<IsotopicEnvelope>();
+
+            MzRange range = new MzRange(IsolationRange.Minimum - isolationPadding, IsolationRange.Maximum + isolationPadding);
```
nbollis: This one should throw just like inside of Deconvoluter

nbollis (commented):
The reason for implementing in such a way is so that MetaMorpheusTask._GetMs2Scans(...) can function identically if the decon is from file or if we are doing it ourselves. 

The current approach is a great first step in making this all work together. In the ideal world, the parameter object will handle the file reading as well. This will allow for a more universal system. That means that the FromFileDecon parameters and algorithm will need to be moved to the readers project. 

The parameters will then take in the file path and load up the objects that are needed. When deconvoluter.deconvolute runs through the fromfile deconvoluter, we can simply return the proper isotopic envelops as we are now. The main change here is where the input ISingleCharge.... come from

Alexander-Sol (approved):

nbollis (approved):

PR #1066, opened 2026-05-12
PSM Reading allows empty name and gene when splititng an ambiguous PSM 

nbollis:
Had a weird crash while reading in a custom PSM file. 

Parser is now more robust.


mzLib/Readers/InternalResults/IndividualResultRecords/SpectrumMatchFromTsv.cs
```diff
@@ -7,6 +7,7 @@
 using Omics.SpectrumMatch;
 using MzLibUtil;
 using System.Numerics;
+using ThermoFisher.CommonCore.Data.Business;
```
trishorts: Minor: `ThermoFisher.CommonCore.Data.Business` is the Thermo RAW-file reading namespace — nothing in the new name/gene logic references it. Looks like an accidental IDE auto-import; worth removing so the PR stays scoped to the name/gene fix.

mzLib/Readers/InternalResults/IndividualResultRecords/SpectrumMatchFromTsv.cs
```diff
@@ -222,8 +223,28 @@ protected SpectrumMatchFromTsv(SpectrumMatchFromTsv psm, string fullSequence, in
                 BaseSeq = baseSequence == "" ? psm.BaseSeq.Split("|")[index] : baseSequence;
                 StartAndEndResiduesInParentSequence = psm.StartAndEndResiduesInParentSequence.Split("|")[index];
                 Accession = psm.Accession.Split("|")[index];
-                Name = psm.Name.Split("|")[index];
-                GeneName = psm.GeneName.Split("|")[index];
+
+                if (psm.Name is null)
```
trishorts, line 226: Informational: the new fixture `TDGPTMDSearchResults_NoNameAndGene.psmtsv` has the Name/Gene columns *present but empty*, which parses to `""` (not `null`) — so it flows through the `else` -> `Length == 1` -> `[0]` path. The `is null` branch here (and the `GeneName` equivalent) isn't actually exercised by the new test. Fine if it's just a defensive guard; if you want it covered you'd need a fixture with the columns absent entirely.

mzLib/Readers/InternalResults/IndividualResultRecords/SpectrumMatchFromTsv.cs
```diff
@@ -222,8 +223,28 @@ protected SpectrumMatchFromTsv(SpectrumMatchFromTsv psm, string fullSequence, in
...
+                    Name = string.Empty;
+                else
+                {
+                    var nameSplits = psm.Name.Split("|");
+                    if (nameSplits.Length == 1)
+                        Name = nameSplits[0];
+                    else 
+                        Name = nameSplits[index];
```
trishorts, line 234: Informational: this still throws `IndexOutOfRangeException` if `Name` splits into more than one value but fewer than `index + 1` (e.g. `Name` has 2 parts while `Accession` split into 4). The `null` and single-value cases are now handled; the partial-split case isn't. May be intentionally out of scope since the PR targets *empty* name/gene — just flagging it for a conscious call.

Nit: line 234 (`else `) has a trailing space.

trishorts (dismissed):
Reviewed the two code files directly. The core fix is correct and well-scoped — null and single-value name/gene no longer crash the ambiguous-PSM split, and parameterizing `TestPsmFromTsvDisambiguatingConstructor` with both fixtures keeps the old path covered.

One thing worth fixing before merge: the stray `using ThermoFisher.CommonCore.Data.Business;` (unrelated to the change). The other two comments are informational — author's-judgment calls.

Approving.

mzLib/Readers/InternalResults/IndividualResultRecords/SpectrumMatchFromTsv.cs
nbollis: This is intentional. If we mess up the PSM writer to be inconsistent with ambiguity writing conventions, we will know from the crash.

mzLib/Readers/InternalResults/IndividualResultRecords/SpectrumMatchFromTsv.cs
nbollis: Good catch, it is simply a defensive guard.

mzLib/Readers/InternalResults/IndividualResultRecords/SpectrumMatchFromTsv.cs
nbollis: Removed.

pcruzparri (approved):

trishorts (approved):

PR #1067, opened 2026-05-13
Accept newer-TopFD column names in `Ms1Feature`

trishorts:
## Summary

[PR 1064 (FromFile deconvolution algorithm)](https://github.com/smith-chem-wisc/mzLib/pull/1064) ships an `Ms1FeatureFile` reader keyed on the older FlashDeconv / TopFD‑v1.6.2 `_ms1.feature` schema (`Sample_ID, ID, Time_begin, Time_end, Minimum_charge_state, Maximum_charge_state, Minimum_fraction_id, Maximum_fraction_id`). Newer TopFD versions write the same `_ms1.feature` extension but use a different header set — `File_name, Fraction_ID, Feature_ID, Min_time, Max_time, Min_charge, Max_charge`, plus `Apex_time / Apex_scan / Apex_intensity / Rep_charge / Rep_average_mz / Envelope_num / EC_score`. `SupportedFileType` picks the right reader by extension, but `CsvHelper` then throws on the missing `Sample_ID / Time_begin / Minimum_charge_state / …` columns because none of the expected `[Name(…)]` headers are present.

This PR makes `Ms1Feature` tolerate **both** schemas by adding column-name aliases on the existing record and marking fields that are absent from the newer schema `[Optional]`. Downstream `GetSingleChargeFeatures()` reads only `Mass`, `ChargeStateMin/Max`, `RetentionTimeBegin/End`, and `IntensityApex` — all of which the newer schema carries, just under different column names — so the FromFile join algorithm behaves identically regardless of which producer wrote the file.

Discovered while integrating PR 1064's `FromFileDeconvolutionParameters` into MetaMorpheus (branch [`ExternalMs1Features`](https://github.com/smith-chem-wisc/MetaMorpheus/tree/ExternalMs1Features)) and pointing it at a real TopFD `_ms1.feature` from a top-down yeast run — the file parsed by hand looks identical in shape to the old schema, just relabelled. Without these aliases the reader throws immediately and the FromFile path can't load anything from current TopFD output.

## Per-field mapping

| Field on `Ms1Feature` | Old schema (existing) | Newer schema (added) | Used by `GetSingleChargeFeatures`? | Resolution |
|---|---|---|---|---|
| `SampleId` (int) | `Sample_ID` | newer schema has `File_name` (string path) — type-incompatible | no | mark `[Optional]`; default 0 when missing |
| `Id` (int) | `ID` | `Feature_ID` | no | add alias |
| `Mass` (double) | `Mass` | `Mass` | yes | unchanged |
| `Intensity` (double) | `Intensity` | `Intensity` | no | unchanged |
| `RetentionTimeBegin` (double) | `Time_begin` | `Min_time` | yes | add alias |
| `RetentionTimeEnd` (double) | `Time_end` | `Max_time` | yes | add alias |
| `RetentionTimeApex` (double) | `Time_apex` / `Apex_time` | `Apex_time` | no | already aliased |
| `IntensityApex` (double?) | `Apex_intensity` / `Intensity_Apex` | `Apex_intensity` | yes (when present) | already aliased + `[Optional]` |
| `ChargeStateMin` (int) | `Minimum_charge_state` | `Min_charge` | yes | add alias |
| `ChargeStateMax` (int) | `Maximum_charge_state` | `Max_charge` | yes | add alias |
| `FractionIdMin` (int) | `Minimum_fraction_id` | newer schema has a single `Fraction_ID`, not a min/max pair | no | mark `[Optional]` |
| `FractionIdMax` (int) | `Maximum_fraction_id` | not present | no | mark `[Optional]` |

`Sample_ID` and `Fraction_ID min/max` are deliberately **not aliased to the newer schema's columns**. `Sample_ID` is an `int` field but the newer column at that semantic position (`File_name`) is a string path — coercing it would silently corrupt the value. `Minimum_fraction_id / Maximum_fraction_id` could have aliased to the newer `Fraction_ID`, but `Fraction_ID` is a single value while the field semantics imply a range; and since the field isn't consumed downstream, marking it `[Optional]` is the honest choice. All four `[Optional]` fields take the type default (0) when the column is missing.

## Why this is safe for existing callers

- Every `[Name(…)]` attribute keeps the original schema name as the **first** alias, so CsvHelper's writer side (`csv.WriteHeader<Ms1Feature>()`) still emits the canonical FlashDeconv / TopFDv1.6.2 column names. Existing files round-trip byte-for-byte the same as today.
- No public surface change: the record's properties retain the same names, types, and visibility.
- `SupportedFileType.GetFileExtension()` is unchanged; format detection by extension still picks `Ms1FeatureFile` for any `_ms1.feature` regardless of internal schema.
- `Software` detection in `Ms1FeatureFile`'s constructor uses `firstLine.Contains("\tApex_intensity\t")` to flag TopFD output. Newer TopFD has `Apex_intensity` too, so detection continues to return `Software.TopFD` for both schemas.

## File-level changes

| File | Change |
|---|---|
| `mzLib/Readers/ExternalResults/IndividualResultRecords/Ms1Feature.cs` | Six `[Name(…)]` attributes gain a second alias; four fields gain `[Optional]`. ~15 lines including doc comments. |
| `mzLib/Test/FileReadingTests/ExternalFileTypes/Ms1Feature_TopFDvLatest_ms1.feature` | **New.** 4-row fixture captured from real TopFD output (header + 4 features including a single-charge edge case). |
| `mzLib/Test/FileReadingTests/ExternalFileReading/TestMsFeature.cs` | New `[TestCase]` on `TestFeaturesLoadAndCountIsCorrect`; new `TestTopFDLatestMs1FeatureFirstAndLastAreCorrect`; new `TestTopFDLatestMs1GetSingleChargeFeatureFunctions`; new `[TestCase]` on `TestMs1FeatureReadWrite`. |
| `mzLib/Test/Test.csproj` | `<CopyToOutputDirectory>` entry for the new fixture. |

Total: **4 files, +130 / −10** across two commits.

## Test coverage

| Test | Asserts |
|---|---|
| `TestFeaturesLoadAndCountIsCorrect` (new TestCase) | The newer-schema fixture loads end-to-end via `FileReader.ReadFile<Ms1FeatureFile>` and produces exactly 4 features. |
| `TestTopFDLatestMs1FeatureFirstAndLastAreCorrect` | Per-field mapping is correct for both the aliased columns (`Time_begin` ↔ `Min_time`, `Minimum_charge_state` ↔ `Min_charge`, etc.) and the `[Optional]` fields (`SampleId / FractionIdMin / FractionIdMax` default to 0 when absent). Covers the single-charge edge case (`Min_charge == Max_charge == 1`) in the last record. |
| `TestTopFDLatestMs1GetSingleChargeFeatureFunctions` | `GetSingleChargeFeatures()` expands the charge range identically across the two schemas: charges 6-14 yield 9 envelopes; a 1-1 single-charge feature yields exactly one; `GetMs1Features()` flattens to 9 + 12 + 10 + 1 = 32 across all four fixture features. |
| `TestMs1FeatureReadWrite` (new TestCase) | Read → write → re-read round-trip preserves every field consumed downstream. Note in the test explains that the writer emits the **older** schema headers (newer columns aren't on the record), so the round-trip converts schema-newer → schema-older + `[Optional]` defaults — exactly the columns that don't survive are the ones marked `[Optional]` and never read by `GetSingleChargeFeatures`. |

### Local test results

```
TestMsFeature suite:           Passed!  Failed: 0, Passed: 17, Skipped: 0   (318 ms)
Ms1Feature-related sweep:      Passed!  Failed: 0, Passed: 148, Skipped: 0  (2 s)
```

The broader sweep covers `TestMsFeature`, `TestFromFileDeconvolution`, `TestResultFile`, `TestSupportedFileExtensions`, and `TestDinosaurTsv` — every test file that touches the `Ms1Feature` / `FromFile` / supported-file-extensions surface area.

## Test plan

- [x] New TopFD-newer-schema fixture loads through `Ms1FeatureFile` and produces correct field values.
- [x] Existing FlashDeconv and TopFDv1.6.2 fixtures still load (no regression — first alias on each `[Name(…)]` is unchanged).
- [x] `GetSingleChargeFeatures()` behaves identically across the two TopFD schemas.
- [x] Read/write round-trip preserves every downstream-consumed field.
- [x] All 148 tests in the related surface area pass.

## Notes

- **Fixture provenance.** The 4-row fixture is real TopFD output trimmed for size; the embedded `File_name` reflects the original capture path (`E:/Projects/LVS_TD_Yeast/…`). Matches the convention of the existing test fixtures (`D:/Averaging/…` in `Ms2Feature_FlashDeconvOpenMs3.0.0_ms2.feature`).
- **Version label.** The fixture is named `Ms1Feature_TopFDvLatest_ms1.feature` because the exact TopFD version that introduced the newer schema isn't pinned in this PR. Reviewers who know the version (TopFD 1.7+?) are welcome to rename it; the test paths reference the filename so a rename is one line in `TestMsFeature.cs` plus one line in `Test.csproj`.
- **Discovery context.** Found while integrating [PR 1064](https://github.com/smith-chem-wisc/mzLib/pull/1064)'s `FromFileDeconvolutionParameters` into MetaMorpheus. The MetaMorpheus side of that integration is on branch [`ExternalMs1Features`](https://github.com/smith-chem-wisc/MetaMorpheus/tree/ExternalMs1Features) and depends on this mzLib PR before it can run end-to-end against current TopFD output.

## Commits

- `ce24771c` — Accept newer-TopFD column names in Ms1Feature (production change, ~15 line delta on `Ms1Feature.cs`).
- `3aa2fe04` — Add test coverage for newer-TopFD `_ms1.feature` schema (fixture + 4 test extensions).


mzLib/Test/FileReadingTests/ExternalFileReading/TestMsFeature.cs
```diff
@@ -36,6 +36,7 @@ public void TearDown()
         [TestCase(@"FileReadingTests\ExternalFileTypes\Ms2Feature_FlashDeconvOpenMs3.0.0_ms2.feature", 2, 1)]
         [TestCase(@"FileReadingTests\ExternalFileTypes\Ms1Feature_TopFDv1.6.2_ms1.feature", 1, 4)]
         [TestCase(@"FileReadingTests\ExternalFileTypes\Ms2Feature_TopFDv1.6.2_ms2.feature", 2, 5)]
+        [TestCase(@"FileReadingTests\ExternalFileTypes\Ms1Feature_TopFDvLatest_ms1.feature", 1, 4)]
```
nbollis: Latest is not a version, please revise.

nbollis (changes requested):
Please specify which version of TopFD instead of saying 'latest' in the testing file and update the supported file types text file to reflect the new versions we support.

mzLib/Test/FileReadingTests/ExternalFileReading/TestMsFeature.cs
trishorts: Fixed in 57de0ec4. Confirmed the version as **TopFD 1.7.0** — that's the TopPIC-suite 1.7.0 release (Dec 2023) whose ML feature-detection rework introduced the ECScore, which is the `EC_score` column in this newer `_ms1.feature` schema.

- Renamed the fixture → `Ms1Feature_TopFDv1.7.0_ms1.feature`
- `TestMsFeature.cs`: updated the 4 path references and renamed the two `TestTopFDLatest*` methods → `TestTopFDv1_7_0*`
- `Test.csproj`: repointed the copy entry
- `SupportedVersions.txt`: TopFD now lists `1.6.2, 1.7.0`

Build clean, `TestMsFeature` suite 17/17 green.

nbollis (approved):

zhuoxinshi (approved):

PR #1068, opened 2026-05-13
Multiple Decon

nbollis:
New decon algorithm which wraps up multiple for use. 

Future work could make a ConsensusDecon which inherits from this and does some filtering and reconciliation between the two algorithms.


mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
```diff
@@ -0,0 +1,442 @@
...
+        [Test]
+        public void MultipleDecon_WithMzRange_RespectsRange()
+        {
+            var spectrum = LoadProteoformSpectrum();
+            var narrowRange = new MzRange(700, 800);
+            var classicParams = DefaultClassicParams();
+            var multipleParams = new MultipleDeconParameters(
+                new DeconvolutionParameters[] { classicParams }, 1, 60);
+
+            var directNarrow = Deconvoluter.Deconvolute(spectrum, classicParams, narrowRange).ToList();
```
trishorts, line 217: **Minor** - MultipleDecon_WithMzRange_RespectsRange can pass vacuously when range is empty

**Problem:**
`MultipleDecon_WithMzRange_RespectsRange` asserts `wrappedNarrow.Count == directNarrow.Count` and then loops `foreach (var env in wrappedNarrow)` asserting each peak falls in `narrowRange`. If the narrow `MzRange(700, 800)` happens to yield zero envelopes, both counts are `0` (equal) and the `foreach` body never executes, so the test passes without verifying that the range filter is actually forwarded to the sub-algorithm — which is the stated purpose of the test (coverage item 5).

**Impact:**
A regression that causes the wrapped algorithm to drop or ignore the `MzRange` could go undetected if results are empty for this particular range, defeating the test's purpose.

**Suggested fix:**
Add an explicit `Assert.That(wrappedNarrow, Is.Not.Empty)` before the per-peak loop so the range-forwarding assertion is guaranteed to run, or assert a known non-zero expected count for this data file.

**Affected test scenarios:**
- MultipleDecon_WithMzRange_RespectsRange - narrow range yields envelopes - non-empty result and every peak within range
- MultipleDecon_WithMzRange_RespectsRange - narrow range yields no envelopes - test should fail or be explicitly meaningless rather than pass silently

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
```diff
@@ -0,0 +1,442 @@
...
+
+            Assert.That(wrapped.Count, Is.EqualTo(direct.Count));
+        }
+
+        // ?? 4. Results from all sub-algorithms are concatenated ???????????????
+
+        [Test]
+        public void MultipleDecon_ClassicAndIsoDec_CountIsAtLeastSumOfIndividual()
+        {
+            // The combined result may contain duplicate envelopes (the same peak found
```
trishorts: **Minor** - Test name and comment say "at least sum" but assertion is exact equality

**Problem:**
`MultipleDecon_ClassicAndIsoDec_CountIsAtLeastSumOfIndividual` is named for an "at least" relationship, and its comment states the combined result "may contain duplicate envelopes ... but must contain at least as many as each alone." The actual assertion is `Assert.That(multipleCount, Is.EqualTo(classicCount + isoDecCount))` — exact equality with the sum. The exact-equality assertion is the correct/stronger one for a pure `Concat`, but the name and comment describe a weaker, different contract.

**Impact:**
A future maintainer reading the name/comment may believe the algorithm is allowed to deduplicate or vary its count, and could weaken the assertion to match the misleading documentation, losing the strong check.

**Suggested fix:**
Rename the test to reflect exact-sum semantics (e.g. `MultipleDecon_ClassicAndIsoDec_CountEqualsSumOfIndividual`) and rewrite the comment to state that `Concat` performs no deduplication so the count is exactly the sum.

**Affected test scenarios:**
- MultipleDecon_ClassicAndIsoDec_CountEqualsSumOfIndividual - Classic + IsoDec wrapped - combined count equals classicCount + isoDecCount exactly

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
```diff
@@ -0,0 +1,442 @@
...
+            return new MzSpectrum(mzs, intensities, false);
+        }
+
+        private static MsDataScan MakeProteoformScan(MzSpectrum spectrum)
+            => new MsDataScan(spectrum, 1, 1, false, Polarity.Positive, 1.0,
+                new MzRange(495, 1617), "scan", MZAnalyzerType.Unknown,
+                spectrum.SumOfAllY, null, null, null,
+                740.372202090153, 19, 108419280, 740.37, 4);
+
+        private static ClassicDeconvolutionParameters DefaultClassicParams()
```
trishorts, line 64: **Minor** - No test coverage for null or empty sub-parameter collection

**Problem:**
`MultipleDeconParameters`'s constructor does `Parameters = deconParameters.ToArray()`, which throws `NullReferenceException` if `deconParameters` is null, and produces an empty `Parameters` array if an empty collection is passed. An empty array makes `MultipleDeconvolutionAlgorithm.Deconvolute` return an empty enumerable silently. None of the 12 tests exercises a null or empty sub-parameter collection, so the behaviour (throw vs. silent empty result) is neither pinned nor validated.

**Impact:**
A misconfigured `MultipleDeconParameters` (no wrapped algorithms) silently yields zero envelopes with no diagnostic, and a null argument throws an unhelpful `NullReferenceException` instead of an `ArgumentNullException`. Either way the contract is undefined and untested.

**Suggested fix:**
Add tests asserting the chosen behaviour for an empty collection (either a meaningful exception at construction or a documented empty result) and for a null argument. If guard behaviour is desired, also add the argument validation in the constructor.

**Affected test scenarios:**
- MultipleDeconParameters_EmptyParameters - empty collection passed - constructor behaviour is defined and asserted
- MultipleDeconParameters_NullParameters - null passed - throws ArgumentNullException (or documented behaviour)
- MultipleDecon_EmptyParameters_Deconvolute - algorithm run with no sub-params - returns empty enumerable

mzLib/MassSpectrometry/Deconvolution/Parameters/MultipleDeconParameters.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        if (_decoyParams is not null)
+            return _decoyParams;
+
+        lock (_decoyParamsLock)
+        {
+            if (_decoyParams is not null)
+                return _decoyParams;
+
+            var decoySubParams = Parameters.Select(p => p.ToDecoyParameters()).ToArray();
```
trishorts, line 36: **Minor** - ToDecoyParameters does not cache the null (no-decoy-support) result

**Problem:**
In `ToDecoyParameters()`, when `decoySubParams.Any(p => p is null)` the method `return null` without assigning `_decoyParams`. Every subsequent call therefore re-enters the lock and re-runs `Parameters.Select(p => p.ToDecoyParameters()).ToArray()`, recomputing each sub-parameter's decoy conversion. The success path is cached (and `MultipleDecon_ToDecoyParameters_ReturnsCachedInstance` verifies it), but the null path is not, making the caching behaviour inconsistent.

**Impact:**
Repeated `ToDecoyParameters()` calls on a `MultipleDeconParameters` that contains a non-decoy-supporting sub-param (e.g. `FromFile`) repeatedly allocate and recompute, and no test detects the inconsistency.

**Suggested fix:**
Either cache the null outcome explicitly (e.g. with a separate `bool _decoyComputed` flag, since `_decoyParams == null` cannot itself distinguish "not computed" from "computed as null"), or document that the null path is intentionally not cached. Add a test that calls `ToDecoyParameters()` twice on a FromFile-containing instance and asserts consistent behaviour.

**Affected test scenarios:**
- MultipleDecon_ToDecoyParameters_FromFileSubParam_RepeatedCalls - called twice - returns null both times without redundant recomputation

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
```diff
@@ -0,0 +1,442 @@
...
+            // Every envelope produced by the direct call should appear in the wrapped result.
+            Assert.That(wrapped.Count, Is.EqualTo(direct.Count));
+            Assert.That(
+                direct.Select(e => e.MonoisotopicMass).OrderBy(m => m),
+                Is.EquivalentTo(wrapped.Select(e => e.MonoisotopicMass).OrderBy(m => m)));
+        }
+
+        [Test]
+        public void MultipleDecon_SingleWrappedIsoDec_MatchesDirectIsoDec()
+        {
```
trishorts, line 151: **Minor** - Wrapped-vs-direct equivalence tested only by count for IsoDec and the scan overload

**Problem:**
`MultipleDecon_SingleWrappedIsoDec_MatchesDirectIsoDec` and `MultipleDecon_ScanOverload_MatchesSpectrumOverload` assert only `Is.EqualTo` on `.Count`. The Classic counterpart (`MultipleDecon_SingleWrappedClassic_MatchesDirectClassic`) additionally asserts mass-level equivalence via `Is.EquivalentTo` on `MonoisotopicMass`. A count-only check passes even if the wrapped path returns a completely different set of envelopes that happens to have the same cardinality.

**Impact:**
A defect where the wrapper forwards the wrong spectrum, range, or parameters — but still produces the same number of envelopes — would not be caught for the IsoDec wrap path or the `MsDataScan` overload.

**Suggested fix:**
Strengthen both tests to compare envelope content (e.g. `MonoisotopicMass` sequences via `Is.EquivalentTo`) as the Classic test already does, not just the count.

**Affected test scenarios:**
- MultipleDecon_SingleWrappedIsoDec_MatchesDirectIsoDec - single IsoDec wrapped - envelope masses equivalent to direct call, not just equal count
- MultipleDecon_ScanOverload_MatchesSpectrumOverload - scan vs spectrum overload - envelope masses equivalent, not just equal count

trishorts (dismissed):
Automated code review - 5 finding(s) at severity >= Minor.

pcruzparri (dismissed):

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
nbollis: Added `Assert.That(wrappedNarrow, Is.Not.Empty)` before the per-peak `foreach` loop so the range-forwarding assertion is guaranteed to execute even if the narrow range yields zero envelopes.

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
nbollis: Renamed to `MultipleDecon_ClassicAndIsoDec_CountIsExactlySumOfIndividual` and rewrote the comment to state that `Concat` performs no deduplication so the count is exactly the sum.

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
nbollis: Added `ArgumentNullException.ThrowIfNull` guard and `ArgumentException` for empty collections in the constructor. Tests added: `MultipleDeconParameters_NullParameters_ThrowsArgumentNullException` and `MultipleDeconParameters_EmptyParameters_ThrowsArgumentException`.

mzLib/MassSpectrometry/Deconvolution/Parameters/MultipleDeconParameters.cs
nbollis: Added a `volatile bool _decoyComputed` flag alongside `_decoyParams` so the null return (when sub-params don\'t support decoys) is cached like the success path. Test added: `MultipleDecon_ToDecoyParameters_FromFileSubParam_ReturnsNullOnRepeatedCalls` verifies consistent null returns without redundant recomputation.

mzLib/Test/MassSpectrometryTests/Deconvolution/TestMultipleDeconvolution.cs
nbollis: Added `Is.EquivalentTo` assertions on `MonoisotopicMass` sequences to both `MultipleDecon_SingleWrappedIsoDec_MatchesDirectIsoDec` and `MultipleDecon_ScanOverload_MatchesSpectrumOverload`, matching the existing stronger assertion pattern in the Classic variant.

PR #1069, opened 2026-05-13
Promote consensus mass-tracing types + add Ms1Feature writer for MassFeature

trishorts:
This PRis for a consensus mass-tracing classic deconvolution post-processor (per-charge trace grouping, off-by-one correction, cross-charge feature stitching). The writer serializes its output as a FLASHDeconv-style `_ms1.feature` file MetaMorpheus already reads — bridging consensus deconvolution into search with no new wire format. Off-by-one correction is now resolution-aware: it derives its window from per-trace mass scatter and splits resolvable co-grouped species such as deamidation into their own features instead of erasing them, which bottom-up search requires. Feature output is deterministic, inputs are guarded, and the promoted types ship with 100%-line-covered unit and performance tests.

---

- **Resolution-aware off-by-one correction (behaviour change).** `TraceCorrector.Correct` now returns `List<CorrectedTrace>` and derives its window from per-trace mass scatter (3σ via MAD, capped below half the isotope spacing) instead of a fixed ±0.05 Da. An envelope at a resolvable, non-isotope offset — e.g. deamidation (+0.98402 Da), which the old window snapped to the unmodified mass — is **split into its own trace and surfaced as a distinct feature**; when scatter is too large to resolve it from the 1.00335 Da isotope spacing it conservatively merges. Bottom-up search needs this (deamidation detection is a requirement). The only constants are the 3σ multiple and the physics constant 1.00335 — the window is data-derived, not hardwired. *Note: this supersedes the flat ±0.05 used for the benchmark numbers below, so those warrant an end-to-end re-run.*

## Summary

Two unrelated-looking but tightly-coupled goals in one PR:

1. **Promote the consensus mass-tracing post-processor** out of `Development.Deconvolution.ConsensusTracing` (research-only, see [NOTES.md on the consensus-tracing branch](https://github.com/trishorts/mzLib/blob/consensus-tracing/NOTES.md) for the 1300-line research log behind the design) into a production namespace `MassSpectrometry.Deconvolution.Consensus`. Per-charge trace grouping, median off-by-one correction, and cross-charge feature stitching are now available to any caller without going through the `Development` project.

2. **Add a writer that turns the consensus pipeline's output into a FLASHDeconv-style `_ms1.feature` file** via a static `Ms1FeatureFile.FromMassFeatures(...)` factory + a `MassFeature.ToMs1Feature(...)` extension method. The factory builds a populated `Ms1FeatureFile` instance; calling its existing `WriteResults(path)` emits the wire format that any `Ms1FeatureFile` consumer already understands — notably MetaMorpheus's `FromFileDeconvolutionParameters` from [mzLib PR 1064](https://github.com/smith-chem-wisc/mzLib/pull/1064), which the [MetaMorpheus ExternalMs1Features PR](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2650) has now wired into the precursor pipeline.

Combined effect: a consensus pipeline can persist its result to a file that MetaMorpheus already reads, with zero new mzLib types crossing into MetaMorpheus and zero new MetaMorpheus surface area. The previously-designed `Ms2PrecursorEmitter` + `Ms2ScanWithPrecursorCandidate` two-PR plan (see NOTES.md "Pending design — MS2 emitter" section) is **obviated** by this approach and can be dropped.

## Why now

The consensus pipeline (Classic-decon → trace grouping → off-by-one correction → cross-charge clustering → optional Phase 13 FastTree scoring) was developed over Phases 0–13 of the research arc with the types parked in `Development\DeconvolutionDevelopment\ConsensusTracing\`. The design has now stabilised: 82–84% recall against the independent R7 reference, mean Phase-13 CV AUC 0.68. NOTES.md flagged "WholeFile move of consensus-tracing types" as the next move once the design settled. With the MetaMorpheus FromFile path landing and the consensus pipeline ready to be consumed, this PR is that move plus the matching writer.

## Phase A: Type promotion (commit [`3d734cc1`](https://github.com/trishorts/mzLib/commit/3d734cc1))

Six types moved out of `Development.Deconvolution.ConsensusTracing` into new `MassSpectrometry.Deconvolution.Consensus`:

| Type | New file |
|---|---|
| `MassTrace` | `mzLib/MassSpectrometry/Deconvolution/Consensus/MassTrace.cs` |
| `MassTraceBuilder` (static) | `mzLib/MassSpectrometry/Deconvolution/Consensus/MassTraceBuilder.cs` |
| `CorrectedEnvelope` | `mzLib/MassSpectrometry/Deconvolution/Consensus/CorrectedEnvelope.cs` |
| `CorrectedTrace` | `mzLib/MassSpectrometry/Deconvolution/Consensus/CorrectedTrace.cs` |
| `TraceCorrector` (static, + `EnvelopeWeight` delegate) | `mzLib/MassSpectrometry/Deconvolution/Consensus/TraceCorrector.cs` |
| `MassFeature` | `mzLib/MassSpectrometry/Deconvolution/Consensus/MassFeature.cs` |
| `MassFeatureBuilder` (static, extracted from `Phase4_CrossChargeConsensus`) | `mzLib/MassSpectrometry/Deconvolution/Consensus/MassFeatureBuilder.cs` |

Each type now lives in its own file (matching the existing `MassSpectrometry/Deconvolution/Algorithms/`, `Parameters/`, `Scoring/` folder discipline). Doc comments updated to drop research-only references; method bodies and signatures are byte-identical to the research-namespace versions. No behaviour change.

The research scaffolding (`Phase*.cs` fixtures, `MakeConsensusSnips`, `LargeTestDataLocator`) stays on the `consensus-tracing` branch. That branch needs a rebase after this PR merges to point at the new namespace — a one-line `using MassSpectrometry.Deconvolution.Consensus;` per phase file plus deletion of the now-redundant source under `Development\`. Out of scope here.

## Phase B: `Ms1FeatureFile.FromMassFeatures` factory (commit [`61029402`](https://github.com/trishorts/mzLib/commit/61029402))

Two pieces:

**Extension method** in `Readers/ExternalResults/IndividualResultRecords/Ms1FeatureFromMassFeatureExtensions.cs`:

```csharp
public static Ms1Feature ToMs1Feature(
    this MassFeature feature,
    int sequentialId,
    int sampleId = 0,
    int fractionId = 0)
```

Field mapping (FLASHDeconv canonical schema):

| `Ms1Feature` field | Source |
|---|---|
| `SampleId` | caller (default 0) |
| `Id` | caller's `sequentialId` |
| `Mass` | `MassFeature.ConsensusMass` (intensity-weighted mean across constituent traces) |
| `Intensity` | `MassFeature.SummedIntensity` |
| `RetentionTimeBegin` | `MassFeature.RTStart` |
| `RetentionTimeEnd` | `MassFeature.RTEnd` |
| `RetentionTimeApex` | RT of the highest-intensity envelope on the highest-summed-intensity constituent trace ("apex of the dominant charge state at its most intense scan") |
| `IntensityApex` | intensity of that same envelope |
| `ChargeStateMin` / `ChargeStateMax` | `min(MassFeature.Charges)` / `max(MassFeature.Charges)` |
| `FractionIdMin` / `FractionIdMax` | caller's `fractionId` (default 0) |

**Static factory** added directly on `Ms1FeatureFile` in `Readers/ExternalResults/ResultFiles/Ms1FeatureFile.cs`:

```csharp
public static Ms1FeatureFile FromMassFeatures(
    IEnumerable<MassFeature> features,
    int sampleId = 0,
    int fractionId = 0,
    Software software = Software.FLASHDeconv)
```

Builds the records via `ToMs1Feature`, populates a fresh `Ms1FeatureFile`, returns it. The caller invokes the existing `.WriteResults(path)` to emit bytes — no new IO path, the existing reader's writer is reused. `Software` defaults to `FLASHDeconv` because that's the canonical schema; callers can override.

**No filter applied.** Every `MassFeature` becomes one row. Downstream consumers (MetaMorpheus's precursor `HashSet` dedupe + FDR machinery) handle noise. This was an explicit design decision — three alternatives (multi-charge only, score-thresholded, score-as-side-artifact) were considered and rejected for v1 to keep maximum recall and let downstream policy decide.

## Phase C: Round-trip tests (commit [`9bc26a27`](https://github.com/trishorts/mzLib/commit/9bc26a27))

Five NUnit cases in `Test/FileReadingTests/ExternalFileReading/TestMs1FeatureFromMassFeature.cs`:

| Test | Asserts |
|---|---|
| `ToMs1Feature_SingletonFeature_MapsAllFields` | One-trace, one-envelope feature locks the full field mapping for the trivial case. |
| `ToMs1Feature_MultiChargeFeature_ApexIsDominantTraceMaxEnvelope` | Two-charge, five-envelope feature confirms apex selection and that cross-charge consensus mass is the intensity-weighted mean of per-trace consensus masses. |
| `ToMs1Feature_SampleIdAndFractionIdHonoured` | Caller overrides for `Sample_ID` and `Fraction_ID` propagate to the row. |
| `FromMassFeatures_WriteThenRead_AllFieldsSurvive` | Three features → write → re-read with `FileReader.ReadFile<Ms1FeatureFile>` → field-by-field equality. Catches serialisation drift between writer and reader. |
| `FromMassFeatures_AssignsSequentialIdsFromZero` | Producer's internal `MassFeature.Id` is ignored; written rows get `0..N-1` regardless of upstream filtering holes. |

### Bug fixed while writing the tests

The original `FromMassFeatures` factory used `file.Results.Add(...)` in a loop. `ResultFile<T>.Results.get` calls `LoadResults()` whenever the list is empty; for an in-memory-constructed `Ms1FeatureFile` with no `FilePath`, that throws `ArgumentException` on the empty path. Reshaped the factory to build the list up front and assign once via the public setter, which bypasses the lazy-load. Inline comment in the factory explains why. The third round-trip test caught this before the writer shipped.

## Results

```
TestMs1FeatureFromMassFeature suite          : 5/5 pass
Broader sweep (Ms1Feature + FromFile + ...)   : 148/148 pass
Build (Release|x64, MassSpectrometry + Readers): 0 errors, ~650 pre-existing warnings
```

The "broader sweep" covers `TestMsFeature`, `TestMs1FeatureFromMassFeature` (new), `TestFromFileDeconvolution`, `TestResultFile`, `TestSupportedFileExtensions`, `TestDinosaurTsv` — every test fixture that touches the `Ms1Feature` / `FromFile` / supported-file-extensions surface.

## Files changed

| File | +/− | Change |
|---|---|---|
| `mzLib/MassSpectrometry/Deconvolution/Consensus/MassTrace.cs` | +37 | new |
| `mzLib/MassSpectrometry/Deconvolution/Consensus/MassTraceBuilder.cs` | +91 | new |
| `mzLib/MassSpectrometry/Deconvolution/Consensus/CorrectedEnvelope.cs` | +21 | new |
| `mzLib/MassSpectrometry/Deconvolution/Consensus/CorrectedTrace.cs` | +31 | new |
| `mzLib/MassSpectrometry/Deconvolution/Consensus/TraceCorrector.cs` | +124 | new |
| `mzLib/MassSpectrometry/Deconvolution/Consensus/MassFeature.cs` | +61 | new |
| `mzLib/MassSpectrometry/Deconvolution/Consensus/MassFeatureBuilder.cs` | +90 | new |
| `mzLib/Readers/ExternalResults/IndividualResultRecords/Ms1FeatureFromMassFeatureExtensions.cs` | +65 | new |
| `mzLib/Readers/ExternalResults/ResultFiles/Ms1FeatureFile.cs` | +50 / −1 | added `FromMassFeatures` static + `using MassSpectrometry.Deconvolution.Consensus;` |
| `mzLib/Test/FileReadingTests/ExternalFileReading/TestMs1FeatureFromMassFeature.cs` | +260 | new |

Total: **10 files, +830 / −3** across 3 commits.

## How a consumer uses this

```csharp
using MassSpectrometry.Deconvolution.Consensus;
using Readers;

// Caller has a list of MassFeature objects from the consensus pipeline.
// Finalise() must have been called on each.
IEnumerable<MassFeature> features = RunConsensusPipeline(...);

// Build and write the .ms1.feature file.
Ms1FeatureFile.FromMassFeatures(features)
    .WriteResults(@"E:\path\to\my_sample_ms1.feature");

// Place the .ms1.feature file next to the corresponding raw/mzML and
// MetaMorpheus's auto-discovery (PR #2650) picks it up automatically.
```

End-to-end on the yeast top-down test data (per the [MetaMorpheus PR](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2650) results table): adding the FromFile source to the today-baseline classic pipeline yields **+44% PSMs and +10% protein groups at 1% FDR**.

## Architectural notes

- **The writer is decoupled from the producer.** It takes `IEnumerable<MassFeature>`; whoever produces the features (the consensus pipeline today, possibly other producers in the future) doesn't matter.
- **No new wire format.** The output is exactly the same FLASHDeconv-style `_ms1.feature` schema that `Ms1FeatureFile.WriteResults` already emits. The writer just constructs the records in memory.
- **The MS2 pairing happens inside `FromFileDeconvolutionAlgorithm`** (mzLib PR 1064), which the MetaMorpheus FromFile path invokes per MS2 scan. No new pairing code in this PR.
- **The `Ms2PrecursorEmitter` + `Ms2ScanWithPrecursorCandidate` design** (see consensus-tracing branch's NOTES.md, "Pending design — MS2 emitter") is **obviated** by this PR: file-based integration achieves the same end-to-end behaviour without the new types crossing repos.

## Out of scope

- **Filter policy.** No quality filter is applied (every `MassFeature` becomes a row). If a downstream consumer wants score-based filtering, that's a separate change — either at the producer (filter the `MassFeature` list before calling `FromMassFeatures`) or downstream (filter `Ms1Feature` records after `LoadResults`).
- **Phase 13 FastTree scoring** (mean CV AUC 0.68 per snippet, NOTES.md Phase 13). Productising the trained model is separate work; the writer carries no score column. If a follow-up wants to add one, the cleanest path is probably a sibling `.scores.tsv` file written alongside, leaving `_ms1.feature` schema unchanged.
- **`CorrectedTrace` implementing `ISingleChargeMs1Feature`.** Would enable an in-memory path (skipping the disk round-trip), which the file-based integration intentionally avoids. Possible follow-up if any consumer ever wants it.
- **Rebase of the `consensus-tracing` research branch** to consume the new namespace. The research Phase fixtures still compile against `Development.Deconvolution.ConsensusTracing`; that branch needs `using MassSpectrometry.Deconvolution.Consensus;` added and the duplicate source files in `Development\` deleted, but that work belongs on `consensus-tracing`, not here.

## Test plan

- [x] All 6 promoted types compile clean in the new namespace (`MassSpectrometry.csproj` build, 0 errors).
- [x] Reader project builds with the new `Ms1FeatureFromMassFeatureExtensions` and updated `Ms1FeatureFile` (`Readers.csproj` build, 0 errors).
- [x] 5/5 NUnit cases in `TestMs1FeatureFromMassFeature` pass.
- [x] 148/148 cases pass across the broader Ms1Feature / FromFile / supported-file-extensions / DinosaurTsv surface — no regression.
- [x] Bug in initial `FromMassFeatures` factory (lazy-load via `.Results.Add` on empty FilePath) caught by the round-trip test and fixed before commit.
- [ ] Land [trishorts/mzLib `ce24771c` + `3aa2fe04`](https://github.com/trishorts/mzLib/commits/master) upstream first (schema aliases for newer TopFD `_ms1.feature` files — strictly speaking unrelated to this PR's writer, but anyone reproducing the MetaMorpheus integration needs them too).
- [ ] Rebase `consensus-tracing` branch on top of this PR once merged; update Phase fixtures to use `MassSpectrometry.Deconvolution.Consensus`; delete the now-duplicate source under `Development\DeconvolutionDevelopment\ConsensusTracing\`.

## Companion PR

[smith-chem-wisc/MetaMorpheus#2650](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2650) (`ExternalMs1Features`) is the consumer side. It wires `Ms1FeatureFilePath` through `FileSpecificParameters` and uses `FromFileDeconvolutionParameters` (from mzLib PR 1064) to pair the features with MS2 scans. That PR's CI is currently red against `smith-chem-wisc/MetaMorpheus`'s pinned mzLib, but will go green once mzLib publishes a version containing PR 1064 + the schema aliases.


mzLib/Readers/ExternalResults/ResultFiles/Ms1FeatureFile.cs
```diff
@@ -14,6 +15,13 @@ public class Ms1FeatureFile : ResultFile<Ms1Feature>, IResultFile, IMs1FeatureFi
 
         public sealed override Software Software { get; set; }
 
+        // Records supplied by the in-memory factory (FromMassFeatures). When set,
+        // WriteResults writes these directly instead of going through the base Results
+        // getter, whose lazy-load would try to read from the empty FilePath of a
+        // factory-built file (crashing for a zero-feature input). Null for files read
+        // from disk, which load on demand as usual.
+        private List<Ms1Feature> _factoryRecords;
```
nbollis: This concerns me, why not have the factor method simply set the results so we do not have two sources of truth? 

A simple fix would be to add a File.Exists to the lazy load check in the base class

mzLib/MassSpectrometry/Deconvolution/Consensus/TraceCorrector.cs
```diff
@@ -0,0 +1,211 @@
...
+    ///   - intensity (TotalIntensity per envelope) -- trusts intense scans.
+    ///   - scorer (e.g., DeconvolutionScorer.ScoreEnvelope) -- trusts well-shaped envelopes.
+    /// </summary>
+    public static class TraceCorrector
+    {
+        /// <summary>
+        /// Average mass spacing between adjacent C12/C13 isotope peaks. An off-by-one
+        /// anchor error shifts the reported monoisotope by an integer multiple of this.
+        /// </summary>
+        public const double IsotopeSpacingDa = 1.00335;
```
nbollis: Reference an existing constant instead of redefining it. 

Even better yet would be to pass it in upon construction so that the decoy deconvolution mode you have built could also be used for decoy peak traces.

nbollis (commented):
Looks pretty good. One concern should be addressed in the handling of the ms1 feature file and it's record storage. 

I see a bunch of new code inside of mass spec/deconvolution. Consider if any of our existing classes for peak tracing (such as the XIC section) and our interfaces can be reused here. It would be great if the mass trace builder could operate on IHasSingle... instead of this custom object. It would leave more door open in the future. That said, I would be happy with a pass that conciders the objects/interfaces we have and tell's me why it is a bad idea to reuse them here.

trishorts:
Thanks for the push to reuse the existing peak-tracing code — I took the pass you asked for. The candidate is exactly the XIC stack: `ExtractedIonChromatogram` + `MassIndexingEngine`/`IndexingEngine.GetAllXics`, with `IndexedMass` (the `IIndexedPeak` you were thinking of) as a ready-made envelope→peak bridge. To test it properly I wrote a variant generator that swaps **only** the grouping — `GetAllXics` in place of `MassTraceBuilder` — and keeps the entire downstream (`TraceCorrector` → `MassFeatureBuilder` → writer) byte-identical, reading each envelope's mass/intensity/charge straight off `IndexedMass.IsotopicEnvelope`. So any difference is attributable to the grouping algorithm alone.

**Yield (the decisive metric).** Identical MM FromFile-only pipeline, same 19 Jurkat top-down files, q≤0.01:

| grouping | proteoforms | PSMs | protein groups |
|---|---|---|---|
| `MassTraceBuilder` | **1,126** | 39,656 | 146 |
| `GetAllXics` | **927** | 32,518 | 118 |

Reusing `GetAllXics` **loses ~18% of proteoforms**. It fragments into more, shorter traces (162k vs 138k on one fraction, +38% singletons) because of apex-intensity seeding + a hard claimed-peak set vs. the consensus tracer's scan-order, fixed-anchor grouping — and that fragmentation propagates into worse off-by-one correction and cross-charge stitching.

**Robustness.** `GetAllXics` actually *throws* (`ArgumentException`, duplicate key) on 1 of the 20 files under the wide 1.5 Da window consensus tracing needs: the seed peak is added without the claimed-peak guard that follow-on peaks get (`IndexingEngine.cs:136-137`), so under a wide tolerance two in-range envelopes at a seed scan collide. That's a latent bug worth a separate issue regardless of this PR — flagging it here.

**Perf.** `GetAllXics` is ~2.6× faster but ~4× the allocations; not a deciding factor when it drops 18% of the IDs.

So I'd like to keep `MassTrace`/`MassTraceBuilder`: the grouping is purpose-built for this and measurably better at the metric that matters. A tuned `GetAllXics` might close the gap, but matching it (seed policy, claim-set, the dedup bug) is enough surgery that it's a fork, not reuse. Happy to revisit if you'd like.

The two inline comments are addressed: `Ms1FeatureFile` now relies on a `File.Exists` guard in the base `ResultFile.Results` getter and drops `_factoryRecords` (single source of truth); `TraceCorrector` now aliases `Constants.C13MinusC12` instead of redefining the spacing. The "pass the spacing in for decoy traces" idea is a nice one — I left it as a follow-up since `TraceCorrector` is static today.

trishorts:
Filed the `GetAllXics` seed-peak dedup bug mentioned above as #1074.

mzLib/Readers/ExternalResults/ResultFiles/Ms1FeatureFile.cs
trishorts: Done in 083d0867 — added a `File.Exists(FilePath)` guard to the base `ResultFile<T>.Results` getter and removed `_factoryRecords`; the factory now just sets `Results`, so there's a single source of truth. Verified safe across all readers (full FileReadingTests suite green).

mzLib/MassSpectrometry/Deconvolution/Consensus/TraceCorrector.cs
trishorts: Done in 9ffdb85d — `IsotopeSpacingDa` now aliases `Constants.C13MinusC12` instead of redefining 1.00335. Good call on passing it in so the decoy-decon mode could correct decoy traces; deferring that to a follow-up since `TraceCorrector` is static today.

Alexander-Sol (dismissed):
Really cool stuff, excited about the results. Code looks great. 

One request: I got a little lost in the sauce trying to figure out how the various envelopes, features, and traces relate to one another. It would be helpful to have the hierarchy defined in a high-level overview comment somewhere.

nbollis (dismissed):

pcruzparri (dismissed):

pcruzparri (approved):

nbollis (approved):

PR #1070, opened 2026-05-18
Decon Param Equality: IEquatable on DeconParams and AverageResidue

nbollis:
This pull request introduces a standardized implementation of equality and cloning for deconvolution parameter classes and average residue models in the mass spectrometry codebase. By implementing `IEquatable<T>`, overriding `Equals`, `GetHashCode`, and adding `Clone` methods, the code now supports robust value-based comparisons and deep copying of parameter objects. This will help with testing, caching, and use in collections, and makes the codebase more maintainable and less error-prone.

Key changes include:

### Equality and Hashing for Parameter Classes

* Added `IEquatable<T>` implementations, along with `Equals`, `GetHashCode`, and protected helper methods (`EqualProperties`, `AddHashCodes`) to `DeconvolutionParameters` and all its derived classes, such as `ClassicDeconvolutionParameters`, `IsoDecDeconvolutionParameters`, `MultipleDeconParameters`, `FromFileDeconvolutionParameters`, and example/test/template classes. This ensures value-based equality and correct behavior in collections. [[1]](diffhunk://#diff-5fe7ce6846f5976234579fc08d46ff262d092a9238e0c4a5e4210bb211389d14R2-R7) [[2]](diffhunk://#diff-b835fe906b448c3dc997866be4c031a9dc399036ce7eeb40871c5f1b7487e112R31-R58) [[3]](diffhunk://#diff-37803275a695b6a864ce335345f55e4ae8011542e978d3b42d588aae2ee78620R230-R281) [[4]](diffhunk://#diff-ffae961f53898782bdf1c8dfd98e7c0c95e2c641508da119cf1aafee53e89c92R23-R49) [[5]](diffhunk://#diff-65869d87ffa59d40f9299ce95a35e06940abff0089bd26174be927ba48373382R109-R136) [[6]](diffhunk://#diff-bcbe594da92e3b8e99fb2d87b0c6f720f7a2400dc09688cfbd0dd32f08669f8aR18-R30) [[7]](diffhunk://#diff-584bfdced466e0b4272ccbcf96c0c0d351aa1d31294fc9a70d9a44f5a5bc9702R355-R363)

* Implemented `Clone` methods for all parameter classes, enabling deep copying of parameter objects, which is useful for thread safety and avoiding side effects. [[1]](diffhunk://#diff-b835fe906b448c3dc997866be4c031a9dc399036ce7eeb40871c5f1b7487e112R31-R58) [[2]](diffhunk://#diff-37803275a695b6a864ce335345f55e4ae8011542e978d3b42d588aae2ee78620R230-R281) [[3]](diffhunk://#diff-ffae961f53898782bdf1c8dfd98e7c0c95e2c641508da119cf1aafee53e89c92R23-R49) [[4]](diffhunk://#diff-65869d87ffa59d40f9299ce95a35e06940abff0089bd26174be927ba48373382R109-R136) [[5]](diffhunk://#diff-bcbe594da92e3b8e99fb2d87b0c6f720f7a2400dc09688cfbd0dd32f08669f8aR18-R30) [[6]](diffhunk://#diff-584bfdced466e0b4272ccbcf96c0c0d351aa1d31294fc9a70d9a44f5a5bc9702R355-R363)

### Equality and Hashing for Average Residue Models

* Implemented `IEquatable<AverageResidue>`, `Equals`, and `GetHashCode` for the `AverageResidue` base class and its derived classes (`EmpiricalAverageResidue`, `DecoyAveragine`), including support for value-based equality and hash codes. [[1]](diffhunk://#diff-d44215af6b73a429f2de4acf6fa4b10dea32eb73f52cadbe99e82c023038e671L1-R9) [[2]](diffhunk://#diff-d44215af6b73a429f2de4acf6fa4b10dea32eb73f52cadbe99e82c023038e671R18-R43) [[3]](diffhunk://#diff-cc609e2b41ff6ec41ff47edbfd1970f8beb019de56555d3a4d823323009fa190R91-R101) [[4]](diffhunk://#diff-33676cc533a506dcc2f444e6b096cc73958e94d99241560f840fc2811a6dc3bfR119-R135)

* Added an `AverageResidueComposition` property to `EmpiricalAverageResidue` and included it in equality and hash code calculations. [[1]](diffhunk://#diff-cc609e2b41ff6ec41ff47edbfd1970f8beb019de56555d3a4d823323009fa190R13-R14) [[2]](diffhunk://#diff-cc609e2b41ff6ec41ff47edbfd1970f8beb019de56555d3a4d823323009fa190R37) [[3]](diffhunk://#diff-cc609e2b41ff6ec41ff47edbfd1970f8beb019de56555d3a4d823323009fa190R91-R101)

### Miscellaneous

* Added missing `using` directives for `System` and `System.Linq` in several files to support new functionality. [[1]](diffhunk://#diff-d44215af6b73a429f2de4acf6fa4b10dea32eb73f52cadbe99e82c023038e671L1-R9) [[2]](diffhunk://#diff-33676cc533a506dcc2f444e6b096cc73958e94d99241560f840fc2811a6dc3bfL1-R5) [[3]](diffhunk://#diff-b835fe906b448c3dc997866be4c031a9dc399036ce7eeb40871c5f1b7487e112L1-R2) Fa72e3d4L1, [[4]](diffhunk://#diff-cc609e2b41ff6ec41ff47edbfd1970f8beb019de56555d3a4d823323009fa190L1-R2) [[5]](diffhunk://#diff-bcbe594da92e3b8e99fb2d87b0c6f720f7a2400dc09688cfbd0dd32f08669f8aL1-R1) [[6]](diffhunk://#diff-584bfdced466e0b4272ccbcf96c0c0d351aa1d31294fc9a70d9a44f5a5bc9702R1)
* Added an `OriginalFilePath` property to `FromFileDeconvolutionParameters` and included it in equality and hash code logic. [[1]](diffhunk://#diff-65869d87ffa59d40f9299ce95a35e06940abff0089bd26174be927ba48373382R40-R42) [[2]](diffhunk://#diff-65869d87ffa59d40f9299ce95a35e06940abff0089bd26174be927ba48373382R68-R69) [[3]](diffhunk://#diff-65869d87ffa59d40f9299ce95a35e06940abff0089bd26174be927ba48373382R109-R136)

These changes collectively improve the robustness and maintainability of the codebase, especially in scenarios involving parameter comparison, caching, and testing.

### DIFF
--- a/mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
@@ -1,9 +1,10 @@
 #nullable enable
+using System;
 using Chemistry;
 
 namespace MassSpectrometry
 {
-    public abstract class DeconvolutionParameters
+    public abstract class DeconvolutionParameters : IEquatable<DeconvolutionParameters>
     {
         public abstract DeconvolutionType DeconvolutionType { get; protected set; }
         public int MinAssumedChargeState { get; set; }
@@ -62,5 +63,44 @@ namespace MassSpectrometry
         /// subclasses keep dispatching via <see cref="DeconvolutionType"/>.
         /// </summary>
         public virtual DeconvolutionAlgorithm? CreateAlgorithm() => null;
+
+        #region IEquatable<DeconvolutionParameters>
+
+        public bool Equals(DeconvolutionParameters? other)
+        {
+            if (other is null) return false;
+            if (ReferenceEquals(this, other)) return true;
+            if (GetType() != other.GetType()) return false;
+            return MinAssumedChargeState == other.MinAssumedChargeState
+                && MaxAssumedChargeState == other.MaxAssumedChargeState
+                && Polarity == other.Polarity
+                && ExpectedIsotopeSpacing.Equals(other.ExpectedIsotopeSpacing)
+                && UseGenericScore == other.UseGenericScore
+                && AverageResidueModel.GetType() == other.AverageResidueModel.GetType()
+                && EqualProperties(other);
+        }
+
+        public override bool Equals(object? obj) => Equals(obj as DeconvolutionParameters);
+
+        public override int GetHashCode()
+        {
+            var hash = new HashCode();
+            hash.Add(MinAssumedChargeState);
+            hash.Add(MaxAssumedChargeState);
+            hash.Add(Polarity);
+            hash.Add(ExpectedIsotopeSpacing);
+            hash.Add(UseGenericScore);
+            hash.Add(AverageResidueModel.GetType());
+            AddHashCodes(hash);
+            return hash.ToHashCode();
+        }
+
+        protected abstract bool EqualProperties(DeconvolutionParameters other);
+
+        protected abstract void AddHashCodes(HashCode hash);
+
+        #endregion
+
+        public abstract DeconvolutionParameters Clone();
     }
 }
--- a/mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
@@ -1,5 +1,6 @@
 #nullable enable
 using System;
+using System.Collections.Generic;
 using Chemistry;
 
 namespace MassSpectrometry
@@ -76,7 +77,7 @@ namespace MassSpectrometry
                 && Polarity == other.Polarity
                 && ExpectedIsotopeSpacing.Equals(other.ExpectedIsotopeSpacing)
                 && UseGenericScore == other.UseGenericScore
-                && AverageResidueModel.Equals(other.AverageResidueModel)
+                && EqualityComparer<AverageResidue>.Default.Equals(AverageResidueModel, other.AverageResidueModel)
                 && EqualProperties(other);
         }
 
--- a/mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
@@ -76,7 +76,7 @@ namespace MassSpectrometry
                 && Polarity == other.Polarity
                 && ExpectedIsotopeSpacing.Equals(other.ExpectedIsotopeSpacing)
                 && UseGenericScore == other.UseGenericScore
-                && AverageResidueModel.GetType() == other.AverageResidueModel.GetType()
+                && AverageResidueModel.Equals(other.AverageResidueModel)
                 && EqualProperties(other);
         }
 
@@ -90,7 +90,7 @@ namespace MassSpectrometry
             hash.Add(Polarity);
             hash.Add(ExpectedIsotopeSpacing);
             hash.Add(UseGenericScore);
-            hash.Add(AverageResidueModel.GetType());
+            hash.Add(AverageResidueModel);
             AddHashCodes(hash);
             return hash.ToHashCode();
         }
--- a/mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
@@ -253,7 +253,8 @@ public class IsoDecDeconvolutionParameters : DeconvolutionParameters
         hash.Add(MinAreaCovered);
         hash.Add(DataThreshold);
         hash.Add(ReportMulitpleMonoisos);
-        hash.Add(MzWindow.GetHashCode());
+        foreach (var mz in MzWindow)
+            hash.Add(mz);
     }
 
     public override IsoDecDeconvolutionParameters Clone()
--- a/mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
@@ -1,4 +1,6 @@
 #nullable enable
+using System;
+using System.Linq;
 using System.Runtime.InteropServices;
 using Chemistry;
 
@@ -224,6 +226,58 @@ public class IsoDecDeconvolutionParameters : DeconvolutionParameters
         if (Polarity == Polarity.Negative)
             AdductMass = (float)-1.00727276467;
     }
+
+    #region IEquatable<IsoDecDeconvolutionParameters>
+
+    protected override bool EqualProperties(DeconvolutionParameters other)
+    {
+        var o = (IsoDecDeconvolutionParameters)other;
+        return PhaseRes == o.PhaseRes
+            && CssThreshold.Equals(o.CssThreshold)
+            && MatchTolerance.Equals(o.MatchTolerance)
+            && MaxShift == o.MaxShift
+            && KnockdownRounds == o.KnockdownRounds
+            && MinAreaCovered.Equals(o.MinAreaCovered)
+            && DataThreshold.Equals(o.DataThreshold)
+            && ReportMulitpleMonoisos == o.ReportMulitpleMonoisos
+            && MzWindow.SequenceEqual(o.MzWindow);
+    }
+
+    protected override void AddHashCodes(HashCode hash)
+    {
+        hash.Add(PhaseRes);
+        hash.Add(CssThreshold);
+        hash.Add(MatchTolerance);
+        hash.Add(MaxShift);
+        hash.Add(KnockdownRounds);
+        hash.Add(MinAreaCovered);
+        hash.Add(DataThreshold);
+        hash.Add(ReportMulitpleMonoisos);
+        hash.Add(MzWindow.GetHashCode());
+    }
+
+    public override IsoDecDeconvolutionParameters Clone()
+    {
+        var clone = new IsoDecDeconvolutionParameters(
+            polarity: Polarity,
+            phaseRes: PhaseRes,
+            reportMultipleMonoisos: ReportMulitpleMonoisos,
+            cssThreshold: CssThreshold,
+            matchTolerance: MatchTolerance,
+            maxShift: MaxShift,
+            mzWindow: (float[])MzWindow.Clone(),
+            knockdownRounds: KnockdownRounds,
+            minAreaCovered: MinAreaCovered,
+            relativeDataThreshold: DataThreshold,
+            averageResidueModel: AverageResidueModel,
+            expectedIsotopeSpacing: ExpectedIsotopeSpacing)
+        {
+            UseGenericScore = UseGenericScore
+        };
+        return clone;
+    }
+
+    #endregion
 
     // Thread-safe lazy caching of decoy parameters using double-checked locking.
     // See ClassicDeconvolutionParameters for the full rationale: ??= alone is not
--- a/mzLib/Omics/EmpiricalAverageResidue.cs
+++ b/mzLib/Omics/EmpiricalAverageResidue.cs
@@ -1,4 +1,5 @@
-﻿using Chemistry;
+using System;
+using Chemistry;
 using MassSpectrometry;
 using Omics.Digestion;
 using Omics.Modifications;
@@ -9,6 +10,8 @@ public record struct AverageResidueComposition(double C, double H, double O, dou
 
 public class EmpiricalAverageResidue : AverageResidue
 {
+    public AverageResidueComposition Composition { get; }
+
     public readonly double[][] AllMasses = new double[NumAveraginesToGenerate][];
     public readonly double[][] AllIntensities = new double[NumAveraginesToGenerate][];
     public readonly double[] MostIntenseMasses = new double[NumAveraginesToGenerate];
@@ -31,6 +34,7 @@ public class EmpiricalAverageResidue : AverageResidue
 
     public EmpiricalAverageResidue(AverageResidueComposition composition)
     {
+        Composition = composition;
         for (int i = 0; i < NumAveraginesToGenerate; i++)
         {
             double averagineMultiplier = (i + 1) / 4.0;
@@ -83,5 +87,16 @@ public class EmpiricalAverageResidue : AverageResidue
         double averageSe = totalChemicalFormula.CountWithIsotopes(PeriodicTable.GetElement("Se")) / (double)totalResidues;
 
         return new AverageResidueComposition(averageC, averageH, averageO, averageN, averageP, averageS, averageSe);
+    }
+
+    protected override bool EqualProperties(AverageResidue other)
+    {
+        var o = (EmpiricalAverageResidue)other;
+        return Composition.Equals(o.Composition);
+    }
+
+    protected override void AddHashCodes(HashCode hash)
+    {
+        hash.Add(Composition);
     }
 }
--- a/mzLib/Omics/EmpiricalAverageResidue.cs
+++ b/mzLib/Omics/EmpiricalAverageResidue.cs
@@ -8,6 +8,14 @@ namespace Omics;
 
 public record struct AverageResidueComposition(double C, double H, double O, double N, double P, double S, double Se);
 
+/// <summary>
+/// Average residue model derived from empirical data. Equality is composition-based:
+/// two instances constructed from equal <see cref="Composition"/> values compare equal,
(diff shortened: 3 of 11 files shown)


mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
```diff
@@ -101,6 +106,34 @@ internal int FindFirstIndexAtOrAbove(double minMz)
...
+
+        protected override bool EqualProperties(DeconvolutionParameters other)
+        {
+            var o = (FromFileDeconvolutionParameters)other;
+            if (Features.Count != o.Features.Count) return false;
+            if (OriginalFilePath != o.OriginalFilePath) return false;
+            return true;
+        }
+
+        protected override void AddHashCodes(HashCode hash)
```
trishorts, line 119: **Critical** - Clone() does not preserve OriginalFilePath — cloned object is not equal to source

**Problem:**
`Clone()` constructs the new instance via the `internal FromFileDeconvolutionParameters(IEnumerable<ISingleChargeMs1Feature>, ...)` overload, which never sets `OriginalFilePath`. As a result, the cloned object's `OriginalFilePath` is always `null` (or empty), while the source's is the original file path. `EqualProperties` includes `OriginalFilePath` in its check, so `original.Equals(original.Clone())` will return `false` whenever the source was constructed via the public file-path constructor.

**Impact:**
Breaks the fundamental contract that a cloned object should be equal to its source. Any caching, deduplication, or "has this changed?" check downstream that relies on `Equals` after a `Clone` will malfunction — exactly the scenarios the PR claims to enable. Hash codes will likewise differ, so cloned instances collide in dictionaries/sets in surprising ways.

**Suggested fix:**
After invoking the internal constructor in `Clone()`, propagate `OriginalFilePath` to the new instance (the setter is already accessible). Alternatively, expose a private/protected ctor that accepts the path and pre-loaded features so `Clone` can populate both atomically.

**Affected test scenarios:**
- Clone_RoundTrip_Equals_Source — clone a FromFileDeconvolutionParameters built from a real feature file; original.Equals(clone) must be true and hash codes must match.
- Clone_PreservesOriginalFilePath — assert clone.OriginalFilePath (via reflection or a test-only accessor) equals source path.
- Equals_AfterClone_UsableAsDictionaryKey — insert original as key, retrieve with clone; must succeed.

mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
```diff
@@ -101,6 +106,34 @@ internal int FindFirstIndexAtOrAbove(double minMz)
             return idx < 0 ? ~idx : idx;
         }
 
+        #region IEquatable<FromFileDeconvolutionParameters>
+
```
trishorts, line 110: **Major** - EqualProperties compares only Features.Count, not the actual features

**Problem:**
`EqualProperties` checks `Features.Count` and `OriginalFilePath` but never compares the actual `ISingleChargeMs1Feature` contents. Two instances constructed via the internal constructor (used by tests and `Clone`) with totally different features but the same count and identical `OriginalFilePath` (both null) will be reported as equal. Even via the public path, two parameter objects loaded from two files that happen to have the same path string and same count will be considered equal regardless of actual feature data, which the path-based identity does not fully justify because the file can change on disk.

**Impact:**
Two semantically distinct parameter objects compare equal, which is exactly the failure mode that caching/dedup/test-assertion use cases must avoid. Tests written against this `Equals` will pass on accidentally-matching counts and silently fail to catch regressions where the wrong feature set was loaded.

**Suggested fix:**
Either (a) compare the feature sequences element-wise (e.g., `Features.SequenceEqual(o.Features)`, optionally using a value-based comparer), or (b) document explicitly that equality is identity-by-OriginalFilePath and reject the internal-constructor case (e.g., return `false` or throw) when `OriginalFilePath` is null on either side. Whichever approach is chosen, the equality contract should be stated in an XML doc on the class.

**Affected test scenarios:**
- Equals_DifferentFeaturesSameCount_ReturnsFalse — build two instances via the internal ctor with different features but identical count; assert not equal.
- Equals_SameFilePathDifferentFeatureLists_BehavesAsDocumented — pin down whichever semantic is chosen.
- GetHashCode_DistinctFeatureSets_ProducesDistinctHashes — when feature contents differ, hashes should differ (low collision rate, not strict guarantee).

mzLib/Omics/EmpiricalAverageResidue.cs
```diff
@@ -84,4 +88,15 @@ private static AverageResidueComposition GetAverageChemicalFormula(IEnumerable<I
 
         return new AverageResidueComposition(averageC, averageH, averageO, averageN, averageP, averageS, averageSe);
     }
+
+    protected override bool EqualProperties(AverageResidue other)
```
trishorts, line 100: **Major** - EqualProperties casts without type check; throws InvalidCastException on mixed types

**Problem:**
`EqualProperties` performs an unchecked cast `(EmpiricalAverageResidue)other`. If the base-class `Equals` calls into derived `EqualProperties` after only confirming `other is AverageResidue` (e.g., another derived type like `DecoyAveragine`), this cast will throw `InvalidCastException`. The override does not first verify that `other` is the same runtime type, and relies entirely on the base contract to filter types. If the base only checks `GetType() == other.GetType()` this is fine, but if it only checks the abstract type, this throws.

**Impact:**
Comparing an `EmpiricalAverageResidue` with a `DecoyAveragine` (or any other `AverageResidue`-derived type) via `Equals` would throw rather than returning `false`. This violates the documented `IEquatable<T>.Equals` contract that mandates returning `false` for different types and breaks collections (`HashSet`, `Dictionary`) containing mixed residue types.

**Suggested fix:**
Either guarantee in the base `Equals` that `EqualProperties` is only invoked when `GetType() == other.GetType()`, or have each derived `EqualProperties` use a safe `as`/`is` pattern returning `false` when types differ.

**Affected test scenarios:**
- Equals_DifferentDerivedTypes - comparing `EmpiricalAverageResidue` to `DecoyAveragine` - returns false without throwing
- Equals_NullOther - returns false without throwing
- HashSet_MixedResidueTypes - both instances coexist and lookups behave correctly

mzLib/Omics/EmpiricalAverageResidue.cs
```diff
@@ -9,6 +10,8 @@ public record struct AverageResidueComposition(double C, double H, double O, dou
 
 public class EmpiricalAverageResidue : AverageResidue
 {
+    public AverageResidueComposition Composition { get; }
+
```
trishorts, line 22: **Major** - Public `Composition` property duplicates pre-existing `AverageResidueComposition` semantics, missing in PR description claim

**Problem:**
The PR description claims an `AverageResidueComposition` property was added to `EmpiricalAverageResidue` and included in equality/hash. The diff actually adds a property named `Composition` (of type `AverageResidueComposition`). Equality uses `Composition.Equals(o.Composition)`, which is correct for the record struct, but downstream consumers expecting a property named `AverageResidueComposition` (per the PR write-up) will not find one. This is a naming/contract mismatch between the PR description and the implementation.

**Impact:**
External tests/docs/callers written to the PR description's name (`AverageResidueComposition`) will fail to compile or won't bind. Reviewers and future maintainers will have trouble reconciling the description with the actual surface.

**Suggested fix:**
Either rename the property to `AverageResidueComposition` to match the PR description and the type name, or update the PR description to reflect the actual `Composition` name. Pick one source of truth and align.

**Affected test scenarios:**
- EmpiricalAverageResidue_PropertyName - reflection lookup for `AverageResidueComposition` - resolves to a real property
- EmpiricalAverageResidue_EqualsByComposition - two instances built from equal compositions - are equal

mzLib/Omics/EmpiricalAverageResidue.cs
```diff
@@ -9,6 +10,8 @@ public record struct AverageResidueComposition(double C, double H, double O, dou
...
     public readonly double[][] AllMasses = new double[NumAveraginesToGenerate][];
```
trishorts, line 23: **Major** - Equality ignores instance fields containing actual computed isotopic data

**Problem:**
`EqualProperties` only compares `Composition`. However, the class also exposes mutable public `readonly` array fields `AllMasses`, `AllIntensities`, `MostIntenseMasses`, and `DiffToMonoisotopic` (declared `readonly` on the reference, but the arrays' contents are mutable). Two instances constructed from the same composition will hash and compare equal, but if a consumer mutates one instance's `AllMasses[i]` afterwards, equality will silently report `true` despite divergent state. Conversely, if a future constructor path bypasses the `composition`-based constructor, two semantically different instances could compare equal.

**Impact:**
Equality is a "constructor input" comparison rather than a state comparison. Caching or de-duplication logic relying on this `Equals` will incorrectly treat mutated instances as identical, leading to subtle bugs when these objects are used as dictionary keys or in `HashSet`.

**Suggested fix:**
Document the equality contract as "by construction composition only" with a comment, OR include a representative state comparison (e.g., `MostIntenseMasses.SequenceEqual`). Also consider making the data fields properly immutable (private setters / `IReadOnlyList`).

**Affected test scenarios:**
- Equals_AfterMutatingAllMasses - two instances from same composition, mutate one - documents and asserts intended behavior
- HashCode_Stability - hash equals before and after mutation

mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
```diff
@@ -225,6 +227,58 @@ public IsoDecDeconvolutionParameters(
...
+            && KnockdownRounds == o.KnockdownRounds
+            && MinAreaCovered.Equals(o.MinAreaCovered)
+            && DataThreshold.Equals(o.DataThreshold)
+            && ReportMulitpleMonoisos == o.ReportMulitpleMonoisos
+            && MzWindow.SequenceEqual(o.MzWindow);
+    }
+
+    protected override void AddHashCodes(HashCode hash)
+    {
+        hash.Add(PhaseRes);
```
trishorts, line 248: **Major** - MzWindow hash uses array reference identity, breaking GetHashCode contract

**Problem:**
In `AddHashCodes`, the call `hash.Add(MzWindow.GetHashCode())` hashes the result of `Array.GetHashCode()`, which is reference-based and not value-based. `EqualProperties` compares `MzWindow` with `SequenceEqual` (value-based), so two equal `IsoDecDeconvolutionParameters` whose `MzWindow` arrays have identical contents but different references will report `Equals == true` but produce different hash codes. This violates the `Equals`/`GetHashCode` contract.

**Impact:**
Using these parameters as keys in `Dictionary`/`HashSet` will silently lose entries — equal instances land in different buckets. Caching keyed on parameter equality will misbehave, which is one of the stated motivations for the PR.

**Suggested fix:**
Hash each element of `MzWindow` individually (e.g. iterate and `hash.Add(element)`), or combine each element into the `HashCode`. Do not hash the array's reference identity.

**Affected test scenarios:**
- IsoDec_GetHashCode_EqualMzWindowDifferentReference_HashesMatch — two instances built with separately allocated `MzWindow` arrays holding identical values — `Assert.AreEqual(a.GetHashCode(), b.GetHashCode())`
- IsoDec_GetHashCode_DefaultConstruction_TwoInstancesHashEqually — `new IsoDecDeconvolutionParameters()` twice — same hash code

---

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
```diff
@@ -62,5 +63,44 @@ protected DeconvolutionParameters(int minCharge, int maxCharge,
...
+
+        public bool Equals(DeconvolutionParameters? other)
+        {
+            if (other is null) return false;
+            if (ReferenceEquals(this, other)) return true;
+            if (GetType() != other.GetType()) return false;
+            return MinAssumedChargeState == other.MinAssumedChargeState
+                && MaxAssumedChargeState == other.MaxAssumedChargeState
+                && Polarity == other.Polarity
+                && ExpectedIsotopeSpacing.Equals(other.ExpectedIsotopeSpacing)
```
trishorts, line 78: **Major** - Equals throws NullReferenceException when AverageResidueModel is null

**Problem:**
`Equals(DeconvolutionParameters? other)` calls `AverageResidueModel.Equals(other.AverageResidueModel)` unconditionally. While the base constructor defaults `AverageResidueModel` to `new Averagine()`, the property has a public setter, so a caller can legitimately assign `null` after construction. In that case `Equals` will throw `NullReferenceException` rather than returning a sensible bool. The same risk applies to `GetHashCode` calling `hash.Add(AverageResidueModel)` — `HashCode.Add` handles null, but `Equals` does not.

**Impact:**
Any equality comparison on a parameter object whose `AverageResidueModel` has been nulled out throws instead of returning false, which can break collections, caches, and assertions in tests.

**Suggested fix:**
Use null-tolerant comparison — e.g. `EqualityComparer<AverageResidue>.Default.Equals(AverageResidueModel, other.AverageResidueModel)` or an explicit null check before delegating to `.Equals`.

**Affected test scenarios:**
- Base_Equal_BothNullAverageResidueModel_AreEqual — set `AverageResidueModel = null` on two otherwise-equal instances, assert `Equals` returns true and does not throw
- Base_Equal_OneNullAverageResidueModel_AreNotEqual — one with null, one with default Averagine, assert `Equals` returns false and does not throw

---

trishorts (changes requested):
Automated code review - 7 finding(s) at severity >= Major.

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
nbollis: Fixed. Replaced AverageResidueModel.Equals(...) with EqualityComparer<AverageResidue>.Default.Equals(...) so Equals handles null AverageResidueModel without throwing.

mzLib/MassSpectrometry/Deconvolution/Parameters/IsoDecDeconvolutionParameters.cs
nbollis: Fixed. AddHashCodes now hashes each MzWindow element individually instead of using Array.GetHashCode(), so equal arrays with different references produce matching hash codes.

mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
nbollis: Fixed. Clone() now propagates OriginalFilePath via the object initializer so that original.Equals(original.Clone()) returns true when the source was constructed via the public file-path constructor.

mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
nbollis: Feature-by-feature comparison was intentionally not added. SingleChargeMs1Feature contents are deterministically loaded from the file identified by OriginalFilePath, so Count + OriginalFilePath is sufficient for the intended scenarios. Full element-wise comparison would be prohibitively expensive for no correctness gain when the same file path is used. The Clone() fix ensures OriginalFilePath is always propagated, closing the main gap.

mzLib/Omics/EmpiricalAverageResidue.cs
nbollis: Added an XML doc comment on EmpiricalAverageResidue stating that equality is composition-based -- two instances constructed from equal Composition values are equal regardless of derived array state, which is deterministically computed from Composition at construction time.

mzLib/Omics/EmpiricalAverageResidue.cs
nbollis: The base class AverageResidue.Equals already guards with GetType() != other.GetType() before calling EqualProperties, so the cast (EmpiricalAverageResidue)other is always safe. No code change needed.

mzLib/Omics/EmpiricalAverageResidue.cs
nbollis: The property is named Composition (of type AverageResidueComposition). The PR description will be updated to match, since the type name on the property already carries the full qualifier.

mzLib/MassSpectrometry/Deconvolution/Parameters/DeconvolutionParameters.cs
nbollis: Fixed. Replaced AverageResidueModel.Equals(...) with EqualityComparer<AverageResidue>.Default.Equals(...) so Equals handles null AverageResidueModel without throwing.

Alexander-Sol (approved):

trishorts (approved):

PR #1073, opened 2026-05-25
Adding remaining Koina models

pcruzparri:
## TLDR
Add 21 Koina prediction models, refactor model layer, introduce null/empty parameter convention
## Summary
Adds 21 new concrete Koina model implementations across fragment intensity, CCS, crosslink, detectability, and retention time categories. Refactors the shared model infrastructure: deduplicates request batching, adds neutral loss parsing, switches from positional to name-based output extraction, and introduces a three-tier null/empty/populated convention for parameter validation (collision energy, instrument type, fragmentation type). The spline models are not included in this PR due to 1) lack of necessity and 2) different output convention. 
### Fragment Intensity (16)
| Model | Koina Model Name | Required Inputs | Notes |
|---|---|---|---|
| Prosit2019Intensity | `Prosit_2019_intensity` | seq, charge, CE | CE any value |
| Prosit2020IntensityHCD | `Prosit_2020_intensity_HCD` | seq, charge, CE | CE any value |
| Prosit2020IntensityCID | `Prosit_2020_intensity_CID` | seq, charge | Fixed NCE=35 |
| Prosit2020IntensityTMT | `Prosit_2020_intensity_TMT` | seq, charge, CE, frag_type | Frag: `HCD`, `CID` |
| Prosit2023IntensityTimsTOF | `Prosit_2023_intensity_timsTOF` | seq, charge, CE | CE any value |
| Prosit2024IntensityCit | `Prosit_2024_intensity_cit` | seq, charge, CE, frag_type | Frag: `HCD`, `CID` |
| Prosit2024IntensityPTMsGl | `Prosit_2024_intensity_PTMs_gl` | seq, charge, CE, frag_type | Frag: `HCD`, `CID`; 34 UNIMODs |
| Prosit2025IntensityLac | `Prosit_2025_intensity_lac` | seq, charge, CE, frag_type, inst_type | Frag: `HCD`/`CID`; Inst: `eclipse`/`astral`/`lumos` |
| Prosit2025Intensity22PTM | `Prosit_2025_intensity_22PTM` | seq, charge, CE, frag_type | All UNIMOD accepted |
| Prosit2025Intensity40PTM | `Prosit_2025_intensity_40PTM` | seq, charge, CE, frag_type | All UNIMOD accepted |
| Prosit2025IntensityPtm2 | `Prosit_2025_intensity_ptm2` | seq, charge, CE, frag_type | Frag: `HCD`, `CID` |
| Altimeter2024Intensities | `Altimeter_2024_intensities` | seq, charge, CE (20–40) | CE range-restricted; `^` charge delimiter |
| UniSpec | `UniSpec` | seq, charge, CE, inst_type | Inst: `QE`/`QEHFX`/`LUMOS`/`ELITE`/`VELOS`/`NONE` |
| AlphaPeptDeepMs2Generic | `AlphaPeptDeep_ms2_generic` | seq, charge, CE, inst_type | Inst: `QE`/`LUMOS`/`TIMSTOF`/`SCIEXTOF` |
| Ms2Pip (7 models) | `ms2pip_*` | seq, charge | Fixed CE, no CE input |
### CCS (2)
| Model | Koina Model Name | Required Inputs | Notes |
|---|---|---|---|
| AlphaPeptDeepCCSGeneric | `AlphaPeptDeep_ccs_generic` | seq, charge | No charge constraint |
| IM2Deep | `IM2Deep` | seq, charge | No charge constraint |
### Crosslink (3)
| Model | Koina Model Name | Required Inputs | Notes |
|---|---|---|---|
| Prosit2023IntensityXLCMS2 | `Prosit_2023_intensity_XL_CMS2` | alpha_seq, beta_seq, charge, CE | CE any value |
| Prosit2024IntensityXLNMS2 | `Prosit_2024_intensity_XL_NMS2` | alpha_seq, beta_seq, charge, CE | CE any value |
| Prosit2023IntensityXLCMS3 | `Prosit_2023_intensity_XL_CMS3` | seq (alpha slot), charge | Fixed NCE=35; single sequence |
### Detectability (1)
| Model | Koina Model Name | Required Inputs | Notes |
|---|---|---|---|
| PFly2024FineTuned | `pfly_2024_fine_tuned` | seq only | No modifications |
### Retention Time (10, 9 new + existing Prosit2019iRT updated)
| Model | Koina Model Name | Required Inputs | Notes |
|---|---|---|---|
| Prosit2019iRT | `Prosit_2019_irt` | seq only | iRT |
| Prosit2020iRTTMT | `Prosit_2020_irt_TMT` | seq only | iRT; requires N-term TMT/iTRAQ |
| Prosit2024iRTCit | `Prosit_2024_irt_cit` | seq only | iRT |
| Prosit2024iRTPTMsGl | `Prosit_2024_irt_PTMs_gl` | seq only | iRT; 34 UNIMODs |
| Prosit2025iRTLac | `Prosit_2025_irt_lac` | seq only | iRT |
| Prosit2025iRT22PTM | `Prosit_2025_irt_22_PTM` | seq only | iRT; all UNIMOD |
| Prosit2025iRT40PTM | `Prosit_2025_irt_40_PTM` | seq only | iRT; all UNIMOD |
| AlphaPeptDeepRTGeneric | `AlphaPeptDeep_rt_generic` | seq only | iRT; seq up to 500 |
| DeeplcHelaHf | `Deeplc_hela_hf` | seq only | iRT |
| ChronologerRT | `Chronologer_RT` | seq only | **Absolute RT** (not iRT) |
### Null/empty convention for Allowed* parameter collections
`AllowedCollisionEnergies`, `AllowedInstrumentTypes`, and `AllowedFragmentationTypes` changed from `HashSet<T>` to `HashSet<T>?`. Previously an empty set meant "bypass validation entirely." Now:
| Collection state | Meaning |
|---|---|
| `null` | Parameter **not applicable** — validation skipped |
| `{}` (empty) | Parameter IS **required** — must be non-null, any value accepted |
| `{a, b, c}` | Parameter IS **required** — value must be in the set |
Models that don't send a parameter to Koina have its allowed set = `null` (e.g., Ms2Pip models for CE). Models that do send to Koina use `{}` (Prosit2019-2025 intensity models, UniSpec, AlphaPeptDeep) or a populated set (Altimeter with range 20–40, Prosit TMT/Cit/Lac/PTM with specific fragmentation types).
### Existing models updated
- `PFly2024FineTuned`: namespace fix; null-forgiving operator in `ToBatchedRequests`
- `Prosit2019iRT`, `Prosit2020iRTTMT`: updated to use new base infrastructure
- `RetentionTimeModel`: removed unnecessary `.ToArray()` call
## Infrastructure Changes
- **`KoinaModelBase`**: new `BuildBatchedRequest` helper and `InputField` record eliminates duplicated request construction across 20+ models
- **Output extraction**: by output name (`"ccs"`, `"annotation"`, etc.) instead of positional `Outputs[0]` — more resilient to API changes
- **Neutral loss**: `ParseFragmentAnnotation` handles NL suffixes (`-H2O`, `-NH3`); `BaseFragmentIdentifier` uses first-dash stripping
- **Thread-safety docs**: documented on all abstract model types
- **Switch exhaustiveness**: all `ParameterHandlingMode` switches include `default: throw`
- **Crosslink beta validation**: moved null-check into `ValidateModelSpecificInputs`


mzLib/PredictionClients/Koina/AbstractClasses/FragmentIntensityModel.cs
```diff
@@ -338,23 +461,36 @@ protected virtual List<PeptideFragmentIntensityPrediction> ResponseToPredictions
...
+                            try
+                            {
+                                parsed = ParseFragmentAnnotation(fragmentIon);
+                            }
+                            catch
+                            {
+                                continue;
+                            }
                             fragmentIons.Add(fragmentIon);
-                            fragmentMZs.Add(tpLookup[fragmentIon.Substring(0, plusIndex)].ToMz(fragmentCharge));
```
trishorts, line 383: **Critical** - MapToInputFullSequence adds annotation before lookup guard, desyncing parallel lists

**Problem:**
In `ResponseToPredictions`, the `FragmentIonMappingMode.MapToInputFullSequence` branch executes `fragmentIons.Add(fragmentIon);` and only afterward runs `if (!tpLookup.TryGetValue(parsed.BaseFragmentIdentifier, out var tp)) { continue; }`. When the base fragment identifier is not found in the input peptide's theoretical products, the loop `continue`s after the annotation was already appended but before `fragmentMZs.Add(...)` and `predictedIntensities.Add(intensity)` run. The three parallel lists (`fragmentIons`, `fragmentMZs`, `predictedIntensities`) that populate `PeptideFragmentIntensityPrediction` therefore drift out of index alignment: `fragmentIons` ends up longer than the other two. This is a regression introduced by the rewrite — the base version added all three together (it threw on a missing key via direct indexing rather than producing misaligned data).

**Impact:**
For every fragment whose base identifier is absent from the regenerated theoretical products (e.g., neutral-loss ions or ion series mzLib's default HCD fragmentation does not emit), all subsequent annotations in that peptide are paired with the wrong m/z and intensity — silent scientific data corruption in spectral output. Downstream, `GenerateLibrarySpectraFromPredictions` iterates `for (i < prediction.FragmentAnnotations.Count)` and indexes `prediction.FragmentIntensities[i]`, which now throws `ArgumentOutOfRangeException` because the intensity list is shorter than the annotation list.

**Suggested fix:**
Move `fragmentIons.Add(fragmentIon);` to after the successful `tpLookup.TryGetValue` so all three lists are appended together only when a theoretical product is found (mirroring the `MapToValidatedFullSequence` branch).

**Affected test scenarios:**
- ResponseToPredictions_MapToInputFullSequence - response contains an annotation whose base id is not in the input peptide's theoretical products - resulting prediction's FragmentAnnotations, FragmentMZs, FragmentIntensities all have equal Count and correct pairwise correspondence
- GenerateLibrarySpectraFromPredictions - prediction produced via MapToInputFullSequence with at least one unmappable fragment - completes without throwing and each peak's m/z matches its annotation
- ResponseToPredictions_MapToInputFullSequence - all fragments map cleanly - output unchanged vs. expected (regression guard)

mzLib/PredictionClients/Koina/SupportedModels/CrosslinkIntensityModels/Prosit2024IntensityXLNMS2.cs
```diff
@@ -0,0 +1,65 @@
...
+        public override string ModelName => "Prosit_2024_intensity_XL_NMS2";
+        public override int MaxBatchSize => 1000;
+        public override int MaxNumberOfBatchesPerRequest { get; init; }
+        public override int ThrottlingDelayInMilliseconds { get; init; }
+        public override int BenchmarkedTimeForOneMaxBatchSizeInMilliseconds => 1500;
+        public override int MaxPeptideLength => 30;
+        public override int MinPeptideLength => 1;
+        public override HashSet<int> AllowedPrecursorCharges => new() { 1, 2, 3, 4, 5, 6 };
+        public override HashSet<int>? AllowedCollisionEnergies => new HashSet<int>();
+        public override int NumberOfPredictedFragmentIons => 174;
```
trishorts: **Major** - NMS2 NumberOfPredictedFragmentIons=174 looks copy-pasted from single-sequence CMS3

**Problem:**
`Prosit2024IntensityXLNMS2` declares `NumberOfPredictedFragmentIons => 174`, the same value as the single-sequence `Prosit2023IntensityXLCMS3` (174). But NMS2 is structurally identical to `Prosit2023IntensityXLCMS2`: it takes *two* full sequences (`peptide_sequences_1` alpha + `peptide_sequences_2` beta), `precursor_charges`, and `collision_energies`, and CMS2 declares double that count (`=> 348`, i.e. 2 × 174 = two peptides' worth of b/y ions over 29 cleavage positions × 3 charges). A non-cleavable MS2 crosslink fragments both peptide backbones, so the predicted-ion dimension would be expected to scale with two peptides (≈348), not one. The author clearly hand-tuned the per-model crosslinker UNIMOD IDs (1896 / 1881 / 1898), which makes the duplicated 174 between the two-sequence NMS2 and the one-sequence CMS3 stand out as a likely oversight.

**Impact:**
`NumberOfPredictedFragmentIons` drives client-side reshaping of the flat Koina intensity array into per-fragment values. If Koina actually returns 348 values for `Prosit_2024_intensity_XL_NMS2`, declaring 174 will misalign intensities to fragment annotations (silently wrong scientific results) or throw on array indexing — in production XL-MS2 intensity prediction.

**Suggested fix:**
Verify the expected output dimension of `Prosit_2024_intensity_XL_NMS2` against the Koina model card / a live response. If NMS2 returns two-peptide output like CMS2, change the value to 348 (or whatever the response length actually is); if 174 is genuinely correct, add a brief comment documenting why a two-sequence model outputs a single peptide's worth of ions, so the next reader does not read it as a copy-paste bug.

**Affected test scenarios:**
- ToBatchedRequests/output-reshape for NMS2 - assert the reshaped per-fragment count equals the length of a real (or fixture) Koina response for `Prosit_2024_intensity_XL_NMS2`
- Cross-model dimension test - assert each crosslink model's `NumberOfPredictedFragmentIons` matches its actual Koina output length (catches CMS2=348, CMS3=174, NMS2=?) so future copy-paste between models is caught

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Prosit2020IntensityTMT.cs
```diff
@@ -0,0 +1,118 @@
...
+                    case SequenceConversionHandlingMode.ThrowException:
+                        throw new ArgumentException(message);
+                    case SequenceConversionHandlingMode.ReturnNull:
+                        warning = new WarningException(message);
+                        return null;
+                    case SequenceConversionHandlingMode.RemoveIncompatibleElements:
+                    case SequenceConversionHandlingMode.UsePrimarySequence:
+                        warning = new WarningException(message);
+                        return null;
+                }
```
trishorts, line 104: **Major** - TMT TryCleanSequence returns apiSequence on success instead of sanitized

**Problem:**
`TryCleanSequence` calls `base.TryCleanSequence(sequence, out apiSequence, out warning)` and captures the base return value in `sanitized`. On the base early-return path it correctly returns `sanitized`, but on the success path (label present) it ends with `return apiSequence;`. The two variables are distinct outputs of the base method — `sanitized` is the cleaned full/display sequence (the base method's documented return), while `apiSequence` is the Koina API-format string (e.g. `[UNIMOD:737]-...`). Returning `apiSequence` diverges from the base contract and from this method's own early-return branch.

**Impact:**
Every successfully validated TMT peptide gets the API-format string returned as its "cleaned" sequence. Since the default `FragmentIonMappingMode` is `MapToValidatedFullSequence`, downstream ion mapping would key off the API-format sequence rather than the validated full sequence, corrupting fragment mapping for all TMT predictions.

**Suggested fix:**
Return `sanitized` on the success path to match the base method's contract, keeping `apiSequence` solely as the out parameter.

**Affected test scenarios:**
- TryCleanSequence - valid TMT-labeled sequence - returned cleaned sequence equals the base sanitized full sequence, not the API-format string
- Fragment mapping - TMT peptide with N-term label - mapped products align to the validated full sequence

trishorts (changes requested):
Automated code review - 3 finding(s) at severity >= Major.

mzLib/PredictionClients/Koina/AbstractClasses/FragmentIntensityModel.cs
pcruzparri: Fixed by moving fragmentIons.Add(fragmentIon) to after the successful TryGetValue, so all three parallel lists are appended together only when a theoretical product is found. This mirrors the MapToValidatedFullSequence branch behavior.

mzLib/PredictionClients/Koina/SupportedModels/CrosslinkIntensityModels/Prosit2024IntensityXLNMS2.cs
pcruzparri: Changed NumberOfPredictedFragmentIons from 174 to 348.

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Prosit2020IntensityTMT.cs
pcruzparri: Fixed by returning sanitized on the success path, keeping apiSequence solely as the out parameter.

mzLib/PredictionClients/Koina/AbstractClasses/CrosslinkFragmentIntensityModel.cs
```diff
@@ -0,0 +1,428 @@
...
+        /// Whether this Koina model requires a separate beta sequence input.
+        /// Paired-sequence models (CMS2, NMS2) leave this true; single-sequence helpers
+        /// (CMS3) override to false and consume only <see cref="CrosslinkIntensityPredictionInput.AlphaSequence"/>.
+        /// </summary>
+        public virtual bool RequiresBetaSequence => true;
+        public override abstract IReadOnlySet<int> AllowedUnimodIds { get; }
+        public override abstract SequenceConversionHandlingMode ModHandlingMode { get; init; }
+        public abstract IncompatibleParameterHandlingMode ParameterHandlingMode { get; init; }
+        public abstract FragmentIonMappingMode FragmentIonMappingMode { get; init; }
+
```
nbollis, line 87: [review-oop] Crosslink mapping mode is configurable but never applied

The new crosslink base class adds a `FragmentIonMappingMode` contract and every concrete crosslink model now accepts that option, but this implementation never branches on it anywhere in the added class. As written, callers can request `MapToInputFullSequence`, yet crosslink predictions always behave like validated-sequence mapping because only a single response path exists.

Suggestion: Either implement distinct `MapToInputFullSequence` behavior in `CrosslinkFragmentIntensityModel`, or remove `FragmentIonMappingMode` from the crosslink hierarchy until the second mode actually exists.

mzLib/Test/KoinaTests/KoinaRequestBuildingTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            AssertBuildsBatches(Instantiate(modelType), inputs);
+        }
+
+        [TestCaseSource(nameof(CrosslinkModels))]
+        public void Crosslink_ToBatchedRequests_BuildsBatches(Type modelType)
+        {
+            var inputs = new List<CrosslinkIntensityPredictionInput>
+            {
+                new("PEPTIDEK[UNIMOD:1896]", "ACDEK[UNIMOD:1896]", 2, 30)
+                {
```
nbollis, line 92: [review-generic] Crosslink request-building smoke test bypasses the new model-specific UNIMOD validation

The added crosslink request-building test seeds `ValidatedAlphaSequence` and `ValidatedBetaSequence` directly with `UNIMOD:1896` values and then calls `ToBatchedRequests`. That means it never exercises the real XLNMS2 validation path introduced in this PR, even though `Prosit2024IntensityXLNMS2` now accepts only `{4, 35, 1898}`.

Suggestion: Add a non-network XLNMS2 test that goes through `TryCleanSequence`/`Predict` with one accepted `UNIMOD:1898` case and one rejected `UNIMOD:1896` case, instead of bypassing validation via pre-populated `Validated*Sequence` fields.

mzLib/Test/KoinaTests/KoinaRequestBuildingTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            {
+                new("PEPTIDEK") { ValidatedFullSequence = "PEPTIDEK" }
+            };
+            AssertBuildsBatches(Instantiate(modelType), inputs);
+        }
+
+        // ── Model-specific request-building overrides ───────────────────────────────
+
+        [Test]
+        public void Tmt_TryCleanSequence_RejectsSequenceWithoutNTerminalLabel()
```
nbollis, line 113: [review-generic] Only the negative TMT label path is protected offline

The new TMT-specific offline test checks only that an unlabeled peptide is rejected. It does not cover the successful labeled path or request-building behavior for the supported HCD/CID cases, so the refactored TMT acceptance path in this PR is still guarded only by live Koina tests.

Suggestion: Add an offline test that proves a supported N-terminal TMT/iTRAQ label survives `TryCleanSequence`, and a request-building assertion for the HCD/CID path, so the positive branch is protected in CI.

nbollis (commented):
## Review Synthesis

### Major (2)
- `mzLib/PredictionClients/Koina/AbstractClasses/CrosslinkFragmentIntensityModel.cs:88` **Crosslink mapping mode is configurable but never applied** — the new crosslink base class adds a `FragmentIonMappingMode` contract, but the implementation never branches on it, so callers can request `MapToInputFullSequence` without getting distinct behavior.
- **Summary-only:** `mzLib/PredictionClients/Koina/AbstractClasses/CrosslinkFragmentIntensityModel.cs:383` **Crosslink response parsing still accepts mismatched output arrays** — the new parser checks only annotation divisibility before indexing `annotations`, `mz`, and `intensities` in lockstep; it does not validate equal array lengths or enforce the fixed-size `NumberOfPredictedFragmentIons` contract.

### Minor (2)
- `mzLib/Test/KoinaTests/KoinaRequestBuildingTests.cs:91` **Crosslink request-building smoke test bypasses the new model-specific UNIMOD validation** — the test seeds `Validated*Sequence` directly with `UNIMOD:1896`, so it never exercises XLNMS2's actual `{4, 35, 1898}` validation path.
- `mzLib/Test/KoinaTests/KoinaRequestBuildingTests.cs:112` **Only the negative TMT label path is protected offline** — the new offline TMT test covers rejection only, not the successful labeled/request-building path.

### Additional summary-only concern
- `mzLib/PredictionClients/Koina/AbstractClasses/FragmentIntensityModel.cs:737` **Generated library spectra keep the original sequence after masses were rebuilt from the validated sequence** — under `MapToValidatedFullSequence`, precursor and fragment masses are rebuilt from `ValidatedFullSequence`, but `LibrarySpectrum.Sequence`/`Name` still use `FullSequence`, which can desynchronize the library identity from the actual spectral content.

**Grouping:** 5 approved concerns total → 3 inline findings submitted | 2 summary-only findings

mzLib/Test/KoinaTests/KoinaRequestBuildingTests.cs
pcruzparri: Added two offline tests against the real Prosit2024IntensityXLNMS2 that go through TryCleanSequence: one accepts the model's crosslinker (UNIMOD:1898) and one rejects a foreign one (UNIMOD:1896, CMS2's, asserting the warning names it). That exercises the actual {4, 35, 1898} path the shared smoke test skipped by pre-seeding the Validated* fields.

mzLib/Test/KoinaTests/KoinaRequestBuildingTests.cs
pcruzparri: Added a positive offline test confirming a supported N-terminal TMT label survives TryCleanSequence and serializes to UNIMOD:737, plus a parametrized request-building test asserting fragmentation_types flows through for both HCD and CID. One note: in the base TryCleanSequence the return value and the apiSequence out-param are the same serialized string, so the "sanitized vs API-format" distinction doesn't actually exist in the code — the method returns the API-format sequence either way, and the new test asserts that.

mzLib/PredictionClients/Koina/AbstractClasses/CrosslinkFragmentIntensityModel.cs
pcruzparri: Good catch — removed FragmentIonMappingMode from the crosslink base and the three concrete models rather than carry a knob that does nothing.

To clarify why it can't just be wired up like the linear models: the crosslink path takes m/z straight from Koina's response because there's no way to recompute it on our side. MapToInputFullSequence works for linear peptides by re-fragmenting the input sequence with mzLib's PeptideWithSetModifications.Fragment(...) and mapping the model's annotations onto those theoretical products. mzLib only does linear fragmentation though — it has no theoretical crosslink fragment generation (no concept of a fragment that spans the crosslink and carries the crosslinker mass plus the partner peptide). That logic lives in MetaMorpheus's EngineLayer (CrosslinkedPeptide.XlGetTheoreticalFragments), not in mzLib.

So implementing the second mode for crosslinks would mean porting crosslink fragmentation into mzLib first — a real feature with a meaningful dependency, out of scope for this PR. The use case is legitimate (the input alpha/beta sequences can diverge chemically from the validated ones after mod cleaning, same as the linear case), so this is worth doing as a deliberate follow-up rather than a dead config flag.

mzLib/PredictionClients/Koina/AbstractClasses/CollisionalCrossSectionModel.cs
```diff
@@ -0,0 +1,226 @@
...
+            return Predictions;
+        }
+
+        public List<PeptideCCSPrediction> Predict(List<CCSPredictionInput> modelInputs)
+        {
+            return AsyncThrottledPredictor(modelInputs).GetAwaiter().GetResult();
+        }
+
+        protected virtual bool ValidateModelSpecificInputs(CCSPredictionInput input, out WarningException? warning)
+        {
```
trishorts, line 156: **Major** - New CCS/Crosslink Predict() blocks with GetResult() — reintroduces documented deadlock

**Problem:**
`CollisionalCrossSectionModel.Predict` calls `AsyncThrottledPredictor(modelInputs).GetAwaiter().GetResult()` directly, and `CrosslinkFragmentIntensityModel.Predict` does the same. Both are new files in this PR. The sibling `FragmentIntensityModel.Predict` (same PR) deliberately wraps the identical call in `Task.Run(() => AsyncThrottledPredictor(...))` and carries a long comment explaining that calling `.GetAwaiter().GetResult()` directly on `AsyncThrottledPredictor()` "would deadlock on any single-threaded SynchronizationContext (WinForms, WPF, ASP.NET non-Core)" because `GetResult()` blocks the context thread while the awaited `Task.WhenAll`/`Task.Delay` continuations need that same thread. The two new model categories replicate exactly the pattern the PR documents as deadlock-prone.

**Impact:**
Any caller invoking `Predict` for a CCS or crosslink model from a thread with a single-threaded SynchronizationContext can hang permanently. Because these are new public entry points added by this PR, the deadlock risk is newly introduced, not inherited.

**Suggested fix:**
Apply the same mitigation `FragmentIntensityModel.Predict` uses for the new `CollisionalCrossSectionModel` and `CrosslinkFragmentIntensityModel` `Predict` methods — offload to `Task.Run(...)` (and add the lock if shared-state corruption across concurrent callers is a concern, consistent with the documented thread-safety contract).

**Affected test scenarios:**
- CollisionalCrossSectionModel.Predict - invoked under a single-threaded SynchronizationContext - completes without deadlock
- CrosslinkFragmentIntensityModel.Predict - invoked under a single-threaded SynchronizationContext - completes without deadlock

mzLib/PredictionClients/Koina/Client/HTTP.cs
```diff
@@ -1,56 +1,46 @@
...
-        private bool _disposed = false;
 
-        public HTTP(int timeoutInMinutes = 1)
-        {
-            Client = new HttpClient { Timeout = TimeSpan.FromMinutes(timeoutInMinutes) };
-        }
+        private static readonly HttpClient Client = new() { Timeout = Timeout.InfiniteTimeSpan };
 
-        public async Task<string> InferenceRequest(string modelName, Dictionary<string, object> request)
+        public static async Task<string> InferenceRequest(string modelName, Dictionary<string, object> request, CancellationToken cancellationToken = default)
```
trishorts: **Major** - Infinite client timeout + default cancellation token allows unbounded hangs

**Problem:**
The shared `HttpClient` is constructed with `Timeout = Timeout.InfiniteTimeSpan`, and both `InferenceRequest` and `TestConnectionAsync` accept `CancellationToken cancellationToken = default`. The base implementation enforced a real wall-clock bound (`TimeSpan.FromMinutes(timeoutInMinutes)`, default 1 minute). With the new design, any caller that invokes `InferenceRequest(modelName, request)` without supplying a token gets `CancellationToken.None`, so there is no timeout from either the client or the token — the request can block forever on a slow/stalled Koina endpoint.

**Impact:**
A hung or non-responsive Koina server (or a dropped TCP connection that never resets) will cause `await Client.PostAsync(...)` or `ReadToEndAsync(...)` to never complete for callers that don't pass a token. The class XML comment claims "each request is bounded by a caller-supplied token," but nothing enforces that a token is actually supplied, so the safety guarantee is only as good as every call site remembering to pass one.

**Suggested fix:**
Either keep a finite default `HttpClient.Timeout` as a backstop alongside the per-request token, or make the token non-optional (remove the `= default`) so call sites are forced to provide a bound. Document the expectation and verify all in-PR call sites pass a real token.

**Affected test scenarios:**
- InferenceRequest - called with no cancellation token against an unresponsive endpoint - should not block indefinitely (should observe a bounded failure)
- InferenceRequest - called with a short-deadline token - request is cancelled near the deadline

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Prosit2025Intensity22PTM.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// - Handles precursor charges 1-6
+    /// - Predicts up to 174 fragment ions per peptide
+    /// - Supports 22 PTM types
+    /// - Requires fragmentation type input
+    /// 
+    /// API Documentation: https://koina.wilhelmlab.org/docs#post-/Prosit_2025_intensity_22PTM/infer
+    /// </remarks>
+    public class Prosit2025Intensity22PTM : FragmentIntensityModel
+    {
+        private static readonly UnimodSequenceFormatSchema PTMSchema = new(UnimodLabelStyle.UpperCase, '[', ']', "-", "-");
```
trishorts, line 22: **Major** - ModelName disagrees with PR-documented Koina endpoint name for 22PTM/40PTM

**Problem:**
The PR description's "Models Added" table states the Koina model names are `Prosit_2025_intensity_22_PTM` and `Prosit_2025_intensity_40_PTM` (underscore before `PTM`). The code, however, sets `ModelName => "Prosit_2025_intensity_22PTM"` (and `"Prosit_2025_intensity_40PTM"` in `Prosit2025Intensity40PTM.cs`) with no underscore, and the embedded API-doc URLs and source comments use the same no-underscore form. The `ModelName` string is the request endpoint path, so the code and the PR's own contract directly contradict each other. Exactly one of them is correct.

**Impact:**
If the live Koina endpoint actually uses the underscore form (`..._22_PTM`), then every prediction request for these two models hits a non-existent endpoint and fails (404 / unknown-model error), making both models entirely non-functional. This would be Critical; severity here is Major only because which spelling is correct cannot be confirmed from the diff alone.

**Suggested fix:**
Verify the exact model identifier against the live Koina `/docs` listing for both models and make `ModelName`, the doc-URL comment, and the PR description agree. Add a discovery/round-trip test that asserts the configured `ModelName` matches the registered Koina model name.

**Affected test scenarios:**
- ModelDiscovery_Prosit2025Intensity22PTM - queried against Koina model registry - ModelName resolves to a real endpoint
- ModelDiscovery_Prosit2025Intensity40PTM - queried against Koina model registry - ModelName resolves to a real endpoint

trishorts (commented):
Automated code review - 3 finding(s) at severity >= Major, posted as inline comments.

mzLib/PredictionClients/Koina/AbstractClasses/CollisionalCrossSectionModel.cs
pcruzparri: Good catch. Same thing was true for the RetentionTime and Detectability models, not just CCS/Crosslink, so I updated all four to match FragmentIntensityModel.Predict — the lock plus Task.Run(() => AsyncThrottledPredictor(...)) so the continuations don't try to resume on a blocked single-threaded context.

I also added KoinaPredictDeadlockTests, which runs Predict under a single-threaded SynchronizationContext for each family. To make sure it actually catches the problem, I reverted one of the models to the direct .GetResult() and the test hung and failed on the timeout as expected, then put it back.

mzLib/PredictionClients/Koina/Client/HTTP.cs
pcruzparri: Makes sense. I dropped the = default on the token parameters for InferenceRequest and TestConnectionAsync so a CancellationToken is now required — all the call sites already pass a real session-deadline token, this just stops anyone from silently getting None. I left a fixed 10-minute timeout on the client as a backstop so an explicit None still can't hang forever, and updated the doc comment.

I didn't go back to the old per-session new HTTP(timeoutInMinutes: ...) — that spun up a new HttpClient per session, which is the socket-exhaustion problem this PR was fixing. The real bound stays on the per-request token; the client timeout is just a safety net.

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Prosit2025Intensity22PTM.cs
pcruzparri: Checked this against the live registry — the code is right. GET /v2/models/Prosit_2025_intensity_22PTM is a 200 and ..._22_PTM is a 400 (same for 40PTM), so the no-underscore name is the real endpoint. The underscore version was just a typo in the PR description, which I've fixed.

I added a [Category("Koina")] test that reflects over every model and checks its ModelName resolves to a real /v2/models/{name} endpoint. It passes for all 37 models, so this is covered going forward.

nbollis (dismissed):

nbollis (dismissed):

mzLib/PredictionClients/Koina/AbstractClasses/FragmentIntensityModel.cs
```diff
@@ -410,40 +551,74 @@ protected virtual bool ValidateModelSpecificInputs(FragmentIntensityPredictionIn
...
                         throw new ArgumentException(exceptionMessage);
                     case IncompatibleParameterHandlingMode.ReturnNull:
                         warning = new WarningException(exceptionMessage);
                         return false;
+                    default:
+                        throw new ArgumentException($"Unhandled ParameterHandlingMode: {ParameterHandlingMode}");
                 }
             }
 
             // Check collision energy constraints
```
trishorts, line 620: **Major** - Neutral-loss peaks are stored with the base (non-loss) theoretical product

**Problem:**
The new neutral-loss support in `GenerateLibrarySpectraFromPredictions` (and likewise in `ResponseToPredictions`'s `MapToInputFullSequence` branch) computes the experimental m/z with the loss subtracted — `experMz = (tp.NeutralMass - parsed.NeutralLossMass.Value).ToMz(charge)` — but then constructs the `MatchedFragmentIon` with `neutralTheoreticalProduct: tp`, where `tp` was looked up by `parsed.BaseFragmentIdentifier` (e.g. `"y3"` for `"y3-H2O"`). The matched ion therefore carries the m/z of the loss fragment while its theoretical product is the non-loss base ion, so its annotation/neutral mass no longer corresponds to `experMz`.

**Impact:**
When a peptide predicts both a base ion and its neutral-loss variant (e.g. `y3+1` and `y3-H2O+1`), the library will contain two peaks both annotated as `y3`, one at the correct y3 m/z and one shifted by the water loss. The neutral-loss annotation is silently lost and the library entry is internally inconsistent (theoretical mass ≠ experimental m/z), which can mislead or break downstream consumers that re-derive or validate m/z from the product.

**Suggested fix:**
Represent the neutral loss in the `Product`/annotation passed to `MatchedFragmentIon` (e.g. construct a loss-bearing product or carry the loss in the annotation) so the stored ion's theoretical mass and annotation match its loss-shifted `experMz`, rather than reusing the base product.

**Affected test scenarios:**
- GenerateLibrarySpectraFromPredictions - peptide with both `y3+1` and `y3-H2O+1` predicted - both peaks present with distinct, correct annotations and matching m/z
- ResponseToPredictions (MapToInputFullSequence) - annotation with neutral loss - resulting fragment annotation still identifies the loss

mzLib/PredictionClients/Koina/AbstractClasses/CollisionalCrossSectionModel.cs
```diff
@@ -0,0 +1,226 @@
...
+                }
+
+                predictions = ResponseToPredictions(responses, validInputs);
+            }
+
+            var realignedPredictions = new List<PeptideCCSPrediction>();
+            int predictionIndex = 0;
+            for (int i = 0; i < ValidInputsMask.Length; i++)
+            {
+                if (ValidInputsMask[i])
```
trishorts, line 118: **Major** - New CCS/Crosslink Predict() can deadlock on single-threaded SynchronizationContext

**Problem:**
The newly added `CollisionalCrossSectionModel.Predict` and `CrosslinkFragmentIntensityModel.Predict` call `AsyncThrottledPredictor(modelInputs).GetAwaiter().GetResult()` directly. `AsyncThrottledPredictor` awaits `Task.WhenAll(...)` and `Task.Delay(...)` without `ConfigureAwait(false)`. This is exactly the pattern `FragmentIntensityModel.Predict` documents at length and deliberately wraps in `Task.Run(() => ...)` to avoid: blocking the context thread with `GetResult()` while the awaited continuations need that same thread to resume.

**Impact:**
If either of these new abstract models' `Predict` is invoked from a single-threaded SynchronizationContext (WPF/WinForms UI thread, classic ASP.NET), the call deadlocks permanently — the precise failure mode the FragmentIntensity refactor was written to prevent. Because these are brand-new files in this PR, the safe pattern was not carried over.

**Suggested fix:**
Wrap the synchronous bridge in `Task.Run(() => AsyncThrottledPredictor(modelInputs)).GetAwaiter().GetResult()` (as `FragmentIntensityModel` does), or add `ConfigureAwait(false)` to the awaits inside `AsyncThrottledPredictor`, for both the new CCS and Crosslink models.

**Affected test scenarios:**
- CCSModel.Predict - invoked under a single-threaded SynchronizationContext - returns without deadlock
- CrosslinkModel.Predict - invoked under a single-threaded SynchronizationContext - returns without deadlock

mzLib/PredictionClients/Koina/Client/HTTP.cs
```diff
@@ -1,56 +1,46 @@
...
     {
         public static readonly string ModelsURL = "https://koina.wilhelmlab.org:443/v2/models/";
-        public readonly HttpClient Client;
-        private bool _disposed = false;
 
-        public HTTP(int timeoutInMinutes = 1)
-        {
-            Client = new HttpClient { Timeout = TimeSpan.FromMinutes(timeoutInMinutes) };
-        }
+        private static readonly HttpClient Client = new() { Timeout = Timeout.InfiniteTimeSpan };
```
trishorts: **Major** - Infinite client timeout + optional cancellation token allows requests to hang forever

**Problem:**
The shared `HttpClient` is constructed with `Timeout = Timeout.InfiniteTimeSpan`, and `InferenceRequest`/`TestConnectionAsync` accept the cancellation token as an optional parameter (`CancellationToken cancellationToken = default`). When a caller invokes `InferenceRequest(modelName, request)` without supplying a token, the effective token is `CancellationToken.None`, so there is no timeout from the client and no cancellation from the caller — the `await Client.PostAsync(...)`/`ReadToEndAsync(...)` can block indefinitely. The base implementation previously enforced a finite default (`TimeSpan.FromMinutes(1)`), so any existing call site not updated to pass a token loses its only time bound.

**Impact:**
A stalled or half-open connection to the Koina endpoint will hang the calling task forever instead of failing after ~1 minute. In a long-running session this leaks tasks/connections and can silently stall prediction pipelines with no error surfaced.

**Suggested fix:**
Either make the `CancellationToken` parameter required (removing the `= default`) so the XML contract ("each request is bounded by a caller-supplied token") is actually enforced, or retain a finite default request timeout (e.g., apply a `CancellationTokenSource` with a default timeout when no token is provided) so an omitted token still bounds the request.

**Affected test scenarios:**
- InferenceRequest - called without a cancellation token against an unresponsive endpoint - should fail/timeout within a bounded interval rather than hang
- InferenceRequest - called with a pre-canceled token - should throw OperationCanceledException promptly
- TestConnectionAsync - called with a canceled token - should observe cancellation rather than block

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Altimeter2024Intensities.cs
```diff
@@ -0,0 +1,144 @@
...
+            if (caretIndex > 0)
+            {
+                var fragmentIdentifier = annotation.Substring(0, caretIndex);
+                var chargeStr = annotation.Substring(caretIndex + 1);
+                if (!int.TryParse(chargeStr, out int charge))
+                {
+                    throw new Exception($"Cannot parse charge from Altimeter fragment annotation '{annotation}'. Charge value: '{chargeStr}'.");
+                }
+
+                if (TryExtractNeutralLoss(fragmentIdentifier, out var baseId, out var nlFormula, out var nlMass))
```
trishorts: **Major** - Neutral-loss base identifier computed but discarded; full suffixed annotation passed instead

**Problem:**
In `ParseFragmentAnnotation`, when `TryExtractNeutralLoss` succeeds it correctly computes `baseId` (the fragment identifier with the `-H2O`/`-NH3` suffix stripped, e.g. `"y8"` from `"y8-H2O"`). But both success branches return `new ParsedFragmentAnnotation(fragmentIdentifier, charge, nlFormula, nlMass)` and `new ParsedFragmentAnnotation(annotation, 1, nlFormulaNoCharge, nlMassNoCharge)` — passing the *un-stripped* string (`fragmentIdentifier`/`annotation`, which still contains the `-H2O` suffix) as the base identifier. The `out baseId` value is never read. This contradicts the PR's stated intent that "`BaseFragmentIdentifier` uses first-dash stripping."

**Impact:**
The fragment identifier handed downstream for theoretical-product matching / m/z recalculation will be `"y8-H2O"` instead of `"y8"`, while the neutral loss is *also* reported separately via `nlFormula`/`nlMass`. Depending on the consumer this either fails to match any theoretical ion (mapping silently drops the fragment) or double-counts the neutral loss. Every neutral-loss fragment Altimeter returns is mis-identified.

**Suggested fix:**
Return `baseId` (the stripped identifier) as the first argument to `ParsedFragmentAnnotation` in both neutral-loss branches, instead of the original `fragmentIdentifier`/`annotation` string.

**Affected test scenarios:**
- ParseFragmentAnnotation - annotation `"y8-H2O^2"` - base identifier is `"y8"`, charge 2, neutral loss H2O
- ParseFragmentAnnotation - annotation `"IY-NH3"` (no charge delimiter) - base identifier is `"IY"`, charge 1, neutral loss NH3
- ParseFragmentAnnotation - annotation `"y8^2"` (no NL) - base identifier `"y8"`, charge 2, no neutral loss (regression guard)

mzLib/PredictionClients/Koina/SupportedModels/RetentionTimeModels/DeeplcHelaHf.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// <summary>
+    /// Implementation of the DeepLC HeLa HF retention time prediction model.
+    /// Uses deep learning with extensive feature engineering for accurate RT prediction.
+    /// </summary>
+    public class DeeplcHelaHf : RetentionTimeModel
+    {
+        private static readonly ISequenceConverter Converter = CreateUnimodConverterAcceptAll(
+            UnimodSequenceFormatSchema.Instance);
+
```
trishorts, line 14: **Major** - DeeplcHelaHf ModelName does not match the Koina name in the PR spec

**Problem:**
`ModelName` is hard-coded to `"Deeplc_hela_hf"`, but the PR description table maps `DeeplcHelaHf` to the Koina model name `DeepLC_Helacell_HF`. The `ModelName` string is the endpoint identifier sent to the Koina `/{model}/infer` route, so the implementation and the stated spec disagree on which remote model is actually invoked. One of the two is wrong, and the discrepancy is invisible at compile time.

**Impact:**
If `DeepLC_Helacell_HF` is the correct Koina identifier, every prediction request from this class resolves to a non-existent endpoint and fails (404 / model-not-found) at runtime, making the model unusable. If instead `Deeplc_hela_hf` is correct, the PR documentation is misleading for downstream users.

**Suggested fix:**
Confirm the exact, case-sensitive Koina model name against the live Koina model list and make the code and the PR table agree. Add a discovery/round-trip test that asserts this class's `ModelName` is present in the Koina model registry so the mismatch is caught automatically.

**Affected test scenarios:**
- DeeplcHelaHf_ModelName - compared against live Koina model list - name must resolve to a real endpoint
- DeeplcHelaHf_ToBatchedRequests - submitted to Koina infer route - returns a successful (non-404) response

trishorts (commented):
Automated code review - 5 finding(s) at severity >= Major, posted as inline comments.

mzLib/PredictionClients/Koina/AbstractClasses/FragmentIntensityModel.cs
pcruzparri: Good catch on the library path. In `GenerateLibrarySpectraFromPredictions` a neutral-loss fragment was getting the loss-shifted m/z but the base (non-loss) `Product` as its theoretical product, so the peak came out annotated as the base ion and its theoretical mass no longer matched the m/z — a `y3-H2O` would land as a second `y3`. Fixed by building a loss-bearing `Product` (same backbone, `NeutralLoss` set to the parsed loss) when `parsed.NeutralLossMass` is present, so the annotation shows the loss and the mass lines up with `experMz`. Added `GenerateLibrarySpectra_NeutralLossFragment_CarriesLossOnProduct`, which predicts both `y3` and `y3-H2O` and asserts two distinct peaks with the loss carried on the product and ~0 mass error (it fails on the old code).

One note on the `ResponseToPredictions`/`MapToInputFullSequence` branch you also mention: there the annotation is stored as the raw model string (`y3-H2O^2`) and only the m/z is recomputed, so the loss is already preserved in that path — this fix is specific to the `MatchedFragmentIon` built in `GenerateLibrarySpectra`.

mzLib/PredictionClients/Koina/AbstractClasses/CollisionalCrossSectionModel.cs
pcruzparri: This was already handled in d262303. `CollisionalCrossSectionModel.Predict` and `CrosslinkFragmentIntensityModel.Predict` both wrap the bridge in `lock` + `Task.Run(() => AsyncThrottledPredictor(...)).GetAwaiter().GetResult()`, same as `FragmentIntensityModel.Predict`, and `KoinaPredictDeadlockTests` exercises each family under a single-threaded `SynchronizationContext`. Line 118 here is the realignment loop inside `AsyncThrottledPredictor`, not a direct `.GetResult()` call — the earlier thread on this exact issue is resolved.

mzLib/PredictionClients/Koina/Client/HTTP.cs
pcruzparri: Also handled in d262303. The shared `HttpClient` no longer uses `Timeout.InfiniteTimeSpan` — it has a fixed 10-minute backstop — and the `CancellationToken` is now required on both `InferenceRequest` and `TestConnectionAsync` (the `= default` was removed), so a caller can't silently get `None`. The precise per-session bound is still the caller's token; the client timeout is just a floor. Prior thread on this is resolved.

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Altimeter2024Intensities.cs
pcruzparri: The first argument to `ParsedFragmentAnnotation` is the full `FragmentIdentifier` by design, not the stripped base — `BaseFragmentIdentifier` derives the stripped form itself (it cuts at the first dash when `NeutralLossFormula` is set). So passing `"y8-H2O"` is correct: downstream matching uses `parsed.BaseFragmentIdentifier`, which returns `"y8"`, while the loss is carried separately in `NeutralLossFormula`/`NeutralLossMass` and the raw annotation string is preserved. The `baseId` out of `TryExtractNeutralLoss` does end up unused since `BaseFragmentIdentifier` re-derives it — happy to drop it for clarity, but there's no mis-identification here.

mzLib/PredictionClients/Koina/SupportedModels/RetentionTimeModels/DeeplcHelaHf.cs
pcruzparri: The code is right and the PR description was wrong — I've fixed the table. Checked against the live registry: `GET /v2/models/Deeplc_hela_hf` returns 200 and `DeepLC_Helacell_HF` returns 400, so `Deeplc_hela_hf` is the real endpoint. Same shape as the 22PTM/40PTM note earlier, and the model-discovery test from d262303 reflects over every model's `ModelName` against `/v2/models/{name}`, so a wrong name would fail there.

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Altimeter2024Intensities.cs
trishorts: Follow-up after the 06-30 commits (`afba1ca1`..`523e3df1`): commit `11a47cc4` fixes the base-class neutral-loss handling in `FragmentIntensityModel.GenerateLibrarySpectraFromPredictions`, but this finding is about `Altimeter2024Intensities.ParseFragmentAnnotation` itself, which the new commits don't touch — so it's still open. The `out baseId` (stripped identifier, e.g. `y8` from `y8-H2O`) is still computed and discarded; both NL branches still return the un-stripped `fragmentIdentifier`/`annotation` as the base id. Worth confirming whether the base-class fix now compensates, or whether Altimeter's parser still needs to return `baseId`.

mzLib/PredictionClients/Koina/SupportedModels/RetentionTimeModels/DeeplcHelaHf.cs
trishorts: Follow-up after the 06-30 commits: this one is untouched — `DeeplcHelaHf.cs` isn't in `afba1ca1`..`523e3df1`, so `ModelName` is still `"Deeplc_hela_hf"` vs the PR table's `DeepLC_Helacell_HF`. Still needs the code/spec reconciled against the live Koina registry (and ideally a discovery test asserting the name resolves, which would also have caught this automatically).

trishorts:
@pcruzparri Thanks for the work on the 06-30 commits — the neutral-loss base-product fix (`11a47cc4`), the bounded HTTP timeout, and the deadlock test all look good from my side.

I did want to check on two of the Major threads that are now marked **resolved** but whose files don't appear in the latest commits (`afba1ca1`..`523e3df1`), so I couldn't see what addressed them:

- **Altimeter `ParseFragmentAnnotation`** ([thread](https://github.com/smith-chem-wisc/mzLib/pull/1073#discussion_r3498923104)) — the stripped `baseId` still looks computed-then-discarded in `Altimeter2024Intensities.cs`. Does the base-class fix in `FragmentIntensityModel` now compensate, making the override a no-op, or is there a separate change I'm missing?
- **`DeeplcHelaHf.ModelName`** ([thread](https://github.com/smith-chem-wisc/mzLib/pull/1073#discussion_r3498923387)) — still reads `"Deeplc_hela_hf"` vs the PR table's `DeepLC_Helacell_HF`. Was the name confirmed correct against the live Koina registry (i.e. the table is the typo)?

No rush — if these are genuinely no-ops or already-correct, a one-line note on each thread explaining why would let me close them out with confidence. Thanks!

mzLib/PredictionClients/Koina/SupportedModels/RetentionTimeModels/DeeplcHelaHf.cs
pcruzparri: Reconciled — the spec side just isn't a commit. Deeplc_hela_hf is the correct endpoint (GET /v2/models/Deeplc_hela_hf → 200, DeepLC_Helacell_HF → 400), so I fixed the PR description table instead of the code (PR-body edit, hence not in afba1ca..523e3df). And the discovery test already exists: KoinaModelDiscoveryTests.EveryConcreteModel_ModelNameResolvesToRegisteredKoinaEndpoint (d262303) reflects over every concrete model and asserts its ModelName resolves at /v2/models/{name}, so DeeplcHelaHf is covered.

mzLib/PredictionClients/Koina/SupportedModels/FragmentIntensityModels/Altimeter2024Intensities.cs
pcruzparri: The base-class design already compensates, independent of 11a47cc. ParsedFragmentAnnotation.FragmentIdentifier holds the full suffixed string (y8-H2O) by contract, and BaseFragmentIdentifier derives the stripped y8 when NeutralLossFormula is set. Both consumers that match theoretical products — ResponseToPredictions (FragmentIntensityModel.cs:516) and GenerateLibrarySpectraFromPredictions (:749) — key on parsed.BaseFragmentIdentifier, and nothing matches on the raw FragmentIdentifier, so y8-H2O resolves to y8 while the loss rides in NeutralLossFormula/NeutralLossMass. No mis-identification. The out baseId does end up unused — I can drop it so it stops reading as load-bearing, but that's cleanup, not a behavior change.

trishorts (approved):

nbollis (approved):

PR #1077, opened 2026-06-15
Batched Chronologer retention-time inference (PredictRetentionTimeEquivalents override)

trishorts:
## Summary

Adds a **true tensor-batched** `PredictRetentionTimeEquivalents` override for the Chronologer retention-time predictor. This is extracted from #1036 so it can be reviewed and merged independently of the `.msl` spectral-library format work — it's a self-contained RT-prediction performance change (2 files, +92/−3) that depends only on what's already on `master`.

## Why an override (and not an edit to the existing batch method)

`RetentionTimePredictor` already has a `PredictRetentionTimeEquivalents(...)` batch method on the base class — but it does **CPU thread-parallelism over single-peptide predictions**: it partitions the peptides across `maxThreads` workers and calls `PredictCore` once per peptide. Each peptide is therefore still its own **lock-held, batch-of-1** libtorch forward pass.

The base method can't be changed to do real batching, because batching is **model-specific** — only Chronologer has a libtorch tensor model to run an `[m, 1]` batch through; `SSRCalc`/`CZE` don't. Putting libtorch batching in the shared base would leak model internals into the abstraction. The base class even documents the intended pattern:

> *"Predictors that support true batched inference (e.g. Chronologer) should **override** `PredictRetentionTimeEquivalents` instead of this method."*

This PR does exactly that.

## What it does

- **Batched override** (`ChronologerRetentionTimePredictor`): format + integer-encode the peptides in parallel on the CPU (no model access), then run the model in **large batched forward passes** — one `[m, 1]` tensor per chunk, with the model lock taken **once per chunk instead of once per peptide**. Far fewer lock acquisitions and kernel launches → much faster when predicting many peptides (e.g. building spectral libraries over millions).
- **Numerical equivalence**: the model is in eval mode (BatchNorm uses running statistics), so each peptide's prediction is independent of the batch. Results match the single-peptide `PredictCore` to **float32 precision** — not bit-identical, because libtorch selects different conv/matmul kernels at batch size `m` vs `1`. The batch test tolerance is widened to allow that drift.
- **Deterministic native cleanup**: wraps the forward pass in `NewDisposeScope()` so `forward()`'s intermediate tensors (residual clones, conv/norm outputs) are freed promptly instead of accumulating on the native heap.

## Tests

`RetentionTimeEquivalentBatchTests` (20 tests) pass on `master` + this change; the only test edit is the float32 tolerance noted above.

## Relationship to #1036

This is the foundational, independent piece of #1036. Merging it first shrinks #1036 (these files drop out of its diff with identical content) and gives the batched-inference work its own focused review.

### DIFF
--- a/mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
+++ b/mzLib/Chromatography/RetentionTimePrediction/Chronologer/ChronologerRetentionTimePredictor.cs
@@ -75,9 +75,94 @@ public class ChronologerRetentionTimePredictor : RetentionTimePredictor
         // Predict retention time - keep both prediction AND disposal inside lock
         lock (_modelLock)
         {
-            using Tensor prediction = _model.Predict(sequenceTensor);
+            using var scope = NewDisposeScope();   // dispose forward()'s intermediates deterministically
+            using Tensor prediction = _model.Predict(sequenceTensor).cpu();
             return prediction[0].ToDouble();
         }
+    }
+
+    /// <summary>
+    /// Batched override: formats/encodes the peptides in parallel (CPU) and runs the Chronologer model in
+    /// large batched forward passes instead of one locked batch-1 call per peptide. The model is in eval
+    /// mode (BatchNorm uses running statistics), so each peptide's prediction is independent of the batch —
+    /// results match <see cref="PredictCore"/> to float32 precision (libtorch selects different conv/matmul
+    /// kernels at batch size m vs 1, so the two paths are not bit-identical), just far faster for many peptides.
+    /// </summary>
+    public override IReadOnlyList<(double? PredictedValue, IRetentionPredictable Peptide, RetentionTimeFailureReason? FailureReason)>
+        PredictRetentionTimeEquivalents(IEnumerable<IRetentionPredictable> peptides, int maxThreads = 1)
+    {
+        if (_disposed)
+            throw new ObjectDisposedException(nameof(ChronologerRetentionTimePredictor));
+
+        var list = peptides as IReadOnlyList<IRetentionPredictable> ?? peptides.ToList();
+        int n = list.Count;
+        int encodedLength = ChronologerSequenceFormatSchema.EncodedLength;
+        var results = new (double?, IRetentionPredictable, RetentionTimeFailureReason?)[n];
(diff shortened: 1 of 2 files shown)


mzLib/Chromatography/RetentionTimePrediction/Chronologer/Chronologer.cs
```diff
@@ -57,6 +64,16 @@ public Chronologer(string? weightsPath = null, bool evalMode = true)
             eval(); // Evaluation mode doesn't update weights
             train(false);
         }
+
+        // Optional CUDA acceleration (opt-in via env). The Chronologer CNN is tiny but is called for
+        // millions of peptides when building spectral libraries; moving it to the GPU turns iRT from the
+        // build bottleneck into a non-issue. Default stays CPU so existing consumers are unaffected.
+        Device = torch.CPU;
+        if (Environment.GetEnvironmentVariable("MZLIB_CHRONOLOGER_CUDA") == "1" && torch.cuda.is_available())
```
nbollis: What is this environmental variable and where does it get set? We should be able to accomplish the CPU/GPU switching without an environmental variable

trishorts: Good call — removed it. The `MZLIB_CHRONOLOGER_CUDA` env var (read in the `Chronologer` constructor) was an opt-in to move the model to CUDA, but it was never actually used: the production library build runs Chronologer on CPU (the win there is the batched CPU override here plus a sharded iRT cache), and an env-gated execution mode is the wrong surface for it anyway. I've dropped the whole CUDA/`Device` path — `Chronologer.cs` is now byte-identical to `master`, and the predictor no longer moves tensors to a device. The PR is down to the batched-CPU override + the `NewDisposeScope()` fix (2 files, +92/−3), tests still green.

If we ever want GPU inference, it should come back as an explicit constructor parameter (e.g. a `torch.Device` on `Chronologer`, defaulting to CPU) with benchmarks behind it — not an env var. Happy to do that as a separate change if/when there's a concrete need.

pcruzparri (approved):
The _model.Predict() call inside the lock can be corrupted, since the model can still be disposed before this call. It's a preexisting issue, so won't hold the PR. The fix is simple though, wrap the stuff inside the Dispose() method in a lock(_modelLock) {...} block.

nbollis (approved):

PR #1078, opened 2026-06-16
Expose the most-abundant observed mass on IsotopicEnvelope (resolved)

trishorts:
## Summary
Adds the mzLib-side support for selecting top-down proteoform candidates by the **most abundant isotopic peak** instead of the often-undetectable monoisotopic peak (which causes off-by-N precursor errors at high mass). Companion MetaMorpheus PR consumes this.

## Changes
- `IsotopicEnvelope`: new public `MostAbundantObservedMass` (proton-corrected neutral mass of the most intense peak) and `AverageObservedMass` (intensity-weighted centroid neutral mass), plus an `EnvelopeResolution { Resolved, Unresolved }` flag and `SetResolution(...)`. Computed in every constructor. The existing (deliberately un-proton-corrected) `MostAbundantObservedIsotopicMass` is retained for back-compat.
- `AverageResidue`: new public `GetMostAbundantOffset(mono)` and `GetAverageOffset(mono)` — the averagine-derived offset from a monoisotopic mass to the most-abundant / average mass, used downstream to place each candidate at its theoretical apex.

## Tests
New `TestMostAbundantMass` fixture (10 tests): offset monotonicity/magnitude, proton-corrected mass, mono+offset round-trip, resolution default/equality. All existing deconvolution/scorer tests pass (493 related).


mzLib/MassSpectrometry/Deconvolution/AverageResidue/AverageResidue.cs
```diff
@@ -16,29 +14,42 @@ public abstract class AverageResidue : IEquatable<AverageResidue>
...
+
+    /// <summary>
+    /// The number of daltons to add to a monoisotopic mass to reach the intensity-weighted average
+    /// (centroid) mass of the isotopic envelope, according to this average-residue model. Used for
+    /// isotopically unresolved (high-mass) species, where the observed precursor is a centroid rather
+    /// than a resolved most-abundant peak.
+    /// </summary>
+    /// <param name="monoisotopicMass">A monoisotopic mass, in daltons.</param>
+    public double GetAverageOffset(double monoisotopicMass)
     {
```
nbollis, line 41: Isn't this the same as diff to monoisotopic?

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
```diff
@@ -15,6 +26,29 @@ public class IsotopicEnvelope : IHasMass, IEquatable<IsotopicEnvelope>
...
         /// </summary>
         internal double MostAbundantObservedIsotopicMass { get; private set; }
+
+        /// <summary>
+        /// Neutral mass of the most abundant observed isotopic peak (proton-corrected, i.e. the most
+        /// intense peak's m/z converted to neutral mass via the envelope charge). This is the
+        /// experimentally most detectable peak in each charge envelope and is used as the precursor
+        /// mass for candidate selection when running in most-abundant precursor-mass mode.
+        /// </summary>
+        public double MostAbundantObservedMass { get; private set; }
```
nbollis: Isn't this handled by MostAbundantObservedIsotopicPeak?

mzLib/MassSpectrometry/Deconvolution/AverageResidue/AverageResidue.cs
trishorts: Good catch — you're right, it was just `GetDiffToMonoisotopic(GetMostIntenseMassIndex(mass))` keyed by mass instead of index, so it added no real logic. Removed it in c144c38b; the consumers (the MetaMorpheus most-abundant acceptors) now compose the existing public methods directly, so there's no extra public API here.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
trishorts: Not quite the same — `MostAbundantObservedIsotopicMass` is `mz × |charge|` and is deliberately *not* proton-corrected (per its doc), i.e. a charge-scaled m/z rather than a neutral mass. `MostAbundantObservedMass` is the proton-corrected neutral mass (`mz.ToMass(charge)`), which is what's directly comparable to theoretical proteoform masses for candidate selection; the two differ by charge × proton mass. I expanded both docstrings to make the distinction explicit in c144c38b. Happy to rename either if you'd prefer clearer names.

mzLib/MassSpectrometry/Deconvolution/AverageResidue/AverageResidue.cs
```diff
@@ -16,29 +14,31 @@ public abstract class AverageResidue : IEquatable<AverageResidue>
...
-        if (ReferenceEquals(this, other)) return true;
-        if (GetType() != other.GetType()) return false;
-        return EqualProperties(other);
+        int index = GetMostIntenseMassIndex(monoisotopicMass);
+        double[] masses = GetAllTheoreticalMasses(index);
+        double[] intensities = GetAllTheoreticalIntensities(index);
+
+        double totalIntensity = 0;
+        double weightedMassSum = 0;
+        for (int i = 0; i < masses.Length; i++)
```
nbollis: Is the intensity weighted mass intentional? That seems like a silly way to arrive at the diff to most intense. Something like 

```
    public virtual double GetDiffToMostAbundant(double testMass)
    {
        int index = GetMostIntenseMassIndex(testMass);
        return GetDiffToMonoisotopic(index) + testMass - GetAllTheoreticalMasses(index)[0];
    }
``` 

may get you there more precisely. Push back if you want the centroided here.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
nbollis: Then I do not see the point of the preexisting MostAbyndantObservedIsotopicMass if it is calculated incorrectly. Could we simply fix the calculation and reuse the property?

it does not appear to be used outside of this class and testing

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
```diff
@@ -80,6 +120,36 @@ public IsotopicEnvelope(int id, List<(double mz, double intensity)> peaks, doubl
...
             MostAbundantObservedIsotopicMass = peaks.MaxBy(p => p.intensity).mz * Math.Abs(chargestate);
+            ComputeObservedMasses();
+        }
+
+        /// <summary>
+        /// Computes the proton-corrected neutral most-abundant and average (centroid) observed masses
+        /// from <see cref="Peaks"/> and <see cref="Charge"/>. Called from every constructor after the
+        /// peak list and charge are set.
+        /// </summary>
+        private void ComputeObservedMasses()
```
nbollis: MostAbundantObservedMass and AverageObservedMass should be computed properties — These are fully derivable from Peaks + Charge (both public readonly). Storing them as { get; private set; } fields creates a consistency burden (must call ComputeObservedMasses() in every constructor) and risks staleness if Peaks is mutated externally. Computed get-only properties would eliminate constructor boilerplate and match the data. MostAbundantObservedIsotopicMass (existing internal property) has the same concern.
   - Consensus across 2 specialists (review-backend, review-oop)

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
nbollis, line 1: ComputeObservedMasses() runs unconditionally in the single-peak file-read path — Constructor 2 (neutal mass from deconvoluted file) calls ComputeObservedMasses() on a single-element peak list where AverageObservedMass == MostAbundantObservedMass == MonoisotopicMass. The entire loop is a no-op. Minor — cheap but redundant.
   - review-backend

mzLib/MassSpectrometry/Deconvolution/AverageResidue/AverageResidue.cs
```diff
@@ -1,12 +1,10 @@
...
-
-namespace MassSpectrometry;
+﻿namespace MassSpectrometry;
 
 /// <summary>
 /// Represents the average residue and is used for most abundant isotopic peak to monoisotopic peak difference
 /// </summary>
 /// <remarks>CAREFUL: This is called a lot during <see cref="ClassicDeconvolutionAlgorithm"/> and care should be taken to precalculate as many values as possible when implementing a new type</remarks>
-public abstract class AverageResidue : IEquatable<AverageResidue>
+public abstract class AverageResidue
```
nbollis: Breaking change: IEquatable<AverageResidue> removed — The interface was added in PR #1070 and removed here. Zero in-codebase consumers (verified by grep), but removing an interface from a public abstract class is technically a SemVer-breaking change. Document in release notes.
   - review-oop

nbollis (changes requested):
I believe our main disconnect here is in terminology. In my head the masses are defined as follows:

Average Mass = tallest peak m/z.ToMass(charge)
Centroided Mass = Intensity weighted sum of each isotopic peak m/z.ToMass(charge)

Other findings besides my inline comments. 

AverageResidue.cs:24 Centroid intensity-weighted average math appears in 3 places — The same weighted-average formula exists in GetAverageOffset (new), ComputeObservedMasses (new), and DeconvolutionScorer.ComputeAvgPpmError (existing). Consider extracting a static ComputeCentroidMass(double[] masses, double[] intensities) helper to reduce duplication.

mzLib/MassSpectrometry/Deconvolution/AverageResidue/AverageResidue.cs
trishorts: You're right that #1070 added `IEquatable<AverageResidue>` — that's exactly what GitHub was flagging as a conflict. This branch was based on a pre-#1070 `master`, so it never had the interface to begin with (nothing here removed it). I've merged the latest `master` in and resolved the `AverageResidue.cs` conflict by keeping **both** the `IEquatable` region from #1070 and the new `GetAverageOffset`. The interface is preserved and the full deconvolution suite passes (661/0).

mzLib/MassSpectrometry/Deconvolution/AverageResidue/AverageResidue.cs
trishorts: The method this pointed at (the most-abundant offset, which used the centroid math) was removed in c144c38b — consumers now compose the existing index-based public methods directly. The only intensity-weighted method left is `GetAverageOffset`, where the centroid *is* the intent: it computes the average (centroid) mass, not the most-abundant one.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
trishorts: Resolved by making `MostAbundantObservedMass`/`AverageObservedMass` computed get-only properties (per your line-131 comment). `ComputeObservedMasses()` and all of its constructor calls are gone, so there's no longer any per-constructor work to be a no-op on the single-peak file-read path.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
trishorts: I'd like to keep these as two separate properties. `MostAbundantObservedIsotopicMass` is deliberately `m/z × |charge|` (not proton-corrected) rather than incorrect — the existing deconvolution tests rely on that exact semantic, dividing it by charge to recover the selected-ion m/z (e.g. `StandardDeconvolutionTest.cs:217/236/271`, `TestDeconvolution.cs:111/212/360`) and pinning hardcoded values like `12367.44`. "Fixing" it to a neutral mass would break those tests (out of scope for this PR) and conflate two distinct quantities. `MostAbundantObservedMass` is the proton-corrected neutral mass used for candidate selection; the docstrings (c144c38b) spell out the distinction. Happy to rename either for clarity if you'd prefer.

mzLib/MassSpectrometry/MzSpectra/IsotopicEnvelope.cs
trishorts: Done — converted both `MostAbundantObservedMass` and `AverageObservedMass` to computed get-only properties and deleted `ComputeObservedMasses()` and its three constructor calls. Since `Peaks` and `Charge` are readonly there's no staleness risk. I left the legacy `MostAbundantObservedIsotopicMass` as-is (discussed in the other thread).

trishorts:
Two follow-ups from the review summary:

**Terminology (average vs centroid).** I think we're using "average mass" in the standard mass-spec sense — the abundance-weighted mass over the isotope distribution (the centroid), as the counterpart to monoisotopic mass — while "most abundant mass" is the tallest peak. That's why `AverageObservedMass` is the intensity-weighted centroid and `MostAbundantObservedMass` is the tallest peak. The docstrings state the formulas explicitly. I'd lean on the standard convention for the names to avoid churning the public API and the companion MetaMorpheus PR, but I'm happy to surface the word "centroid" more prominently in the docs if that's clearer.

**Helper extraction (`ComputeCentroidMass`).** Looking at the three sites: `DeconvolutionScorer.ComputeAvgPpmError` isn't actually a centroid (it anchors at the apex m/z and averages |ppm error| unweighted, `/peaks.Count`), and `ComputeObservedMasses` was removed in the computed-properties change above. That leaves two genuine centroid computations on different data shapes — theoretical `double[]/double[]` in mass space vs. the `Peaks` tuple-list centroided in m/z — so a shared `ComputeCentroidMass(double[], double[])` would fit only one and would force array allocation in a hot getter. I left them inline, but can revisit if you feel strongly.

pcruzparri (dismissed):

nbollis (approved):

pcruzparri (approved):

PR #1085, opened 2026-07-07
perf(FromFileDecon): lazy-load MS1 features; I/O-free equality

nbollis:
Human Explanation: 

I want to use feature files for deconvolution in MetaMorpheus. This PR enables that by lazy-loading the features so that loading the decon params in the UI of MetaMorpheus does not read and parse the feature file. (An earlier revision also added TOML round-trip support via Nett; that has been dropped from this PR — see the reverted commit `b8bf59fe` — to keep the scope focused on the lazy-load and I/O-free-equality refactor.)

## Motivation

`FromFileDeconvolutionParameters` previously read, parsed, and materialized the entire per-charge feature expansion in the constructor. This was a problem for two distinct workflows:

1. **Configuration-only code paths** — when parameters are loaded from a config (TOML, XML, programmatically), the file-path-based constructor is invoked for *every* parameter object, even when the consumer only inspects a few fields and never actually runs deconvolution. For workflows that build many parameter objects up-front (e.g. configuration grids, batched engine setup, the MetaMorpheus UI), the feature file was being parsed once per object — wasted memory and disk I/O.
2. **Multi-scan reuse** — when the same parameters are reused across many MS2 scans, the cache is correctly amortized, but any *configuration-only* code path (equality checks, hash computation, `Clone()` for parameter-set construction) still paid the up-front load cost.

This change makes the feature load lazy (double-checked locking, idempotent) and makes `GetHashCode` and `Equals` I/O-free so they are safe to call on an instance whose feature cache has not yet been materialized.

## Summary of Changes

- **`mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs`**
  - Replaced eager `_featuresMzAscending` / `_mzKeys` fields with a private `FeatureCache` value type, populated on first access via `GetOrLoadFeatureCache()` using double-checked locking.
  - Added `FilePath` property (replaces `OriginalFilePath`); setting it invalidates the cache so a new file can be assigned to an existing instance.
  - Refactored `EqualProperties` and `AddHashCodes` to key on `FilePath` alone (no `Features.Count` read). `GetHashCode` and `Equals` are now I/O-free and cannot throw, even when `FilePath` points at a missing or non-feature file. The base class still contributes the other config fields (`MinAssumedChargeState`, `MaxAssumedChargeState`, `Polarity`, `ExpectedIsotopeSpacing`, `UseGenericScore`, `AverageResidueModel`), so equality is still a complete config-only identity.
  - Hardened `Clone()`: preserves `ExpectedIsotopeSpacing`, `AverageResidueModel`, and `UseGenericScore`; uses the file-path constructor when `FilePath` is set (avoids re-loading on clone).

## Testing Notes

- Existing `FromFileDeconvolutionParameters` / `FromFileDeconvolutionAlgorithm` tests should continue to pass — the public API is preserved (constructor signatures, `Features` property, `DeconvolutionType`).
- The PR adds five new tests to `TestFromFileDeconvolution.cs`:
  - `GetHashCode_OnUnloadedInstance_DoesNotLoadFile` — constructs from a missing path and asserts `GetHashCode()` does not throw and returns a non-trivial value.
  - `Equals_OnUnloadedInstances_DoesNotLoadFile` — two instances from the same path compare equal without ever touching `.Features`.
  - `SetFilePath_InvalidatesCache` — materializes the cache from path A, assigns `FilePath = pathB`, accesses `.Features` again, asserts the second access reflects path B (cache was dropped, the new file was re-read).
  - `RepeatedFeaturesAccess_LoadsOnce` — three sequential `.Features` accesses return the same list reference (double-checked locking, no re-read).
  - `NonExistentFile_ConstructorDoesNotThrow_LoadDoes` — `Assume.That(File.Exists(...), Is.False)` guard on a non-existent path; constructor returns silently, `.Features` throws. Uses the constraint-style `Assert.That(..., Throws.Exception)` because the load path can surface `FileNotFoundException` or `MzLibException` depending on extension.
- The existing `NonFeatureFile_ThrowsMzLibException` test was rewritten to split the constructor out of the `try`, so a regression that re-introduces eager loading in the constructor fails the test immediately.
- `TestDeconvolutionParametersEqualityAndClone.FromFile_Equal_DifferentFeatureCount_AreNotEqual` was updated to reflect the new contract (now `…AreEqualWhenNoFilePath`) and a new `…DifferentFilePaths_AreNotEqual` test was added to lock the FilePath-distinguishes-them half.
- Full test suite: 4675/4675 pass.

## Migration Steps

None. The change is backward-compatible at the public API level:
- The two existing public constructors (`string filePath`, `IEnumerable<ISingleChargeMs1Feature>`) are unchanged.
- `Features` and `DeconvolutionType` are unchanged.
- `OriginalFilePath` was `protected` — only derived/internal code could have observed it; no public callers exist in this repo. If any external consumer was reading it, switch to the new public `FilePath` property.
- One semantic change to note: two `FromFileDeconvolutionParameters` instances built with the in-memory feature ctor (no `FilePath`) are now `Equals`-equal under the config-only contract, even if their feature counts differ. They were `not-equal` under the old contract. Inserting a `FilePath` on either distinguishes them, which is the only config field the type contributes to the base class's equality scheme.

## Screenshots

_N/A — this is a non-visual performance/refactor change. No UI screenshots._


mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
```diff
@@ -55,27 +116,16 @@ public class FromFileDeconvolutionParameters : DeconvolutionParameters
...
-                    "Expected an Ms1Feature (.ms1.feature) or Dinosaur (.feature.tsv) file.");
-
-            OriginalFilePath = filePath;
-
-            resultFile.LoadResults();
-            (_featuresMzAscending, _mzKeys) = SortAndIndex(NormaliseRetentionTimes(featureFile.GetMs1Features()));
+            FilePath = filePath;
         }
 
         /// <summary>
```
trishorts, line 127: **Major** - Equals/GetHashCode still force full feature-file load, defeating the lazy-load goal

**Problem:**
`EqualProperties` and `AddHashCodes` both read `Features.Count` (lines 145 and 152). `Features` is now the lazy property `GetOrLoadFeatureCache().FeaturesMzAscending`, so any equality check or hash computation triggers `LoadFeaturesFromFile()` — reading and parsing the entire feature file from disk. The PR's own Motivation section names "equality checks, hash computation" as *configuration-only* paths that should no longer "pay the up-front load cost," but this code still pays it. Worse, `GetHashCode` can now perform disk I/O and can throw `MzLibException`/`InvalidOperationException` (from `LoadFeaturesFromFile`) if the path is unset or the file is not a recognized feature file — hashing should never throw or do I/O.

**Impact:**
The stated performance benefit is not realized for equality/hashing: inserting these parameter objects into a `HashSet`/`Dictionary`, or comparing two of them (e.g. during parameter-set dedup), reads every feature file eagerly. A `GetHashCode` call on a params object whose `FilePath` points at a missing/invalid file throws, which can destabilize any hashed collection containing it.

**Suggested fix:**
Key equality and hashing on `FilePath` alone (plus other config fields), or on a lightweight already-known count, so no load is forced. Only fall back to comparing `Features.Count` when both instances already have a materialized cache. Ensure `GetHashCode` is I/O-free and cannot throw.

**Affected test scenarios:**
- GetHashCode - instance constructed from FilePath, Features never accessed - should not read the file from disk
- Equals - two instances constructed from paths, Features never accessed - comparison should not trigger LoadFeaturesFromFile
- GetHashCode - instance with FilePath pointing at a non-feature file - must not throw

mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
```diff
@@ -33,18 +31,81 @@ public class FromFileDeconvolutionParameters : DeconvolutionParameters
...
         /// </summary>
-        public IReadOnlyList<ISingleChargeMs1Feature> Features => _featuresMzAscending;
+        [TomlIgnore]
+        public IReadOnlyList<ISingleChargeMs1Feature> Features => GetOrLoadFeatureCache().FeaturesMzAscending;
+
+        private string? _filePath;
+        public string? FilePath
+        {
+            get => _filePath;
+            set
```
trishorts, line 96: **Major** - `RetentionTimeNormalizedFromSeconds` is not `[TomlIgnore]` and breaks TOML round-trip

**Problem:**
This PR newly enables TOML (de)serialization of `FromFileDeconvolutionParameters` and applies `[TomlIgnore]` to the cache, lock, and `Features`. However `RetentionTimeNormalizedFromSeconds` (public getter, `private set`) is transient runtime state computed during feature load, yet it has no `[TomlIgnore]`. Nett serializes public-getter properties, so it will be written to the TOML output, and on deserialization its `private set` cannot be assigned by the serializer. This is exactly the "not configuration state and would be incorrectly round-tripped" category the PR set out to prevent.

**Impact:**
Serialized TOML contains a meaningless load-derived flag, and depending on Nett's behavior for read-restricted setters, deserialization either silently drops it or throws when mapping the property back — undermining the "FilePath is preserved, plumbing is absent" guarantee the PR claims.

**Suggested fix:**
Add `[TomlIgnore]` to `RetentionTimeNormalizedFromSeconds` (and audit other public-getter/restricted-setter members such as the overridden `DeconvolutionType` for the same round-trip hazard) so only genuine configuration state — notably `FilePath` — is serialized.

**Affected test scenarios:**
- TOML round-trip - serialize an instance after a seconds→minutes normalization - RetentionTimeNormalizedFromSeconds must not appear in output
- TOML round-trip - deserialize a config embedding these params - must not throw on the private-set property

mzLib/Test/MassSpectrometryTests/TestFromFileDeconvolution.cs
```diff
@@ -6,6 +6,7 @@
 using MassSpectrometry;
 using MzLibUtil;
 using NUnit.Framework;
+using NUnit.Framework.Legacy;
```
trishorts, line 9: **Major** - New FilePath setter / cache-invalidation behavior has no test

**Problem:**
The PR adds a public `FilePath` property whose setter "invalidates the cache so a new file can be assigned to an existing instance," and adds a lazy `FeatureCache` populated via double-checked locking. None of the tests in this file read `FilePath`, set `FilePath`, or verify that re-assigning it forces a re-read on the next `Features` access. The scope explicitly lists "FilePath set/invalidate" as an in-scope test behavior, but the diff adds no such coverage.

**Impact:**
Cache-invalidation bugs (stale features after `FilePath` reassignment, or failure to clear `_featureCache`) would pass CI undetected. The idempotent/double-checked-locking load path also has no test asserting the file is loaded exactly once.

**Suggested fix:**
Add a test that constructs from path A, accesses `Features`, sets `FilePath = pathB`, accesses `Features` again, and asserts the second access reflects path B's features (cache was invalidated and re-read).

**Affected test scenarios:**
- Set FilePath on an existing instance - after first Features access - next Features access re-reads the new file
- Repeated Features access on same instance - unchanged FilePath - features loaded once, same reference/content

trishorts, line 9: **Major** - TOML round-trip and [TomlIgnore] behavior is untested despite being a stated PR goal

**Problem:**
The PR's second stated motivation is TOML serializability: `FilePath` must round-trip through TOML while the `FeatureCache`, lock, and `Features` property are excluded via `[TomlIgnore]`. The `Nett 0.15.0` dependency was added to both `Readers.csproj` and `mzLib.nuspec` specifically to support this. No test serializes/deserializes a `FromFileDeconvolutionParameters` and asserts (a) `FilePath` survives the round-trip and (b) the cache/lock/`Features` are absent from the emitted TOML.

**Impact:**
A missing or misapplied `[TomlIgnore]` (e.g. the feature cache leaking into serialized output, or `FilePath` not being written) would ship undetected, directly defeating the MetaMorpheus regeneration workflow this PR is built for.

**Suggested fix:**
Add a TOML round-trip test using `Nett`: serialize an instance built from a file path, assert the emitted text contains `FilePath` and does not contain the cache/lock/features keys, then deserialize and assert `FilePath` is preserved.

**Affected test scenarios:**
- Serialize params from file-path ctor - emitted TOML - contains FilePath, omits cache/lock/Features
- Deserialize TOML - reconstructed params - FilePath preserved and features re-loadable

trishorts (commented):
Automated code review - 4 finding(s) at severity >= Major, posted as inline comments.

mzLib/Test/MassSpectrometryTests/TestFromFileDeconvolution.cs
```diff
@@ -439,8 +440,18 @@ public void FromFileDeconvolutionParameters_NonFeatureFile_ThrowsMzLibException(
             // message rather than silently producing an empty feature set.
             var psmTsv = Path.Combine(TestContext.CurrentContext.TestDirectory,
                 @"FileReadingTests\SearchResults\BottomUpExample.psmtsv");
-            Assert.Throws<MzLibUtil.MzLibException>(
-                () => new FromFileDeconvolutionParameters(psmTsv, minCharge: 1, maxCharge: 60));
+            try
```
trishorts: **Major** — Lazy-load deferral (the PR's central claim) is never actually tested

Both the constructor call and the `.Features` access sit inside the *same* `try` block here, so this test cannot distinguish whether the `MzLibException` was thrown **eagerly in the constructor** or **lazily on first `Features` access**. The PR's core claim — "no file I/O occurs in the constructor; the feature file is read/parsed only on first access" — has no assertion that proves the constructor did not already do the work. A regression that reverts to eager loading would leave every test in this file green, since they all eventually touch `Features`.

**Suggested fix:** split construction out of the `try` so the throw is provably from `.Features`, and add a test that constructs from a bad/non-existent path and asserts the constructor itself does **not** throw / does no I/O — then asserts the throw (or load) occurs only on first `Features` access.

mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
nbollis: Fixed in `1313a9a5` ("test(FromFileDecon): lock lazy-load + I/O-free equality + FilePath cache contracts"). `EqualProperties` and `AddHashCodes` now key on `FilePath` alone, dropping the `Features.Count` read. `GetHashCode` and `Equals` are now I/O-free and cannot throw, even on an instance whose `FilePath` points at a missing or non-feature file. The base class still contributes the other config fields (`MinAssumedChargeState`, `MaxAssumedChargeState`, `Polarity`, `ExpectedIsotopeSpacing`, `UseGenericScore`, `AverageResidueModel`), so equality is still a complete config-only identity.

Two new tests lock this in:
- `FromFileDeconvolutionParameters_GetHashCode_OnUnloadedInstance_DoesNotLoadFile` constructs from a missing path and asserts `GetHashCode()` does not throw and returns a non-trivial value.
- `FromFileDeconvolutionParameters_Equals_OnUnloadedInstances_DoesNotLoadFile` compares two instances from the same path without ever touching `.Features`.

Also updated `TestDeconvolutionParametersEqualityAndClone.FromFile_Equal_DifferentFeatureCount_AreNotEqual` to reflect the new contract (now `…AreEqualWhenNoFilePath`) and added `…DifferentFilePaths_AreNotEqual` for the FilePath-distinguishes-them half. Full test suite (4675 tests) passes.

mzLib/Readers/Deconvolution/FromFileDeconvolutionParameters.cs
nbollis: No longer applicable — the entire TOML surface has been removed from this PR in `b8bf59fe` ("revert(FromFileDecon): remove TOML/Nett surface from PR"). The Nett dependency is gone from `Readers.csproj` and `mzLib.nuspec`, the `using Nett;` is gone from `FromFileDeconvolutionParameters.cs`, and no `[TomlIgnore]` attributes are applied anywhere on the class. Because nothing in this PR is serialized to TOML, the load-derived `RetentionTimeNormalizedFromSeconds` and the overridden `DeconvolutionType` are no longer at risk of leaking into a TOML output. The PR is now a pure lazy-load + I/O-free-equality + FilePath refactor.

mzLib/Test/MassSpectrometryTests/TestFromFileDeconvolution.cs
nbollis: Fixed in `1313a9a5` ("test(FromFileDecon): lock lazy-load + I/O-free equality + FilePath cache contracts"). Two new tests cover the FilePath setter:

- `FromFileDeconvolutionParameters_SetFilePath_InvalidatesCache` materializes the cache from path A, assigns `FilePath = pathB`, accesses `.Features` again, and asserts the second access reflects path B (the cache was dropped, the new file was re-read).
- `FromFileDeconvolutionParameters_RepeatedFeaturesAccess_LoadsOnce` locks the double-checked-locking contract: three sequential `.Features` accesses on the same instance return the same list reference (no re-read, no re-parse).

nbollis: Withdrawn — the TOML round-trip motivation is no longer part of this PR. In `b8bf59fe` ("revert(FromFileDecon): remove TOML/Nett surface from PR"): the Nett package reference is gone from `Readers.csproj` and `mzLib.nuspec`, the `using Nett;` is gone from `FromFileDeconvolutionParameters.cs`, no `[TomlIgnore]` attributes are applied on the class, and the three TOML round-trip tests have been deleted from `TestFromFileDeconvolution.cs`. Since nothing in the PR is (de)serialized to TOML, the "FilePath must round-trip through TOML" goal is gone with it. The PR is now a pure lazy-load + I/O-free-equality + FilePath refactor. Full test suite (4675 tests) passes. If a TOML round-trip story is wanted in the future, it should be its own PR (it will need either a `[TomlConstructor]` factory or a parameterless ctor on `FromFileDeconvolutionParameters`, neither of which is in scope here).

mzLib/Test/MassSpectrometryTests/TestFromFileDeconvolution.cs
nbollis: Fixed in `1313a9a5` ("test(FromFileDecon): lock lazy-load + I/O-free equality + FilePath cache contracts"). `FromFileDeconvolutionParameters_NonFeatureFile_ThrowsMzLibException` is rewritten so the constructor and the `.Features` access are no longer in the same `try`:

```csharp
var parameters = new FromFileDeconvolutionParameters(psmTsv, minCharge: 1, maxCharge: 60);
// Reaching this line proves the constructor did not throw and did no I/O.
var ex = Assert.Throws<MzLibException>(() => _ = parameters.Features, ...);
StringAssert.Contains("MS1 feature file", ex.Message, ...);
```

So a regression that re-introduces eager loading in the constructor fails this test immediately. I also added a second test, `FromFileDeconvolutionParameters_NonExistentFile_ConstructorDoesNotThrow_LoadDoes`, using a path guaranteed not to exist (`Assume.That(File.Exists(...), Is.False)` as a guard) and asserting both halves: constructor returns silently, `.Features` throws. The exact exception type is relaxed to `Throws.Exception` (constraint syntax) because the load path can surface `FileNotFoundException` from `FileReader` or a downstream `MzLibException` depending on extension — the lazy-load contract only requires that the throw comes from the load, not the ctor.

trishorts (approved):
Approving — the substantive finding (Equals/GetHashCode forcing a full feature-file load) is genuinely fixed: equality and hashing now key on `FilePath` alone, are I/O-free, and can't throw on a missing/invalid path. The two TOML findings were honestly retired by reverting the Nett surface (`b8bf59fe`), and the lazy-load deferral is now provably tested (ctor split out of the `try`). Nice work.

One non-blocking simplification to consider (now, or in a follow-up):

Now that the TOML surface is gone, does `FilePath` still need to be a public *mutable* setter? The reason to make it settable was serializer-driven construction (parameterless ctor + property assignment during TOML deserialization) — and that's exactly what `b8bf59fe` removed. In the actual deconvolution workflows, a params object is constructed from a path and used; nothing reassigns `FilePath` to a *different* path mid-life. So the `_featureCache = null` invalidation logic (and the `SetFilePath_InvalidatesCache` test guarding it) is now machinery for a path nobody exercises.

Making `FilePath` `get`-only / `init`-only would delete the invalidation branch *and* the test that guards it — less surface, less to get wrong.

If you'd rather keep the mutable setter, one small thing: `SetFilePath_InvalidatesCache` doesn't actually assert invalidation. It reassigns the path and then only checks `secondFeatures.Count > 0`, so it would still pass if `_featureCache = null` were deleted from the setter (the stale cache is also non-empty). The test's own comment promises a content-fingerprint check that was never written — asserting `secondFeatures[0].Mz != firstFeatures[0].Mz` would actually lock it. Not a blocker either way.

RayMSMS (approved):

PR #1088, opened 2026-07-08
feat(usefulproteomicsdatabases): add PRIDE Archive REST client

trishorts:
**The nut:** mzLib can't reach the PRIDE Archive, the EBI proteomics data repository. This PR adds `PrideArchiveClient` — a small async REST client whose first capability answers the most useful question about a public dataset: *given a PRIDE accession, what files does it contain?* You hand it `PXD012345` and get back the complete, typed manifest of that project's files.

Builds on the now-merged `MzLibUtil.CvParam` (#1087) and adopts the `ExternalService` test convention from #1090.

## What it does
```csharp
using var pride = new PrideArchiveClient();
List<PrideArchiveFile> files = await pride.GetProjectFilesAsync("PXD012345");
// each file: name, size, checksum, category, download locations, dates
var ftp = files[0].PublicFileLocations.FirstOrDefault(l => l.Accession == "PRIDE:0000469"); // FTP by CV accession
```
- Calls PRIDE Archive REST **v3** (`projects/{accession}/files`).
- **Auto-pages** to return the whole manifest — paging is hidden from the caller (stops on the `total_records` header, a short page, or an empty page; a `MaxPages` cap guards a server that ignores paging).
- Deserializes into `PrideArchiveFile`, whose controlled-vocabulary fields (`FileCategory`, `PublicFileLocations`) use `MzLibUtil.CvParam`.

## Design
- **Instance `IDisposable` client owning one reusable `HttpClient`** (mirrors the Koina client), with a constructor overload that accepts an `HttpClient` for testing/config. The injected client is **caller-owned** — not disposed by us.
- **Newtonsoft.Json** (13.0.4, matching `PredictionClients`) with `NullValueHandling.Ignore`, so an explicit JSON `null` never clobbers a DTO's non-null default.
- **Error contract:** blank accession → `ArgumentException`; non-positive page size → `ArgumentOutOfRangeException`; non-success HTTP → `HttpRequestException`; unknown/empty project → **empty list, never null**.
- **Keys download locations on the stable CV accession** (`PRIDE:0000469` FTP, `PRIDE:0000468` Aspera), not the display name — not every file exposes HTTPS.

## Tests
`Test/DatabaseTests/PrideArchiveClientTests.cs`:
- **13 offline tests** driven by a stub `HttpMessageHandler` over recorded PRIDE wire payloads (deserialization, FTP-by-accession, multi-page concatenation via `total_records`, short-page/unparseable-header termination, the `MaxPages` cap, empty/null bodies, and every error path). No network — they run in the **required** unit-test job.
- **1 live canary** under `[Category("ExternalService")]`, routed through `ExternalServiceTestHelper.RunAsync("PRIDE", …)` so a PRIDE outage **skips** (not fails) in the dedicated non-blocking external-service job, while a genuine contract break still fails.

## Scope
New types + tests and one package reference; no existing code changed. Downloading file bytes, project metadata, and dataset search are intentionally left for follow-ups — this lands the manifest primitive they all build on.

https://claude.ai/code/session_01CaQVBed3CG6HuF8Nkpdzoo


RayMSMS (dismissed):

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="accession">The PRIDE project accession, e.g. "PXD012345".</param>
+        /// <param name="pageSize">Files requested per page (default 100). The full manifest is returned regardless.</param>
+        /// <returns>
+        /// The project's files. Empty if the project has no files or the accession is unknown (PRIDE
+        /// returns an empty result for an unknown accession). Never null.
+        /// </returns>
+        /// <exception cref="ArgumentException">The accession is null, empty, or whitespace.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">The page size is not positive.</exception>
+        /// <exception cref="HttpRequestException">The API returned a non-success status code.</exception>
+        public async Task<List<PrideArchiveFile>> GetProjectFilesAsync(string accession, int pageSize = 100)
```
nbollis: [review-oop] GetProjectFilesAsync lacks CancellationToken support

GetProjectFilesAsync does not accept a CancellationToken parameter. Callers cannot cooperatively cancel a long-running pagination sequence. The underlying HttpClient.GetAsync is called without a token (line 88), so the request itself cannot be cancelled.

Suggestion: Add an optional CancellationToken parameter (default CancellationToken.None) and pass it to _httpClient.GetAsync(requestUri, cancellationToken). Insert cancellationToken.ThrowIfCancellationRequested() before each page iteration.

*Consensus across 2 specialists (review-oop, review-generic).*

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                foreach (string value in values)
+                {
+                    if (long.TryParse(value, out total))
+                        return true;
+                }
+            }
+            return false;
+        }
+
+        public void Dispose()
```
nbollis, line 140: [review-oop] Dispose not virtual on non-sealed class violates standard Dispose pattern

PrideArchiveClient is public and non-sealed (line 19) but Dispose() is not virtual. The standard .NET IDisposable pattern (CA1063) requires non-sealed classes to implement 'protected virtual void Dispose(bool disposing)' so derived classes can participate in cleanup. A subclass that needs to release its own managed resources cannot override Dispose().

Suggestion: Either seal the class (add 'sealed' on line 19) or refactor to the standard Dispose(bool) pattern: add 'protected virtual void Dispose(bool disposing)', move HttpClient disposal logic there, and have Dispose() call Dispose(true) + GC.SuppressFinalize(this).

nbollis, line 140: [review-oop] Missing GC.SuppressFinalize(this) in Dispose method

The Dispose method does not call GC.SuppressFinalize(this). While the class currently has no finalizer, the standard pattern (CA1816) recommends calling it to protect against future changes where a subclass might add a finalizer.

Suggestion: Add 'GC.SuppressFinalize(this);' at the end of the Dispose method, after _disposed = true.

*Consensus across 2 specialists (review-oop, review-generic).*

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            if (_disposed)
```
nbollis, line 142: [review-oop] Thread-unsafe _disposed flag can cause double-dispose race

The _disposed check-and-set at lines 136-140 is not thread-safe. Two concurrent calls to Dispose() can both observe _disposed == false and both attempt to execute HttpClient.Dispose().

Suggestion: Use Interlocked.CompareExchange(ref _disposed, true, false) for atomic check-and-set, or guard with a lock statement. Alternatively, document this limitation.

*Consensus across 2 specialists (review-oop, review-generic).*

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                string requestUri = $"projects/{Uri.EscapeDataString(accession)}/files?pageSize={pageSize}&page={page}";
+                using HttpResponseMessage response = await _httpClient.GetAsync(requestUri).ConfigureAwait(false);
+
+                if (!response.IsSuccessStatusCode)
+                    throw new HttpRequestException(
+                        $"PRIDE Archive request failed with status {(int)response.StatusCode} {response.ReasonPhrase} for '{requestUri}'.");
+
+                string content = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                List<PrideArchiveFile> pageFiles =
```
nbollis, line 100: [review-oop] Direct coupling to Newtonsoft.Json violates Dependency Inversion Principle

Lines 95-96 call JsonConvert.DeserializeObject directly, creating tight coupling to Newtonsoft.Json. Makes it impossible to swap serialization strategies without modifying this method.

Suggestion: Introduce an IJsonSerializer abstraction with Deserialize<T>(string) and inject it via the private constructor. Default implementation uses Newtonsoft.Json; tests can supply a stub.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// A hard upper bound on the number of pages fetched, guarding against a misbehaving server that
+        /// ignores the paging parameters. No real PRIDE project approaches this; exceeding it throws.
+        /// </summary>
+        public int MaxPages { get; init; } = 10000;
+
+        /// <summary>Creates a client with its own <see cref="HttpClient"/> pointed at the PRIDE Archive API.</summary>
+        public PrideArchiveClient()
+            : this(new HttpClient { BaseAddress = new Uri(DefaultBaseAddress), Timeout = TimeSpan.FromSeconds(100) }, ownsHttpClient: true)
```
nbollis, line 40: [review-oop] Default constructor creates HttpClient directly without IHttpClientFactory

Line 39 uses 'new HttpClient { ... }' directly. The recommended pattern in .NET Core/.NET 5+ is IHttpClientFactory to manage HttpClient lifecycle and avoid socket exhaustion in long-running services.

Suggestion: For the default constructor, either accept an IHttpClientFactory via constructor injection, or document that callers should use the HttpClient overload in DI-managed scenarios.

mzLib/UsefulProteomicsDatabases/PrideArchiveFile.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>The file category as a controlled-vocabulary term (e.g. value "RAW", "SEARCH", "PEAK").</summary>
+        public CvParam FileCategory { get; set; } = new();
+
+        /// <summary>
+        /// The public download locations as controlled-vocabulary terms; each term's
+        /// <see cref="CvParam.Value"/> is a URL and its <see cref="CvParam.Accession"/> identifies the
+        /// protocol (FTP "PRIDE:0000469", Aspera "PRIDE:0000468").
+        /// </summary>
+        public List<CvParam> PublicFileLocations { get; set; } = new();
```
nbollis, line 36: [review-oop] Public setter on List<CvParam> collection allows external replacement

PublicFileLocations exposes a public setter on List<CvParam>, allowing external code to replace the entire collection reference. This breaks encapsulation and can cause unexpected null reference issues.

Suggestion: Change the setter to 'private set' and consider exposing the getter as IReadOnlyList<CvParam> to signal intent.

nbollis (commented):
## Code Review: PRIDE Archive REST Client

Automated multi-specialist review (review-oop, review-generic). 16 inline findings across 2 severity tiers.

**Scope:** New `PrideArchiveClient` (async REST client), `PrideArchiveFile` DTO, and 14 tests. All-new code, no existing code changed.

**Key findings:**
- 2 major: missing CancellationToken support, non-virtual Dispose on non-sealed class
- 10 minor: missing GC.SuppressFinalize, thread-unsafe _disposed, Newtonsoft.Json coupling, no IHttpClientFactory, public List setter, missing XML doc on Dispose, error message missing accession, 5 test coverage gaps
- 4 suggestions: no interface for DI, missing User-Agent header, 2 more test coverage gaps

**Note:** review-backend specialist timed out; findings from review-oop and review-generic comprehensively cover API design, error handling, Dispose pattern, thread safety, and test gaps.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// A client for the PRIDE Archive REST API (https://www.ebi.ac.uk/pride/ws/archive/v3/) that makes
+    /// EBI PRIDE proteomics dataset data available to mzLib. The initial capability, given a project
+    /// accession (e.g. "PXD012345"), is to return the complete manifest of that project's files.
+    /// </summary>
+    /// <remarks>
+    /// Holds one reusable <see cref="HttpClient"/> and is <see cref="IDisposable"/>. Use one instance
+    /// per unit of work and dispose it. A constructor overload accepts an <see cref="HttpClient"/> to
+    /// support testing and custom configuration.
+    /// </remarks>
+    public class PrideArchiveClient : IDisposable
```
nbollis: [review-oop] No interface for PrideArchiveClient limits testability and DI

PrideArchiveClient implements IDisposable but defines no domain-specific interface (e.g. IPrideArchiveClient). Callers must depend on the concrete class, making it harder to mock in unit tests.

Suggestion: Extract an IPrideArchiveClient interface with the public members. Have PrideArchiveClient implement it.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
nbollis, line 140: [review-generic] Missing XML documentation on Dispose method

All other public members have XML doc comments. Dispose() is the only public member without documentation or <inheritdoc />, which is inconsistent.

Suggestion: Add an <inheritdoc /> tag on Dispose() so it inherits the IDisposable interface documentation.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            var files = new List<PrideArchiveFile>();
+            int page = 0;
+
+            while (true)
+            {
+                string requestUri = $"projects/{Uri.EscapeDataString(accession)}/files?pageSize={pageSize}&page={page}";
+                using HttpResponseMessage response = await _httpClient.GetAsync(requestUri).ConfigureAwait(false);
+
+                if (!response.IsSuccessStatusCode)
+                    throw new HttpRequestException(
```
nbollis, line 96: [review-generic] HTTP error message omits raw accession parameter

The error message includes the requestUri (which has the URL-encoded accession) but not the original accession parameter. Having the raw accession explicitly would make log parsing and debugging easier.

Suggestion: Include the accession parameter in the error message, e.g.: "PRIDE Archive request failed with status {code} for accession '{accession}' (URI: '{requestUri}')."

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
nbollis, line 40: [review-generic] HTTP requests lack a User-Agent header

The default HttpClient does not set a User-Agent header. Setting a User-Agent is best practice for API clients so server operators can identify the client software.

Suggestion: Set a default User-Agent header, e.g. client.DefaultRequestHeaders.UserAgent.ParseAdd("mzLib-PrideArchiveClient/1.0").

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
```diff
@@ -0,0 +1,259 @@
...
+    // ---- construction / lifetime -------------------------------------------
+
+    [Test]
+    public void Constructor_NullHttpClient_ThrowsArgumentNullException()
+    {
+        Assert.That(() => new PrideArchiveClient(null), Throws.ArgumentNullException);
+    }
+
+    [Test]
+    public void DefaultConstructor_And_Dispose_DoNotThrow()
```
nbollis, line 230: [review-generic] Missing test: injected HttpClient survives Dispose

The constructor overload PrideArchiveClient(HttpClient) documents that 'The caller retains ownership: the supplied client is NOT disposed.' However, there is no test verifying this contract.

Suggestion: Add a test that creates a client with an injected HttpClient, calls Dispose, and verifies the injected HttpClient is still usable.

nbollis, line 230: [review-generic] Missing test: double Dispose is safe

The Dispose method has a guard for double calls (if (_disposed) return), but there is no test verifying that calling Dispose twice does not throw.

Suggestion: Add a test (Dispose_CalledTwice_DoesNotThrow) that calls Dispose twice and asserts no exception.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
```diff
@@ -0,0 +1,259 @@
...
+
+    [Test]
+    public void GetProjectFilesAsync_BlankAccession_ThrowsArgumentException()
+    {
+        using var client = new PrideArchiveClient(new HttpClient(new StubHandler(_ => JsonResponse("[]"))));
+        Assert.That(async () => await client.GetProjectFilesAsync("   "), Throws.ArgumentException);
+    }
+
+    [Test]
+    public void GetProjectFilesAsync_NonPositivePageSize_ThrowsArgumentOutOfRange()
```
nbollis: [review-generic] PageSize validation test covers 0 but not negative values

GetProjectFilesAsync_NonPositivePageSize_ThrowsArgumentOutOfRange tests pageSize=0, but the validation check is pageSize <= 0. Negative values are also rejected but not covered.

Suggestion: Use a parameterized test ([TestCase(0)], [TestCase(-1)], [TestCase(-100)]) to cover the full range of non-positive page sizes.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
```diff
@@ -0,0 +1,259 @@
...
+    public void GetProjectFilesAsync_NonSuccessStatus_ThrowsHttpRequestException()
+    {
+        var handler = new StubHandler(_ => JsonResponse("server error", HttpStatusCode.InternalServerError));
+        using var client = new PrideArchiveClient(new HttpClient(handler));
+        Assert.That(async () => await client.GetProjectFilesAsync("PXD012345"), Throws.InstanceOf<HttpRequestException>());
+    }
+
+    // ---- construction / lifetime -------------------------------------------
+
+    [Test]
```
nbollis, line 223: [review-generic] Missing test: injected HttpClient with null BaseAddress gets default

The private constructor sets _httpClient.BaseAddress to DefaultBaseAddress when the injected HttpClient has no BaseAddress. This fallback path is untested.

Suggestion: Add a test that injects an HttpClient without a BaseAddress, then verifies that the request URI sent is rooted at DefaultBaseAddress.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
```diff
@@ -0,0 +1,259 @@
...
+        });
+        using var client = new PrideArchiveClient(new HttpClient(handler));
+
+        var files = await client.GetProjectFilesAsync("PXD012345", pageSize: 2);
+
+        Assert.That(files.Count, Is.EqualTo(1));
+    }
+
+    [Test]
+    public void GetProjectFilesAsync_ServerIgnoresPaging_ThrowsAfterMaxPages()
```
nbollis, line 147: [review-generic] Missing test for MaxPages edge values (0, 1)

MaxPages is only tested with a value of 3. Edge cases like MaxPages = 0 (silently returns empty list) and MaxPages = 1 (only page 0) are untested. A value of 0 may silently return an empty list rather than throwing.

Suggestion: Add tests for MaxPages = 0 and MaxPages = 1. Consider adding validation (ArgumentOutOfRangeException) for MaxPages <= 0.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Fair point. The download methods added in the stacked #1093 (`DownloadFileAsync` / `DownloadProjectFilesAsync`) already thread a `CancellationToken`; extending it to `GetProjectFilesAsync` — passing it to `GetAsync` and adding a `ThrowIfCancellationRequested` per page — is a reasonable consistency improvement. Leaving this open to track.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Agreed the standard pattern applies. This client isn't designed for inheritance (it owns exactly one `HttpClient`), so the cleaner resolution is to `seal` the class rather than adopt a virtual `Dispose(bool)`. Leaving open to seal it.

trishorts: The class has no finalizer, so `GC.SuppressFinalize(this)` is a no-op today. Keeping `Dispose` minimal and will revisit only if a finalizer is ever introduced. Resolving.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: By contract this client is used per unit of work and disposed once (see the class remarks), not shared for concurrent disposal — and `HttpClient.Dispose` is itself idempotent — so the check-and-set race is benign. Leaving the simple flag as-is. Resolving.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: mzLib uses Newtonsoft.Json directly throughout; introducing an `IJsonSerializer` abstraction here would be inconsistent with the rest of the codebase for no practical gain, and deserialization is already isolated to this one method. Resolving as won't-fix.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: mzLib is a library, not a DI-hosted service, so `IHttpClientFactory` isn't part of its model. For managed-lifetime scenarios the class already accepts an injected `HttpClient` (documented on that overload); the default constructor is a convenience for simple callers. Resolving.

mzLib/UsefulProteomicsDatabases/PrideArchiveFile.cs
trishorts: `PrideArchiveFile` is a wire DTO populated by Newtonsoft deserialization, which needs a settable concrete collection; public setters match how mzLib's other deserialized models are written. Resolving as intended.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: mzLib doesn't extract an interface per client. Testability is already covered: the `HttpClient`-injection constructor lets tests supply a stub `HttpMessageHandler` (see `PrideArchiveClientTests`), which is how the suite mocks the transport. Resolving.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Trivially fair — will add `<inheritdoc/>` to `Dispose()`. Leaving open.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: The message already includes the request URI, which carries the (escaped) accession, so the failure is diagnosable. Resolving; the raw accession can be added later if log-grepping needs it.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: PRIDE's API doesn't require a `User-Agent`, and mzLib's other HTTP clients don't set one, so this isn't needed for correctness. A courtesy UA is a reasonable nicety but optional. Resolving as won't-fix.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Good gap — the "injected `HttpClient` is not disposed by the client" contract deserves an explicit test (dispose the client, then confirm the injected `HttpClient` is still usable). Leaving open to add.

trishorts: Agreed; `DefaultConstructor_And_Dispose_DoNotThrow` disposes once — a second `Dispose()` should be asserted no-throw as well. Leaving open to add.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Fair — will extend `GetProjectFilesAsync_NonPositivePageSize_ThrowsArgumentOutOfRange` with a `[TestCase(-1)]` alongside 0. Leaving open.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Reasonable — will add a test that an injected `HttpClient` with no `BaseAddress` gets the PRIDE default set by the constructor. Leaving open.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: The cap is exercised at `MaxPages = 3`; adding 0/1 edge cases is a reasonable boundary check. Leaving open to add.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Done in 643a10d7 — `GetProjectFilesAsync` now takes an optional `CancellationToken`, checks it per page, and passes it to `GetAsync` / `ReadAsStringAsync`.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Done in 643a10d7 — sealed `PrideArchiveClient` (it is not designed as a base type).

trishorts: Done in 643a10d7 — added `<inheritdoc/>` to `Dispose()`.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Done in 643a10d7 — added `Dispose_InjectedHttpClient_IsNotDisposed_AndRemainsUsable` (disposes the client, then confirms the injected `HttpClient` still serves a request).

trishorts: Done in 643a10d7 — added `Dispose_CalledTwice_DoesNotThrow`.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Done in 643a10d7 — `GetProjectFilesAsync_NonPositivePageSize_ThrowsArgumentOutOfRange` now runs `[TestCase(0)]` and `[TestCase(-1)]`.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Done in 643a10d7 — added `Constructor_InjectedHttpClientWithoutBaseAddress_SetsPrideDefault`.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Done in 643a10d7 — added `GetProjectFilesAsync_ServerIgnoresPaging_LowMaxPages_Throws` with `[TestCase(0)]` / `[TestCase(1)]`.

nbollis (approved):

RayMSMS (approved):

PR #1089, opened 2026-07-09
Revert MCDF upgrade

nbollis:
ThermoRawFileReader uses OpenMCDF version 2.3.X 

Dependabot upgraded this and we approved. 

Need to revert to restore full file reader functionality.

### DIFF
--- a/mzLib/Readers/Readers.csproj
+++ b/mzLib/Readers/Readers.csproj
@@ -10,7 +10,7 @@
 
 	<ItemGroup>
 		<PackageReference Include="CsvHelper" Version="32.0.3" />
-		<PackageReference Include="OpenMcdf" Version="3.1.3" />
+		<PackageReference Include="OpenMcdf" Version="2.3.1" />
 		<PackageReference Include="OpenMcdf.Extensions" Version="2.3.1" />
 		<PackageReference Include="System.Data.SQLite" Version="1.0.118" />
 		<PackageReference Include="System.Runtime.InteropServices" Version="4.3.0" />
--- a/mzLib/mzLib.nuspec
+++ b/mzLib/mzLib.nuspec
@@ -21,7 +21,7 @@
     <dependency id="Microsoft.ML.FastTree" version="2.0.0" />
     <dependency id="Microsoft.Win32.Registry" version="5.0.0" />
     <dependency id="NetSerializer" version="4.1.1" />
-    <dependency id="OpenMcdf" version="3.1.3" />
+    <dependency id="OpenMcdf" version="2.3.1" />
     <dependency id="OpenMcdf.Extensions" version="2.3.1" />
     <dependency id="SharpLearning.Optimization" version="0.28.0" />
     <dependency id="SharpZipLib" version="1.4.2" />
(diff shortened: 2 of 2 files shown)


trishorts (approved):

trishorts:
This is the right call, Nic — thanks for catching it. Adding some context for the record, plus two small suggestions so Dependabot doesn't walk us back into it.

### Why the revert is correct (and not just a stopgap)

The Thermo reader binary `ThermoFisher.CommonCore.RawFileReader.dll` is **compiled against OpenMcdf 2.x** (I checked the DLL's referenced assemblies — it links `OpenMcdf 2.1.6` / `OpenMcdf.Extensions 2.1.6`). OpenMcdf **3.x is an API-incompatible rewrite** with a different namespace, so overriding to 3.x means the Thermo reader calls into methods that no longer exist → raw file reading breaks. You can't swap a checked-in binary's dependency across a major version, which is exactly why Dependabot's 3.x bump wasn't actually usable.

Worth knowing: there is **no OpenMcdf version that is both patched and Thermo-compatible**. The advisory ([GHSA-5qwm-7pvp-w988](https://github.com/advisories/GHSA-5qwm-7pvp-w988) / CVE-2026-45785) is fixed only in **3.1.4**, but even Thermo's newest release (`8.0.37`, net8) still ships against **OpenMcdf 2.3.0**. So 2.x is the only option until Thermo themselves recompile RawFileReader against 3.x. `2.3.1` is the right pin — it's the highest 2.x and API-compatible with what the Thermo binary expects.

And the advisory doesn't really apply to us anyway: it's a DoS that only triggers on **crafted, untrusted CFB files**. We only ever open trusted local instrument files through this path, never attacker-controlled input.

### Suggestion 1 — leave a breadcrumb in `Readers.csproj`

So the next person (or the next Dependabot PR) doesn't "helpfully" re-bump it:

```xml
<!-- OpenMcdf is intentionally pinned to the 2.x line. DO NOT upgrade to 3.x.
     ThermoFisher.CommonCore.RawFileReader.dll is compiled against OpenMcdf 2.x, and
     OpenMcdf 3.x is an API-incompatible rewrite that breaks Thermo .raw reading.
     Thermo's latest release (8.0.37) still ships against OpenMcdf 2.3.0, so there is
     no 3.x-based option today. GHSA-5qwm-7pvp-w988 (fixed in 3.1.4) is a DoS on
     crafted/untrusted CFB files and does not apply — we only read trusted local files. -->
<PackageReference Include="OpenMcdf" Version="2.3.1" />
<PackageReference Include="OpenMcdf.Extensions" Version="2.3.1" />
```

### Suggestion 2 — stop the recurring security PR

Since the upgrade came from a **Dependabot security update** (there's no `dependabot.yml` in the repo), the surgical fix is to **dismiss the OpenMcdf alert** under *Security → Dependabot alerts* with reason *"No fix available that's compatible"* — that stops the auto-PR without any config changes.

If you'd rather encode it in-repo, a `.github/dependabot.yml` ignore works too (note this also switches on scheduled NuGet version-update PRs for the repo):

```yaml
version: 2
updates:
  - package-ecosystem: "nuget"
    directory: "/mzLib"
    schedule:
      interval: "weekly"
    ignore:
      # Must stay on OpenMcdf 2.x — Thermo RawFileReader is compiled against 2.x (see Readers.csproj)
      - dependency-name: "OpenMcdf"
        versions: [">=3.0.0"]
      - dependency-name: "OpenMcdf.Extensions"
        versions: [">=3.0.0"]
```

Either one is fine — dismissing the alert is the lower-friction option. 👍

trishorts (approved):
Approving — this is the correct fix. OpenMcdf must stay on 2.x because the Thermo RawFileReader binary is compiled against it and 3.x is an API-incompatible rewrite that breaks .raw reading; there's no patched-and-compatible OpenMcdf until Thermo recompiles (their latest 8.0.37 still ships against 2.3.0). The reverted-in advisory (GHSA-5qwm-7pvp-w988) is a DoS on untrusted CFB files and doesn't apply to our trusted-local-file path. See my comment above for the optional csproj breadcrumb + Dependabot dismissal so this doesn't recur.

zhuoxinshi (approved):

PR #1091, opened 2026-07-09
Upgrade Thermo CommonCore RawFileReader to 8.0.37 (net8)

trishorts:
## What this does

Upgrades the checked-in Thermo CommonCore RawFileReader binaries from **5.0.0.7** (netstandard2.0) to the current **8.0.37** native **.NET 8** build, and adds tests that guard against any behavioral regression from the swap.

Sourced from Thermo's official [thermofisherlsms/RawFileReader](https://github.com/thermofisherlsms/RawFileReader) repo (`Libs/NetCore/Net8`, released 2026-05-19).

## Why

- **Native net8 alignment.** We currently ship netstandard2.0 shim DLLs into a net8.0 project; 8.0.37 is a true net8 assembly and is Thermo's recommended target.
- **Newer instrument/format support.** Thermo added new scan-filter fields since our old build — better metadata fidelity for data from newer instruments.
- **Perf.** The net8 build uses memory-mapped file I/O for raw access.
- **Cleaner deps.** Adds no runtime packages (CommandLineParser is example-only). The DLL binds OpenMcdf **2.3.0**, closer to our pin than the old 2.1.6.

## OpenMcdf pin (overlaps #1089)

8.0.37 is compiled against OpenMcdf 2.x, so this PR also pins `OpenMcdf` back to **2.3.1** — the same change as #1089. OpenMcdf **3.x is an API-incompatible rewrite** that breaks `.raw` reading, and there is no 3.x-compatible Thermo build. [GHSA-5qwm-7pvp-w988](https://github.com/advisories/GHSA-5qwm-7pvp-w988) (fixed in 3.1.4) is a DoS on **crafted/untrusted** CFB files and does not apply — we only read trusted local instrument files. A code comment in `Readers.csproj` now records this so it isn't "helpfully" re-bumped. If #1089 merges first, the OpenMcdf lines resolve trivially (identical change).

## Tests

New `TestRawFileReaderVersionCompat` exercises **every** CommonCore API surface `ThermoRawFileReader` depends on and asserts version-independent invariants, so a bad reader bump fails loudly instead of silently corrupting spectra:

- **Deployed-version canary** — fails if the upgraded 8.x DLL didn't reach the output dir.
- **Static-load invariant sweep** across 6 representative `.raw` files — MS-order/polarity/mass-analyzer enum mappings, centroid + `PreferredMasses` extraction, zero-intensity stripping, TIC, and MS2 precursor resolution.
- **Trailer-extra parsing** across the corpus (injection time, charge state, monoisotopic m/z) — the most format-sensitive surface.
- **Static vs. dynamic read parity** — two independent CommonCore code paths must agree scan-for-scan.

### Verification (local)

| Reader | Result |
|---|---|
| 5.0.0.7 baseline | `TestRawFileReader` — **18/18 pass** |
| 8.0.37 (this PR) | `TestRawFileReader*` — **29/29 pass** (18 existing + 11 new) |

No behavioral regressions; raw reads passing also confirms the OpenMcdf 2.3.1 binding works at runtime.


trishorts:
Upstream tracking: I've filed an issue asking Thermo to move CommonCore onto a patched OpenMcdf (3.1.4+) so the transitive advisory can be cleared downstream — thermofisherlsms/RawFileReader#22. Until they recompile against 3.x (an API-incompatible rewrite), the OpenMcdf 2.3.1 pin in this PR is the correct state.

nbollis (approved):

PR #1095, opened 2026-07-10
fix(variants): read UniProt-native deletions (empty <variation>)

trishorts:
## What

Fixes the `ProteinXmlEntry` reader so it reads **UniProt-native deletions**, which UniProt encodes as an empty `<variation/>` (e.g. `<original>P</original><variation/>`). The reader previously admitted a `sequence variant` feature only when `<variation>` was non-empty, so **every UniProt-native deletion was silently dropped** — no `SequenceVariation` was ever created. This is not something we can push onto the encoding: it is how UniProt writes a deletion.

The guard now admits any feature that encodes a real change (not both `<original>` and `<variation>` empty), so deletions are read alongside substitutions and insertions. `SequenceVariation`, the writer, and the round-trip already support an empty `VariantSequence`, so no other production change is needed.

## Corpus (L1 deletions — executable spec, oracle derived from the domain model, red-before-fix)

New nodes in `Test/DatabaseTests/VariantCorpus`:

- **D00–D01** — single- and multi-base deletion, unmodified
- **D02–D04** — deletion **before / on / after** a PTM residue (phospho re-anchors ↓ / drops / stays)
- **D05–D06** — deletion **out of range** / straddling the C-terminus → pruned
- **D07–D08** — the **VCF depth cutoff** (`minAlleleDepth`), same deletion passing at cutoff 5 (2 forms) / filtered at 7 (1 form)
- **MD00–MD02** — **pairs** of separate deletions (unmodified, flanking a PTM, one-on-the-PTM), capped at 2 to stay under the combination cap

D00–D04 fail on master (reproducing the reader gap) and go green with the fix. The harness gains `BEGIN`/`END`, empty-`VAR` (deletion), `SRC=vcf` (GT/AD/DP) support, and a per-case `MinAlleleDepth`.

## Buggy fixture the drop was masking

Admitting the deletions surfaced a latent error in `seqvartests.xml`: the `YPSE` deletion declared `begin=960`, but `YPSE` sits at **961–964** (the base is `NYPSE` at 960–964), violating `SeqVarXmlTest`'s own invariant that `OriginalSequence` equals the residues at `[begin,end]`. It never ran because the variant was being dropped. Corrected `begin` 960→961.

## Verification

- 34/34 `VariantCorpus` (incl. `ProteinDbWriter` round-trip + run-twice determinism)
- 666/666 `DatabaseTests` + `Transcriptomics`


trishorts:
### Manual unit-test review

All 14 unit test(s) added by this PR were reviewed individually by the author (not only by automated/AI-generated checks): 13 approved as written, 1 edited, 0 flagged.

This satisfies the author's standing AI-use policy that AI-assisted tests receive human review before relying on them. Recorded 2026-07-11 06:54.

zhuoxinshi (approved):

Alexander-Sol (approved):

PR #1099, opened 2026-07-18
feat: fetch PRIDE spectra by USI via PROXI (MzSpectrum + MsDataScan)

trishorts:
## The nut

Given a **USI** (Universal Spectrum Identifier), fetch a spectrum from PRIDE and get back something that **is** an mzLib `MzSpectrum` — and, when you want the metadata too, a full `MsDataScan`. The next capability on the PRIDE client after the merged manifest MVP (#1088) and file download (#1093).

## What's here

`PrideProxiSpectrum` **derives from `MzSpectrum`**. A spectrum fetched from PRIDE can be handed straight to anything in mzLib that takes an `MzSpectrum`, with no conversion step. On top of the peaks it carries what an `MzSpectrum` cannot: the `Usi` it was resolved from, its PROXI `Status`, and the controlled-vocabulary `Attributes` (`MzLibUtil.CvParam`, exactly as `PrideArchiveFile` uses them).

New API on `PrideArchiveClient`:
- **`GetProxiSpectrumAsync(usi)` → `PrideProxiSpectrum`** — the spectrum plus its CV attributes.
- **`GetSpectrumAsync(usi)` → `MzSpectrum`** — the simple case; returns that same object narrowed to the base type.

And the piece that turns the CV terms into something mzLib can use:
- **`PrideArchiveExtensions.ToMsDataScan()`** — a pure, offline transform reading the PROXI attributes into a fully-populated `MsDataScan`: scan number, ms level, polarity, retention time, centroid-ness, precursor m/z and charge, isolation window, injection time, filter string. The spectrum itself is carried through as the scan's `MassSpectrum` rather than re-wrapped.

```csharp
using var client = new PrideArchiveClient();
MsDataScan scan = (await client.GetProxiSpectrumAsync(usi)).ToMsDataScan();
```

## Design notes

**Why `[JsonConstructor]`.** `MzSpectrum` has no parameterless constructor and keeps `XArray`/`YArray` private to set, so the peaks must reach the base constructor rather than being assigned by the deserializer afterwards. The DTO is immutable as a result. It sorts ascending on construction (`MzSpectrum` assumes sorted m/z and does not sort itself — the `LibrarySpectrum` pattern) and still rejects a `mzs`/`intensities` length mismatch. That guard now runs *inside* a constructor Newtonsoft invokes, so the contract was verified rather than assumed: the `MzLibException` **surfaces unwrapped**, not as a `JsonSerializationException`. There is a test pinning it.

**The CV mapping was built against a live response, not the spec** — which changed the design:

| Field | PSI form expected | What PRIDE actually sends |
|---|---|---|
| Polarity | `MS:1000130` / `MS:1000129` | `MS:1000465` carrying the value `"positive scan"` |
| Mass analyzer | `MS:1000484` etc. | *nothing* — only `MS:1000463 instrument` |
| Dissociation | `MS:1000422` etc. | *nothing* — only inside the vendor filter string |

Both spellings of polarity and centroid-ness are read, since PROXI is repository-agnostic and another server (PeptideAtlas, MassIVE) may send the bare terms.

**`ToMsDataScan` lives beside the DTO in `UsefulProteomicsDatabases`, not in `Readers`.** `Readers` references `UsefulProteomicsDatabases`, not the reverse, so `Readers.Mzml`'s CV dictionaries are not reachable from here — and per the table above, all three of them miss on PRIDE's payload anyway, so there is essentially nothing to reuse and nothing duplicated.

**Fields PROXI does not state are left `Unknown`/`null` rather than guessed** — `MZAnalyzerType.Unknown`, `DissociationType.Unknown`, no noise data, no precursor scan number. The filter string does encode the dissociation method (`hcd32.00`), but parsing vendor filter strings is too fragile to bury in a converter; it is preserved verbatim as `ScanFilter` so a caller can do that deliberately.

**Retention time is normalized to minutes**, mzLib's convention everywhere — per its unit term when one is present, and as seconds when absent, which is what PRIDE serves. A live scan reports `4662.32`; read as minutes that would be 77 hours rather than 77 minutes.

**Defaults where a term is absent:** ms level 2 and scan number 1. A USI addresses a fragment spectrum, so 2 is the meaningful default (the same one `Mgf` hardcodes). PRIDE sends `MS:1000511` in practice.

**Other notes.** PROXI lives under a different path root than the v3 archive API (`/pride/proxi/archive/v0.1/` vs `/pride/ws/archive/v3/`), so the request is issued as an absolute URI rather than resolving against the client's archive `BaseAddress`. Adds a `MassSpectrometry` project reference to `UsefulProteomicsDatabases` (no cycle — `MassSpectrometry` depends only on `Chemistry` + `MzLibUtil`). A repeated attribute is read first-wins, since PRIDE legitimately sends one `instrument` term per instrument in the originating project.

## Verified against the live API

Probed the real endpoint before writing the DTO and again before writing the CV mapping: `resultType=full` returns a **1-element JSON array** of `{status, usi, mzs[], intensities[], attributes[cvParam]}`; an unknown/unreadable USI returns **404** and a malformed USI **400**, both surfacing as `HttpRequestException` through the existing non-success path.

## Tests

- **22 offline** fixture-based tests: deserialization, URL construction, the error contract, the constructor's sort/length-mismatch/null/empty edges, and `ToMsDataScan` across the full realistic attribute set, the unstated-field defaults, both polarity/centroid spellings, retention-time units, a repeated term, and non-numeric CV values.
- **2 live canaries** carrying `[Category("ExternalService")]` + `[Category("Pride")]`, routed through `ExternalServiceTestHelper` so a PRIDE/EBI outage **skips** while a genuine contract break **fails** — same convention as #1090/#1088/#1093. The canary now also asserts `ToMsDataScan` against the real payload (scan 17555, charge 2, positive polarity, RT > 0), so a PRIDE contract change fails loudly instead of silently degrading to defaults.

All 81 offline PRIDE tests and both live canaries green; full solution builds clean.

## Review history

Revised after review: `PrideProxiSpectrum` was originally a plain DTO with a `ToMzSpectrum()` extension. Per @Alexander-Sol's feedback it is now an `MzSpectrum` subclass, and `ToMzSpectrum` is removed as unnecessary. `ToMsDataScan` is new in the same revision. **The DTO's shape changed (it is now immutable, constructed rather than object-initialized), so this needs a fresh look from earlier approvers.**

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
+++ b/mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using MassSpectrometry;
 using Newtonsoft.Json;
 
 namespace UsefulProteomicsDatabases
@@ -23,6 +24,14 @@ namespace UsefulProteomicsDatabases
     {
         /// <summary>The base address of the PRIDE Archive REST API (v3).</summary>
         public const string DefaultBaseAddress = "https://www.ebi.ac.uk/pride/ws/archive/v3/";
+
+        /// <summary>
+        /// The base address of PRIDE's PROXI spectrum API. PROXI (the PSI standard for retrieving a spectrum by
+        /// USI) lives under a DIFFERENT path root than the v3 archive API (<see cref="DefaultBaseAddress"/>), so
+        /// PROXI requests are issued as absolute URIs and do not resolve against the client's archive
+        /// <see cref="HttpClient.BaseAddress"/>.
+        /// </summary>
+        public const string DefaultProxiBaseAddress = "https://www.ebi.ac.uk/pride/proxi/archive/v0.1/";
 
         // Explicit JSON nulls must not clobber the non-null string defaults on the DTOs.
         private static readonly JsonSerializerSettings JsonSettings = new() { NullValueHandling = NullValueHandling.Ignore };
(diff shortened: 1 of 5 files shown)


zhuoxinshi (dismissed):

pcruzparri (dismissed):

Alexander-Sol (commented):
I would prefer to see the PrideProxiSpectrum as a child class of MzSpectrum. If implemented this way, the ToMzSpectrum function becomes unnecessary.

Everything else looks good though

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -234,6 +243,71 @@ public async Task<IReadOnlyList<string>> DownloadProjectFilesAsync(string access
...
+        /// <see cref="MsDataScan"/>.
+        /// </summary>
+        /// <param name="usi">The Universal Spectrum Identifier.</param>
+        /// <param name="cancellationToken">Cancels the fetch.</param>
+        /// <returns>The spectrum's peaks as an <see cref="MzSpectrum"/> (m/z ascending). Never null.</returns>
+        /// <exception cref="ArgumentException">The USI is null, empty, or whitespace.</exception>
+        /// <exception cref="HttpRequestException">The API returned a non-success status or an empty result for the USI.</exception>
+        /// <exception cref="MzLibUtil.MzLibException">The returned spectrum's peak arrays are not parallel.</exception>
+        /// <exception cref="OperationCanceledException">The operation was cancelled via <paramref name="cancellationToken"/>.</exception>
+        public async Task<MzSpectrum> GetSpectrumAsync(string usi, CancellationToken cancellationToken = default) =>
```
Alexander-Sol, line 308: Is this redundant now that PrideProxiSpectrum inherits from mzSpectrum?

Alexander-Sol (approved):
lgtm

zhuoxinshi (approved):

PR #1101, opened 2026-07-21
Read PRIDE project metadata by accession (PrideProject + Try/GetProjectAsync)

trishorts:
Given a PRIDE accession, you can now read the dataset's metadata — title, protocols, instruments, species, publications, submitters — **without downloading any of it**. That is the "should I even want this data?" step that has been missing: `GetProjectFilesAsync` already tells you what files a project has, but nothing told you what the project *is*.

```csharp
using var client = new PrideArchiveClient();

PrideProject project = await client.GetProjectAsync("PXD012345");
Console.WriteLine(project.Title);
Console.WriteLine(project.Instruments.Single().Name);   // "LTQ Orbitrap"
Console.WriteLine(project.References.Single().Doi);

// ...or, when the accession came from a user and may not exist:
var (found, p) = await client.TryGetProjectAsync(accession);
if (!found) return $"No PRIDE project '{accession}'.";
```

New types: `PrideProject`, plus `PrideContact` (submitters / lab PIs), `PrideReference` (publications) and `PrideSampleAttribute` (sample characteristics). New methods on the existing `PrideArchiveClient`: `GetProjectAsync` and `TryGetProjectAsync`.

## Why it is shaped this way

**Nothing was reusable, and I checked hard.** mzIdentML's `PersonType` and `BibliographicReferenceType` look like the types I needed and are not: they are `xsd`-generated (`// <auto-generated> ... will be lost if regenerated`), triplicated across three schema namespaces (`mzIdentML110/111/120.Generated`), field-mismatched (`PersonType` covers 3 of the 8 fields PRIDE returns — no email, country or ORCID, and its `affiliation` is an `organization_ref` pointer, not text; `BibliographicReferenceType` has no `pubmedID`), and unreachable regardless — `MzIdentML` and `Readers` reference `UsefulProteomicsDatabases`, not the reverse, so using them would mean adding a new project-reference edge to borrow two XML classes. This is the same reasoning that produced the hand-written `MzLibUtil.CvParam` in #1087, which this PR reuses heavily for all nine controlled-vocabulary groups.

**The error contract is a deliberate pair, because this endpoint disagrees with its sibling.** Live-verified 2026-07-21:

| Request | Response |
|---|---|
| `projects/PXD999999999/files` | `200 []` |
| `projects/PXD999999999` | **`404`, empty body** |
| `projects/NOTANACCESSION` | **`404`, empty body** (no `400`, no error payload) |

So an unknown accession cannot be modeled the way the manifest models it. This ships `TryGetProjectAsync`, which reports a missing project as a value, and `GetProjectAsync`, which throws and delegates to it — the `int.Parse`/`int.TryParse` convention, so callers who already trust their accession are not taxed with a tuple.

**`Found` is false for a 404 and *only* a 404.** 5xx, timeouts, rate limits and transport faults still throw. Collapsing an outage into "no such project" would send a user hunting for a typo in a perfectly valid accession — and it would break #1090: `ExternalServiceTestHelper` skips a live test on `HttpRequestException`, so swallowing that exception would make the live canary **fail red during an EBI outage** instead of skipping. This also matches every other network path in the repo (`GetProxiSpectrumAsync`, Koina `InferenceRequest`); the lone counterexample, `ProteinDbRetriever`, returns null and cannot distinguish a bad ID from an outage.

**The two failure kinds get two exception types, and that distinction is load-bearing.** Getting the above right still left the mirror-image hole: if `GetProjectAsync` threw `HttpRequestException` for a genuine 404, then a *withdrawn accession* — a real contract regression — would be converted to `Assert.Ignore` by `ExternalServiceTestHelper` and pass unnoticed. So:

| Condition | Exception | Live test behavior |
|---|---|---|
| Unreachable, timeout, 5xx, non-404 status | `HttpRequestException` | **skips** — not our bug |
| No such project; empty body; payload with no accession | `MzLibException` | **fails** — contract broke |

A test pins that the not-found case is *not* an `HttpRequestException`, so the distinction cannot silently regress.

**Dates are `DateTime`, not the `DateTimeOffset` used on `PrideArchiveFile`.** The files endpoint sends `"2019-01-15T09:42:57.000+00:00"`; this one sends a bare `"2019-01-15"` — no time, no offset. Parsing that into a `DateTimeOffset` silently attaches whatever offset the executing machine is in, so an assertion would pass locally and fail in CI. A test pins both values.

**The DTOs copy `PrideArchiveFile` exactly** — plain class, `{ get; set; }`, inline `= string.Empty` / `= new()` defaults, no `[JsonProperty]` (Newtonsoft matches case-insensitively, per the policy documented on `CvParam`), CV fields as `CvParam`. Deliberately *not* `PrideProxiSpectrum`'s get-only `[JsonConstructor]` shape, which was forced by `MzSpectrum` lacking a parameterless ctor — no base class here, so that constraint does not apply. Deliberately not a `record` either: the generated equality would compare a `List<CvParam>` member by reference, so two identical projects would compare unequal.

## Tests

33 offline fixture tests over a payload captured live from `PXD012345` (real wire shape, including the nested `sampleAttributes` tuple and the `"@type"` discriminators), plus 2 live canaries under `[Category("ExternalService")][Category("Pride")]` routed through `ExternalServiceTestHelper.RunAsync`.

The exhaustive error-contract coverage is **offline**, deliberately: `RunAsync` converts `HttpRequestException` into `Assert.Ignore`, so a live-only negative test would be silently skipped rather than passed. The live canary is documented with what it *can* catch (a 404 becoming a 200-with-project → fails) and what it cannot (a 404 becoming 5xx → indistinguishable from an outage).

114 PRIDE offline tests green; solution builds clean. Coverlet/Release reports **100% patch coverage** on the changed lines, branches included.

## Scope

`PrideArchiveClient.cs` is touched **additively only** — two new methods and one new `using`. No existing member's body, signature or name changes.

Follows #1087 (`CvParam`), #1088 (manifest), #1093 (download) and #1099 (PROXI spectrum).

### DIFF
--- a/mzLib/Test/DatabaseTests/PrideProjectTests.cs
+++ b/mzLib/Test/DatabaseTests/PrideProjectTests.cs
@@ -182,7 +182,7 @@ public class PrideProjectTests
     }
 
     [Test]
-    public async Task GetProjectAsync_CvParamGroups_DeserializeAndKeyOnAccession()
+    public async Task GetProjectAsync_CvParamGroups_Deserialize()
     {
         using var client = ClientReturning(ProjectJson);
 
@@ -210,10 +210,10 @@ public class PrideProjectTests
 
         Assert.Multiple(() =>
         {
-            Assert.That(project.QuantificationMethods, Is.Empty);
-            Assert.That(project.OrganismParts, Is.Empty);
-            Assert.That(project.Diseases, Is.Empty);
-            Assert.That(project.AdditionalAttributes, Is.Empty);
+            Assert.That(project.QuantificationMethods, Is.Not.Null.And.Empty);
+            Assert.That(project.OrganismParts, Is.Not.Null.And.Empty);
+            Assert.That(project.Diseases, Is.Not.Null.And.Empty);
+            Assert.That(project.AdditionalAttributes, Is.Not.Null.And.Empty);
         });
     }
 
@@ -304,11 +304,11 @@ public class PrideProjectTests
             Assert.That(project.Accession, Is.EqualTo("PXD000001"));
             Assert.That(project.Title, Is.Empty);
             Assert.That(project.Doi, Is.Empty);
-            Assert.That(project.Instruments, Is.Empty);
-            Assert.That(project.Submitters, Is.Empty);
-            Assert.That(project.References, Is.Empty);
-            Assert.That(project.SampleAttributes, Is.Empty);
-            Assert.That(project.ProjectTags, Is.Empty);
+            Assert.That(project.Instruments, Is.Not.Null.And.Empty);
+            Assert.That(project.Submitters, Is.Not.Null.And.Empty);
+            Assert.That(project.References, Is.Not.Null.And.Empty);
+            Assert.That(project.SampleAttributes, Is.Not.Null.And.Empty);
+            Assert.That(project.ProjectTags, Is.Not.Null.And.Empty);
         });
     }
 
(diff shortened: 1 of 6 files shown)


trishorts:
### Manual unit-test review

All 26 unit test(s) added by this PR were reviewed individually by the author (not only by automated/AI-generated checks): 15 approved as written, 8 edited, 3 flagged.

Flagged for follow-up:
- `mzLib/Test/DatabaseTests/PrideProjectTests.cs` :: GetProjectAsync_StringArrays_Deserialize -- Asserts Newtonsoft deserializes string arrays - exercises the JSON library, not our code. The only code-specific claim (case-insensitive binding without [JsonProperty]) is already proven by test 1.
- `mzLib/Test/DatabaseTests/PrideProjectTests.cs` :: GetProjectAsync_LabPIs_Deserialize -- Near-duplicate of test 6: same PrideContact type, every field already asserted there. Only new information is that the labPIs property binds.
- `mzLib/Test/DatabaseTests/PrideProjectTests.cs` :: GetProjectAsync_Cancelled_ThrowsOperationCanceledException -- StubHandler ignores its CancellationToken, so this exercises HttpClient pre-flight cancellation rather than proving our code threads the token. Proving that would require diverging one of four verbatim StubHandler copies.

This satisfies the author's standing AI-use policy that AI-assisted tests receive human review before relying on them. Recorded 2026-07-21 20:57.

Alexander-Sol (approved):

Alexander-Sol (approved):

pcruzparri (approved):

PR #1102, opened 2026-07-23
Return the full PRIDE manifest when pageSize exceeds the server cap

trishorts:
`GetProjectFilesAsync("PXD012345", pageSize: 500)` returns **100 of that project's 131 files, and reports no error** — silently contradicting the method's own doc promise that "the full manifest is returned regardless" of page size. This PR fixes that, and pins a second latent hazard next to it with a test.

Both are hardening changes to `PrideArchiveClient`; no new public API.

## Why it happens

PRIDE caps `pageSize` server-side at **100** and then pages by the *capped* size, not the requested one. Verified live 2026-07-23:

| request | elements returned | `total_records` |
|---|---|---|
| `projects/PXD012345/files?pageSize=100&page=0` | 100 | 131 |
| `projects/PXD012345/files?pageSize=500&page=0` | 100 | 131 |
| `projects/PXD012345/files?pageSize=500&page=1` | 31 | 131 |

`pageSize=500&page=1` returns exactly the same records as `pageSize=100&page=1`, so the cap is applied before paging. An over-large request therefore comes back "short" on **every** page while more records remain — and the paging loop treated the first short page as terminal:

```csharp
// Stop when we have collected everything the server reported.
if (TryGetTotalRecords(response, out long total) && files.Count >= total)
    break;
// Safety net: a short page is the last page even if the total header is missing.
if (pageFiles.Count < pageSize)   // 100 < 500 -> break, with 31 files still unfetched
    break;
```

The `total_records` guard above it does not save the call: `files.Count (100) >= total (131)` is false, so control falls through to the short-page check.

The same cap applies to `search/projects`, which is how this surfaced — it was found while grounding a future dataset-search capability against the live API.

## The fix

Make the server's own record count authoritative, and demote the short-page heuristic to the fallback used **only** when that header is absent:

```csharp
if (TryGetTotalRecords(response, out long total))
{
    if (files.Count >= total)
        break;
    // more remain -- keep paging even though the page may look short
}
else if (pageFiles.Count < pageSize)
{
    break;   // no total to trust: a short page is the last page
}
```

All four termination paths still hold: empty page, total reached, no-header short page, and the `MaxPages` cap for a server that ignores paging entirely. A `total_records` that overstates the truth still terminates on the empty-page check (paging past the end returns a zero-byte body, which deserializes to an empty list).

**Behaviour is unchanged for every existing caller.** `DownloadProjectFilesAsync` and all other call sites use the default `pageSize` of 100, at or below the cap, where the short-page check was already correct.

### Alternatives rejected

- **Clamp `pageSize` to 100.** Bakes a server constant into the client; wrong the day EBI changes it.
- **Learn the effective page size from page 0** and compare against that. Cannot distinguish a *capped* first page from a *genuinely short last* page — it breaks `GetProjectFilesAsync_UnparseableTotalHeader_StillTerminates`, where a 1-file first page really is the whole answer.

## Second commit: the null sample-attribute key

Callers dereference `attribute.Key.Accession` directly (the live canary does), so a `PrideSampleAttribute` arriving with a null `Key` would NRE rather than read as an absent term. Nothing in `RemoveNullElements` guards it.

What actually defends it is `NullValueHandling.Ignore` dropping the explicit null and leaving the DTO's `= new()` default standing — a property of `JsonSettings`, not of the DTO, which a later serializer change could remove without touching either file.

The obvious fix — a null-`Key` guard inside the `RemoveNullElements` foreach — **was written and then discarded**: with `NullValueHandling.Ignore` in place that branch is unreachable through the public API, so it is dead code whose null path never fires and cannot be covered. Verified empirically both ways: the new test passes against unguarded source, and fails with a `NullReferenceException` when `NullValueHandling.Ignore` is removed. The test is therefore the guard, and a comment in the client records why there is no second check.

## Tests

- `GetProjectFilesAsync_ServerCapsPageSize_StillReturnsFullManifest` — offline regression, **verified red against unfixed source** (2 of 3 files, 1 request instead of 2).
- `GetProjectFilesAsync_TotalRecordsOverstated_StopsOnEmptyPage` — an overstated `total_records` terminates on the empty page (2 requests), not by throwing.
- `GetProjectFilesAsync_TotalRecordsExactlyMatched_DoesNotFetchExtraPage` — total reached on page 0 stops after exactly 1 request.
- `GetProjectFilesAsync_LivePageSizeAboveServerCap_ReturnsSameManifestAsDefault` — live canary under `[Category("ExternalService")][Category("Pride")]`, routed through `ExternalServiceTestHelper` so an EBI outage skips rather than fails. Compares against the default-page-size call rather than a literal, because the project's file count drifts while the invariant — page size must not change the answer — does not.
- `TryGetProjectAsync_ExplicitJsonNullKeyOrValue_DoNotClobberSampleAttributeDefaults` — the serializer-settings pin described above.

Full offline suite: **5272 passed, 0 failed**, 29 skipped. Live PRIDE canaries: **7/7**.

Touches `PrideArchiveClient.GetProjectFilesAsync` and `RemoveNullElements` (both existing), `PrideArchiveClientTests`, and `PrideProjectTests`.

## Review follow-up (commit `568584d3`)

A review round closed the one real gap and tightened three tests, with the core fix unchanged:

- **The `MaxPages` backstop was unreachable once `total_records` parses.** A server that honours the capped `pageSize` but ignores `page` re-serves the same records every request; `files.Count` accumulated the duplicates and tripped `files.Count >= total` after a couple of identical pages, returning a manifest padded with duplicates and silently missing the records past the cap — instead of failing at `MaxPages` as the four-termination-path claim asserted. The loop now collects files through a `HashSet` keyed on `FileName` (a manifest lists each file once, so a repeated name is a paging artefact) and throws — like `MaxPages` — when a page advances nothing while `total_records` says records remain. New test `GetProjectFilesAsync_ServerIgnoresPageButReportsTotal_ThrowsInsteadOfReturningDuplicates` pins it.
- **The `<summary>` still promised an unconditional "complete manifest"** that the qualified `<param>` block had already walked back; the qualification is carried up so IntelliSense's first line no longer overstates the guarantee.
- **The live canary made an exactly-at-cap manifest a permanent hard failure** (`Assume` ≥ cap then `Assert` > cap leaves count == cap red). A genuinely 100-file project and a truncated one are indistinguishable from the count alone, so that case is now inconclusive (skip); a real truncation still fails **red** through the `aboveCap`-vs-`atDefault` equivalence assertion, which was always the true check.
- **The capped-pageSize regression test asserted exact query-string text**, coupling it to URL construction this PR does not own; it now parses each request and asserts the paging semantics (two requests, both at the caller's page size, pages 0 then 1).

Patch coverage remains 100%; all PRIDE unit tests pass.

---

## Note on CI history

CI on this PR went red on 2026-08-09 for reasons unrelated to it: a UniProt outage failed the required `build` job through ~150 XML-parsing, variant, decoy and digestion tests that download an ontology during fixture setup, and the job was cancelled at its 15-minute cap. Nothing here touches that path — this PR is three PRIDE files.

The fix for that was briefly committed to this branch and has been **split out to #1140**, which also hardens `Loaders.DownloadContent` and moves the live Koina fixtures to the non-blocking job. This branch has been rewound to its PRIDE-only state; if you saw those commits here earlier, that is where they went.


trishorts:
### Manual unit-test review

All 5 unit tests added by this PR were reviewed individually by the author, line by line, rather than being trusted because they passed: **0 approved as written, 5 edited, 0 flagged.**

The review changed something substantive in every one, and two of the five were not testing what they claimed:

- **`GetProjectFilesAsync_TotalRecordsOverstated_StopsOnEmptyPage` pinned nothing.** It requested `pageSize: 2` against a stub serving 2 files, which made the old short-page check a no-op (`2 < 2` is false) — so the pre-fix code produced an identical result and the test would have passed either way. It now requests 4 against a stub that serves 2, so the request count differs (1 before the fix, 2 after), and it is verified red against master `220f67f2`. It also now pins the *cost* of the fix, so an optimisation that restores the early break fails here.

- **The live canary could have hidden the regression it exists to catch.** Truncation returns *exactly* the server cap, so the `Assume.That(count > cap)` precondition would have reported **skipped** in precisely that case — and a skip in the external-service job reads as an EBI outage. Split into `Assume(Is.Not.LessThan(cap))` for a project that genuinely shrank and `Assert(Is.GreaterThan(cap))` for the at-cap truncation signature.

The other three:

- `..._NullKeyOrValueOnSampleAttribute_AreReplacedWithEmpties` was renamed to `TryGetProjectAsync_ExplicitJsonNullKeyOrValue_DoNotClobberSampleAttributeDefaults`. The old name described the implementation that was deliberately **rejected** (an active `??=` guard); nothing replaces anything, the defaults are simply never clobbered. A reader grepping the old name would have gone looking for replacement logic and concluded the guard was lost.
- `..._ServerCapsPageSize_StillReturnsFullManifest` now asserts query strings rather than absolute URIs — the base address is already pinned by `GetProjectAsync_BuildsRelativeV3Uri`, and repeating it here fails as an opaque two-long-strings diff.
- `..._TotalRecordsExactlyMatched_DoesNotFetchExtraPage` now records that its stub never returns an empty page, which is what makes the `total_records` check the only possible terminator there — so weakening that check does not cost one wasted request, it runs to the `MaxPages` throw.

Verified after the edits: full offline suite 5274 passed / 0 failed (29 pre-existing skips); live PRIDE canaries 7/7 with 0 skipped, confirming the `Assume` change did not silently disable the canary.

This satisfies the author's standing AI-use policy that AI-assisted tests receive human review before being relied upon. Recorded 2026-07-23.

trishorts:
**Follow-up commit `568584d3` — review fixes (one real gap + test hardening)**

A local review round closed the one finding with a runtime consequence and tightened three tests. The core fix (making `total_records` authoritative, short-page heuristic as fallback) is unchanged.

**1. `MaxPages` no longer backstopped a page-ignoring server once `total_records` parses.** A server that honours the capped `pageSize` but ignores `page` re-serves the same records each request; `files.Count` accumulated the duplicates and tripped `files.Count >= total` after a couple of identical pages — returning a manifest padded with duplicates and silently missing the records past the cap, instead of failing at `MaxPages` as the PR's four-termination-path claim asserted. The loop now collects through a `HashSet` keyed on `FileName` and throws — like `MaxPages` — when a page advances nothing while `total_records` says records remain. New test `GetProjectFilesAsync_ServerIgnoresPageButReportsTotal_ThrowsInsteadOfReturningDuplicates` pins it.

**2. `<summary>` doc contradiction.** The summary still promised an unconditional "complete manifest" that the qualified `<param>` block had walked back; the qualification is carried up.

**3. Live canary went permanently red on an exactly-100-file manifest.** `Assume` ≥ cap then `Assert` > cap made `count == cap` a hard failure, though a genuinely 100-file project and a truncated one are indistinguishable from the count alone. That case is now inconclusive (skip); a real truncation still fails **red** via the `aboveCap`-vs-`atDefault` equivalence assertion, which was always the true check.

**4. Exact query-string assertion** in the capped-pageSize test is replaced with semantic parsing (two requests, both at the caller's page size, pages 0 then 1), decoupling it from URL construction this PR does not own.

**Not changed on judgment:** the `RemoveNullElements` comment already cites the real test (`TryGetProjectAsync_ExplicitJsonNullKeyOrValue_DoNotClobberSampleAttributeDefaults`) — it was the PR body that was stale, now reconciled; `ExternalServiceTestHelper.RunAsync` already lets `Assume`'s `InconclusiveException` through (it catches only outage types), so no change was needed there; and the two near-identical paging stubs were left separate to keep their distinct assertions legible.

Patch coverage remains 100% (14/14 changed lines); all 57 PRIDE unit tests pass.

trishorts:
**Follow-up commit `2c3cb532` — round-two review found my own previous commit reintroduced the bug it fixed**

A second review pass over `568584d3` (the earlier round of fixes) turned up three Major findings, all in that commit rather than in the original PR. Worth writing up plainly, because the first one is the same class of defect this PR exists to remove.

**1. The `FileName` dedup silently dropped real records.** `568584d3` detected a stalled server by deduplicating on `FileName`, which made *manifest membership* depend on name uniqueness. Two consequences, both silent:
- An entry whose `fileName` is absent from the JSON keeps the DTO's non-null default, so every such entry collapses onto the first — and that is a case this class explicitly expects, since `DownloadFileAsync` guards `string.IsNullOrWhiteSpace(file.FileName)`.
- Two distinct records sharing a leaf name (same name under different `publicFileLocations`) collapse the same way.

It also falsified the claim that behaviour was unchanged for existing callers: at the default `pageSize` the base code returned every entry via `AddRange`, and the deduplicating version did not.

**Fix:** progress is a property of the **page**, not of individual files. Every returned entry goes into the manifest again, and a stalled server is detected by comparing a page against its immediate predecessor. That check now runs **before** the `total_records` comparison — otherwise duplicate records push `files.Count` past `total` and the loop breaks out as an apparent success, hiding the fault. Comparing whole pages is also the stricter signal the adjacent empty-page comment argues for: a project whose files change mid-fetch can legitimately return a page that merely *overlaps* its predecessor, so only an exactly-repeated page is treated as a fault, and the message now names mid-fetch mutation as an alternative cause.

**2. The guard threw the outage exception type.** It threw `HttpRequestException`, but this class's own remarks reserve that for "unreachable, timing out or answering 5xx" and assign a successful answer that breaks the contract to `MzLibException` *"so it fails loudly instead of being mistaken for an outage"* — and `ExternalServiceTestHelper` converts `HttpRequestException` into a **skip**. So a real PRIDE paging regression would have passed the new live canary **green**, which is precisely the silence this PR set out to remove. Now throws `MzLibException`, with the missing `<exception>` element added.

**3. The guard's test passed with the guard deleted.** It asserted only the exception *type* — which the pre-existing `MaxPages` backstop also throws — so removing the guard left it green. It now pins the type **and** that the fault aborts on the **second** request rather than after `MaxPages` futile round-trips, which is the behaviour that distinguishes the guard from the backstop.

**Also in this commit:** tests that no entry is dropped for a repeated file name or for entries with no file name at all; a test for the documented boundary of the narrowed guarantee (a server cap *plus* a missing `total_records` header still truncates — that is the limitation the summary now admits to); the stale `MaxPages` comment in the exactly-matched test corrected; and the query assertions read through a lookup so a renamed paging parameter fails with the intended message instead of `KeyNotFoundException`.

**Not changed on judgment:** the reviewer suggested asserting `Is.Unique` on the live manifest's file names to protect the dedup's premise. That premise no longer exists — nothing depends on name uniqueness now — and asserting it would fail if PRIDE legitimately serves a repeated leaf name, so it was dropped rather than applied.

Full suite: **5342 passed, 0 failed**. Patch coverage **100% (23/23 changed lines)**.

zhuoxinshi (approved):

Alexander-Sol (approved):
Took an independent second look at the paging logic since the follow-ups reworked it. The fix is right — I walked every exit path (empty page, total reached, no-header short page, identical-page throw, `MaxPages`) and couldn't construct an input that loops forever or returns early with records unfetched. Throwing `MzLibException` rather than `HttpRequestException` is the right call so the external-service helper fails instead of skipping.

Two things to note, neither blocking:

- The PR description is now stale on the mechanism. `2c3cb53` and `54c9960` replaced the FileName-keyed dedup with page-progress keyed on the raw response body, and there's no per-record dedup anymore. That's the correct choice — `PrideArchiveFile` has no unique key (`FileName` defaults to `""` and can legitimately repeat, `Checksum` is often empty), so per-record dedup would drop real files — but the body text still describes the old approach.
- An *understated* `total_records` would silently truncate the tail at `files.Count >= total` — the same failure class this PR fixes, mirrored. Not reachable with honest PRIDE behavior, so I wouldn't hold the PR for it, but it's the one truncation path that's neither documented nor pinned. A one-line docstring caveat next to the overstated-total case (which you *do* document) would close it.

Minor: relying on `NullValueHandling.Ignore` to keep `RemoveNullElements` from NRE-ing is defensible since the test pins it, but a `?.` at the call site would cost nothing and drop the cross-file coupling.

Issue #1113, opened 2026-07-23
Digest applies modifications after cleavage, producing peptides whose C-terminal K is modified in a way that blocks trypsin


trishorts:
## What happens

`Protein.Digest` cleaves first and places modifications afterwards, with no check that the modification abolishes the cleavage site it sits on. The result is peptidoforms ending in a **modified lysine that trypsin could not have cleaved after** — reported as ordinary peptides, some with `MissedCleavages = 0`.

## Why it is wrong

Trypsin's specificity comes from the positively charged K/R side chain binding Asp189 in the S1 pocket. Acylation of the lysine ε-amine removes that charge:

- **N6-succinyllysine** — neutralises *and* reverses the charge (adds a carboxylate). Trypsin does not cleave. This is why succinylome and acetylome workflows are dominated by missed cleavages at the modified residue.
- **N6-methyllysine** — retains the charge but is sterically impaired; cleavage is strongly reduced and in practice usually shows as a missed cleavage.

Either way, a peptidoform whose C-terminal residue is a succinylated lysine, reported at zero missed cleavages, describes an event that does not occur.

## Measured

Human serum albumin P02768, `trypsin|P`, 2 missed cleavages, min length 7, `--max-mods 2` → 303 peptidoforms, of which **17 (5.6%)** end in a modified C-terminal K:

| peptide | modification on the C-terminal K | reported missed cleavages |
|---|---|---|
| `EFNAETFTFHADICTLSEK` | N6-succinyllysine | **0** |
| `QTALVELVK` | N6-methyllysine | **0** |
| `LKCASLQK` | N6-succinyllysine | 1 |
| `NLGKVGSK` | N6-succinyllysine | 1 |
| `CCKADDK` | N6-succinyllysine | 1 |
| `KQTALVELVK` | N6-methyllysine | 1 |
| … | | |

The first two are the clearest: a fully-tryptic, zero-missed-cleavage peptide terminating in a succinylated or methylated lysine.

## Impact

- **Peptidoform enumeration** (what a "what peptides does this protein give" call returns) is ~6% chemically impossible entries on this protein. On a heavily acetylated/succinylated substrate — a histone, which is the case people care about — the proportion is far higher, because that is where these modifications concentrate.
- **Search space** is inflated with candidates that cannot be generated, and the corresponding *real* peptide (the longer form reading through the modified K) is present only as a missed-cleavage variant, so the two compete.
[...]

Issue #1112, opened 2026-07-23
ProteinDbLoader silently drops glycosylation-site annotations that have a defined formula and mass (22 of 24 on albumin)


trishorts:
## What happens

`ProteinXmlEntry.ParseFeatureEndElement` turns a UniProt feature into a modification only for two feature types:

```csharp
if (FeatureType == "modified residue") { ... AnnotatedMods.Add(...); }
else if (FeatureType == "lipid moiety-binding region") { ... AnnotatedMods.Add(...); }
```

`glycosylation site` (UniProt `CARBOHYD`) is not in the list, so those annotations are discarded — **including the ones UniProt gives a complete chemical formula and monoisotopic mass**.

## Why this is not just "glycans are ambiguous"

The usual justification is that a glycosylation annotation like `N-linked (GlcNAc...) asparagine` has no fixed composition, so there is no mass to add. That is true, and correct to exclude. But it does not describe most of them.

Human serum albumin, P02768, annotates **24 glycosylation sites**:

| description | n | usable? |
|---|---|---|
| `N-linked (Glc) (glycation) lysine` (incl. `; alternate`, `; in vitro`) | **22** | **yes** |
| `N-linked (GlcNAc...) asparagine` | 2 | no — composition genuinely unspecified |

`N-linked (Glc) (glycation) lysine` is glycation — glucose adduction — and UniProt's own `ptmlist.txt`, the file mzLib already loads, defines it fully:

```
ID   N-linked (Glc) (glycation) lysine
FT   CARBOHYD
TG   Lysine.
PP   Anywhere.
CF   C6 H10 O5
MM   162.052823
MA   162.14
DR   PSI-MOD; MOD:01347.
DR   Unimod; 41.
```

Defined formula, defined monoisotopic mass, cross-referenced to both PSI-MOD and Unimod. There is nothing ambiguous about it, and glycation of albumin lysines is a clinically important modification — it is the basis of the glycated-albumin assay.

So on this protein mzLib loads 14 of 38 annotations, when 36 of 38 are chemically usable. The 22 it drops are dropped because of their **feature type**, not because of anything about their chemistry.

## The silent part

The exclusion happens before any modification lookup, so these never appear in `unknownModifications` either. Nothing anywhere reports that 22 annotated, mass-defined modifications were discarded. A caller enumerating albumin's modifications gets 14 and no indication that a rule was applied.

[...]

Issue #1111, opened 2026-07-23
FlashLFQ peptide roll-up nondeterministically drops MBR intensities, making protein quant irreproducible run-to-run


trishorts:
## What happens

Running `FlashLfqEngine` twice on identical inputs, with `MatchBetweenRuns = true` and default `MaxThreads = -1`, produces **different peptide intensities and different protein results**.

The chromatographic peaks are stable. The roll-up on top of them is not.

## Measured

Inputs: mzLib's own FlashLFQ test data — `AllPSMs.psmtsv` with `20100614_Velos1_TaGe_SA_K562_3.mzML` and `..._4.mzML`. Six consecutive runs, byte-identical inputs, `--mbr`, `MaxThreads = -1`:

| invariant across all 6 runs | value |
|---|---|
| `Peaks` | 647 |
| `PeptideModifiedSequences` | 354 |
| MBR peaks | 140 |

| varies | |
|---|---|
| protein groups whose intensity is NaN in both runs | **2 in five runs, 1 in the sixth** |

Diffing the divergent pair: **the peak lists are byte-identical** — every `(file, sequence)` key carries the same peaks with the same intensities and detection types. But **6 peptides differ in the roll-up**:

| peptide | run 4 intensity, run A | run B |
|---|---|---|
| `DEESGGGSNPFQHLEK` | **0** | 16,821.66 |
| `ISEMALEHELK` | **0** | 53,353.23 |
| `KQEQEAAK` | **0** | 21,410.28 |
| `LAEVEAALEK` | **0** | 41,674.70 |
| `LNGWYMEK` | **0** | 56,217.99 |
| `YDEC[Carbamidomethyl]L…` | **0** | 41,646.49 |

Each of those peptides has **exactly one peak** in that run, an MBR peak with a real intensity:

```
DEESGGGSNPFQHLEK in K562_4 -> 1 peak
    intensity=16821.655  detection=MBR  rt=79.0399  num_identifications=1
  roll-up says: intensity=0  detection=NotDetected
```

So the peptide table reports `0` / `NotDetected` for a peptide that has a single, unambiguous, real MBR peak — **and whether it does so changes between runs.**

Downstream, `CalculateProteinResultsMedianPolish` consumes those peptide intensities, so a peptide flipping between `0` and a real value flips a borderline protein between NaN and a number. `Q15149` (PLEC) was NaN in 5 of 6 runs and quantified in the 6th.

## Why it matters

[...]

Issue #1110, opened 2026-07-23
ETD/ECD suppress z• ions N-terminal to proline but still emit the complementary c ions


trishorts:
## What happens

`PeptideWithSetModifications.Fragment` deliberately suppresses `zDot` ions at proline for the electron-transfer dissociation types:

```csharp
// skip zDot ions for proline residues for ETD/ECD/EThcD
if (cTermResidue == 'P'
    && (dissociationType == DissociationType.ECD || dissociationType == DissociationType.ETD || dissociationType == DissociationType.EThcD)
    && cTermProductTypes[i] == ProductType.zDot)
{
    continue;
}
```

There is **no corresponding check in the N-terminal branch**, so the `c` ion from the *same cleavage site* is still generated.

## Why that is inconsistent

ETD/ECD cleave the **N–Cα** bond. At a proline the nitrogen is inside the pyrrolidine ring, so even when that bond breaks the two halves stay tethered by the ring's Cδ–N bond and **no fragment pair separates**. That is why "ETD does not cleave N-terminal to proline" is a standard rule.

The consequence is that at an Xxx–Pro bond **neither** the c nor the z• ion is observed. mzLib currently drops one and keeps the other, so every c ion N-terminal to a proline is a theoretical ion that cannot appear in a real spectrum.

The existing suppression is already site-correct — for cleavage site `i` (between residues `i` and `i+1`), `c_i` and `z_(L-i)` are complementary, and the code suppresses `z_(L-i)` when residue `i+1` is proline. The same condition simply needs to gate `c_i`.

## Measured

Human serum albumin P02768, `trypsin|P`, 2 missed cleavages, min length 7, ETD, `FragmentationTerminus.Both`, no modifications — 195 peptides:

| | count |
|---|---|
| cleavage sites whose following residue is proline | 138 |
| **z• ions suppressed** | **138** (100% of them, 0 elsewhere) |
| **c ions suppressed** | **0** |
| total c ions generated | 3,253 |

A perfect 1:1 between proline sites and suppressed z• ions, and not one suppressed c ion. So ~4% of the c-ion list for this protein consists of ions ETD cannot produce. Example — `ALVLIAFAQYLQQCPFEDHVK`, site 14 (residue 15 is P): `c14` is present, its partner z• is absent.

## Suggested fix

Mirror the existing guard in the N-terminal product loop, keying on the residue *following* the cleavage site (`BaseSequence[r + 1]`), for the same three dissociation types.

[...]

PR #1114, opened 2026-07-24
Remove y ions from the ETD and ECD product sets (#1109)

### LINKED ISSUE #1109: ETD and ECD generate y ions but not b ions, which has no fragmentation-mechanism basis
## What happens

`ProductsFromDissociationType` in `Omics/Fragmentation/Peptide/DissociationTypeCollection.cs` maps ETD and ECD to a product set that includes **`y` but not `b`**:

```csharp
{ DissociationType.CID,   new List<ProductType>{ ProductType.b, ProductType.y } },
{ DissociationType.HCD,   new List<ProductType>{ ProductType.b, ProductType.y } },
{ DissociationType.ECD,   new List<ProductType>{ ProductType.c, ProductType.y, ProductType.zDot } },
{ DissociationType.ETD,   new List<ProductType>{ ProductType.c, ProductType.y, ProductType.zDot } },
{ DissociationType.EThcD, new List<ProductType>{ ProductType.b, ProductType.y, ProductType.c, ProductType.zDot } },
```

## Why it looks wrong

ETD and ECD cleave the **N–Cα** bond, giving c and z• ions. b and y ions come from **amide (C–N)** cleavage under vibrational activation. So:

- If the `y` ions are there to model residual/supplemental vibrational activation, then **`b` should be there too** — which is exactly what the `EThcD` row does, correctly.
- There is no mechanism that produces y ions *without* b ions. The asymmetry is the tell.

The ETD and ECD rows also read like an edit of a `{ b, y }` row in which `b` was replaced by `c, zDot` and `y` was left behind.

## Impact

Every ETD/ECD theoretical fragment list carries roughly a third spurious ions. Fragmenting human serum albumin (P02768), tryptic, `FragmentationTerminus.Both`, via `PeptideWithSetModifications.Fragment(DissociationType.ETD, ...)`:

| peptide (31-mer) | c | zDot | y |
|---|---|---|---|
| `ALVLIAFAQYLQQCPFEDHVK...` | 30 | 30 | **30** |

30 of 90 products, 33%, are y ions that ETD should not produce. In a search these are matchable theoretical ions, so they can inflate matched-ion counts and scores for ETD/ECD data.

## Suggested fix

```csharp
{ DissociationType.ECD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
{ DissociationType.ETD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
```

leaving `EThcD` as-is, since that is where supplemental activation genuinely belongs.

If the `y` ions **are** intentional — e.g. modelling activation that in practice accompanies ETD on a particular instrument — then a comment saying so would help, and the absence of `b` still needs explaining.

## How it was found

Building [mzLibRust](https://github.com/smith-chem-wisc/mzLibRust), the Rust sibling of pyMzLib over the same language-neutral bridge. A parity test asserting "ETD produces c and z• ions, not b and y" failed against a recorded albumin digest, which surfaced the y ions.

trishorts:
Closes #1109.

## The change

```diff
-{ DissociationType.ECD, new List<ProductType>{ ProductType.c, ProductType.y, ProductType.zDot } },
+{ DissociationType.ECD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
-{ DissociationType.ETD, new List<ProductType>{ ProductType.c, ProductType.y, ProductType.zDot } },
+{ DissociationType.ETD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
```

`EThcD` is deliberately left alone.

## Why

ETD and ECD cleave the **N–Cα** bond and yield c and z• ions. b and y come from **amide** cleavage under vibrational activation. If the y ions were modelling residual activation then b ions would come with them — which is exactly what the `EThcD` row does. **y without b has no fragmentation mechanism behind it**, and the ETD/ECD rows read like an edit of a `{ b, y }` row in which `b` became `c, zDot` and `y` was left behind.

## Impact

Human serum albumin P02768, tryptic, `FragmentationTerminus.Both`, ETD — a 31-mer peptide:

| series | before | after |
|---|---|---|
| c | 30 | 30 |
| zDot | 30 | 30 |
| **y** | **30** | **0** |

A third of every ETD/ECD theoretical fragment list was ions ETD does not produce. In a search those are matchable, so they can inflate matched-ion counts and scores on ETD and ECD data.

## Tests

Updated rather than deleted, so the diff shows exactly what behaviour changes:

- the ETD/ECD product-set `TestCase`s
- `Test_ETD_ECD_EThcD_Fragmentation_No_FragmentsAtProline`: **17 → 11** for ETD and ECD. `PEPTIDE` gives 6 c + 5 zDot, one z• already suppressed at the proline. **HCD 12 and EThcD 23 are unchanged**, which is a useful check that only the intended rows moved.
- `TestCTermEtd`: its subject is the extra zDot neutral loss; the y assertions were incidental. 9 → 0, total 20 → 11.
- `Test_UnmodifiedPeptide_AllProductType_fragmentMasses`: the ECD/ETD m/z sets lose `120` and `249`, which are the two y ions — the CID set still contains them.
- `Test_TerminusSpecificProductTypes`: ETD `Both` `{c, y, zDot}` → `{c, zDot}`, ETD `C` `{y, zDot}` → `{zDot}`.
- `Test_TerminusSpecificProductTypesFromPeptideWithSetMods`: same two expectations at the fragmentation level.
- the O-glycopeptide fragment count: 22 → 15.

Two new tests pin the intent so this cannot silently regress:

- `ElectronTransferDissociationProducesCAndZDotOnly` — ETD and ECD are exactly `{ c, zDot }`
- `EThcDKeepsBAndYTogetherBecauseSupplementalActivationProducesBoth` — EThcD keeps b **and** y alongside c and z•, pinning the asymmetry as deliberate

**Full suite: 5334 passed, 0 failed, 29 skipped.**

## Review follow-up (commit `bdabc170`)

A review round caught that the y removal was incomplete and tightened three assertions, with the scientific call unchanged:

- **The y-series neutral-loss path was still live.** `GetWaterAndAmmoniaLossProductTypesFromDissociation` had a `case ECD: case ETD:` arm that returned `yWaterLoss` and `yAmmoniaLoss` — coherent only while those rows carried `y`. Those are y-series ions, so a y-derived loss with no parent y has exactly the missing mechanism this PR uses to reject y. An ETD/ECD search with loss ions enabled kept generating them. The arm is deleted (falls through to an empty list), and `Test_WaterAndAmmoniaLossFragmentProductIons` now asserts **empty** for ECD/ETD instead of `{ yWaterLoss, yAmmoniaLoss }`; EThcD/CID/HCD/IRMPD are unchanged controls.
- **`ElectronTransferDissociationProducesCAndZDotOnly`** now also asserts the loss helper returns nothing for ETD/ECD across all three termini, so this second product-type entry point can no longer reintroduce a y-series ion with the suite green.
- **`Test_ETD_ECD_EThcD_Fragmentation_No_FragmentsAtProline`** asserted only a total count; for ETD, 11 is equally satisfied by `6 y + 5 zDot`, so it now pins the `ProductType` set per dissociation type (`ETD/ECD = { c, zDot }`).
- Added the N–Cα rationale as inline comments on the ECD/ETD rows and the loss-helper default arm (matching the existing HCD comment convention), and reverted the UTF-8 BOM this branch had prepended to two test files.

`TestDissociationProductTypes`' subset assertion was left as-is deliberately: tightening it to an exact set would break its intentionally-partial cases (e.g. the EThcD case lists only `{b, y}` though the row has four types), and `ElectronTransferDissociationProducesCAndZDotOnly` already carries the exact check.

## Note for reviewers

This changes search results for ETD/ECD data, so it is a scientific call as much as a code one — worth a second opinion from someone who runs ETD routinely before merging. The mechanism argument seems clear to me, but the consequence is real and I would rather it were confirmed than assumed.

Found while building [mzLibRust](https://github.com/smith-chem-wisc/mzLibRust); a parity test asserting "ETD produces c and z•, not b and y" failed against a recorded albumin digest. Related: #1110, #1111, #1112, #1113.

### DIFF
--- a/mzLib/Omics/Fragmentation/Peptide/DissociationTypeCollection.cs
+++ b/mzLib/Omics/Fragmentation/Peptide/DissociationTypeCollection.cs
@@ -15,6 +15,9 @@ namespace Omics.Fragmentation.Peptide
             { DissociationType.CID, new List<ProductType>{ ProductType.b, ProductType.y } },
             { DissociationType.LowCID, new List<ProductType>{ ProductType.b, ProductType.y, ProductType.aStar, ProductType.bAmmoniaLoss, ProductType.yAmmoniaLoss, ProductType.aDegree, ProductType.bWaterLoss, ProductType.yWaterLoss } },
             { DissociationType.IRMPD, new List<ProductType>{ ProductType.b, ProductType.y } },
+            // ECD/ETD cleave the N-Cα backbone bond and produce c and z• ions only; the amide-cleavage b/y series
+            // requires vibrational activation. EThcD below deliberately keeps b and y because its supplemental
+            // collisional activation produces both together (b without y, or y without b, has no mechanism).
             { DissociationType.ECD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
             { DissociationType.PQD, new List<ProductType>() },
             { DissociationType.ETD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
@@ -87,14 +90,10 @@ namespace Omics.Fragmentation.Peptide
                         productList.Add(ProductType.yAmmoniaLoss);
                     }
                     break;
-                case DissociationType.ECD:
-                case DissociationType.ETD:
-                    if (fragmentationTerminus == FragmentationTerminus.C || fragmentationTerminus == FragmentationTerminus.Both)
-                    {
-                        productList.Add(ProductType.yWaterLoss);
-                        productList.Add(ProductType.yAmmoniaLoss);
-                    }
-                    break;
+                // ECD/ETD produce c and z• only (N-Cα cleavage), so there is no y series and therefore no
+                // y water/ammonia loss to report -- a y-derived loss ion with no parent y has no mechanism,
+                // the same argument that removed ProductType.y from the ECD/ETD product sets above. Falls
+                // through to default and returns an empty list.
                 default:
                     break;
             }
--- a/mzLib/Omics/Fragmentation/Peptide/DissociationTypeCollection.cs
+++ b/mzLib/Omics/Fragmentation/Peptide/DissociationTypeCollection.cs
@@ -15,9 +15,9 @@ namespace Omics.Fragmentation.Peptide
             { DissociationType.CID, new List<ProductType>{ ProductType.b, ProductType.y } },
             { DissociationType.LowCID, new List<ProductType>{ ProductType.b, ProductType.y, ProductType.aStar, ProductType.bAmmoniaLoss, ProductType.yAmmoniaLoss, ProductType.aDegree, ProductType.bWaterLoss, ProductType.yWaterLoss } },
             { DissociationType.IRMPD, new List<ProductType>{ ProductType.b, ProductType.y } },
-            { DissociationType.ECD, new List<ProductType>{ ProductType.c, ProductType.y, ProductType.zDot } },
+            { DissociationType.ECD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
             { DissociationType.PQD, new List<ProductType>() },
-            { DissociationType.ETD, new List<ProductType>{ ProductType.c, ProductType.y, ProductType.zDot } },
+            { DissociationType.ETD, new List<ProductType>{ ProductType.c, ProductType.zDot } },
             { DissociationType.HCD, new List<ProductType>{ ProductType.b, ProductType.y } },//HCD often creates a-, aStar, and aDegree-ions and we should examine what other prominent algoroithms do to see if that would benefit our search results
             { DissociationType.AnyActivationType, new List<ProductType>{ ProductType.b, ProductType.y } },
             { DissociationType.EThcD, new List<ProductType>{ ProductType.b, ProductType.y, ProductType.c, ProductType.zDot } },
--- a/mzLib/Omics/Fragmentation/Peptide/DissociationTypeCollection.cs
+++ b/mzLib/Omics/Fragmentation/Peptide/DissociationTypeCollection.cs
@@ -92,8 +92,10 @@ namespace Omics.Fragmentation.Peptide
                     break;
                 // ECD/ETD produce c and z• only (N-Cα cleavage), so there is no y series and therefore no
                 // y water/ammonia loss to report -- a y-derived loss ion with no parent y has no mechanism,
-                // the same argument that removed ProductType.y from the ECD/ETD product sets above. Falls
-                // through to default and returns an empty list.
+                // the same argument that removed ProductType.y from the ECD/ETD product sets above.
+                case DissociationType.ECD:
+                case DissociationType.ETD:
+                    break;
                 default:
                     break;
             }
--- a/mzLib/Test/Omics/FragmentationTests/FragmentGenerationTests.cs
+++ b/mzLib/Test/Omics/FragmentationTests/FragmentGenerationTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2012, 2013, 2014 Derek J. Bailey
+// Copyright 2012, 2013, 2014 Derek J. Bailey
 // Modified work copyright 2016 Stefan Solntsev
 //
 // This file (FragmentGenerationTests.cs) is part of Proteomics.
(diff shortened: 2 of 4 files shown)


trishorts:
**Follow-up commit `bdabc170` — completing the y removal, plus a fragmentation-chemistry + computational second opinion**

The PR body asks for a mechanism check from someone who runs ETD. Here's that assessment alongside the review fixes.

## The change is scientifically right — and review found it was incomplete

**The mechanism supports removing y.** ECD (Zubarev/McLafferty 1998) and ETD (Syka/Coon/Hunt 2004) are radical-driven cleavages of the **N–Cα** backbone bond, producing **c (even-electron)** and **z• (radical)** ions. b and y come from **amide (C–N)** cleavage under *vibrational* activation (CID/HCD/IRMPD) — a different bond, a different mechanism. So the base ETD/ECD set is `{c, z•}`, exactly as the PR sets it.

**"y without b has no mechanism" is the crux, and it's correct.** b and y are complementary termini of *one* amide-cleavage event; any process that genuinely makes amide-cleavage ions in an electron experiment makes b **and** y together. That process is real and named — supplemental collisional activation of the charge-reduced ETnoD population and unreacted precursor, i.e. **EThcD/ETciD** — which is precisely why leaving `EThcD = {b, y, c, z•}` untouched is not just acceptable but is the *confirming* evidence.

**But the removal wasn't complete** (the review's two Major findings, and the reason for this commit): `GetWaterAndAmmoniaLossProductTypesFromDissociation` still returned `yWaterLoss`/`yAmmoniaLoss` for ECD/ETD. Those are y-series ions — a y-derived neutral loss with no parent y carries the *exact* mechanism problem the PR rejects. So an ETD/ECD search with loss ions enabled kept generating them, and a test in an edited file was pinned to expect them. That arm is now deleted (empty list), the test asserts empty, and the intent test pins the loss helper across all termini so it can't regress.

**Honest caveats** (don't undermine the change): real ETD spectra can carry *minor* b/y from imperfect isolation or trace activation — but they come as b **and** y, and the right tool for that population is EThcD, which the PR preserves. ETD/ECD also make minor **w**-type (side-chain radical loss, e.g. Cys/Leu/Ile) and z+1/c• variants — none of which is y, and all out of scope here. Net: the change is strictly *more* correct than the status quo.

**Verdict: support merging.** The base set is `{c, z•}`; y (and its water/ammonia losses) had no mechanism in these methods.

## Computational impact — strictly favorable

Path: `ProductsFromDissociationType[type]` → `GetTerminusSpecificProductTypesFromDissociation` (result **cached** per `(type, terminus)`) → the `Fragment()` loop emits one `Product` per `(productType, cleavage site)`, so fragment count ≈ (# product types) × (peptide length).

- **~⅓ fewer theoretical fragments for ETD/ECD** (`{c, y, z•}` → `{c, z•}`; 90 → 60 on the albumin 31-mer), plus the y water/ammonia-loss ions now gone on the loss path — less allocation, less GC pressure.
- **Search matching gets proportionally cheaper** on ETD/ECD data — fewer theoreticals to match, never more. Strictly faster.
- **No algorithmic or structural change**; the dictionary edit is evaluated once and cached, so zero per-call overhead. No new hot-path branches.
- **The real win is correctness/scoring, not speed.** The removed y (and y-loss) ions were *matchable*; coincidental matches inflated matched-ion counts and Morpheus-style scores specifically on ETD/ECD, and should now tighten — improving target/decoy FDR discrimination. There is no computational cost to weigh against it.

## What changed in `bdabc170`

- Deleted the `case ECD: case ETD:` arm in `GetWaterAndAmmoniaLossProductTypesFromDissociation` (→ empty); `Test_WaterAndAmmoniaLossFragmentProductIons` asserts empty for ECD/ETD.
- Extended `ElectronTransferDissociationProducesCAndZDotOnly` to pin the loss helper (no y-series, all termini).
- Pinned the proline test's `ProductType` set per dissociation type (its total of 11 was equally satisfied by dropping the c series).
- Added the N–Cα rationale as inline comments; reverted the accidental UTF-8 BOM on two test files.
- Left `TestDissociationProductTypes`' subset assertion as-is on purpose (exact would break its intentionally-partial cases; the dedicated test carries the exact check). PR body reconciled with the diff (two `TerminusSpecific*` tests added).

All 83 fragmentation tests pass; patch coverage unaffected (the change deletes production lines and adds comments — no new coverable lines).

_Sources: [Riley & Coon, Anal. Chem. 2018](https://pubs.acs.org/doi/10.1021/acs.analchem.7b04810) · [ETD overview, ScienceDirect](https://www.sciencedirect.com/topics/medicine-and-dentistry/electron-transfer-dissociation) · [AI-ETD / supplemental activation, Riley 2017](https://pubmed.ncbi.nlm.nih.gov/28383247/) · [w-type ions in ETD, Kempkes 2018](https://analyticalsciencejournals.onlinelibrary.wiley.com/doi/10.1002/jms.4298)._

Alexander-Sol (approved):

nbollis (approved):

PR #1116, opened 2026-07-24
Convert MSFragger retention time from seconds to minutes

### LINKED ISSUE #1115: FlashLFQ silently mis-quantifies MSFragger results: retention time is read as minutes but MSFragger writes seconds
### Summary

`IQuantifiableRecord.RetentionTime` is documented as minutes, and FlashLFQ consumes it as minutes,
but the MSFragger reader passes the tool's `Retention` column through unconverted — and MSFragger
writes seconds. Quantifying an MSFragger `psm.tsv` therefore searches for every peptide at roughly
1/60th of its true retention time. The run does not fail; it returns intensities near zero, which is
the most damaging way for this to be wrong.

MetaMorpheus `.psmtsv` and `.osmtsv` are unaffected (they are genuinely in minutes). Since those
three are the only types implementing `IQuantifiableResultFile`, MSFragger is the sole affected
input — and it is the main reason a user would reach for the interface at all.

### The chain

1. `Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs:48-49` — the column is mapped
   verbatim, with no conversion:

   ```csharp
   [Name("Retention")]
   public double RetentionTime { get; set; }
   ```

2. This property satisfies `IQuantifiableRecord.RetentionTime`, whose contract is explicit —
   `Readers/BaseClasses/IQuantifiableRecord.cs:36-38`:

   ```csharp
   /// The retention time (in minutes) associated with the result
   public double RetentionTime { get; }
   ```

3. `FlashLFQ/ResultsReading/MzLibExtensions.cs:25` takes it at its word — the variable is *named*
   for the unit it is not in:

   ```csharp
   double ms2RetentionTimeInMinutes = record.RetentionTime;
   ```

4. `FlashLfqEngine.cs:510` and `:528` use it for peak finding, and `:1839-1840` builds the search
   window in minutes:

   ```csharp
   double rightWindow = mostCommonChargeIdGroup.Select(p => p.Ms2RetentionTimeInMinutes).Max() + 2;
   double leftWindow  = mostCommonChargeIdGroup.Select(p => p.Ms2RetentionTimeInMinutes).Min() - 2;
   ```

   So a peptide eluting at 60 min is searched at 1 min ± 2 min, and is not found.

### Evidence that the column is seconds

From the committed fixture
`Test/FileReadingTests/ExternalFileTypes/FraggerPsm_FragPipev21.1_psm.tsv`:

| scan | 3 | 4 | 5 | 9 | 18 |
|---|---|---|---|---|---|
| `Retention` | 1.9398 | 2.9446 | 4.1263 | 8.8589 | 19.114 |

That is ~1.1 units per scan on an LTQ Orbitrap Velos — i.e. ~1.1 **seconds** per scan, which is a
normal duty cycle. Read as minutes it would imply about 80 scans in a whole Hela run. The existing
test `TestMsFraggerResultFiles.cs:103` pins `RetentionTime == 1.9398` for scan 3, so the parsing is
correct; only the unit contract is violated.

TopPIC has the same shape (`ToppicPrsm.cs:55-56`; `TestToppicResultFiles.cs:246` pins 340.12 for the
first PrSM), though `ToppicPrsm` does not implement `IQuantifiableRecord`, so it does not reach
FlashLFQ today.

### Why this is easy to miss

mzLib already normalizes retention time on the **spectra** side and not on the **result-file** side:

- `Readers/Bruker/BrukerFileReader.cs:318-321` — *"Bruker records retention time in seconds; mzLib's
  convention (mzML/Thermo) is minutes"*, then divides by 60
- `MsAlign.cs:435-436`, `Mgf.cs:191`, `Mzml.cs:459-469` — same

and mzLib already *knows* result files disagree, working around it with a magnitude heuristic rather
than at the reader — `Readers/Deconvolution/FromFileDeconvolutionParameters.cs:128-141`:

> *Heuristic: if the largest RetentionTimeEnd in the file exceeds 500, it can't be minutes — no
> realistic LC run is over 8 hours long — so the file is in seconds and every RT value is divided
> by 60.*

So there are two conventions in one library, and nothing enforces the documented one. No test
asserts units for any result-file format, which is why this passes CI.

### Related

Not the subject of this issue. Two items originally noted here have been removed:

- **`MsFraggerPsm.MonoisotopicMass`** — retracted in the comment below; both formats return the
  theoretical peptide mass and agree, so there is nothing to fix.
- **`MsFraggerPsm.IsDecoy` hardcoded `false`** — not a defect. MSFragger `psm.tsv` carries no
  target/decoy column (FragPipe/Philosopher strip decoys before writing), so every row is genuinely
  a target and `false` is operationally correct for FlashLFQ. The real question it raises — how the
  interface should express a value a source format simply does not provide, so the Python/Rust/R
  bridges don't each reinvent a stop-gap — is an interface/bridge **design** matter, not a bug on
  this issue, and is being handled separately.

### Possible fixes

Roughly in order of how much I would trust them:

1. **Convert in the reader**, matching what the spectra readers already do — divide MSFragger's
   `Retention` by 60 in the `IQuantifiableRecord` implementation, leaving the raw CsvHelper-mapped
   property untouched so round-trip writing is unaffected. Smallest change, makes the documented
   contract true, and no consumer changes.
2. **Make the unit explicit on the interface** — e.g. add `RetentionTimeUnit` and have consumers
   convert. More honest across the 29 formats (TopFD changed units between v1.6.2 and v1.7.0, so a
   per-format constant is not sufficient in general), but a breaking interface change.
3. Reuse the existing `>500` heuristic from `FromFileDeconvolutionParameters`. Cheap, but it is a
   guess, and it fails for short gradients where seconds and minutes are both plausible.

I would suggest (1) now plus a test asserting units per format, and (2) considered separately.

### Impact note

This surfaced while building a Python binding over `Readers`/`FlashLFQ`
([pyMzLib](https://github.com/smith-chem-wisc/pyMzLib)), which is why it is written up from the
outside. Any mzLib caller quantifying MSFragger results is affected, including directly in C#.

trishorts:
Fixes #1115.

`IQuantifiableRecord.RetentionTime` is documented as minutes and every consumer treats it as minutes, but `MsFraggerPsm` passed MSFragger's `Retention` column through unconverted — and MSFragger writes **seconds**.

`MzLibExtensions.MakeIdentifications` then assigns it to a variable literally named `ms2RetentionTimeInMinutes`, and `FlashLfqEngine` builds a **±2 minute** peak-finding window around it. So a peptide eluting at 60 minutes was searched for at 1 minute ± 2. Quantification collapses toward zero rather than failing — the most damaging way for this to be wrong, because the run still "succeeds".

Of the three types implementing `IQuantifiableResultFile`, MetaMorpheus `.psmtsv`/`.osmtsv` are genuinely in minutes and were always fine. MSFragger was the only one wrong, and it is the main reason someone reaches for the interface from outside MetaMorpheus.

## The fix

The raw column keeps a property of its own, now **named for its unit**:

```csharp
[Name("Retention")]
public double RetentionTimeInSeconds { get; set; }      // raw, as MSFragger wrote it

[Ignore] public double RetentionTime => RetentionTimeInSeconds / 60.0;   // the interface contract
```

Keeping the raw value means the file still round-trips on write (the existing round-trip test covers this) and anyone wanting the tool's verbatim number still has it. The conversion happens at the interface boundary — which is exactly what the spectra readers already do: `BrukerFileReader.cs:318` ("Bruker records retention time in seconds; mzLib's convention is minutes"), `MsAlign.cs:435`, `Mgf.cs:191`, `Mzml.cs:459`. Both halves of the library now agree on the unit.

## The bug was pinned by tests

Worth flagging, because it explains how this survived:

- `TestIdentificationAdapter.cs:58` asserted `Ms2RetentionTimeInMinutes == 1.9398` — the seconds value, sitting in a property named for minutes.
- `TestMsFraggerResultFiles.cs:103` asserted the same number on the record.

Both are corrected. Two new tests pin the *contract* rather than the number: that the interface value is the raw one over 60, and a gradient sanity check on the whole fixture (the five scans are ~1.1 s apart on an LTQ Orbitrap Velos, which read as minutes would imply a whole HeLa run of about 80 scans). The gradient test was tightened in review — see the follow-up below.

## Not in this PR

Two more unenforced contracts on the same interface, found alongside this and described in #1115. Both change semantics FlashLFQ consumes, so they deserve their own review:

- `MsFraggerPsm.IsDecoy` is hardcoded `false` — consumers cannot tell "target" from "unknown".
- `MsFraggerPsm.MonoisotopicMass` returns `CalculatedPeptideMass` (theoretical) while the psmtsv readers return the observed mass.

Also worth considering separately: no test asserted retention-time units for *any* result format, which is why this passed CI. TopPIC has the same seconds convention (`ToppicPrsm.cs:55`) and only escapes the bug because it does not implement `IQuantifiableRecord`.

## Review follow-up (commit `6c04e01f`)

A local review round hardened the tests and trimmed documentation, with no change to production behavior:

- **The gradient test now pins the divisor.** As first written, `MsFraggerPsm_RetentionTimesAreAPlausibleGradientInMinutes` only asserted `RetentionTime < 1.0` (true for *any* divisor ≥ 60, so a `/600` or `/3600` bug would still pass) and `RetentionTime == RetentionTimeInSeconds / 60.0` (a restatement of the getter, so it can't fail), with its assertions inside a `foreach` that an empty fixture would satisfy vacuously. It now guards `Has.Count.EqualTo(5)` and asserts against fixture literals external to the property under test — the last scan equals `19.114 / 60` and the first-to-last span equals `(19.114 − 1.9398) / 60` — so a wrong conversion factor fails the test.
- **Trimmed documentation that would rot.** The new `<remarks>` on `MsFraggerPsm` had narrated the bug history and hard-coded FlashLFQ internals (a consumer method name, one of its locals, the ±2-minute window) into a `Readers` record; they're reduced to the unit contract in the spectra-readers' style, and non-ASCII typography removed. The change-narration comment in `TestIdentificationAdapter` is trimmed to the durable fact that the column is seconds. That history stays in the commit message and this PR body, where it belongs.

Patch coverage remains 100% (2/2 changed lines).

## Testing

833 tests pass across the file-reading, quantifiable, FlashLFQ and identification-adapter suites.

Found while building a Python binding over `Readers`/`FlashLFQ` ([pyMzLib](https://github.com/smith-chem-wisc/pyMzLib)), which is why it is written up from the outside — but it affects any mzLib caller quantifying MSFragger results, including directly in C#.

https://claude.ai/code/session_01Jp7uYMFdRCEWvg7sWZHFr2

### DIFF
--- a/mzLib/Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs
+++ b/mzLib/Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs
@@ -46,16 +46,10 @@ namespace Readers
         public int Charge { get; set; }
 
         /// <summary>
-        /// The raw "Retention" column, in SECONDS — the unit MSFragger writes.
+        /// The raw "Retention" column, in seconds, which is the unit MSFragger writes.
+        /// The verbatim value is kept here so the file round-trips on write; conversion to
+        /// minutes for <see cref="IQuantifiableRecord.RetentionTime"/> happens at the interface below.
         /// </summary>
-        /// <remarks>
-        /// Named for its unit because the unit is the whole problem. mzLib's convention for
-        /// retention time is minutes (see <see cref="IQuantifiableRecord.RetentionTime"/>, and the
-        /// spectra readers, which all divide by 60), but this column is not in minutes, and reading
-        /// it as though it were made FlashLFQ search for every peptide at a sixtieth of its true
-        /// elution time. Keep the raw value here so the file round-trips on write, and convert at
-        /// the interface boundary below.
-        /// </remarks>
         [Name("Retention")]
         public double RetentionTimeInSeconds { get; set; }
 
@@ -226,21 +220,11 @@ namespace Readers
         [Ignore] public int ChargeState => Charge;
 
         /// <summary>
-        /// The retention time in MINUTES, converted from the seconds MSFragger writes.
+        /// The retention time in minutes, per the <see cref="IQuantifiableRecord.RetentionTime"/>
+        /// convention, converted from the seconds MSFragger writes in <see cref="RetentionTimeInSeconds"/>.
+        /// This mirrors the seconds-to-minutes conversion the spectra readers already do at their
+        /// own boundary (BrukerFileReader, MsAlign, Mgf, Mzml).
         /// </summary>
-        /// <remarks>
-        /// <see cref="IQuantifiableRecord.RetentionTime"/> is documented as minutes and every
-        /// consumer treats it that way — <c>MzLibExtensions.MakeIdentifications</c> assigns it
-        /// straight to a variable named <c>ms2RetentionTimeInMinutes</c>, and FlashLFQ then builds
-        /// a ±2 minute peak-finding window around it. Passing MSFragger's seconds through
-        /// unconverted therefore looked for a peptide eluting at 60 minutes at the 1 minute mark,
-        /// and quantification collapsed toward zero rather than failing.
-        /// <para>
-        /// This mirrors what the spectra readers already do at their own boundary
-        /// (<c>BrukerFileReader</c>, <c>MsAlign</c>, <c>Mgf</c>, <c>Mzml</c> all normalise to
-        /// minutes), so both halves of the library now agree on the unit.
-        /// </para>
-        /// </remarks>
         [Ignore] public double RetentionTime => RetentionTimeInSeconds / 60.0;
 
         // decoy reading isn't currently supported for MsFragger psms, this will be revisited later
--- a/mzLib/Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs
+++ b/mzLib/Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs
@@ -45,8 +45,19 @@ namespace Readers
         [Name("Charge")]
         public int Charge { get; set; }
 
+        /// <summary>
+        /// The raw "Retention" column, in SECONDS — the unit MSFragger writes.
+        /// </summary>
+        /// <remarks>
+        /// Named for its unit because the unit is the whole problem. mzLib's convention for
+        /// retention time is minutes (see <see cref="IQuantifiableRecord.RetentionTime"/>, and the
+        /// spectra readers, which all divide by 60), but this column is not in minutes, and reading
+        /// it as though it were made FlashLFQ search for every peptide at a sixtieth of its true
+        /// elution time. Keep the raw value here so the file round-trips on write, and convert at
+        /// the interface boundary below.
+        /// </remarks>
         [Name("Retention")]
-        public double RetentionTime { get; set; }
+        public double RetentionTimeInSeconds { get; set; }
 
         [Name("Observed Mass")]
         public double ObservedMass { get; set; }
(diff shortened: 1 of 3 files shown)


trishorts:
**Follow-up commit `6c04e01f` — review fixes (test strength + doc cleanup)**

A local review flagged that the new tests, while green, didn't actually defend the conversion, and that some new documentation would rot. This commit addresses that without touching the production behavior.

**1. The gradient test now pins the divisor.** `MsFraggerPsm_RetentionTimesAreAPlausibleGradientInMinutes` previously asserted only:
- `RetentionTime < 1.0` — holds for *any* divisor ≥ 60, so a `/600` or `/3600` bug would still pass; and
- `RetentionTime == RetentionTimeInSeconds / 60.0` — restates the getter's own body, so it can't fail.

It also had its assertions inside `foreach` with no non-empty guard, so an empty fixture would have passed vacuously. It now:
- guards `Has.Count.EqualTo(5)` (matching `TestMsFraggerPsmLoadsAndCountCorrect`);
- asserts the last scan equals `19.114 / 60` and the first-to-last **span** equals `(19.114 − 1.9398) / 60` — both literals read from the fixture, external to the property under test. A wrong divisor now fails the test.

**2. Trimmed documentation that would go stale.** The two new `<remarks>` blocks on `MsFraggerPsm` narrated the bug history and baked in FlashLFQ internals — a consumer method name, one of its local variables, and the ±2-minute window width — none of which this `Readers` record owns. They're reduced to the unit contract in the same style the spectra readers use, and the non-ASCII typography (`—`, `±`) is gone. The change-narration comment in `TestIdentificationAdapter` ("previously read…") is trimmed to the durable fact that the column is seconds.

**Not changed:** the `RetentionTimeInSeconds` rename + `[Ignore]` computed `RetentionTime` — that's the PR's payload and is correct. The bug-history narrative still lives where it belongs, in the original commit message and PR body.

Patch coverage remains 100% (2/2 changed lines); affected tests pass.

nbollis (approved):

RayMSMS (approved):

PR #1118, opened 2026-07-24
IQuantifiableRecord.IsDecoy → bool? (a format without a decoy column can report "not provided")

### LINKED ISSUE #1117: IQuantifiableRecord.IsDecoy should be bool? so a format without a decoy column can report "not provided"
### Summary

`IQuantifiableRecord.IsDecoy` is typed `bool`, so a source format that has **no target/decoy
column** is forced to report a value it does not have. `MsFraggerPsm.IsDecoy` returns a hardcoded
`false` (`MsFraggerPsm.cs`, "decoy reading isn't currently supported"), and MSFragger `psm.tsv`
genuinely carries no target/decoy column — FragPipe/Philosopher strip decoys before writing it. So
every consumer reads `false` as **"known target"** when the truthful answer is **"not provided."**

Today this happens to be harmless for FlashLFQ (an all-targets file quantifies fine), so it is not a
bug. But it is a contract gap: the interface cannot express *"this format does not say,"* and the
Python / Rust / R bridges over `Readers`/`FlashLFQ` would otherwise each have to reinvent a per-bridge
stop-gap for it. The fix belongs once, in the core contract, where every consumer — C# and every
bridge — inherits it.

### Proposal

Change **`IQuantifiableRecord.IsDecoy` from `bool` to `bool?`**, where `null` = "the source format
does not provide target/decoy information." This matches mzLib's existing way of expressing an
optional record field:

- `Proteomics.UniProtSequenceAttributes.IsPrecursor` is `bool?` (commented `//optional`; `null` when
  UniProt omits it).
- `Omics.SpectralMatch.ISpectralMatch.Intensities` is `double[]?` ("if no quantification was
  performed, this will be null").

Each target language has a first-class "missing" for booleans, so `bool?` projects with **no
per-bridge stop-gap**: C# `bool?`, Python `Optional[bool]`/`None`, Rust `Option<bool>`, R logical `NA`.

### Semantics

| `IsDecoy` | meaning | consumer policy |
|---|---|---|
| `true` | known decoy | exclude from quant |
| `false` | known target | include |
| `null` | format does not provide it | treat as not-a-decoy (`?? false`) — preserves today's behaviour |

### Scope (kept contained to mzLib — no MetaMorpheus break)

`LightWeightSpectralMatch` implements **both** `IQuantifiableRecord` and
`Omics.SpectralMatch.ISpectralMatch` through one `bool IsDecoy`, and `SpectrumMatchFromTsv`'s public
`bool IsDecoy` is consumed by MetaMorpheus. To avoid cascading into `BaseSpectralMatch` and
MetaMorpheus, change **only** `IQuantifiableRecord.IsDecoy`, and give those implementers an **explicit
interface implementation** (`bool? IQuantifiableRecord.IsDecoy => IsDecoy;`) so their public `bool`
API is untouched. The two `ISpectralMatch` interfaces stay `bool` (a parsed spectral match always
knows its decoy status).

Touched:

- `IQuantifiableRecord.IsDecoy` → `bool?` (+ doc: `null` = not provided).
- `MsFraggerPsm.IsDecoy` → returns `null` (no decoy column) instead of `false`.
- `SpectrumMatchFromTsv`, `LightWeightSpectralMatch` → explicit `IQuantifiableRecord.IsDecoy` impl,
  public `bool IsDecoy` unchanged.
- `FlashLFQ/ResultsReading/MzLibExtensions.cs` → collapse to `bool` at the boundary with `?? false`
  (the established decoy idiom from `Protein.cs`); `FlashLFQ.Identification.IsDecoy` stays `bool`.
- Test double `MockQuantifiableRecord.IsDecoy` → `bool?`.

### Not in scope

- The two `ISpectralMatch` interfaces and `BaseSpectralMatch` (parsed matches always know decoy
  status; changing them would break MetaMorpheus and is not needed).
- A per-format "capability descriptor" system — YAGNI while `IsDecoy` is the only availability-varying
  field; revisit if more surface.

### Context

Surfaced from #1115 (originally filed as a "related" bullet, since removed) while building the
[pyMzLib](https://github.com/smith-chem-wisc/pyMzLib) binding and planning the Rust/R siblings: the
recurring shape is *a record value whose availability varies by format*, and the efficient home for
the fix is the shared contract, not each bridge.

trishorts:
Closes #1117.

## The change

`IQuantifiableRecord.IsDecoy` was `bool`, forcing every record to claim a target/decoy status even when its source format has no such column. MSFragger `psm.tsv` has no decoy column (FragPipe/Philosopher strip decoys before writing), so `MsFraggerPsm.IsDecoy` returned a hardcoded `false` — which every consumer reads as **"known target"** when the truthful answer is **"not provided."**

```diff
-public bool IsDecoy { get; }
+public bool? IsDecoy { get; }   // true = decoy, false = target, null = format does not provide it
```

It is harmless for FlashLFQ today (an all-targets file quantifies fine), so this is not a bug fix — it is a contract fix. The interface could not express *"this format does not say,"* and the Python / Rust / R bridges over `Readers`/`FlashLFQ` would otherwise each reinvent the same per-bridge stop-gap. The fix belongs once, in the shared contract.

## Why `bool?`

It matches how mzLib already expresses an optional record field:
- `Proteomics.UniProtSequenceAttributes.IsPrecursor` — `bool?`, commented `//optional`
- `Omics.SpectralMatch.ISpectralMatch.Intensities` — `double[]?`, "null if no quantification was performed"

And each target language has a first-class "missing" for booleans, so it projects with **no per-bridge stop-gap**: C# `bool?`, Python `Optional[bool]`/`None`, Rust `Option<bool>`, R logical `NA`.

| `IsDecoy` | meaning | consumer policy |
|---|---|---|
| `true` | known decoy | exclude from quant |
| `false` | known target | include |
| `null` | format does not provide it | treated as not-a-decoy (`?? false`) — preserves today's behaviour |

## Kept contained to mzLib — no MetaMorpheus break

`LightWeightSpectralMatch` implements **both** `IQuantifiableRecord` and `Omics.SpectralMatch.ISpectralMatch` through one `bool IsDecoy`, and `SpectrumMatchFromTsv`'s public `bool IsDecoy` is consumed by MetaMorpheus. Rather than cascade the nullable into `ISpectralMatch` / `BaseSpectralMatch` (and MM), **only `IQuantifiableRecord.IsDecoy` changes**; those two implementers get an **explicit interface implementation** (`bool? IQuantifiableRecord.IsDecoy => IsDecoy;`) so their public `bool` API is untouched. The two `ISpectralMatch` interfaces stay `bool` — a parsed spectral match always knows its decoy status.

## Touched

- `IQuantifiableRecord.IsDecoy` → `bool?` (+ doc).
- `MsFraggerPsm.IsDecoy` → `null` (no decoy column) instead of `false`.
- `SpectrumMatchFromTsv`, `LightWeightSpectralMatch` → explicit `IQuantifiableRecord` impl; public `bool IsDecoy` unchanged.
- `FlashLFQ/ResultsReading/MzLibExtensions.cs` → collapse to `bool` at the boundary with `?? false` (the established decoy idiom, cf. `Protein.cs`); `FlashLFQ.Identification.IsDecoy` stays `bool`.
- Test double `MockQuantifiableRecord.IsDecoy` → `bool?`.

## Tests

- `MakeIdentifications_IsDecoyNullOrTrueOrFalse_MapsToTargetDecoyCorrectly` — `null` → target (quantified, `UseForProteinQuant` true); `true` → decoy (not used for quant); `false` → target.
- `MsFraggerPsm_IsDecoy_IsNull_BecauseFormatHasNoDecoyColumn` — the reader reports `null` through the interface.

Full solution builds; all FlashLFQ / MsFragger / lightweight-PSM adapter tests pass.

## Context

Surfaced from #1115 while building [pyMzLib](https://github.com/smith-chem-wisc/pyMzLib) and planning the Rust/R siblings: the recurring shape is *a record value whose availability varies by format*, and the efficient home for the fix is the shared contract, not each bridge. A per-format "capability descriptor" was considered and rejected as premature — `IsDecoy` is currently the only availability-varying field; revisit if more surface.

https://claude.ai/code/session_01Jp7uYMFdRCEWvg7sWZHFr2

### DIFF
--- a/mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
+++ b/mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
@@ -69,8 +69,10 @@ namespace FlashLFQ
                     ms2RetentionTimeInMinutes, 
                     precursorChargeState, 
                     proteinGroups, 
-                    useForProteinQuant: !record.IsDecoy, 
-                    decoy: record.IsDecoy,
+                    // A record whose format does not carry decoy info (IsDecoy == null) is treated as
+                    // a target for quantification, matching the pre-nullable behaviour.
+                    useForProteinQuant: !(record.IsDecoy ?? false),
+                    decoy: record.IsDecoy ?? false,
                     psmScore: score,
                     qValue: usePepQValue ? pepQValue : qValue);
                 identifications.Add(id);
--- a/mzLib/Readers/BaseClasses/IQuantifiableRecord.cs
+++ b/mzLib/Readers/BaseClasses/IQuantifiableRecord.cs
@@ -43,9 +43,11 @@ namespace Readers
         public int ChargeState { get; }
 
         /// <summary>
-        /// Defines whether or not the result is a decoy identification
+        /// Whether the result is a decoy identification: <c>true</c> = decoy, <c>false</c> = target,
+        /// <c>null</c> = the source format does not carry target/decoy information (e.g. MSFragger
+        /// psm.tsv, which has no decoy column). Consumers should treat <c>null</c> as not-a-decoy.
         /// </summary>
-        public bool IsDecoy { get; }
+        public bool? IsDecoy { get; }
 
         /// <summary>
         /// The mass of the monoisotopic peptide (i.e., no c13 or n15 atoms are present, the lowest possible mass)
--- a/mzLib/Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs
+++ b/mzLib/Readers/ExternalResults/IndividualResultRecords/MsFraggerPsm.cs
@@ -214,8 +214,9 @@ namespace Readers
 
         [Ignore] public int ChargeState => Charge;
 
-        // decoy reading isn't currently supported for MsFragger psms, this will be revisited later
-        [Ignore] public bool IsDecoy => false;
+        // MSFragger psm.tsv has no target/decoy column -- FragPipe/Philosopher strip decoys before
+        // writing it -- so decoy status is genuinely not provided. null (not a fake false) says so.
+        [Ignore] public bool? IsDecoy => null;
 
         [Ignore] public double MonoisotopicMass => CalculatedPeptideMass;
 
--- a/mzLib/Readers/InternalResults/IndividualResultRecords/LightWeightSpectralMatch.cs
+++ b/mzLib/Readers/InternalResults/IndividualResultRecords/LightWeightSpectralMatch.cs
@@ -33,6 +33,11 @@ namespace Readers
         public int ChargeState { get; }
         public double MonoisotopicMass { get; }
 
+        // The shared bool IsDecoy above satisfies ISpectralMatch (a parsed match always knows its
+        // decoy status). Explicit implementation presents it through the nullable IQuantifiableRecord
+        // contract without forcing the public property, or ISpectralMatch, to nullable.
+        bool? IQuantifiableRecord.IsDecoy => IsDecoy;
+
         // Additional fields for filtering / downstream use
         public double QValue { get; }
         public double PepQValue { get; }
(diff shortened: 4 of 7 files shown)


Alexander-Sol:
I'm against this change. I'm not aware of any software that will report decoys in their results and not mark them as such. I think we can safely assume that reported psms are targets unless they are explicitly labeled decoys

PR #1119, opened 2026-07-24
Opt-in: treat a cleavage-blocking modification as abolishing its cleavage site (Modification.BlocksCleavage)

trishorts:
> **Project of origin:** `cleavageRules`

Trypsin cannot cleave after an acylated lysine. Digestion applies modifications *after* cleavage, so nothing checked whether a modification abolished the very site a peptide was cut at — and `PEPTIDEK[acetyl]` was reported as a clean, zero-missed-cleavage tryptic peptide describing an event that cannot occur, while the real peptide (`PEPTIDEK[acetyl]AAAAAAAR`, reading through the blocked site) was scored as a missed cleavage and often not generated at all.

This PR adds `Modification.BlocksCleavage` and an **opt-in** `DigestionParams.RespectCleavageBlockingModifications`. Default is `false`, which reproduces historical digestion exactly.

## What it does

With the flag on, in a full-specificity search, digestion performs a **trade**:

```
Protein AAAAAARPEPTKPEPTRAAAAAAK, acetyl-K variable, maxMissedCleavages: 0

flag OFF                          flag ON
  PEPTK                             PEPTK
  PEPTK[Acetyl]      <- impossible  PEPTK[Acetyl]PEPTR   <- the real peptide, MC=0
  PEPTR                             PEPTR
```

The impossible peptidoform leaves and the read-through that replaces it arrives, so the peptide stays identifiable. Both halves matter, and they are deliberately coupled — see *Scope* below.

Four pieces make that work:

1. **`Modification.BlocksCleavage`** — classifies a modification as abolishing its cleavage site. Only side-chain (`Anywhere.`) acylations on K, and citrullination on R, qualify. Charge is the criterion: acylation removes the ε-amine's positive charge, so trypsin cannot cleave; methylation retains it and is therefore excluded (it is a genuine missed cleavage, not an impossible peptide). N-terminal α-amine acetylation on K is correctly *not* blocking — the side chain stays charged.
2. **`CleavageBlockingModifications.BlocksCleavageBy(mod, protease)`** — blocking is a property of the PAIR, not of the modification. An acylated lysine abolishes a trypsin site and abolishes nothing in a Glu-C digest, which never cleaved after that lysine. Digestion pairs the chemistry above with the new `DigestionAgent.CleavesCTerminalTo(char)`, read off the protease's own motifs, so the correction applies under trypsin, Lys-C, Arg-C, elastase and subtilisin and is inert under Glu-C, chymotrypsin and CNBr. It is inert for Asp-N and Lys-N too, by direction: they cut *N*-terminal to their recognition residue, where a blocked residue invalidates a peptide's N-terminus instead — the mirror image of the drop below, and not modelled here.
3. **The C-terminal drop** — a peptidoform whose C-terminal residue carries a blocking modification, where that C-terminus is an internal protease cut, is dropped. A peptide ending at the protein's own C-terminus is exempt: no cleavage happened there for a modification to block.
4. **Missed cleavages count cleavage events, not residues** — a blocked residue is not a cleavage site for the peptidoform carrying it, so it is not a missed cleavage either. Digestion enumerates a wider span internally to reach read-through forms, but that slack is discounted again before a peptide is emitted, so `MaxMissedCleavages` keeps its meaning: **no peptide leaves digestion reporting more missed cleavages than the caller asked for.**

## Scope: what is and is not in this PR

**In scope — full-specificity searches only** (`SearchModeType == CleavageSpecificity.Full`).

**Explicitly out of scope — semi-specific and nonspecific searches are left entirely untouched**, and that is a deliberate design decision rather than an oversight.

The drop and the wider generation span are two halves of one exchange. The slack is granted only in a full search, so applying the drop outside one performs half the trade. Measured, flag off vs on:

| mode | maxMC | dropped | added | net |
|---|---|---|---|---|
| Full | 0 | `PEPTK[Acetyl]` | `PEPTK[Acetyl]PEPTR` | 6 → 6 |
| Semi | 0 | `PEPTK[Acetyl]` | *(none)* | 6 → **5** |

In the semi case the peptide becomes **unidentifiable** — the wrong answer is removed and the right one never generated. That is strictly worse than leaving it alone, so the drop now requires `SearchModeType == Full` too. It would have bitten whenever the read-through needs more budget than `MaxMissedCleavages` allows: one blocked site at 0, three consecutive at 2.

Making semi searches benefit properly is follow-up work, and it is **not** simply widening that gate. A semi peptide's C-terminus may be a genuine protease cut or a length-driven truncation, and telling them apart requires the protease's cleavage-site list — which full digestion gets free from its own enumeration and semi digestion does not.

## Known limitations, carried knowingly

- **The classifier is name-keyed.** `BlocksCleavage` matches acyl stems in the modification's name. Broader Unimod coverage is a curation refinement. A user-declarable blocking flag persisted in the custom-modification `.txt` (the `BM`/`BL` precedent) is the real upgrade path.
- **Blocked internal sites are matched at residue level, not by re-deriving the protease's site list for the protein.** Since round 4 the protease *identity* is consulted, so this is now only about sequence context: a blocked K sitting before a proline under `trypsin|P` can still discount a site the `K[P]` rule means it never occupied. The count is clamped so it cannot go negative, and the C-terminal drop is exact.
- **Fixed blocking modifications** remain an unbounded edge: the generation slack is tied to `MaxMods`, which bounds variable modifications only. (A fixed blocking modification does now *buy* the slack — the round-4 gate consults `AllKnownFixedModifications` as well as `VariableModifications` — it is the size of the slack that still assumes variable mods.)
- **Truncation-product ends** beyond the protein C-terminus are not exempted from the drop.
- **Not exposed to MetaMorpheus.** The flag is a `DigestionParams` constructor parameter only; it is not on `IDigestionParams`, has no RNA equivalent, and MetaMorpheus has zero references to it, so every MM path takes the `false` default. Surfacing it needs GUI plumbing and a check that the flag survives the `DigestionParameters.toml` round-trip through Nett given its private setter.

## Review history

**Round 2** (`02bce317`) fixed five real correctness bugs found in a second review pass: N-terminal α-amine acetylation on K misclassified as blocking; `Clone()` dropping the flag on its non-`None` path; `Digestion()` bypassing the mutable missed-cleavage property; a fixed slack of 2 losing peptides when more than two sites were blocked (now `MaxMods`); and the C-terminal drop firing on non-cut termini.

**Round 3** — from @nbollis's two review comments, seconded by @Alexander-Sol:

- *"the number of missed cleavages we get out of digestion may or may not match our digestion parameters"* — **confirmed and fixed.** At `maxMissedCleavages: 0` the emitted peptides reported `< 0, 0, 1 >`. The filter was already computing the blocked-discounted count to decide survival, but the peptide was still constructed with the modification-blind count, so the generation slack leaked out. The discounted count is now what the surviving peptidoform carries — which is the alternative both reviewers proposed.
- *"Cleavage specificity == Full will prevent this from working in nonspecific and semi-specific searches, is that intended?"* — **tested, and it was worse than intended.** It was not inert in semi searches, it was half-applied; see the table above. Now genuinely confined to full searches, with the reasoning documented at the gate.

**Round 4** — this round, from @acesnik's three review comments, plus the generalisation the first two point at:

- *the `methyl` guard runs before the acyl stems* — **confirmed against the shipped `unimod.xml` and fixed.** `Methyl+Acetyl:2H(3)` and `Methyl:2H(3)+Acetyl:2H(3)` are both site K, position Anywhere, and both carry an acetyl on the ε-amine, which is the chemistry the class exists for. The acyl test now has the last word, and the guard is gone rather than reordered: once the stems run first, `Methyl`, `Dimethyl` and `Trimethyl` carry no stem and fall through to false on their own, so keeping it would have left a branch that never decides anything.
- *the generation slack is paid even when nothing can block* — **fixed.** It is now bought only where it can be spent. `AllKnownFixedModifications` is consulted as well as `VariableModifications`, because fixed mods are appended to the variable pattern before the drop is evaluated — so a fixed blocking modification can trigger a drop and has to be able to buy the slack that replaces it.
- *`ToString` is the one member no test asserts* — **fixed**, on all three identity members together. It matters because MetaMorpheus keys its fragment-index cache on `IndexingEngine.ToString()` and `SameSettings` compares it verbatim, so a flag missing there would silently reuse a flag-off index for a flag-on run.
- *nothing consulted the protease* — **a behaviour change beyond the three threads.** The rule fired under any protease at all: in a Glu-C digest an internal acylated lysine discounted a missed cleavage at a site Glu-C never had, both under-reporting the count and letting an over-budget peptide through on the slack. Same shape under Arg-C for an acylated K and under Lys-C for a citrullinated R. Reproduced before being changed. See piece 2 above for what replaced it; the limitation this narrows is the second bullet above.

On **glycosylation**, asked in review and deliberately left out: `BlocksCleavage` is keyed on *charge*, and a glycan sits on N (in N-X-S/T) or S/T — never on a trypsin-family cleavage residue — so there is no charge to remove and the residue gate already rejects it. Where glycans do suppress cleavage the mechanism is *steric* and the effect is *partial*: a raised missed-cleavage rate, not an impossible peptidoform. That calls for the opposite correction to this one — more missed-cleavage tolerance, not a dropped peptidoform — so modelling it here would delete real peptides. For the glyco-proteases the relationship inverts outright (StcE, OpeRATOR/OgpA and flavastacin all *require* the glycan). The reasoning is now in the class remarks rather than only in a thread.

## Tests

New this round, each verified red against the previous source:

- `ReadThroughOfABlockedSite_ReportsZeroMissedCleavages_NotOne` — was 1, now 0.
- `NoEmittedPeptide_EverReportsMoreMissedCleavagesThanRequested(0/1/2)` — was `< 0, 0, 1 >` at a requested max of 0.
- `SemiSpecificSearch_IsEntirelyUnaffectedByTheFlag(0/1/2)` — fails on all three budgets without the new gate. This replaces `SemiSpecific_NonFullPeptides_AreUnaffectedByTheFlag`, which compared only the *non-Full subset* of a semi digest and therefore stayed green while the Full peptides inside that digest were being dropped unreplaced.
- `AnUnblockedMissedCleavage_StillCountsWithTheFlagOn` and `FullSearch_TradesTheImpossibleFormForTheReadThrough` — guard the opposite direction, so the fix cannot degrade into blanket zeroing or into dropping without replacing.
- `AnAcylationIsBlocking_EvenWhenItsNameAlsoCarriesAMethyl` — fails on `Methyl+Acetyl:2H(3)` against the old classifier order, while still pinning plain `Methyl`/`Dimethyl`/`Trimethyl` as non-blocking.
- `WithNoBlockingModificationConfigured_NoGenerationSlackIsPaid` — reads `ProteinDigestion.Digestion` directly and asserts the enumeration width. The slack is invisible in the finished digest (the open-site filter trims the surplus on the way out), so an output comparison passes either way and pins nothing. The opposite direction is pinned too, so the gate cannot pass by disabling the slack outright.
- `AProteaseThatDoesNotCleaveAfterTheModifiedResidue_IsUnaffected(Glu-C/Arg-C)` and `ABlockingModOnAResidueThisProteaseIgnores_DoesNotDiscountAMissedCleavage` — both fail without the protease pairing, at protease and at residue granularity. `AProteaseThatDoesCleaveAfterTheModifiedResidue_StillGetsTheCorrection` is the positive control, so the gate cannot pass by disabling the feature everywhere.
- `CleavesCTerminalTo_ReadsTheProteasesOwnMotifs` — pins the direction distinction that leaves Asp-N and Lys-N inert rather than mirror-imaged.
- `DigestionParams_DifferingOnlyInTheFlag_AreDistinguishableByEveryIdentityMember` — `ToString`, `Equals` and `GetHashCode`, plus the equal-params direction so it cannot pass by everything being unequal.

Plus the round-2 regressions: citrulline-on-R-only, `Clone` preserving the flag on both paths, a blocked residue at the protein C-terminus surviving, and a three-blocked-site read-through surviving at 0 missed cleavages.

Full offline suite: **5580 passed, 0 failed, 29 skipped** (5609 total).

> Note on CI: the `external-service-tests` failure described in earlier rounds (EBI's FTP unreachable from GitHub runners, inherited from master via #1121) has since recovered; all checks were green on the round-3 head.


mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
```diff
@@ -30,8 +31,22 @@ public DigestionParams(string protease = "trypsin", int maxMissedCleavages = 2,
...
+        /// <see cref="RespectCleavageBlockingModifications"/> is on. A peptidoform whose C-terminal
+        /// residue carries a cleavage-blocking modification is impossible, and its real counterpart
+        /// reads THROUGH that residue to the next site -- which costs a missed cleavage under the
+        /// ordinary (modification-blind) count. Without this slack that read-through form would never
+        /// be generated at low limits, and at MaxMissedCleavages = 0 the real peptide would be lost
+        /// entirely rather than merely mis-reported. The surplus is trimmed again by the open-site
+        /// filter in ProteolyticPeptide.GetModifiedPeptides, so only genuinely-reachable peptidoforms
+        /// survive; the cost is enumeration, not correctness.
+        /// </summary>
+        public const int CleavageBlockingReadThroughSlack = 2;
```
nbollis: This concerns me. It means that if we are doing the new mod blocking missed cleavages, then the number of missed cleavages we get out of digestion may or may not match our digestion parameters. This feels bad. 

Is this second property needed? An alternative approach would be to treat the mod blocking the cleavage, say acetyl lysine, as not a cleavage site and not considered for the missed cleavage count.

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
```diff
@@ -47,6 +62,28 @@ public DigestionParams(string protease = "trypsin", int maxMissedCleavages = 2,
...
+        public bool RespectCleavageBlockingModifications { get; private set; }
+
+        /// <summary>
+        /// The missed-cleavage budget used for GENERATION, as opposed to <see cref="MaxMissedCleavages"/>
+        /// which remains the budget the caller asked for and which the open-site filter enforces.
+        /// Only full-specificity digestion is inflated: the semi- and single-terminus modes do not
+        /// enumerate by missed cleavage in the same way, and the C-terminal drop still applies to them.
+        /// </summary>
+        public int EffectiveMaxMissedCleavages =>
+            RespectCleavageBlockingModifications && SearchModeType == CleavageSpecificity.Full
```
nbollis: Cleavage specificity == Full will prevent this from working in nonspecific and semi-specific searches, is that intended?

trishorts:
**Follow-up commit `02bce317` — round-2 review caught five real correctness bugs; all fixed**

A second `/review-pr` pass returned **Request changes** with 8 Major findings, and it was right — **five were genuine correctness bugs in the first commit**, not test nits. Writing them up plainly:

1. **N-terminal α-amine acetylation on K was misclassified as blocking.** The classifier decided from target motif + name substring and never looked at `LocationRestriction`. UniProt ships both `N6-acetyllysine` (ε-amine side chain, genuinely blocking, `Anywhere.`) and `N-acetyllysine` (α-amine, `N-terminal.`) — the terminal form leaves the side chain charged, so trypsin still cleaves, yet both matched. **Fix:** only side-chain (`Anywhere.`) modifications classify.

2. **`Clone()` dropped the flag on the non-None path.** The flag was passed on only the `None`-specificity return, so a cloned `DigestionParams` silently reverted to modification-blind digestion for every other mode (and compared unequal to its source). **Fix:** both paths pass it.

3. **`Digestion()` bypassed the mutable missed-cleavage property.** It read `EffectiveMaxMissedCleavages` off `DigestionParams` instead of the `MaximumMissedCleavages` instance property that `SpeedySemiSpecificDigestion` also reads, so a caller mutating that property had the mutation silently discarded — even flag-off. **Fix:** removed `EffectiveMaxMissedCleavages`; the budget is computed from `MaximumMissedCleavages`, so flag-off is byte-for-byte the historical path.

4. **Slack of 2 lost real peptides when >2 sites were blocked.** With three blocked K's in a row, the read-through spanning them was never generated while the shorter forms were still dropped — the peptide vanished. **Fix:** slack is now `MaxMods` (a peptidoform carries at most that many variable mods, hence blocked sites, so every variable-mod read-through is reachable). Fixed blocking mods remain a documented unbounded edge.

5. **The C-terminal drop fired on non-cut termini.** It ignored specificity, so semi- and single-terminus peptides whose C-terminus is a *length truncation* were wrongly dropped — my "exact, all modes" claim was false. **Fix:** gated on `CleavageSpecificityForFdrCategory == Full`, matching where the slack applies.

Also corrected: **arginine is now reachable and right** — the previously-dead R branch classifies **citrullination** (deimination removes Arg's charge → trypsin doesn't cleave); and the `BlocksCleavage` doc no longer claims a terminal acylated residue is an impossible peptide end (it's a real one — position is digestion's call).

**New regression tests** pin each fix: citrulline-on-R-only, `Clone` preserves the flag on both paths, a blocked residue at the protein C-terminus survives, semi non-Full peptides are untouched by the flag, and a three-blocked-site read-through survives at 0 missed cleavages with `slack = MaxMods`. The residue and location gates are now exercised (a blocking name on a non-cleavage residue; an N-terminal-K), and "reproduces historical behaviour" pins concrete pre-PR output plus that the impossible form is kept — instead of comparing two flag-off runs.

**Deferred, and raised here honestly for a follow-up** (all Minor):
- The surviving read-through still reports the modification-blind `MissedCleavages` count (the blocked site "isn't a cut" but the stored count doesn't reflect that). Recomputing it per-peptidoform is a deeper change.
- Truncation-product ends beyond the protein C-terminus aren't exempted from the drop.
- `BlocksCleavage` recomputes (lowercases + scans the name) on each access rather than caching.
- The name-keyed acyl-stem classifier is inherently a heuristic; broader Unimod coverage and the `contains("methyl")` bluntness are curation refinements. The persisted-custom-`.txt`-field path (`BM`/`BL` precedent) remains the upgrade for user-declarable blocking mods.

Full suite: **5343 passed, 0 failed**.

_This is exactly why the second review pass is worth running — same as #1102, the round-1 hardening had bugs a first review of the original PR could not have seen._

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
Alexander-Sol: I agree with Nic's assessment

trishorts: i'm going to look into this but i it seems to me that a blocked cleavage is not a missed cleavage. It's not cleavable. we are reporting cleavages, not lysines.

trishorts: You and Alexis were right, and I went with the alternative you proposed rather than defending the second property.

Measured first: at `maxMissedCleavages: 0` the emitted peptides were reporting `< 0, 0, 1 >`, so the mismatch you predicted was real and not hypothetical. The cause was that the open-site filter already computed the blocked-discounted count to decide whether a peptidoform survives, but the peptide was then constructed with the modification-blind count — so the generation slack leaked out into the reported value.

What changed:

- `CleavageBlockingReadThroughSlack` is gone (`bcdd9c00`). There is no second property now — just `RespectCleavageBlockingModifications`. The read-through slack is tied to `MaxMods` instead of a fixed 2, since a peptidoform can carry at most `MaxMods` variable modifications and therefore at most that many blocked sites (`ProteinDigestion.cs:251`).
- A blocked residue is no longer counted as a missed cleavage, which is exactly your "treat it as not a cleavage site" framing. The count now reports cleavages that could have happened and did not, rather than Lys/Arg residues.
- `MaxMissedCleavages` keeps its guarantee: no peptide leaves digestion reporting more missed cleavages than were asked for. Digestion still enumerates a wider span internally to reach the read-through forms, but that slack is discounted again before a peptide is emitted, so it is never observable in the result.

Pinned by `ReadThroughOfABlockedSite_ReportsZeroMissedCleavages_NotOne` and `NoEmittedPeptide_EverReportsMoreMissedCleavagesThanRequested`, plus `AnUnblockedMissedCleavage_StillCountsWithTheFlagOn` guarding the opposite direction so the fix cannot degrade into blanket zeroing.

Resolving since the property you objected to no longer exists — reopen if the reported counts still look wrong to you.

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
trishorts: Yes, intended — and it is now stated as a deliberate scope decision in the XML docs rather than being an accident of where the check sits (`a30851ed`).

The reasoning: the drop and the wider generation span are two halves of one exchange. The impossible peptidoform leaves, and the read-through form that replaces it arrives. Applying only the first half would make a peptide *unidentifiable* rather than correctly identified whenever the missed-cleavage budget is too small to reach the read-through. Half a correction is worse than none, so semi and nonspecific searches get none — they behave exactly as they did before, with the flag on or off.

Worth flagging for whoever picks up the follow-up: making semi searches benefit is not just a matter of widening the gate. A semi peptide's C-terminus may be a genuine protease cut or a length-driven truncation, and telling those apart needs the protease's site list — which full digestion gets free from its own enumeration and semi digestion does not.

`SemiSpecificSearch_IsEntirelyUnaffectedByTheFlag` pins that semi is untouched in both flag states.

Resolving as answered, but if you would rather this were not gated at all, that is a design call worth reopening on.

acesnik (commented):
🤖 This is careful work, and the third round shows: default `false`, the new parameter last in the constructor, and `ToString`, `Equals`, `GetHashCode` and `Clone` all updated — `Clone` on both return paths, with a test that loops over `Full` and `None`. I went in expecting to find the cache-key member missed, and it isn't.

I checked the missed-cleavage guarantee rather than taking it on trust. `openMissedCleavages = MissedCleavages - Math.Min(blocked, MissedCleavages)` is a clamped `Max(0, …)`, and survivors must satisfy `<= maxMissedCleavagesAllowed`, so the emitted count is in budget. The one way the slack could still leak is a Full-mode peptidoform that skips the third gate — and `FullDigestion` stamps `CleavageSpecificity.Full` on everything it yields, including the proteolysis-product peptides (`Protease.cs:154, 160, 187, 204, 236`), so nothing does. That makes `CleavageSpecificityForFdrCategory == Full` redundant on that path, but it is harmless and I would leave it.

The position-key arithmetic is exact: residue *i* is key *i*+1, so internal residues 1..len-1 are keys 2..len and the C-terminal residue is key len+1 — which is what the bounds say.

One thing in the classifier I think is worth fixing before this goes in — it misses two shipped K-targeted acylations, and it's a one-line reorder — plus two smaller notes. All inline.

mzLib/Omics/Modifications/CleavageBlockingModifications.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            // Citrullination (deimination) converts arginine's guanidinium to a neutral ureido group,
+            // removing the positive charge trypsin recognises -- the one common Arg-side blocking mod.
+            if (targetResidue == 'R')
+                return id.Contains("citrullin") || id.Contains("deimin");
+
+            // --- Lysine epsilon-amine chemistry below ---
+
+            // Methylation retains the charge -- excluded by design (see class remarks).
+            if (id.Contains("methyl"))
```
acesnik: This guard runs before the acyl stems below, so a K-targeted acylation whose name happens to contain "methyl" is classified as non-blocking.

That is live against the shipped databases rather than theoretical. `ptmlist.txt` has no name containing both "methyl" and one of your stems. `unimod.xml` has five, and two of them target K:

- `Methyl+Acetyl:2H(3)` — site K
- `Methyl:2H(3)+Acetyl:2H(3)` — sites K, T

Both carry an acetyl on the epsilon-amine, which is the chemistry this class exists for. The other three — `Methylmalonylation`, `2-dimethylsuccinyl`, `2-monomethylsuccinyl` — target S/C/T and never reach here.

Testing `ChargeNeutralizingAcylStems` before this guard fixes it without new logic: `Methyl`, `Dimethyl` and `Trimethyl` contain no acyl stem, so they still fall through to false. Checked the position restriction too, rather than leaving it assumed: both are `site="K"` with `position="Anywhere"` in `unimod.xml`, so nothing narrows the misclassification. Both are Unimod class `Isotopic label`, so they are less likely to be selected as variable mods than a biological acetylation would be — the practical exposure is narrower than the count suggests, but the classifier is still wrong for them and the reorder still fixes it.

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
```diff
@@ -237,7 +237,20 @@ public IEnumerable<ProteolyticPeptide> SpeedySemiSpecificDigestion(Protein prote
...
+            // construction. When cleavage-blocking modifications are respected in full-specificity mode we
+            // add slack so the read-through form of a blocked cleavage can be generated (it costs one extra
+            // missed cleavage per blocked site); the surplus is trimmed again by the open-site filter in
+            // ProteolyticPeptide.GetModifiedPeptides. The slack is MaxMods, not a fixed 2: a peptidoform
+            // can carry at most MaxMods variable modifications and therefore at most that many blocked
+            // sites, so this reaches every variable-mod read-through no matter how many co-occur. (A fixed
+            // blocking modification is unbounded and remains a documented limitation.)
+            int generationMaxMissedCleavages = MaximumMissedCleavages;
+            if (DigestionParams.RespectCleavageBlockingModifications && DigestionParams.SearchModeType == CleavageSpecificity.Full)
+                generationMaxMissedCleavages += DigestionParams.MaxMods;
```
acesnik, line 282: This is paid on every protein whenever the flag and Full mode are set, including searches where no configured modification can block anything. At MetaMorpheus defaults (MaxMissedCleavages 2, MaxMods 2) it enumerates at 4, which is roughly 1.7x the unmodified peptides before modification combinatorics, and it carries into indexing.

`VariableModifications` and `AllKnownFixedModifications` are both already on this class from the constructor, so `&& VariableModifications.Any(m => m.BlocksCleavage)` would make the slack free when it cannot fire, with no new plumbing.

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
```diff
@@ -122,7 +158,8 @@ public override string ToString()
         {
             return MaxMissedCleavages + "," + InitiatorMethionineBehavior + "," + MinLength + "," + MaxLength + ","
                    + MaxModificationIsoforms + "," + MaxMods + "," + SpecificProtease.Name + "," + SearchModeType + "," + FragmentationTerminus + ","
-                   + GeneratehUnlabeledProteinsForSilac + "," + KeepNGlycopeptide + "," + KeepOGlycopeptide;
+                   + GeneratehUnlabeledProteinsForSilac + "," + KeepNGlycopeptide + "," + KeepOGlycopeptide + ","
+                   + RespectCleavageBlockingModifications;
```
acesnik, line 162: Right, and worth pinning: this is the one of the four members no test asserts — `Equals` gets covered indirectly by `Clone_PreservesRespectCleavageBlockingModifications`'s `AreEqual`. It matters downstream because MetaMorpheus uses `IndexingEngine.ToString()` as the fragment-index cache key and `SameSettings` compares it verbatim, so a flag that ever went missing here would let a flag-off index be reused silently for a flag-on run. One assertion that two params differing only in this flag stringify differently would hold it.

acesnik:
🤖 Re-read this at the current head, `919db3ec`. That commit is from 06:15 on 2026-08-19 and my review was 22:42 the same day, so nothing has moved since — flagging that rather than assuming it was seen, since the three inline points have no replies and are all still live:

- **`CleavageBlockingModifications.cs:72`** — the `id.Contains("methyl")` guard still returns before `ChargeNeutralizingAcylStems` is tested, so the two K-targeted acylations are still classified non-blocking. I have updated that comment with the position check I had left assumed: both are `site="K"` / `position="Anywhere"`, so nothing narrows it. They are Unimod class `Isotopic label`, which makes them less likely to be picked as variable mods than a biological acetylation — so the practical exposure is smaller than "two shipped mods" suggests, but the classifier is still wrong and the fix is still a one-line reorder.
- **`ProteinDigestion.cs:249-251`** — the slack is still added on `RespectCleavageBlockingModifications && SearchModeType == Full` alone, with nothing checking that a configured modification can actually block. `VariableModifications` is already on the class, so `&& VariableModifications.Any(m => m.BlocksCleavage)` still costs no plumbing and makes the enumeration free when it cannot fire.
- **`DigestionParams.ToString()`** — still the one member of the four with no direct assertion. `RespectCleavageBlockingModifications` appears three times in the test tree, none of them paired with `ToString`. This is the one I would least want to skip, because MetaMorpheus keys its fragment-index cache on `IndexingEngine.ToString()` and compares it verbatim, so a flag missing from that string means a flag-off index silently serves a flag-on run.

Nothing new from this pass — the earlier round's answers to Nic and Alexis both hold up, and I have no further objections beyond the three above. Not approving only because the first one is a live misclassification against shipped data rather than a style point; all three look like small changes.

mzLib/Omics/Modifications/CleavageBlockingModifications.cs
trishorts: Fixed, and confirmed against the shipped `unimod.xml` -- same five names, same two on K.

The guard is gone rather than reordered. Once the acyl stems run first, `Methyl`/`Dimethyl`/`Trimethyl` carry no stem and fall through to false on their own, so keeping it would have left a branch that never decides anything. The intent is now a comment where the branch was.

`AnAcylationIsBlocking_EvenWhenItsNameAlsoCarriesAMethyl` fails on `Methyl+Acetyl:2H(3)` against the old order.

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
trishorts: Gated, with one widening of your suggestion: `AllKnownFixedModifications` is consulted as well as `VariableModifications`. Fixed mods are appended to the variable pattern before the drop is evaluated, so a fixed blocking mod can trigger a drop -- and therefore has to be able to buy the slack that replaces what it dropped. Variable-only would have re-created the semi half-trade inside Full mode.

The cost is invisible in the finished digest, since the open-site filter trims the surplus on the way out -- an output comparison passes either way and pins nothing. So `WithNoBlockingModificationConfigured_NoGenerationSlackIsPaid` reads `ProteinDigestion.Digestion` directly and asserts the enumeration width, with the opposite direction pinned too, so the gate cannot pass by disabling the slack outright.

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
trishorts: Pinned, on all three identity members together rather than just `ToString`, plus the equal-params direction so the assertions cannot pass by everything being unequal. `DigestionParams_DifferingOnlyInTheFlag_AreDistinguishableByEveryIdentityMember`.

trishorts:
## Round 4 also generalises the rule past trypsin, which is a behaviour change beyond what was asked

The two findings above point at the same missing question, so this round answers it rather than patching around it. Flagging it explicitly because it changes behaviour outside the three threads.

**Nothing consulted the protease.** `Modification.BlocksCleavage` hardcoded K/R, and digestion asked only that. So the rule fired under any protease at all: in a **Glu-C** digest an internal acylated lysine discounted a missed cleavage at a site Glu-C never had, both under-reporting the count and letting an over-budget peptide through on the generation slack. Same shape under Arg-C for an acylated K, and under Lys-C for a citrullinated R. Reproduced before changing anything, not read off the source.

The PR body called this out as "harmless for `trypsin`, untested for context-restricted proteases such as `trypsin|P`". That undersold it — the gap was protease *identity*, not sequence context.

**Blocking is a property of the pair.** Digestion now asks `CleavageBlockingModifications.BlocksCleavageBy(mod, protease)`, which pairs the chemistry with a new `DigestionAgent.CleavesCTerminalTo(char)` read off the protease's own `DigestionMotifs`. Across the shipped protease table:

| proteases | why | with the flag on |
|---|---|---|
| trypsin, trypsin\|P, Lys-C\|P, Arg-C, elastase\|P, subtilisin\|P | cut after K and/or R | correction applies |
| Glu-C, Glu-C (with asp), chymotrypsin\|P, CNBr, ProAlanase | cut elsewhere | inert |
| Asp-N (`\|D`), Lys-N (`\|K`) | cut *before* their recognition residue | inert, by direction |

That last row is a deliberate non-fix. Lys-N genuinely has this problem — an acylated K does block it — but mirrored: the blocked residue invalidates the peptide's **N-terminus**, not its C-terminus. Half-applying the C-side drop there would drop real peptides without generating their replacements, which is the same trap the `SearchModeType == Full` gate exists to avoid. Inert and documented beats silently wrong.

**What is left of the approximation.** Sequence context alone: `trypsin|P`'s `K[P]` rule still is not consulted, so a blocked K sitting before a proline can discount a site it never occupied. Bounded (the count is clamped at zero) and the C-terminal drop is still exact.

### On glycosylation

Asked separately, and worth recording since it looks like it belongs in this set and does not.

`BlocksCleavage` is keyed on **charge** — acylation kills the lysine epsilon-amine's positive charge, so trypsin's S1 pocket has nothing to bind. A glycan sits on N (in N-X-S/T) or on S/T, never on a trypsin-family cleavage residue, so there is no charge to remove and the residue gate already rejects it. Where glycans do suppress cleavage the mechanism is **steric**, usually from a neighbouring residue, and the effect is **partial** — the literature describes a raised missed-cleavage rate and a "glycan footprint", not an impossible peptidoform. That calls for the opposite correction to this one: *more* missed-cleavage tolerance, not a dropped peptidoform. Wiring glycans in here would make the drop fire and delete real peptides.

For the glyco-proteases the relationship inverts outright — StcE cleaves C-terminal to an O-glycosylated S/T (glycosylation required at P2), OpeRATOR/OgpA cleaves N-terminal to an O-glycosylated S/T, and flavastacin cleaves C-terminal to N-glycosylated Asn. There the glycan is a *requirement*, not a blocker. So glycosylation stays out of this set, and the reasoning is now in the class remarks rather than only in a thread.

*(Unrelated to this PR, noticed while checking the above: `StcE-trypsin` ships as `TX|T,TX|S,SX|T,SX|S,K|,R|`, which encodes the sequence context but not the glycan requirement, and no C# in mzLib references StcE. Happy to file it separately.)*

Full offline suite green: 5580 passed, 0 failed, 29 skipped.

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
```diff
@@ -237,7 +260,28 @@ public IEnumerable<ProteolyticPeptide> SpeedySemiSpecificDigestion(Protein prote
...
+            // sites, so this reaches every variable-mod read-through no matter how many co-occur. (A fixed
+            // blocking modification is unbounded and remains a documented limitation.)
+            //
+            // The slack is bought with enumeration, so it is only granted when it can actually be spent:
+            // a search whose configured modifications cannot block a cleavage of THIS protease pays
+            // nothing. That gate matters at MetaMorpheus defaults, where MaxMissedCleavages 2 plus
+            // MaxMods 2 enumerates at 4 -- roughly 1.7x the unmodified peptides before modification
+            // combinatorics, and it carries into the fragment index.
+            int generationMaxMissedCleavages = MaximumMissedCleavages;
+            if (DigestionParams.RespectCleavageBlockingModifications
```
nbollis, line 279: [review-oop] Slack gate and drop gate read different sources of truth for protease and missed-cleavage budget

The generation slack in Digestion reads settable instance state: MaximumMissedCleavages (line 278) and Protease via AnyConfiguredModificationCanBlockCleavage (line 281). The drop/discount in ProteolyticPeptide.GetModifiedPeptides reads the immutable digestionParams.MaxMissedCleavages and digestionParams.Protease (ProteolyticPeptide.cs lines 98-99). The two halves of the 'exchange' (slack granted <-> drop applied) therefore depend on two objects holding equal values. Nothing in the codebase mutates these setters today, but the public setters make the feature fragile: a caller that changes ProteinDigestion.Protease or MaximumMissedCleavages after construction silently desynchronizes the slack from the drop (e.g., a mutated Protease that no longer cleaves after K/R still gets the drop applied under the original digestionParams.Protease). The per-protein re-evaluation of AnyConfiguredModificationCanBlockCleavage is a symptom of the same settability.

Suggestion: Read both halves from a single source. Either make Protease, MaximumMissedCleavages, VariableModifications and AllKnownFixedModifications init-only (or private-set) so they cannot diverge from DigestionParams, or have the drop read the same settable properties the slack reads. If settability must remain, hoist AnyConfiguredModificationCanBlockCleavage into a per-instance cached value recomputed on mutation, and state the 'no peptide reports more missed cleavages than requested' invariant in one place that both halves reference.

(Consensus across 2 specialists: review-oop, review-backend)

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
```diff
@@ -237,7 +260,28 @@ public IEnumerable<ProteolyticPeptide> SpeedySemiSpecificDigestion(Protein prote
...
+                && DigestionParams.SearchModeType == CleavageSpecificity.Full
```
nbollis, line 280: [review-backend] Generation slack tied to MaxMods cannot cover fixed blocking modifications

The slack added to the generation budget is DigestionParams.MaxMods, which bounds VARIABLE modifications per peptide. Fixed blocking modifications are appended to the variable pattern by AppendFixedModificationsToVariable regardless of MaxMods, so a fixed blocking mod can trigger the C-terminal drop in ProteolyticPeptide.GetModifiedPeptides while the read-through form that should replace it costs one extra missed cleavage that the slack (MaxMods) may not provide. Concretely, with MaxMods=0 and MaxMissedCleavages=0, a single fixed blocking modification (e.g. fixed N6-succinyllysine on K) causes every peptidoform ending in that residue at an internal cut to be dropped while the read-through is never generated (slack=0), so the peptide is lost entirely rather than mis-reported. This is a sharper edge case than the documented 'fixed blocking mods unbounded slack' limitation (which implies many fixed mods, not one with MaxMods=0), and it is untested: every test in CleavageBlockingModificationTests.cs uses variable blocking mods, so the fixed-mod path of the drop is never exercised.

Suggestion: Either restrict the drop to blocking mods that are variable (so a fixed blocking mod never drops a peptidoform without the slack to replace it), or compute the slack as MaxMods plus the number of configured fixed blocking modifications, or at minimum add a test pinning the MaxMods=0 + fixed blocking mod behavior and document that the peptide is lost in that configuration.

nbollis (commented):
## Automated review of PR #1119 (Opt-in: Modification.BlocksCleavage)

7 of 10 findings approved for inline posting. Core scratching: the digestion-side arithmetic, position keys, slack/drop gating and the flag's identity-member wiring were all verified sound by three parallel specialist reviews (backend, OOP, generic).

### Major
- **ProteinDigestion.cs:281** — Slack gate and drop read different sources of truth for protease and missed-cleavage budget (consensus across review-oop + review-backend). [posted inline]

### Minor
- **ProteinDigestion.cs:282** — Generation slack tied to MaxMods cannot cover fixed blocking modifications (peptide lost at MaxMods=0). [posted inline]
- **Modification.cs:88** — Memoised BlocksCleavage relies on a comment-only immutability contract on a non-sealed DTO. [posted inline]
- **ProteolyticPeptide.cs:162** — Digestion policy embedded as a private method on a data class, split across two classes. [posted inline]

### Suggestions
- **CleavageBlockingModifications.cs:106** — Name-keyed substring heuristic is a leaky abstraction; memoisation makes misclassification sticky. [posted inline]
- **CleavageBlockingModifications.cs:94** — Ubiquitin-remnant spelling 'diGly' is not covered by the classifier. [posted inline]
- **CleavageBlockingModificationTests.cs:568** — Arg-C case in AProteaseThatDoesNotCleaveAfterTheModifiedResidue_IsUnaffected is degenerate (no R in the shared protein). [posted inline]

3 findings were triaged and excluded by the reviewer: circular namespace dependency (CleavageBlockingModifications.cs:131), NoEmittedPeptide docstring overstatement (test:163), 13-positional-parameter constructor (DigestionParams.cs:19).

mzLib/Omics/Modifications/Modification.cs
```diff
@@ -40,6 +40,53 @@ private set
...
+
+                return cached == BlocksCleavageYes;
+            }
+        }
+
+        private const int BlocksCleavageNotYetClassified = 0;
+        private const int BlocksCleavageNo = 1;
+        private const int BlocksCleavageYes = 2;
+
+        private int _blocksCleavageCache;
```
nbollis, line 88: [review-oop] Memoised BlocksCleavage relies on a comment-only immutability contract on a non-sealed DTO

_blocksCleavageCache is memoised from OriginalId, Target and LocationRestriction, which the getter comment asserts are 'set at construction and not changed afterwards'. But Modification is not sealed, those properties have protected set, and subclasses exist (BaseModification, BackboneModification). The immutability is a comment, not a type-enforced invariant: any future subclass, deserialization path, or reflection-based mutation that changes these fields after a first BlocksCleavage read gets a permanently stale answer. Because Modification instances are shared across threads (static lists in Mods.cs), a stale cache is globally wrong, not locally. The benign-race argument for the int cache is sound for the value, but it presupposes the inputs never change; the cache also adds hidden mutable state to a DTO that participates in Equals/GetHashCode and is serialized.

Suggestion: Enforce the contract the memoisation depends on: seal Modification or make the classification inputs (OriginalId, Target, LocationRestriction, IdWithMotif) truly immutable (private set / readonly), or invalidate the cache if any setter is ever added. At minimum, document the invariant on the class declaration (not just the property) so subclasses cannot silently break it.

mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
```diff
@@ -82,5 +117,88 @@ internal IEnumerable<PeptideWithSetModifications> GetModifiedPeptides(List<Modif
...
+        /// The protease is consulted, so a modification only blocks a cleavage this protease would
+        /// have made: an acylated lysine counts under trypsin, Lys-C, elastase or subtilisin, and does
+        /// not under Glu-C, Asp-N, Lys-N or chymotrypsin, none of which cut after a lysine. A protease
+        /// that cuts N-TERMINAL to its recognition residue (Asp-N, Lys-N) is left inert rather than
+        /// half-corrected: there the blocked residue invalidates a peptide's N-terminus, which is the
+        /// mirror image of the drop below and is not modelled here.
+        ///
+        /// Known approximation, now narrowed to sequence context: cleavage sites are matched at the
+        /// residue level without re-deriving the protease's site list for this protein, so a modified
+        /// K or R whose site was prevented by surrounding sequence (trypsin|P's K[P] rule) can still
```
nbollis, line 158: [review-oop] Digestion policy embedded as a private method on a data class, split across two classes

IsUnreachableThroughBlockedCleavage is an instance method on ProteolyticPeptide (a data-ish product class) but is really digestion policy: it takes the protease and the missed-cleavage budget as parameters and mixes them with instance state (MissedCleavages, OneBasedEndResidue, Protein). The policy is split across three places -- the slack decision in ProteinDigestion.Digestion, the drop gate in GetModifiedPeptides' yield loop (lines 95-102), and the computation here -- so the 'exchange' (slack <-> drop) is not a single coherent unit and the guarantee ('no peptide reports more missed cleavages than requested') is enforced by the accidental cooperation of two classes. The out int openMissedCleavages parameter is also an awkward API.

Suggestion: Extract a small digest-policy object (e.g., CleavageBlockingPolicy) that holds the flag, the protease, and the budget, and exposes something like bool IsUnreachable(ProteolyticPeptide peptide, Dictionary<int, Modification> modPattern, out int reportedMissedCleavages). That centralizes the three gates, makes the slack/drop exchange explicit, and lets ProteolyticPeptide stay a data product. Replace the out-parameter with a small result struct or tuple.

mzLib/Omics/Modifications/CleavageBlockingModifications.cs
```diff
@@ -0,0 +1,137 @@
...
+            // The acyl test has the last word, and deliberately runs after nothing that could veto it.
+            // An earlier revision excluded any name containing "methyl" BEFORE reaching here, which
+            // misclassified the two Unimod entries carrying both chemistries on the same lysine --
+            // "Methyl+Acetyl:2H(3)" and "Methyl:2H(3)+Acetyl:2H(3)", both site K, position Anywhere.
+            // The acetyl on the epsilon-amine is what decides: the charge is gone whatever else the
+            // residue also carries.
+            //
+            // The methyl series still classifies as non-blocking, and needs no guard to do it: plain
+            // Methyl, Dimethyl and Trimethyl contain no acyl stem, so they fall through to false here.
+            return ChargeNeutralizingAcylStems.Any(stem => id.Contains(stem));
```
nbollis, line 106: [review-oop] Name-keyed substring heuristic is a leaky abstraction, and memoisation makes misclassification sticky

The classifier matches substrings of the modification id against curated acyl stems (id.Contains("acetyl"), etc.). This is a heuristic, not a chemistry calculation: any mod name containing an acyl stem that sits on K/R with an 'Anywhere.' restriction is classified as blocking. False positives are possible (e.g., a name like 'Acetylglucosamine' on K would match 'acetyl'), and the residue gate is the only defense. Because Modification.BlocksCleavage memoises the result for the instance's lifetime, a misclassification is permanent for that instance, and instances are shared across threads. The curated set is also closed -- users cannot extend it without editing library code.

Suggestion: Document the false-positive risk on the classifier (the residue gate bounds but does not eliminate it), and consider making the classification overridable or configurable (e.g., an extensible registry) rather than a hardcoded private array. If the heuristic must stay, at least ensure the memoised result can be recomputed if the classification rules change.

mzLib/Omics/Modifications/CleavageBlockingModifications.cs
```diff
@@ -0,0 +1,137 @@
...
+            // Citrullination (deimination) converts arginine's guanidinium to a neutral ureido group,
+            // removing the positive charge trypsin recognises -- the one common Arg-side blocking mod.
+            if (targetResidue == 'R')
+                return id.Contains("citrullin") || id.Contains("deimin");
+
+            // --- Lysine epsilon-amine chemistry below ---
+
+            // The ubiquitin / NEDD8 / SUMO remnant left after tryptic digestion is a Gly-Gly stub,
+            // which both neutralises the amine and is sterically prohibitive.
+            if (id == "gg" || id.Contains("gly-gly") || id.Contains("diglycyl") || id.Contains("ubiquit"))
```
nbollis, line 94: [review-generic] Ubiquitin-remnant name 'diGly' is not covered by the classifier

The Gly-Gly stub check covers id == 'gg' (Unimod 'GG'), 'gly-gly' (UniProt 'Glycyl lysine isopeptide (Gly-Gly)'), 'diglycyl' (PSI-MOD) and 'ubiquit', but not 'digly' -- the common MaxQuant/community spelling of the di-glycine ubiquitin remnant. A modification with OriginalId 'diGly' on K would fall through to the acyl-stem test, find no stem, and classify as non-blocking, contradicting the class doc's claim that the ubiquitin/NEDD8/SUMO remnant is included because it both neutralises the amine and is sterically prohibitive. The existing test only pins 'GG', so this gap is untested.

Suggestion: Add 'digly' to the substring checks (e.g., id.Contains("digly")) and extend CleavageBlockingModifications_ClassifyOnlyChargeNeutralizingAcylations with a MakeKModification("diGly") assertion so the name set is pinned.

mzLib/Test/ProteomicsTests/ProteolyticDigestion/CleavageBlockingModificationTests.cs
```diff
@@ -0,0 +1,630 @@
...
+        /// cleaves after a lysine, so an acylated lysine abolishes nothing in a Glu-C digest. Counting
+        /// it as blocked discounted a missed cleavage the peptide genuinely has -- both under-reporting
+        /// the count and letting an over-budget peptide through on the generation slack.
+        /// </summary>
+        [Test]
+        public static void AProteaseThatDoesNotCleaveAfterTheModifiedResidue_IsUnaffected(
+            [Values("Glu-C", "Arg-C")] string protease)
+        {
+            var succinyl = MakeKModification("N6-succinyllysine");
+            var protein = new Protein("AAAAAAAEAAKAAAAAEAAAAAAA", "accession");   // E8, K11, E17
```
nbollis, line 568: [review-generic] Arg-C case in AProteaseThatDoesNotCleaveAfterTheModifiedResidue_IsUnaffected is degenerate

The shared protein 'AAAAAAAEAAKAAAAAEAAAAAAA' (E8, K11, E17) contains no arginine, so for the Arg-C parameter the digest is a single whole-protein peptide with zero missed cleavages and no internal cut. The flag-on vs flag-off comparison passes trivially and would still pass even if the protease half of BlocksCleavageBy were removed entirely: the succinylated K11 is internal, openMissedCleavages stays 0 (clamped), and no C-terminal drop can trigger because the peptide ends at the protein C-terminus. Only the Glu-C case meaningfully exercises the 'protease does not cleave after K, so a blocked K must not discount a missed cleavage' logic. The test therefore gives false confidence for Arg-C.

Suggestion: Give Arg-C its own protein containing R sites with a succinylated K between them (e.g., 'AAAAAAARAAKAAAAARAAAAAAA' with R8, K11, R17), so the Arg-C digest has a real missed cleavage at R8 that must NOT be discounted by the succinylated K11 -- that case would fail if the protease gate in BlocksCleavageBy were dropped.

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
trishorts: Right, and it is gone rather than papered over -- your own #1318 removed the class it lived in.

#1318 merged on the 18th (`smith/master` `5cc7c674`) and deletes `ProteinDigestion.cs`, so the settable `MaximumMissedCleavages` / `Protease` pair the slack was reading no longer exists. Rebased onto it, the slack moves into `Protein.Digest`, which reads `digestionParameters.MaxMissedCleavages` and `digestionParameters.Protease` -- the same immutable object `ProteolyticPeptide` was already reading for the drop. So the divergence you describe is structurally unavailable now.

That alone leaves the guarantee enforced by two classes happening to agree, which was the other half of your point. So both halves now come from one value: `CleavageBlockingPolicy` (`Proteomics/ProteolyticDigestion/CleavageBlockingPolicy.cs`) is built once in `Protein.Digest` from the `DigestionParams` and the variable-modification list, and is **passed into** `GetModifiedPeptides` rather than rebuilt there. It carries `GenerationSlack`, `MaxMissedCleavages` and `Blocks(Modification)`; a default-valued instance is inert, which is what flag-off digestion gets, so the historical path is untouched.

The per-protein re-evaluation you called a symptom went with it: `AnyConfiguredModificationCanBlockCleavage` is now `CleavageBlockingPolicy.AnyCanBlockCleavage`, evaluated once per `Digest` call instead of on every property read.

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
trishorts: Confirmed, and fixed. You were right that it is sharper than the documented limitation: one fixed blocking mod at `MaxMods = 0` loses the peptide outright, which is worse than the behaviour the flag replaces.

I took your first option rather than widening the slack, because the arithmetic does not close. A fixed blocking modification is unavoidable -- it sits on **every** instance of its residue -- so the read-through replacing a dropped peptidoform has to span every such residue in the protein. `MaxMods + (count of configured fixed blocking mods)` is not that bound: one fixed succinyl-K in a protein with nine lysines still needs nine. Any slack derived from a modification *count* is the wrong shape. The mechanism that case actually needs is removing those sites from the protease's site list before enumeration, which is a different change.

So the correction is scoped to variable modifications at both ends:

- `CleavageBlockingPolicy.For` activates on the variable list alone, so a fixed blocking modification buys no slack;
- the drop is asked **before** `AppendFixedModificationsToVariable` merges the fixed mods in, so the pattern it reads is variable-only and a fixed blocking modification cannot spend slack either.

That also resolves the tension with acesnik's round-4 note, which had me widen the gate to include `AllKnownFixedModifications` precisely so a fixed mod could buy the slack it spent. Both are coherent; this is the one that cannot lose a peptide.

`AFixedBlockingModification_NeverLosesAPeptide` pins your configuration exactly -- fixed succinyl-K, `MaxMissedCleavages = 0`, over `MaxMods` 0/1/2 -- and asserts the flag-on digest equals the flag-off digest. Restoring the old semantics (policy built from variable+fixed, drop asked after the merge) fails all three cases. `AVariableBlockingModification_StillGetsTheCorrection` is the counterpart, so the scoping cannot pass by disabling the feature.

mzLib/Omics/Modifications/Modification.cs
trishorts: Enforced by the type now, rather than by the comment.

`OriginalId`, `IdWithMotif`, `Target` and `LocationRestriction` are `private set`, so the constructor is the only writer and neither `BaseModification` nor `BackboneModification` -- nor a future subclass -- can invalidate the memo. Checked before tightening rather than after: nothing in mzLib or MetaMorpheus assigns any of the four outside `Modification`'s own constructor, so it costs no call site.

The invariant is stated on the class declaration as you asked, not only on the property, and `ModificationIdentityIsImmutable` pins all four setters. Widening one back to `protected` fails it -- which nothing else would have noticed, since the staleness it allows only surfaces later as a wrong digest.

On the serialization half of your note, since it was worth checking rather than assuming: the added field is not in any serialized graph. MetaMorpheus's peptide index is field-based NetSerializer, but `PeptideWithSetModifications` keeps `_allModsOneIsNterminus` `[NonSerialized]` and rebuilds it from the full sequence, so no `Modification` is written to that cache and its on-disk layout is unaffected. That is now in the comment so the next reader does not have to re-derive it.

mzLib/Proteomics/ProteolyticDigestion/ProteolyticPeptide.cs
trishorts: Done, close to what you sketched.

`CleavageBlockingPolicy` (`Proteomics/ProteolyticDigestion/CleavageBlockingPolicy.cs`) holds the protease, the caller's budget and the slack, and exposes `GenerationSlack`, `MaxMissedCleavages` and `Blocks(Modification)`. It is built once in `Protein.Digest` and passed into `GetModifiedPeptides`, so the three sites collapse to: one construction that decides whether the exchange can happen at all, one per-peptide condition the policy cannot know (`CleavageSpecificityForFdrCategory == Full` -- is this peptide's C-terminus a protease cut, or a length truncation), and the arithmetic. The exchange itself is stated once, in the type's summary, instead of in two comments that had to agree.

I kept `IsUnreachableThroughBlockedCleavage` on `ProteolyticPeptide` rather than moving it onto the policy: it reads `MissedCleavages`, `OneBasedEndResidue` and `Protein.Length`, so moving it means handing the policy the peptide, which is the same coupling pointing the other way. It takes the policy instead of `(protease, budget)`, which was the part that was actually loose.

The `out int` I left as it is -- it is one value out of a method whose answer is the branch condition, inside the peptidoform loop. That one is a style call rather than a correctness one; say the word and I will make it a result struct.

mzLib/Omics/Modifications/CleavageBlockingModifications.cs
trishorts: Documented, and measured rather than argued.

You are right that this is a heuristic and that the residue gate bounds rather than eliminates the risk. What I could add is how large the residual actually is against what mzLib ships. Running the classifier over both resource databases, restricted to a Lys/Arg target with `Anywhere.`:

- `Omics/Resources/unimod.xml` -- 24 entries classify as blocking
- `Omics/Resources/ptmlist.txt` -- 17 entries classify as blocking

and all 41 are genuine epsilon-amine acylations or citrulline. There is no false positive in shipped data today. Your `Acetylglucosamine` example is the right shape and does not land: the HexNAc-family entries target N/S/T, so the residue gate rejects them before the name is ever read. That measurement is now in the class summary, along with the statement that a user-curated modification file is not covered by it -- which is the honest limit of a name-keyed rule.

On making it extensible: agreed in principle, and it is the upgrade path already noted on this PR (a declared field on `Modification`, `BM`/`BL` precedent). It is more than a registry though, and worth doing properly rather than quickly: a new field on `Modification` is lost on a database round trip unless `ToString()` emits a two-letter key for it *and* `ModificationLoader` reads it back, and `Modification.Equals` compares only `IdWithMotif`/`OriginalId`/`ModificationType`/mass -- so the existing Equals-based round-trip test would stay green while the field was silently dropped. I would rather that be its own change than bolt a mutable static registry onto a type whose instances are shared across threads. Deliberately out of this PR.

On "misclassification is sticky": the memo now derives from four `private set` properties (see the `Modification.cs` thread), so it cannot go stale against its inputs. You are right that it still means a wrong *rule* stays wrong for the process lifetime -- though that is the lifetime of the rule itself, since the stems are `static readonly`.

mzLib/Omics/Modifications/CleavageBlockingModifications.cs
trishorts: Added -- and checking for the same class of miss turned up six more.

`diGly` is in. The deeper problem was that `id == "gg"` is an **exact** test, so every Unimod entry whose id is the stub plus something else was falling through to the acyl test and classifying as non-blocking. All K-targeted, all `position="Anywhere"`, all in the shipped `unimod.xml`:

- `LRGG`, `LRGG+methyl`, `LRGG+dimethyl` -- the longer stub a non-tryptic digest leaves
- `Label:13C(6)+GG`, `Label:13C(6)15N(2)+GG`, `Label:13C(4)15N(2)+GG`, `Label:2H(4)+GG`

So the stubs are now matched as whole **tokens** of the id, split on the separators Unimod composes names with, rather than by `==`. Tokens rather than `Contains`, because a bare `Contains("digly")` is wrong in the other direction: `unimod.xml` also carries `Bisphenol A diglycidyl ether derivative` and `N-acyl diglyceride cysteine`, neither of which is a ubiquitin remnant.

`CleavageBlockingModifications_ClassifyOnlyChargeNeutralizingAcylations` now pins `diGly`, `LRGG`, `LRGG+dimethyl`, `Label:13C(6)+GG` and the UniProt `Glycyl lysine isopeptide (Gly-Gly)` spelling as blocking, and the two `digly`-substring names as not. Reverting the matcher to `id == "gg"` fails it.

mzLib/Test/ProteomicsTests/ProteolyticDigestion/CleavageBlockingModificationTests.cs
trishorts: Correct, and fixed with the protein you proposed.

Arg-C now digests `AAAAAAARAAKAAAAARAAAAAAA` (R8, K11, R17) instead of sharing the Glu-C substrate. The 1-missed-cleavage product `AAAAAAARAAKAAAAAR` has a real missed cleavage at R8 **and** carries the succinylated K11 internally, so a blocked K discounting an Arg-C missed cleavage would report 0 instead of 1 and the flag-on/flag-off comparison fails.

Verified both directions rather than assuming:

- deleting the protease half of `BlocksCleavageBy` now fails **both** parameter cases; before this change only the Glu-C case caught it, exactly as you said;
- putting Arg-C back on the old shared protein fails the new vacuity assertion -- "this substrate must actually produce a peptide with a real missed cleavage and a blocked K inside it" -- which I added so the case cannot silently go degenerate again.

trishorts:
## Round 5: rebased onto #1318, and all seven findings answered

Two of the seven pointed at `ProteinDigestion.cs`, which **#1318 deleted when it merged on the 18th** (`smith/master` `5cc7c674`). So the rebase came first, and it answers the Major finding by itself.

### The rebase

`ProteinDigestion.cs` is gone; its orchestration moves into `Protein.Digest`, which is what #1318 routes digestion through. `DigestionParams` took a trivial constructor union; `DigestionAgent`, `Protease`, the subsite arithmetic and the classifier all carried unchanged. One test constructed `ProteinDigestion` directly and was ported. The relocation sites were pinned by a dry run done before #1318 merged, so this was mechanical rather than exploratory.

It also means the promoting/blocking hooks now sit in the digestion path `Rnase` shares, which is a gift rather than a cost.

### The findings

| # | finding | resolution |
|---|---|---|
| Major | slack gate and drop gate read different sources of truth | one `CleavageBlockingPolicy`, built in `Protein.Digest` and **passed into** `GetModifiedPeptides` |
| Minor | slack tied to `MaxMods` cannot cover fixed blocking mods | correction scoped to **variable** modifications at both ends |
| Minor | memoised `BlocksCleavage` on a comment-only contract | the four classification inputs are `private set`; pinned by test |
| Minor | policy split across a data class and two call sites | same policy type; the peptide-state method takes it instead of `(protease, budget)` |
| Suggestion | name-keyed heuristic is leaky | measured against both shipped databases; documented |
| Suggestion | `diGly` not covered | added -- and six more shipped remnants that were also missing |
| Suggestion | Arg-C test case is degenerate | given its own substrate; mutation-checked |

### Two things worth surfacing here rather than leaving in a thread

**The fixed-modification fix reverses an earlier answer.** Round 4 widened the slack gate to include `AllKnownFixedModifications` specifically so a fixed blocking modification could buy the slack it spends. That was the wrong direction: a fixed modification sits on *every* instance of its residue, so the read-through replacing a dropped peptidoform must span all of them, and no bound derived from a modification count can pay for it. The correction is now variable-only at both ends, and a fixed blocking modification is explicitly outside it with the reason recorded. Modelling that case means removing those sites before enumeration, which is a different mechanism and a different change.

**The classifier gaps were worse than the two reported.** `id == "gg"` is an exact test, so it was missing six K-targeted Unimod remnants besides `diGly` -- `LRGG`, `LRGG+methyl`, `LRGG+dimethyl`, and four `Label:...+GG` isotope pairs. The stubs are matched as whole tokens now. Conversely, running the whole classifier over `unimod.xml` and `ptmlist.txt`: 24 and 17 entries classify as blocking, and all 41 are genuine epsilon-amine acylations or citrulline -- no false positive in shipped data. That measurement is in the class summary, along with the fact that it says nothing about a user-curated modification file.

### Evidence

Every new assertion was mutation-checked against the behaviour it pins, not just observed to pass:

- restoring the fixed-modification semantics fails `AFixedBlockingModification_NeverLosesAPeptide` on all three `MaxMods` values;
- reverting the matcher to `id == "gg"` fails the classifier test;
- widening `OriginalId` back to `protected set` fails `ModificationIdentityIsImmutable`;
- deleting the protease half of `BlocksCleavageBy` now fails **both** cases of `AProteaseThatDoesNotCleaveAfterTheModifiedResidue_IsUnaffected`; it previously failed only Glu-C;
- putting Arg-C back on the shared protein fails the new vacuity assertion.

Full mzLib suite: **6594 passed, 0 failed, 36 skipped**.

### Still open, and still why this is a draft

`RespectCleavageBlockingModifications` does **not** reach MetaMorpheus's fragment-index cache key. `IndexingEngine.ToString()` enumerates `CommonParameters.DigestionParams.<field>` one line at a time and never calls `DigestionParams.ToString()`, so a flag-off index is silently reused for a flag-on run. That is a MetaMorpheus-side one-line fix and cannot ship from here. The `DigestionParams_DifferingOnlyInTheFlag_AreDistinguishableByEveryIdentityMember` test added in round 4 is necessary but not sufficient, and the rationale attached to it overstates what it protects -- correcting that here rather than leaving it.

mzLib/Omics/Modifications/Modification.cs
trishorts: Correcting my answer above: **`private set` was the wrong fix, and you were more right than the finding says.** CI caught it within the hour.

I claimed nothing in mzLib or MetaMorpheus assigns these outside `Modification`'s own constructor. That was wrong. `MetaMorpheus/EngineLayer/GlycoSearch/Glycan.cs` is a real `Modification` subclass and its constructor assigns `OriginalId`, `Target` and `IdWithMotif` **after** `base(...)` has run (`:78`, `:79`, `:85`). So `private set` breaks the MetaMorpheus build outright. My search missed it twice over -- the declaration is `Glycan :  Modification` with two spaces, so a `: Modification\b` search skipped it, and the assignments are unqualified rather than through a receiver, so a `\.OriginalId\s*=` search skipped them too.

Which means the invariant cannot be enforced: there is a subclass that depends on mutating exactly the inputs the memo reads. The finding was not "this contract is only documented" -- it was "this contract is already violated".

So the memo survives mutation instead of assuming it away (`f406651e`). It stores the three inputs beside the answer in one immutable snapshot behind a single reference field, and recomputes when any of them changes by reference. Three reference comparisons per read, cheaper than the lower-casing scan they guard, and correct under precisely the `Glycan` pattern. One reference publish is atomic, so a read cannot see a half-updated pair and the race stays benign -- which was the reason for the int in the first place.

`ModificationIdentityIsImmutable` is gone, replaced by `BlocksCleavage_IsRecomputedWhenASubclassChangesItsInputs`: construct on a serine, read, reassign onto a lysine, read again, reassign back. It drives the `Glycan` shape through a local stand-in subclass, since mzLib has no subclass that does this. It fails against a plain compute-once memo.

The class remarks now say identity is MUTABLE BY A SUBCLASS and name `Glycan`, so nobody re-derives the wrong invariant from the old comment.

The serialization half of my answer stands: `PeptideWithSetModifications` keeps `_allModsOneIsNterminus` `[NonSerialized]` and rebuilds it from the full sequence, so no `Modification` reaches the NetSerializer index and neither field moves its layout.

trishorts:
## Round 6: brought up to master, MetaMorpheus's suite run against it for the first time since #1318, and the body corrected

Head is now **`d65dc13c`**. Two commits since round 5, no behaviour change in either.

- `3d7b19d4` — catch-up merge of master `0b614b4b` (SDRF, mzIdentML, variants, FlashLFQ RNA). Nothing near digestion; clean.
- `d65dc13c` — restores the UTF-8 BOM this branch had stripped from `Modification.cs` and `Protein.cs`, which was putting a spurious line-1 change at the top of both diffs. Cosmetic, but it was cosmetic noise *I* added.

**Full mzLib suite: 6817 passed / 0 failed / 41 skipped.**

### MetaMorpheus's suite has now run against this branch: 2313 passed / 0 failed / 1 skipped

Worth stating plainly, because CI cannot show it. mzLib master's `integration` job has been unable to *compile* MetaMorpheus since #1318 merged, so the last time MetaMorpheus's tests saw this branch was `37d10626` on the 17th — **before** the #1318 rebase relocated the hooks into `Protein.Digest`. Everything since had been verified by mzLib's own suite only.

Run locally: mzLib packed from this head plus **#1330** (without it MetaMorpheus master does not compile against any mzLib carrying #1318's abstract `SpecificDigestionAgent`), with **#2835**'s two `SpecificDigestionAgent` toml ignores applied. MetaMorpheus builds **0 errors**, full suite **2313 / 0 / 1**. So nothing in the rebase disturbs MetaMorpheus, and this PR's constructor-parameter addition breaks no MetaMorpheus source.

### Two preconditions, so nobody merges this ahead of them

Neither affects a flag-off run, which is every run today — the default is `false`. Both bite the first user who turns it on.

1. **MetaMorpheus #2832** — `IndexingEngine.ToString()`, the index cache fingerprint, does not name this flag, so a flag-off index would be silently reused for a flag-on run. Still a draft.
2. **MetaMorpheus #2845** — opened tonight, was not tracked anywhere before. `MetaMorpheusTask.SetAllFileSpecificCommonParams` rebuilds `DigestionParams` field by field, and its own comment warns that a field left off that list "silently falls back to its constructor default for every file with file-specific settings". This flag was left off, so any file with *any* file-specific setting — even one that only changes a precursor tolerance — got the modification-blind digest while the task reported the corrected one.

`FileSpecificDigestionParamsTests` was written to catch exactly that class of bug and did not fire, for a reason worth knowing: a whole-object `Equals` comparison only guards a field the fixture sets **away from its constructor default**, because the rebuild's fallback *is* that default. Fixed there.

### And a correction to this PR's own description

Reading it back before asking for review, four claims in the body contradicted the code. All now fixed in place:

- **Two bullets still described the round-4 fixed-modification decision that round 5 reversed** — the body said `AllKnownFixedModifications` is consulted so a fixed blocking mod can buy its slack. It is not, and it cannot: a fixed modification sits on every instance of its residue, so no bound derived from a modification *count* pays for the read-through, and at `MaxMods = 0` the round-4 shape lost the peptide outright. This is the single most-debated point in the review and the body had the opposite of the code on it.
- **The `ToString` bullet claimed appending the flag there protects the MetaMorpheus index cache.** It does not. `IndexingEngine.ToString()` enumerates `CommonParameters.DigestionParams.<field>` one line at a time and never calls `DigestionParams.ToString()` (`EngineLayer/Indexing/IndexingEngine.cs:65-94`). Appending it is still right for identity-member parity; the cache needs #2832.
- **A test bullet still named `ProteinDigestion.Digestion`**, deleted by #1318.
- The body also left open "a check that the flag survives the toml round-trip through Nett given its private setter." **Answered by measurement**, against Nett 0.15.0: a `{ get; private set; }` property round-trips (Nett sets private setters — `SpecificProtease` has always relied on it), and an unknown toml key is tolerated, so the new key is safe in both directions across versions. A *get-only* property is what throws `Property set method not found`, which is what #1318's `SpecificDigestionAgent` is doing to 26 MetaMorpheus tests today.

### CI

`build`, `external-service-tests` and codecov pass. `integration` fails, and the log shows **only** the two `CS0535` from MetaMorpheus master's test doubles — nothing of this PR's. It cannot go green until **#2835** merges to MetaMorpheus master; #1330 fixes the compile half but not the toml half. Details on [#1330](https://github.com/smith-chem-wisc/mzLib/pull/1330#issuecomment-5768011353).

PR #1120, opened 2026-07-28
Diann reader

Alexander-Sol:
Reads DIA-NN output as IQuantifiableResultFile. This will enable extracting MS1 features from DIA data


nbollis (dismissed):

mzLib/Readers/ExternalResults/ResultFiles/DiaNnReportFile.cs
```diff
@@ -0,0 +1,203 @@
...
+                        break;
+                    }
+
+                    if (!IsSuffixExtensionOf(runName, fileNameWithoutExtension))
+                        continue;
+
+                    // Prefer the longest candidate: given files "sample" and "sample_rep1" for run
+                    // "sample_rep1_centroid", the longer one shares more of the name and is the
+                    // likelier pairing. Equal-length candidates give no reason to prefer either.
+                    if (fileNameWithoutExtension.Length > matchLength)
```
trishorts, line 153: **Longest-wins ranking can silently pick the wrong replicate.**

`IsSuffixExtensionOf` is deliberately symmetric — it accepts a candidate when the *file* name is longer than the run name as well as when it is shorter. But this branch ranks by the absolute length of `fileNameWithoutExtension`, and in the file-longer direction a longer file name shares *less* with the run, not more, so the rationale in the comment above ("the longer one shares more of the name") doesn't hold there.

Concretely, for run `sample` with candidates `sample_rep1.raw` (11) and `sample_rep10.raw` (12): both pass `IsSuffixExtensionOf`, the lengths differ so `ambiguous` is never set, and the method returns `sample` → `sample_rep10.raw`. That's the same `A1`/`A10` plate-naming collision the `IsSuffixExtensionOf` doc comment says the delimiter check guards against — the delimiter check only closes it in the other direction.

FlashLFQ then quantifies that run against the wrong replicate's spectra with no error, which is exactly the failure the comment at the bottom of this method says the design avoids.

Suggestion: rank only in the direction where ranking is meaningful (file name a proper prefix of the run name → longest prefix wins), and treat multiple candidates in the run-name-is-a-prefix direction as ambiguous regardless of length, since none of them shares more of the run name than any other.

mzLib/Readers/ExternalResults/ResultFiles/DiaNnReportFile.cs
```diff
@@ -0,0 +1,203 @@
...
+                foreach (var file in fullFilePath)
+                {
+                    string fileNameWithoutExtension = DiaNnPrecursor.StripPathAndExtension(file);
+
+                    // An empty name is a prefix of everything and would match every run
+                    if (fileNameWithoutExtension.Length == 0)
+                        continue;
+
+                    // An exact match is unambiguous, so take it and stop looking
+                    if (fileNameWithoutExtension.Equals(runName, StringComparison.InvariantCultureIgnoreCase))
```
trishorts, line 140: **The exact-match `break` treats duplicate basenames as unambiguous.**

Candidates are compared after `StripPathAndExtension` removes both the directory and the extension, so two distinct entries in `fullFilePath` can strip to the same name — `C:\batch1\QC.raw` and `C:\batch2\QC.mzML` both strip to `QC`. Both are exact matches for run `QC`, and the `break` takes whichever happens to come first in the caller's list.

The contested-set cleanup at the end of the method doesn't catch this: it detects one *file* claimed by two *runs*, never two files claimed by one run.

Batched acquisitions with per-batch subdirectories hit this routinely, and it violates the stated invariant that ambiguity is always left unmatched.

Suggestion: don't `break` on the first exact match — count exact matches across the whole candidate list, take it if there is exactly one, and fall through to the existing unmatched path when two or more candidates strip to the run name.

mzLib/Readers/ExternalResults/ResultFiles/DiaNnReportFile.cs
```diff
@@ -0,0 +1,203 @@
...
+        private static bool IsSuffixExtensionOf(string runName, string fileName)
+        {
+            // Either side may carry the extra suffix: the caller may hold the converted file instead
+            string longer = runName.Length > fileName.Length ? runName : fileName;
+            string shorter = runName.Length > fileName.Length ? fileName : runName;
+
+            if (!longer.StartsWith(shorter, StringComparison.InvariantCultureIgnoreCase))
+                return false;
+
+            char boundary = longer[shorter.Length];
```
trishorts, line 199: **`longer[shorter.Length]` can throw `IndexOutOfRangeException`, because the comparison is culture-sensitive.**

`StringComparison.InvariantCultureIgnoreCase` is linguistic, ICU-backed matching, which does **not** guarantee the matched prefix occupies exactly `shorter.Length` chars — ICU collapses ignorable code points (soft hyphen U+00AD, ZWJ/ZWNJ). So `StartsWith` can succeed on two *equal-length* strings that aren't equal, and then this line indexes one past the end.

Constructible example: `fileName = "samplex"` (7) and `runName = "sampl\u00ade"` (7, soft hyphen). `longer` = `"samplex"`, `shorter` = `"sampl\u00ade"`; the soft hyphen is ignorable so `shorter` collates as `"sample"`, `StartsWith` returns true, and `longer[7]` throws. The caller's `Equals(..., InvariantCultureIgnoreCase)` guard doesn't catch it, because `"samplex"` and `"sample"` aren't equal.

Low-likelihood input, but it's an unhandled crash out of a public API, and the fix is free.

Suggestion: use `StringComparison.OrdinalIgnoreCase` here and at the exact-match test — file-system identity is ordinal, `LoadResults` already pools these same strings with `StringComparer.Ordinal`, and it's faster inside this O(runs × files) loop. A `if (longer.Length == shorter.Length) return false;` guard makes the intent explicit either way.

While you're in here: `fullFilePath.Distinct()` and the `GroupBy(pair => pair.Value)` contested check both use the default *ordinal* comparer, so `C:\Data\S1.raw` and `c:\data\s1.raw` survive as two candidates, can match two different runs case-insensitively, and then escape the contested-set cleanup. Passing `StringComparer.OrdinalIgnoreCase` to both would make all four sites agree on what equality means.

mzLib/Readers/Util/SupportedFileTypes.cs
```diff
@@ -77,6 +78,7 @@ public static string GetFileExtension(this SupportedFileType type)
                 SupportedFileType.CruxResult => ".txt",
                 SupportedFileType.ExperimentAnnotation => "experiment_annotation.tsv",
                 SupportedFileType.CasanovoMzTab => ".mztab",
+                SupportedFileType.DiaNnReport => "report.tsv",
```
trishorts, line 81: **`DiaNnReport` advertises `"report.tsv"` but `ParseFileType` never matches on it.**

Every other `.tsv` member in this switch (`MsFraggerPsm`, `ExperimentAnnotation`, `Tsv_Dinosaur`, the Toppic members, …) has a corresponding `filePath.EndsWith(type.GetFileExtension(), …)` clause, so `("anything" + type.GetFileExtension()).ParseFileType() == type` holds for them. It doesn't for `DiaNnReport` — the new branch matches on the header only, so a caller resolving a type from the advertised name gets `MzLibException("Tsv file type not supported")` for a file literally named `report.tsv` that isn't on disk yet.

I checked, and this does **not** break CI — there's no enum-wide extension round-trip test, and the fixture `DiaNn_LongFormat_report.tsv` does end with `report.tsv`, so `TestSupportedFileTypeExtensions` passes. It's an API-contract inconsistency rather than a red build.

Header sniffing is clearly the right primary route here (and the `report.pr_matrix.tsv` disambiguation via `File.Name` is a nice touch). Suggestion: either add a `report.tsv` suffix clause after the header sniff so the round-trip holds, or note the asymmetry in a comment on the `GetFileExtension` arm so the next person doesn't assume it.

mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
```diff
@@ -53,13 +53,26 @@ public static List<Identification> MakeIdentifications(this IQuantifiableResultF
...
                     score = psmFromTsv.Score;
                 }
+                else if (record is DiaNnPrecursor diaNnPrecursor)
+                {
+                    // Global.Q.Value, not Q.Value: DIA-NN's Q.Value is scoped to a single run and is
+                    // already filtered below 1%, so it would let every precursor through the
+                    // experiment-wide gates below. Global.Q.Value is the run-spanning figure, which
+                    // is what DonorQValueThreshold is comparing against when it picks MBR donors.
+                    qValue = diaNnPrecursor.GlobalQValue;
+                    pepQValue = diaNnPrecursor.PosteriorErrorProbability;
```
trishorts, line 70: **DIA-NN's raw PEP is being assigned to the PEP *q-value* slot.**

`pepQValue` is handed to `Identification(..., qValue: usePepQValue ? pepQValue : qValue)`. The MetaMorpheus branch two lines up assigns `psmFromTsv.PEP_QValue` — a q-value *derived from* PEP, i.e. a monotone cumulative FDR estimate. DIA-NN's `PEP` column is the local posterior error probability of a single precursor. Different quantities on different scales, landing in the same `Identification.QValue` field.

So at a nominal 1% cutoff, `usePepQValue: true` filters a DIA-NN report on something that isn't an FDR, and the accepted set differs from the MetaMorpheus-equivalent set — silently changing which precursors are quantified and which qualify as MBR donors. (Stricter rather than looser, as it happens, since the q-value at a cutoff is the mean PEP of the accepted set — but still not the same gate.)

I can see from `TestQValuesReachIdentifications` that this is a deliberate choice, but nothing on the public surface says so.

Suggestion: either derive a real PEP-based q-value (cumulative mean of ascending-sorted PEPs over the report), or fall back to `GlobalQValue` here and add a comment on why the `PEP` column isn't used — matching the style of the `Global.Q.Value` rationale you've already written just above, which is genuinely helpful.

mzLib/Test/FileReadingTests/ExternalFileReading/TestDiaNnResultFiles.cs
```diff
@@ -0,0 +1,658 @@
...
+            List<string> fullFilePaths = new()
+            {
+                @"D:\Data\20240315_PBMCs_24min_3Th-1.raw",
+                @"D:\Data\20240316_PBMCs_24min_3Th-1.raw",
+                @"D:\Data\20240317_PBMCs_24min_3Th-1.raw",
+            };
+
+            Dictionary<string, string> matched = file.FileNameToFilePath(fullFilePaths);
+
+            Assert.That(matched.Count, Is.EqualTo(3));
```
trishorts, line 344: **The fuzzy run→file matching tests assert only the dictionary count, so a transposed mapping passes.**

This test asserts `matched.Count == 3` and then verifies exactly one entry. `TestFileNameToFilePathOmitsUnmatchedRuns` asserts `Count == 1` plus one `ContainsKey(...) == false`, and never checks which run matched or to what.

A bug that cross-wires `20240315_..._centroid` → `...20240317....raw` and vice versa produces the same count and leaves the one asserted entry correct — both tests stay green. That's the central mechanism this class exists to implement, and it's also the defect class of the two `FileNameToFilePath` issues I flagged above, so it's worth making these strong.

Suggestion: assert the full expected run→path dictionary in both tests rather than count-plus-one-key, and in the omit test assert the surviving key *and* value explicitly. A case with `20240315` and `20240317` swapped in the input list would also pin order-independence for free.

mzLib/Test/FileReadingTests/ExternalFileReading/TestDiaNnResultFiles.cs
```diff
@@ -0,0 +1,658 @@
...
+            Assert.That(zfp91.GeneName, Is.EqualTo("ZFP91"));
+
+            // The four-accession tropomyosin group carries all four proteins through
+            Identification fourProteins = identifications.First(id => id.BaseSequence == "AADESER");
+            Assert.That(fourProteins.ProteinGroups.Count, Is.EqualTo(4));
+
+            // Decoys are not reported as rows in DIA-NN's main report, so everything is usable
+            // for protein quantification
+            Assert.That(identifications.All(id => !id.IsDecoy));
+            Assert.That(identifications.All(id => id.UseForProteinQuant));
```
trishorts, line 529: **This assertion is tautological, and its justification doesn't match what it pins.**

`UseForProteinQuant` is set from `!record.IsDecoy` in `MakeIdentifications`, and `DiaNnPrecursor.IsDecoy` is hardcoded `false` — so `All(id => id.UseForProteinQuant)` cannot fail regardless of the fixture. The comment above justifies it with "Decoys are not reported as rows in DIA-NN's main report," which is a claim about decoys, not about protein-quant eligibility.

What it actually cements is that the `Proteotypic` column the record already parses is ignored on the FlashLFQ path. The fixture deliberately includes rows where that matters — `AAAAAAAAAAR` (`Q8NFD5;Q9Y651`, `Proteotypic 0`) and `AADESER` (four tropomyosin accessions, `Proteotypic 0`) — and this locks them in as protein-quant-eligible, so any future fix would read as a regression.

Suggestion: split the two claims. Keep `All(id => !id.IsDecoy)` for the decoy statement, and assert protein-quant eligibility per-row against whatever the intended policy is. If ignoring `Proteotypic` is deliberate for now, that's fine — just say *that* in the comment instead.

trishorts (commented):
Nice piece of work overall — the header-sniff detection, the string pooling, the `[Optional]` discipline, and the reasoning in the doc comments are all well above the bar for a new reader in this family, and the test suite is genuinely thorough. All four `SupportedFileTypes` switch arms and both binding tests are wired correctly.

I ran a full review pass over the diff plus a prior-art/conventions sweep against master. Leaving the **major** items inline — most of them cluster in `FileNameToFilePath`, which is the part that can silently quantify against the wrong file. There are also ~28 minor/informational findings I'm not posting to keep the noise down; happy to share the full report if useful.

mzLib/Readers/ExternalResults/ResultFiles/DiaNnReportFile.cs
trishorts: Verified fixed. Ranking on `Math.Min(fileNameWithoutExtension.Length, runName.Length)` is a different formulation than the one I suggested, but it lands in the same place: in the run-is-a-prefix direction every candidate reproduces the whole run name and therefore scores `runName.Length`, so two or more always tie and `fuzzyAmbiguous` trips; in the file-is-a-prefix direction it still ranks by how much of the run name is reproduced, which is where ranking is meaningful. `sample`/`sample_rep1`/`sample_rep10` now returns empty, and `TestFileNameToFilePathDoesNotRankByCandidateLengthWhenRunIsThePrefix` pins it. Passes locally at `8a2829f2`. Resolving.

mzLib/Readers/ExternalResults/ResultFiles/DiaNnReportFile.cs
trishorts: Verified fixed. `exactMatchCount` is accumulated across the whole candidate list and only `== 1` binds, so two paths stripping to the same basename leave the run unmatched instead of taking whichever the caller listed first. `TestFileNameToFilePathLeavesRunUnmatchedWhenTwoFilesStripToItsName` covers the `C:\batch1\QC.raw` / `C:\batch2\QC.mzML` case. Passes locally at `8a2829f2`. Resolving.

mzLib/Readers/ExternalResults/ResultFiles/DiaNnReportFile.cs
trishorts: Verified fixed, and all three parts of it: `OrdinalIgnoreCase` at both the exact-match test and the `StartsWith`, the explicit `longer.Length == shorter.Length` guard before the boundary index, and the case-folding in `Distinct`/`GroupBy` that I'd tacked onto the end of the comment. The comment explaining why ordinal rather than linguistic is the right call is a good addition. Resolving.

mzLib/Readers/Util/SupportedFileTypes.cs
trishorts: Good. Documenting the asymmetry on the `GetFileExtension` arm was one of the two exits I offered, and the comment says exactly the right thing -- that the value is what `WriteResults` appends when naming an output, not something `ParseFileType` keys on. Resolving.

mzLib/Test/FileReadingTests/ExternalFileReading/TestDiaNnResultFiles.cs
trishorts: Verified fixed. Both tests now assert the full run to path dictionary rather than a count plus one key, the `reordered` call pins order-independence, and the two new tests cover the failure modes from the threads above. This is what I was after. Resolving.

mzLib/Test/FileReadingTests/ExternalFileReading/TestDiaNnResultFiles.cs
trishorts: Resolving. The decoy claim is split out, and the comment now says plainly that the FlashLFQ path ignores DIA-NN's `Proteotypic` column and treats every target row as protein-quant eligible -- which is what I asked for if that behaviour is deliberate for now.

One residual, not worth blocking on: `All(id => id.UseForProteinQuant)` still cannot fail, since `UseForProteinQuant` is set from `!IsDecoy` and `DiaNnPrecursor.IsDecoy` is hardcoded `false`. The added `Any(id => id.BaseSequence is "AAAAAAAAAAR" or "AADESER")` pins that the fixture contains those shared-peptide rows, but not what the policy does with them. The comment carries the intent, so I'm fine leaving it.

mzLib/FlashLFQ/ResultsReading/MzLibExtensions.cs
trishorts: Leaving this one open -- it is the only thread from my pass I have not resolved.

The comment you added is accurate and genuinely useful. But the behaviour is unchanged: `pepQValue` is still `diaNnPrecursor.PosteriorErrorProbability`, so `usePepQValue: true` continues to gate on a local posterior error probability rather than an FDR, and at a nominal 1% cutoff still accepts a different set of precursors than the MetaMorpheus-equivalent path -- which in turn changes which precursors are quantified and which qualify as MBR donors.

Both exits I suggested changed that (derive a cumulative PEP-based q-value, or fall back to `GlobalQValue`). Documenting is a reasonable third option and may well be the right call for now. If it is, one part of the original point still stands: a comment in the method body is not the public surface. `MakeIdentifications` carries no XML doc saying any of this, so a caller passing `usePepQValue: true` cannot learn it without reading the implementation.

Happy to resolve if the note moves onto the XML doc for `MakeIdentifications`, or if you would rather carry it as a known limitation into whatever consumes this next -- just say which and I will close it out.

mzLib/Readers/ExternalResults/IndividualResultRecords/DiaNnPrecursor.cs
```diff
@@ -0,0 +1,388 @@
...
+
+        [Name("Global.PG.Q.Value")]
+        public double GlobalProteinGroupQValue { get; set; }
+
+        [Name("GG.Q.Value")]
+        public double GeneGroupQValue { get; set; }
+
+        [Name("Proteotypic")]
+        [TypeConverter(typeof(IntegerBooleanConverter))]
+        public bool Proteotypic { get; set; }
```
trishorts, line 137: **`Translated.Q.Value` is not unused -- deleting it breaks the write round-trip.**

New in `85a3ab22`, so not part of my earlier pass.

Nothing *consumes* the property, but it is a mapped column, so it participates in `WriteResults`. Removing it drops the column from the output and shifts every column after it one place left. In `DiaNn_LongFormat_report.tsv` it sits at index 24, immediately before `Proteotypic` at index 25, so the written file's index 25 now holds `Precursor.Quantity`:

```
TestDiaNnResultFiles: Failed 1, Passed 30

  TestProteotypicRoundTripsAsZeroOrOne
  Expected string length 1 but was 11. Strings differ at index 0.
  Expected: "0"
  But was:  "5359.077148"
```

Control, at head `8a2829f2`: restore this one property and change nothing else -> 31/31 pass. So `85a3ab22` is the sole cause, and `build` should go red on it.

The failing test is the smaller half of this. The effect on the reader is that `WriteResults` no longer round-trips a DIA-NN report faithfully: a `Translated.Q.Value` column present in the input is silently dropped from the output, and every column after it in the written file is misaligned against the original header.

Worth knowing if the intent was pruning dead surface: every other `[Optional] double?` column here is in the same position -- mapped, unconsumed, but written -- so the same deletion is available and equally unsafe for any of them. `TestProteotypicRoundTripsAsZeroOrOne` only caught this one because it samples the `Proteotypic` column and `Translated.Q.Value` happens to sit immediately before it; it fixes `column` once from the original header and checks only that index, so dropping any column *after* index 25 would shift the tail of the file and still pass. If these columns are going to be pruned, the round-trip check probably wants to compare the full header and row split rather than a single column.

trishorts (approved):

RayMSMS (approved):

nbollis (approved):

PR #1121, opened 2026-07-29
Add PrideArchiveClient.GetProjectFilesFromFtpAsync for the complete PRIDE file list

trishorts:
## Why

PRIDE's REST manifest is knowingly incomplete. For **PXD000001**, `GetProjectFilesAsync` returns **8** files while the project's FTP tree holds **13**, omitting the two largest. A caller that must know everything a project actually contains — or its true size — has no way to get it from the REST API today, and is left to open the FTP directory in a browser by hand.

## What

Adds **`PrideArchiveClient.GetProjectFilesFromFtpAsync(accession)`**, which walks the project's FTP directory tree and returns the complete list.

- The tree lives at `https://ftp.pride.ebi.ac.uk/pride/data/archive/{yyyy}/{MM}/{accession}/` (year/month from `PrideProject.PublicationDate`). The PRIDE FTP host also serves that path over **HTTPS** — the same fact `DownloadFileAsync` already relies on — so the whole walk goes over the client's reused `HttpClient`. **Subdirectories are followed.**
- Returns `List<PrideFtpFile>` (`RelativePath`, `FileName`, `Url`, `ApproximateSizeBytes`).
- **Sizes are honest about their precision.** PRIDE's Apache autoindex rounds to ~3 significant figures ("429M"), so the field is named `ApproximateSizeBytes` and its doc points a caller wanting exact transfer bytes at an HTTP HEAD on `Url`.

### Failure contract (matches the rest of the client)
- Blank accession → `ArgumentException`, before any request.
- Unknown accession → `MzLibException` (via `GetProjectAsync`), **not** an `HttpRequestException` a live test would skip as an outage.
- A failed directory listing → `HttpRequestException`.
- Honors `CancellationToken`; `ConfigureAwait(false)` throughout.

## Tests

- **5 offline** (`StubHandler`, no network): the full tree walk, recursion into a subdirectory, a REST-hidden file surfacing, autoindex size parsing, and all three failure contracts.
- **1 live canary** — `[Category("ExternalService")]`/`[Category("Pride")]`, routed through `ExternalServiceTestHelper.RunAsync` so a PRIDE outage skips rather than fails. It asserts the FTP tree is **more complete than the REST manifest** for real PXD000001 (a comparison, not a hard-coded 13, so it survives re-curation) — which also catches EBI changing its autoindex layout, the one drift offline fixtures can't.

Full PRIDE suite (126 tests) green locally, live canary included.

## Note on HTML parsing

The FTP directory is a standard Apache autoindex served over HTTPS; there is no structured (JSON) endpoint for it. Parsing is isolated to two small, commented regex helpers and pinned by both a recorded-fixture offline test and the live canary above.

### DIFF
--- a/mzLib/Test/DatabaseTests/PrideFtpListingTests.cs
+++ b/mzLib/Test/DatabaseTests/PrideFtpListingTests.cs
@@ -179,8 +179,10 @@ public class PrideFtpListingTests
     [Test]
     public void ASelfReferentialSubdirectoryHitsTheDepthCapInsteadOfRecursingForever()
     {
-        // A listing whose only entry links back to itself would recurse until the stack overflows; the
-        // depth cap turns that into a throwable HttpRequestException, and the walk stays bounded.
+        // A listing whose only entry links to a deeper copy of itself grows the URL each level, so the
+        // visited-set cannot prune it — the depth cap is the backstop. It throws MzLibException (a broken
+        // listing is a contract violation, NOT an HttpRequestException a live test would skip as an
+        // outage), and the walk stays bounded to MaxFtpDirectoryDepth (64) levels plus the root = 65.
         int dirFetches = 0;
         var handler = new StubHandler(uri =>
         {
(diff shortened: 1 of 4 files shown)


nbollis (approved):

Alexander-Sol (approved):
Solid PR. One minor remark: It would be nice to know what the difference is between the PRIDE manifest and the complete list. Like, it looks like the unlisted files are all autogenerated file that convert data into alternate formats. If you download all files, you end up getting multiple copies of the data, just presented different ways. It would be cool to include that info in the GetProjectFilesFromFtpAsync summary comment

PR #1122, opened 2026-07-31
feat(flashlfq): export the raw MS1 peaks around each quantified peak

Alexander-Sol:
Adds FlashLfqResults.WritePeakWindows, which writes one row per MS1 peak observed around every quantified chromatographic peak: the scans the peak was traced through, spanning from 0.5 m/z below its lowest observed m/z to 0.5 m/z above its highest. Peaks belonging to other species are included, so the output exposes the co-elution and interference that quantification discards. It is kept out of WriteResults because it re-reads the spectra files, streaming rows to disk so only one file is resident at a time.

Peaks are read straight out of the spectra file rather than from a PeakIndexingEngine. The index is m/z-major and scan-minor, so a contiguous m/z slice of a few consecutive scans costs one binary search per m/z bin, at 100 bins per dalton, where reading the scans costs one binary search per scan. Measured over 500 scans of 2000 peaks and 2000 windows of 10 scans: 33ms vs 16ms for a 1 Da window and 1989ms vs 53ms for a 200 Da one, plus 219ms of index construction avoided per file.

A window holds packed m/z and intensity arrays per scan rather than IIndexedPeak objects, and addresses the data file by one-based scan number translated from the MS1-only index. The two numbering schemes diverge as soon as MS2 scans are interleaved, and GetOneBasedScan is a bare array offset, so conflating them reads the wrong scan.

Also adds IndexingEngine<T>.GetPeaksInRange, the only public way to enumerate every peak in a region of a live index. The writer no longer uses it, but MassIndexingEngine callers have no file-based equivalent because their deconvoluted masses are computed rather than stored.


trishorts (commented):
Reviewed against the live clone at `20262b92`. Solution builds clean and all 13 new tests pass locally, and the tests themselves are unusually good — the binary-search edge cases in `TestPeakWindowHandlesScansWithNothingInRange` and the "the map is not simply index + 1" assertion in `TestMs1ScanNumberMapIsNotTheZeroBasedIndex` are exactly the right things to pin down. The MS1-index-to-scan-number premise the PR is built on is sound: `PeakIndexingEngine.InitializeIndexingEngine` really does index MS1-only in ascending scan-number order, so the alignment holds.

Five things I think need addressing before merge, one of which I'd call blocking. Inline.

1. **Critical** — `PeakWindow.Create` takes `Min`/`Max` of `IndexedPeak.M` over *all* isotopic envelopes, but a `ChromatographicPeak` carries one envelope per (scan, charge state). A peptide seen at two charge states therefore gets a single window spanning between them. Measured here: 226 Da wide, 1356 rows for one peak versus 6 for the same peptide at one charge. On real MS1 data this is unbounded output.
2. **Major** — the `peakId` / QuantifiedPeaks correspondence promised in the comment doesn't hold when two `SpectraFileInfo` share a base filename, and the `File Name` column can't disambiguate them.
3. **Major** — `IndexingEngine<T>.GetPeaksInRange` is new permanent public API on the base class MetaMorpheus consumes, with no production caller (the description says so).
4. **Major** — `BuildMs1ScanNumberMap` rebuilds from the file a map that `ScanInfoArray` already holds, by hand-copying `PeakIndexingEngine`'s LINQ. Two copies of an invariant that silently corrupts every scan number if they drift.
5. **Major** — `LoadAllStaticData()` per file, when `InitiateDynamicConnection` / `GetOneBasedScanFromDynamicConnection` is the API for exactly this access pattern.

Smaller things, not worth inline comments:

- When `Apex == null`, `ChromatographicPeak.ToString()` and `MbrChromatographicPeak.ToString()` both emit `-` for the apex columns, while `ToTsvRows` emits `0` for charge and `NaN` for apex RT — different sentinels for the same condition in the same output family.
- If a spectra file is missing or unreadable, `GetDataFile` throws mid-write and leaves a truncated TSV. That does match the existing contract (nothing in FlashLFQ catches), so it's only worth noting that the graceful no-MS1-scans path sets a different expectation right next to it.
- `WriteResults` takes `silent` as a required trailing parameter; this one defaults it to `false`.
- `WriteSyntheticMzml` leaves four `.mzML` files in the test directory permanently — the `.tsv` outputs are deleted but the synthetic mzMLs are not.
- Two distinct doubles in one scan can narrow to the same `float` and both get flagged as peakfinding peaks (`peakfindingMz.Contains((float)mz[i])`). At m/z 500 the float ULP is ~6e-5, so this is only reachable in profile-mode data. The `PeakWindow.cs:274-275` comment about the narrowing being an exact round-trip is right; it's the collision direction that isn't covered.
- The binning in `GetPeaksInRange` and the `GetClosestIndex` clamp handling in `PeakWindowScan.Create` are both correct — I checked them against `ClassExtensions.GetClosestIndex` and `MassIndexingEngine`'s `BinsPerDalton = 1`, and the comments describing them are accurate.

mzLib/FlashLFQ/PeakWindow.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// <param name="ms1ScanNumberMap"> the map built by BuildMs1ScanNumberMap for that same file </param>
+        /// <param name="mzExpansion"> how far below the lowest and above the highest observed m/z the window extends </param>
+        public static PeakWindow Create(ChromatographicPeak peak, MsDataFile dataFile, int[] ms1ScanNumberMap,
+            double mzExpansion = DefaultMzExpansion)
+        {
+            if (peak == null || dataFile == null || ms1ScanNumberMap == null || !peak.IsotopicEnvelopes.Any())
+                return null;
+
+            double minMz = peak.IsotopicEnvelopes.Min(e => e.IndexedPeak.M) - mzExpansion;
+            double maxMz = peak.IsotopicEnvelopes.Max(e => e.IndexedPeak.M) + mzExpansion;
```
trishorts: **Critical — this is a bounding box across charge states, not a window around the peak.**

`ChromatographicPeak.IsotopicEnvelopes` holds one envelope per (scan, **charge state**) — that is what `NumChargeStatesObserved = IsotopicEnvelopes.Select(p => p.ChargeState).Distinct().Count()` (`ChromatographicPeak.cs:91`) counts. A single `Min`/`Max` over all of them means a peptide observed in more than one charge state gets one window spanning *between* the charge states, rather than a window around each observed m/z.

Measured on this branch: EGFQVADGPLYR (1350.66 Da) observed at z=2 (676.3) and z=3 (451.2), over a synthetic MS1 with one peak every 0.5 m/z:

```
NumChargeStatesObserved = 2
Window    = 450.700 - 676.800 m/z   (width 226.1 Da)
PeakCount = 1356 over 3 scans       (452 peaks/scan)
TSV rows for this ONE peak = 1356

[same peptide, single charge] width 1.0 Da, PeakCount = 6, rows = 6
```

226x the rows, and essentially none of the 226 Da between the two charge states is co-elution "surrounding the quantified precursor" — it is the rest of the spectrum. On a real Orbitrap MS1 (10-50k peaks over 350-1500 m/z) a 226 Da slice is thousands of rows per scan per peak, multiplied over every peak in every file. I suspect the 200 Da window benchmarked in the PR description is exactly this case rather than a pathological one.

Grouping by charge state keeps the span at ~`mzExpansion` regardless of how many charges were traced:

```csharp
foreach (var chargeGroup in peak.IsotopicEnvelopes.GroupBy(e => e.ChargeState))
{
    double minMz = chargeGroup.Min(e => e.IndexedPeak.M) - mzExpansion;
    double maxMz = chargeGroup.Max(e => e.IndexedPeak.M) + mzExpansion;
    ...
}
```

Whichever way it goes, the multi-charge case needs a test — `TestPeakWindowExtraction` quantifies a single identification and `BuildPeak` takes a fixed m/z, so nothing currently exercises more than one charge state.

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -712,6 +713,83 @@ public void WriteResults(string peaksOutputPath, string modPeptideOutputPath, st
...
+                            Console.WriteLine("FlashLFQ Error: The file " + spectraFile.FilenameWithoutExtension
+                                + " contained no MS1 scans; no peak windows were written for it");
+                        }
+                        continue;
+                    }
+
+                    // Ordered to match the QuantifiedPeaks output, so that the nth peak of a file in one file is the
+                    // nth peak of that file in the other
+                    int peakId = 0;
+                    foreach (ChromatographicPeak peak in peaksInFile.OrderByDescending(p => p.Intensity))
```
trishorts: **Major — this correspondence breaks when two `SpectraFileInfo` share a base filename.**

`WriteResults` orders by `FilenameWithoutExtension` then intensity (`FlashLFQResults.cs:599-601`), so every peak sharing a base filename lands in **one** intensity-sorted block. This loop keys on `SpectraFileInfo`, whose identity is `(FullFilePathWithExtension, Condition, BiologicalReplicate, Fraction, TechnicalReplicate)` (`SpectraFileInfo.cs:104-108`), and restarts `peakId` per key.

So for two `SpectraFileInfo` sharing a base name — `runA/sample1.raw` and `runB/sample1.raw`, or one path registered as two fractions, which this repo's own tests construct (`Test/FlashLFQ/TestFlashLFQ.cs:423-426`) — QuantifiedPeaks.tsv emits one merged block while this emits two blocks each restarting at 1. The nth peak of "a file" is then not the same peak in the two outputs.

It is also unjoinable from the output alone, since the `File Name` column is `FilenameWithoutExtension` and those two files are indistinguishable in this TSV.

Grouping `Peaks` by `FilenameWithoutExtension` (and ordering `.OrderBy(filename).ThenByDescending(p => p.Intensity)`) would make the comment true. Otherwise the comment should be narrowed to what is actually guaranteed.

Two things that *are* right and worth keeping: the tie-break is safe (LINQ `OrderBy` is stable and both enumerations start from the same `Peaks[file]` list), and incrementing `peakId` before the `window == null` check is correct for keeping the counters aligned.

mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
```diff
@@ -224,6 +224,80 @@ public virtual List<ExtractedIonChromatogram> GetAllXics(Tolerance peakFindingTo
...
+        /// Returns every indexed peak that falls inside a rectangular region of the m/z by scan-index plane.
+        /// Unlike GetIndexedPeak and GetXic, no attempt is made to pick a single best peak per scan: peaks belonging
+        /// to species other than the one of interest are included, as are multiple peaks from the same scan.
+        /// </summary>
+        /// <param name="minM"> the inclusive lower bound of the m/z (or mass) range </param>
+        /// <param name="maxM"> the inclusive upper bound of the m/z (or mass) range </param>
+        /// <param name="minZeroBasedScanIndex"> the inclusive index of the first scan to search </param>
+        /// <param name="maxZeroBasedScanIndex"> the inclusive index of the last scan to search </param>
+        /// <returns> A list of IIndexedPeak objects, ordered by scan index and then by m/z, both ascending </returns>
+        public List<IIndexedPeak> GetPeaksInRange(double minM, double maxM, int minZeroBasedScanIndex, int maxZeroBasedScanIndex)
```
trishorts: **Major — new permanent public API on a shared base class, with no production caller.**

The PR description says it outright: *"The writer no longer uses it."* Repo-wide the only callers of either overload are the RT overload delegating to the scan-index one (line 298) and `Test/FlashLFQ/TestPeakWindow.cs`. `PeakWindow.Create` deliberately bypasses it (`PeakWindow.cs:18-20`), so this PR ships two independent implementations of "enumerate every peak in an m/z x scan region" and uses the non-indexed one.

`IndexingEngine<T>` is the shared base of `PeakIndexingEngine` and `MassIndexingEngine` and is consumed downstream by MetaMorpheus, so this is permanent surface area. I would either wire `PeakWindow` to it — which also collapses the two implementations into one — or split it into its own PR alongside the `MassIndexingEngine` caller the description says motivates it.

The RT overload additionally re-derives an RT to scan-index bracket that `FlashLfqEngine` already does twice (`FlashLfqEngine.cs:1181-1193` and `1942-1952`).

The binning is correct, for what it's worth: `Floor(minM * BinsPerDalton)` / `Ceiling(maxM * BinsPerDalton)` does bracket `Math.Round(m * BinsPerDalton)` for both `BinsPerDalton = 100` here and `MassIndexingEngine`'s `= 1`.

mzLib/FlashLFQ/PeakWindow.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// Maps the zero based scan indices carried by indexed peaks onto the one based scan numbers of the raw
+        /// file. The two are not interchangeable: indexed peaks are numbered against an MS1-only array, so using a
+        /// zero based index to address the data file lands on the wrong scan as soon as MS2 scans are interleaved.
+        ///
+        /// The MS1 scans are selected and ordered exactly as PeakIndexingEngine.InitializeIndexingEngine selects
+        /// them, so that the indices recorded during quantification line up with this map.
+        /// </summary>
+        public static int[] BuildMs1ScanNumberMap(MsDataFile dataFile)
+        {
+            return dataFile.GetMS1Scans()
```
trishorts: **Major — this map already exists, and hand-copying it duplicates an alignment invariant that silently corrupts the output if it drifts.**

`IndexingEngine.IndexPeaks` builds exactly this map while indexing (`IndexingEngine.cs:34-38`): `ScanInfoArray[i].OneBasedScanNumber` **is** `ms1ScanNumberMap[i]`. `ScanInfoArray[i].RetentionTime` also removes the `GetOneBasedScan(...).RetentionTime` lookups at `PeakWindow.cs:129` and in `PeakWindowScan.Create`.

The LINQ chain here is a literal copy of `PeakIndexingEngine.InitializeIndexingEngine` (`PeakIndexingEngine.cs:50-53`) — the doc comment above says as much. That is two independently maintained copies of a mapping whose entire job is to line up; if either changes, every `Scan Number` in this output is silently wrong and no test would catch it.

Related: the justification in `WritePeakWindows` — *"FlashLfqEngine discards the raw data once quantification finishes"* — holds for the peaks but not for this map. `ClearIndex()` nulls `IndexedPeaks` and leaves `ScanInfoArray` intact (`PeakIndexingEngine.cs:70-74`), and `IndexingEngineDictionary` is `internal` in the same assembly as `FlashLfqResults`, so every file's map is still in memory when this runs. Threading it through removes both the copy and the need to re-read scan metadata from disk.

Minor, while here: `GetMS1Scans()` already filters to `MsnOrder == 1` and already yields in ascending scan-number order (`MsDataFile.cs:88-100`), so both `.Where(...)` and `.OrderBy(...)` are no-ops. The `s != null` guard doesn't protect anything either — `GetMS1Scans` dereferences `scan.MsnOrder` itself, so a sparse `Scans` array throws before this filter ever sees the null.

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -712,6 +713,83 @@ public void WriteResults(string peaksOutputPath, string modPeptideOutputPath, st
...
+
+                foreach (SpectraFileInfo spectraFile in SpectraFiles)
+                {
+                    if (!Peaks.TryGetValue(spectraFile, out var peaksInFile) || !peaksInFile.Any())
+                    {
+                        continue;
+                    }
+
+                    MsDataFile dataFile = MsDataFileReader.GetDataFile(spectraFile.FullFilePathWithExtension);
+                    dataFile.LoadAllStaticData();
```
trishorts: **Major — `LoadAllStaticData()` materializes every scan of every file in order to read a few thousand small windows.**

The doc comment says rows are streamed "so only one file is resident at a time", but the *whole file* is resident: `LoadAllStaticData` materializes every scan and every peak array. FlashLFQ elsewhere goes out of its way not to hold that — `ClearIndex()` is `IndexedPeaks = null; GC.Collect();` (`PeakIndexingEngine.cs:70-74`), called per file and again at the end of the run (`FlashLfqEngine.cs:219-235`, `259`). Here the loaded file is released only when the loop variable is reassigned on the next iteration, with no `= null` / `GC.Collect()` to match that house style.

mzLib already has the API for this exact access pattern — random access to specific scan numbers without materializing the file: `InitiateDynamicConnection()` / `GetOneBasedScanFromDynamicConnection(oneBasedScanNumber)` / `CloseDynamicConnection()` (`MsDataFile.cs:63-67`, implemented by every reader). Combined with taking the scan-number map from `ScanInfoArray` instead of rebuilding it from the file, `WritePeakWindows` would not need to load anything statically at all.

For the record this is about resident memory, not leaked handles: `MsDataFile` is not `IDisposable` and each reader closes its own handle inside `LoadAllStaticData`, so not disposing is the correct convention here.

mzLib/FlashLFQ/PeakWindow.cs
trishorts: Resolved. `PeakWindow` now covers a single charge state and throws on mixed input (`PeakWindow.cs:123-135`), and `QuantifiedMs1Feature.Create` groups the envelopes by `ChargeState` and recombines the windows under the peak (`QuantifiedMs1Feature.cs:123-133`).

`TestChargeStatesGetSeparateWindows` pins the case I measured: a peptide traced at 2+ and 3+ with an unrelated species at 400 sitting between them now yields 9 peaks across two windows rather than the 14 a single 333.5–500.5 rectangle would have swept, and it asserts that neither window contains the other's charge state nor the species between them.

mzLib/FlashLFQ/FlashLFQResults.cs
trishorts: Resolved. `AssignFeatureIds()` numbers peaks within each `FilenameWithoutExtension` group rather than within each `SpectraFileInfo`, which is what makes the ids line up when two files share a base name. `TestFeatureIdsAreUniqueWithinAFileName` covers it with two `sharedName.mzML` in different directories.

I re-checked the correspondence against the actual ordering rather than the comment: `WriteResults` uses `.OrderBy(FilenameWithoutExtension).ThenByDescending(Intensity)` over the same unfiltered `Peaks.SelectMany(...)`, and LINQ's stable `OrderBy`/`GroupBy` make the tie-breaking identical, so the join holds.

Two things worth a line in the doc comment, since a consumer can't infer either:

- the ids are **non-contiguous** — a peak whose `Create` returns null is still numbered, just not written;
- the **file order** still differs. This output iterates `SpectraFiles` in insertion order while the tsv blocks are alphabetical, so `(fileName, featureId)` is the join key and row order is not.

mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
trishorts: Resolved by removal — both overloads are gone from `IndexingEngine`.

Confirmed nothing else had come to depend on them: `git log -S GetPeaksInRange` shows the string only ever existed inside this PR's own commits, so removing it left no gap behind.

mzLib/FlashLFQ/PeakWindow.cs
trishorts: Resolved, and with a better seam than I asked for. `PeakIndexingEngine.GetScansToIndex` is now the single selection that both the index and the map go through, and `FlashLfqEngine` hands `ScanInfoArray` to the results so the common path doesn't re-read the file at all.

I checked the two premises it rests on: `ClearIndex()` nulls only `IndexedPeaks`, so `ScanInfoArray` really does survive the index being cleared; and `IndexingEngineDictionary` can legitimately lack an entry (a file with no MS1 peaks `continue`s before it is added), which the `TryGetValue` + rebuild fallback handles correctly.

One thing to add while you're here: dropping `.Where(i => i != null && i.MsnOrder == 1).OrderBy(...)` is correct against `MsDataFile.GetMS1Scans` as written, but that method is `virtual` and `InitializeIndexingEngine` is public API. Worth a line in the doc comment saying the ascending, MS1-only ordering is now `GetMS1Scans`'s contract — `PeakWindow.MinRetentionTime`/`MaxRetentionTime` read the endpoints of the array as though it were sorted, so a downstream subclass that overrode it unordered would corrupt `rtStart`/`rtEnd` silently.

Two follow-ups, neither of them this PR's job:

- The seam centralises the invariant but retires none of the existing copies. Five files under `Test/MassSpectrometryTests/Deconvolution/` re-derive `GetAllScansList().Where(MsnOrder == 1).OrderBy(OneBasedScanNumber)` to feed an indexing engine and could call `GetScansToIndex` today.
- `Readers/MsDataFileExtensions.cs:24,28` builds `IndexedMassSpectralPeak` with `scan.OneBasedScanNumber - 1` as the zero-based scan index, which is only true for an MS1-only file. That is this same bug, still live, one file over.

mzLib/FlashLFQ/FlashLFQResults.cs
trishorts: Resolved as asked — the writer now seeks through a dynamic connection when the windows cover fewer scans than the file holds, falls back to a static load otherwise, and `TestDynamicAndStaticReadsProduceTheSameOutput` asserts the two paths emit identical bytes.

I'm closing this thread because the ask was implemented, but the implementation leaves loose handles on two of the readers, so I've raised that separately rather than reopening this. Short version: the dynamic-connection contract isn't uniform across readers and the scan-count heuristic can't see the difference — on Thermo every `GetOneBasedScanFromDynamicConnection` call opens a new raw file handle that is never disposed, and on timsTOF the "seek" call loads the entire file anyway. There's also a connection leaked out of `TryInitiateDynamicConnection` when `InitiateDynamicConnection` throws after having already opened one. Both are in the new review.

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -712,6 +726,188 @@ public void WriteResults(string peaksOutputPath, string modPeptideOutputPath, st
...
+                    // load: the file has to be read to be able to say which of its scans are MS1.
+                    int[] ms1ScanNumberMap = GetMs1ScanNumberMap(spectraFile);
+
+                    // A static load reads, and then holds, every scan in the file. Seeking to the scans the windows
+                    // actually cover is worth it while there are fewer of them than that, which is the usual case
+                    // for a file contributing a few peaks and never the case for one contributing thousands.
+                    bool readScanByScan = ms1ScanNumberMap != null
+                        && peaksInFile.Sum(p => (long)QuantifiedMs1Feature.CountScansToRead(p.Peak))
+                            < ms1ScanNumberMap.Length;
+                    bool useDynamicConnection = readScanByScan && TryInitiateDynamicConnection(dataFile);
```
trishorts, line 789: **Major — the scan-by-scan path is only cheap on mzML. On Thermo it leaks a file handle per scan read; on timsTOF it loads the whole file anyway.**

The heuristic counts scans, but the cost of a "seek" is reader-dependent, and two of the readers `MsDataFileReader.GetDataFile` can return make this branch worse than the static load it is avoiding:

- **Thermo** — `GetOneBasedScanFromDynamicConnection` ignores the `dynamicConnection` field that `InitiateDynamicConnection` opened and calls `RawFileReaderAdapter.FileFactory(FilePath)` afresh on *every* call, plus `SelectInstrument` each time (`ThermoRawFileReader.cs:167-168`). `CloseDynamicConnection` disposes only the field, never these. Since the guard bounds total reads at `< ms1ScanNumberMap.Length`, a file with 10k MS1 scans can open on the order of 10k raw-file handles that live until the process exits — and on Windows keep the `.raw` locked.
- **timsTOF** — `GetOneBasedScanFromDynamicConnection` carries the explicit comment `WARNING! This method reads in the entire data file before returning the requested scan!` and calls `LoadAllStaticData(maxThreads: ProcessorCount - 1)` (`TimsTofFileReader.cs:298-312`). So the branch that exists to avoid a static load performs one, and spins up N−1 threads to do it.

Both are pre-existing reader behaviour, but this is their first production caller in a loop. FlashLFQ runs on `.raw` in practice, and `TestDynamicAndStaticReadsProduceTheSameOutput` uses mzML only, so neither is caught.

Narrowest fix is to gate on the reader rather than on scan count alone:

```csharp
// A dynamic read is only cheaper than a static load where the reader actually seeks. Thermo opens
// (and never disposes) a new raw file handle per scan, and timsTOF loads the whole file to serve a
// single scan, so for both the scan-by-scan path costs more than the load it is avoiding.
bool readerSeeksCheaply = dataFile is Mzml;
bool readScanByScan = readerSeeksCheaply
    && ms1ScanNumberMap != null
    && peaksInFile.Sum(p => (long)QuantifiedMs1Feature.CountScansToRead(p.Peak))
        < ms1ScanNumberMap.Length;
```

Teaching `ThermoRawFileReader` to reuse its own connection is the better long-term answer, but that is a load-bearing reader with its own consumers — separate PR, not this one.

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -712,6 +726,188 @@ public void WriteResults(string peaksOutputPath, string modPeptideOutputPath, st
...
+        /// statically instead rather than failing the output.
+        /// </summary>
+        private static bool TryInitiateDynamicConnection(MsDataFile dataFile)
+        {
+            try
+            {
+                dataFile.InitiateDynamicConnection();
+                return true;
+            }
+            catch (Exception)
```
trishorts, line 894: **Major — a partially opened dynamic connection is leaked when `InitiateDynamicConnection` throws.**

`ThermoRawFileReader.InitiateDynamicConnection` assigns `dynamicConnection = RawFileReaderAdapter.FileFactory(FilePath)` (`:136`) and validates only *afterwards*, throwing `MzLibException` for `!IsOpen`, `IsError`, or `InAcquisition` (`:138-151`). This `catch` swallows that and returns false, so `useDynamicConnection` stays false and the `finally` never calls `CloseDynamicConnection()` — the connection opened at `:136` leaks for the life of the process. A `.raw` that is still being acquired is the ordinary way to reach this.

Closing the failed attempt on the way out is enough:

```csharp
catch (Exception)
{
    // InitiateDynamicConnection can throw after it has already opened the connection - Thermo assigns
    // it and only then rejects a file that is still being acquired - so a failed attempt has to be
    // closed rather than abandoned.
    try { dataFile.CloseDynamicConnection(); } catch (Exception) { }
    return false;
}
```

Separately, the comment above this method attributes the catch to "the in-memory ones throw `NotImplementedException`", but `MsDataFileReader.GetDataFile` can only return Thermo/Mzml/Mgf/Bruker/TimsTof/Ms1Align/Ms2Align — `GenericMsDataFile` and `FakeMsDataFile` are unreachable through it. What is actually being swallowed is `FileNotFoundException`, `InvalidDataException`, and `MzLibException("RAW file still being acquired!")`, each of which then resurfaces from the `LoadAllStaticData()` fallback with a less useful message.

trishorts (commented):
Re-reviewed at `e71d161`. All five findings from the last round are addressed and I've resolved those threads with a note on each. The charge-state fix in particular is the real thing rather than a narrowing: `TestChargeStatesGetSeparateWindows` pins the exact case I measured. The solution builds clean here and all 16 tests in `TestQuantifiedMs1Feature` pass locally.

Two **Major** items remain, both in the dynamic-connection machinery that came out of the `LoadAllStaticData` thread — the ask was implemented correctly, but the dynamic-connection contract turns out not to be uniform across readers. Inline.

Two smaller things I'm not filing as findings, but which want a decision rather than a fix:

- **The PR description is stale.** It still describes a TSV of "one row per MS1 peak" and advertises `IndexingEngine<T>.GetPeaksInRange` — which this PR now *deletes* — and doesn't mention `QuantifiedMs1Feature`, the per-charge-state split, JSON Lines, or `FlashLfqResults.Ms1ScanInfo`. Anyone reviewing from the description is reviewing a different PR.
- **JSON Lines is unprecedented here.** This is the first use of `System.Text.Json` as *result serialization* in mzLib. The repo's result idiom is CsvHelper TSV via `IResultFile.WriteResults`, hand-rolled `StreamWriter` TSV, or the `.msl` `BinaryWriter`; JSON exists only as an HTTP wire concern and uses Newtonsoft. No new package, since net8.0 carries STJ — but the repo now has two JSON stacks side by side, and that arrived as a refactor after the last review rather than as a stated goal. Worth an explicit call.

trishorts:
Pushed a branch update (merge of `master`) plus one commit, `feat(peakindexing): record total ion current on ScanInfo`, to get the production half of this PR compiling again. Here is what was wrong and what is still left — I deliberately stopped short of the rest, because it needs your call rather than my guess.

## Fixed: `ScanInfo` (pushed)

`FlashLFQ` has not compiled on this branch since 2026-08-05 — four consecutive red runs:

- `Ms1FeatureData.cs:123-124` constructs `ScanInfo` with **five** arguments
- `SpectraFileHeaderData.cs:196,202` read `scan.TotalIonCurrent`

Neither existed. `MassSpectrometry/PeakIndexing/ScanInfo.cs` had four readonly fields and one four-argument constructor, on `master` **and on this branch**.

Added `TotalIonCurrent` as a fifth readonly field with a five-argument constructor overload. The four-argument constructor is kept and chains to it with `double.NaN` — so all three existing call sites (`IndexingEngine.cs:38`, `MassIndexingEngine.cs:60`, `TestXicVsMassTraceComparison.cs:207`) are untouched, and `SpectraFileHeaderData` already guards on `double.IsNaN`, which makes NaN the right "not recorded" sentinel. I also wired both indexing engines to pass the real value through from `MsDataScan`, otherwise the new header column would have been NaN on every scan.

Verified in isolation: the three-file change applied to clean `master` @ `5ba13155` builds with 0 errors and the indexing + XIC + FlashLFQ subset is **128/128**.

## Not fixed: `Test/FlashLFQ/TestMs1FeatureData.cs` is one API generation behind

`FlashLFQ.csproj` now builds, but `Test.csproj` still does not. The test file was written against an earlier generation of this PR's own types and never updated:

| test expects | production has |
|---|---|
| `QuantifiedMs1Feature` | `Ms1FeatureData` |
| `PeakWindow` | `PeakWindowData` |
| `BuildMs1ScanNumberMap(dataFile)` → `int[]` | `BuildMs1ScanInfo(dataFile)` → `ScanInfo[]` |
| `window.Mz` jagged (`Mz.Sum(a => a.Length)`) | `Mz` is flat `float[]` |
| `window.RetentionTimes` | no such member |

I tried the mechanical rename and backed it out: the `int[]`→`ScanInfo[]` change and the jagged→flat change ripple into the assertions themselves (`map[firstMs1Index]` is compared to a scan number; `window.Mz.Sum(a => a.Length)` is compared to `PeakCount`). Rewriting those means deciding what the assertions were meant to prove under the new shape, which is yours to decide, not mine to infer. Roughly 15 distinct errors, all in that one file.

## Worth settling before this goes further: this PR and #1130 add the same class twice

`PeakWindowData` here and `PeakWindow` in #1130 are the same type, both new, both in namespace `FlashLFQ`:

- both are documented as a rectangular MS1 slice around **one charge state** of a single `ChromatographicPeak`
- both carry `DefaultMzExpansion`, `ChargeState`, `MinMz`, `MaxMz`, `MinRetentionTime`, `MaxRetentionTime`, `ApexRetentionTime`, `PeakCount`
- both enforce the same one-charge-state invariant

And the tell: this PR's test calls `PeakWindow.Create(...)` and `PeakWindow.BuildMs1ScanNumberMap(dataFile)` returning `int[]` — which is **exactly #1130's API**, not this PR's. So this PR looks like the newer redesign of that work, and its test is the older #1130-shaped one left behind.

They also overlap on files: this PR touches `FlashLFQResults.cs` (+218) and `PeakIndexingEngine.cs`; #1130 touches `FlashLFQResults.cs` (+81) and `IndexingEngine.cs` (+74).

Reconciling them looks cheaper than fixing the test file in place — if `PeakWindowData` is the design you want, most of the test drift above is the same reconciliation done once. Happy to do whichever way you call it.

trishorts:
Correcting the last section of my comment above — I framed #1130 as a parallel PR adding the same class, and suggested reconciling the two instead of repairing the test file. That was backwards, and the recommendation flips with it.

#1130 was **stacked on this PR**, not parallel to it. `git merge-base` of the two is `20262b92` (Aug 3); #1130 carried four of this PR's commits plus one of its own. `PeakWindow` and `PeakWindowData` are the same class at two points in its own history, not two designs. And #1130's one distinct fix — one window per charge state — is already here, in `Ms1FeatureData.Create`'s `GroupBy(e => e.ChargeState)` (`Ms1FeatureData.cs:93,150`) and `PeakWindowData.Create`'s mixed-charge guard. I have closed #1130 as superseded; nothing was lost.

So there is nothing to reconcile, and this PR is the trunk. The only thing standing between it and green is the test file — and the cause is narrower than I made it sound. From `53ac091c "Renamed stuff"`:

```
{QuantifiedMs1Feature.cs => Ms1FeatureData.cs}          |  91 +++---
{PeakWindow.cs => PeakWindowData.cs}                    | 251 +++------
{TestQuantifiedMs1Feature.cs => TestMs1FeatureData.cs}  |   0
```

The test file was renamed with **zero content changes** while both classes under it were renamed and reshaped in the same commit. That is the whole reason CI has been red since exactly Aug 5.

The drift table earlier in this thread still stands, and so does the reason I did not just fix it: the rename half is mechanical, but `int[]`→`ScanInfo[]` and jagged→flat `Mz` change what several assertions are actually checking (`map[firstMs1Index]` against a scan number, `Mz.Sum(a => a.Length)` against `PeakCount`). Say the word and I will take a pass at it — otherwise it is yours.

trishorts:
Test file is fixed and pushed — `test(flashlfq): bring the MS1 feature tests up to the renamed API` (+285/−140 in `TestMs1FeatureData.cs`). Local run on this branch: **5557 passed, 0 failed, 29 skipped** (`TestCategory!=ExternalService`), and 16/16 in this fixture.

The drift turned out to be three changes stacked inside `53ac091c`, not one rename:

1. **Renames** — mechanical.
2. **`BuildMs1ScanNumberMap` → `BuildMs1ScanInfo`** — `int[]` of scan numbers became `ScanInfo[]`, so every `map[i]` is now `scanInfo[i].OneBasedScanNumber`.
3. **Jagged → flat** — this was the real work. A window used to hold one array per scan; it now holds m/z, intensity and scan number as three parallel flat arrays with one entry *per peak*. So `ScanNumbers.Length` is a peak count, not a scan count — that is `ScanCount`, with `FirstScanNumber`/`LastScanNumber` replacing `.First()`/`.Last()`. A `GroupByScan` helper regroups the flat arrays so the per-scan assertions still say what they meant, and I pinned the invariants the flat layout creates: a scan's peaks are contiguous and ascending in m/z, and **a scan holding nothing inside the m/z range now contributes no entries at all** rather than an empty array. `TestPeakWindowHandlesScansWithNothingInRange` asserts that directly, since it is a real behaviour change and was the whole point of that fixture.
4. The output also gained a `SpectraFileHeaderData` line per file and moved per-scan retention times onto it, so line counts went 1 → 2 and the window tests resolve retention time through the header — which pins the format's actual contract.

## One thing for you to decide

`TestDynamicAndStaticReadsProduceTheSameOutput` asserted the two read paths emit **identical bytes**, its stated reason being that otherwise "the output would depend on how the results were assembled." The header broke that invariant: `mzAnalyzer` and `totalScanCount` come out **null on the dynamic path and populated on the static one**. Your own XML docs anticipate it — *"a reader seeking to one scan of it never passes"* the mzML head — so the code and that test are asserting opposite things.

I did not paper over it. The test now asserts:

- every **feature** line is byte identical between the two paths, and
- the headers agree on **every** field except those two, which it pins explicitly as a known divergence.

So a regression anywhere else still fails, but the divergence is visible rather than silent. The open question is yours: either the header should not carry reader-dependent fields, or this is accepted and now documented. `TestWriteMs1Features` currently asserts both are null on the dynamic path, so if you change the behaviour that assertion is the one to flip.

Related: wiring `TotalIonCurrent` through both indexing engines in the previous commit turned out to be load-bearing rather than tidy-up. Without it the engine-fed dynamic path would have written null total ion currents against the static path's real numbers, and this same test would have caught it as a third divergence.

PR #1127, opened 2026-08-03
Package architecture-neutral assemblies so mzLib can be used on arm64

acesnik:
The published mzLib package ships x64-stamped assemblies, which cannot be loaded by an arm64 .NET runtime. This is the blocker behind MetaMorpheus issues [#2503](https://github.com/smith-chem-wisc/MetaMorpheus/issues/2503) and [#2504](https://github.com/smith-chem-wisc/MetaMorpheus/issues/2504) — MetaMorpheus failing on Apple Silicon, and the absence of an `osx-arm64` conda-forge build.

## Root cause

An assembly built with `PlatformTarget=x64` carries AMD64 in its PE header, and CoreCLR on arm64 refuses it:

```
System.IO.FileLoadException: Could not load file or assembly
'Omics, Version=1.0.583.0, Culture=neutral, PublicKeyToken=null'.
```

Confirmed to be architecture rather than versioning: the shipped `Omics.dll` loads under an x64 runtime and fails under arm64, while `AssemblyName.GetAssemblyName` reports the expected version under both.

Note that **no csproj sets `PlatformTarget`**. The stamping comes from the build configuration:

1. the release workflow builds the solution with `--configuration Release`
2. the solution defines only `x64` platforms, so `Platform=x64`
3. the SDK derives `PlatformTarget` from `Platform`
4. `mzLib.nuspec` packages out of `bin\x64\Release\`

## Why the change is shaped this way

Two smaller fixes were tried first and don't work:

- Setting `PlatformTarget` to empty in the csprojs — the SDK re-derives it from `Platform`, so the output is still `PE32+ x86-64`.
- Passing `/p:Platform="Any CPU"` to the release build alone — fails with `MSB4126: The specified solution configuration "Release|AnyCPU" is invalid`, because the solution doesn't define that platform.

So the packaged build has to run as Any CPU, and the solution has to offer an Any CPU platform for it to target. This PR:

- adds `Debug|Any CPU` and `Release|Any CPU` to `mzLib.sln`. The existing `x64` and `TestAndRelease` configurations are untouched, but **which platform an unqualified build resolves to does change**: MSBuild prefers `Any CPU` as a solution's default whenever it exists, so a bare `dotnet build` now lands in `bin/Debug` rather than `bin/x64/Debug`. That is benign — it is why the `integration` job's `nuget pack` finds the files where the nuspec now points — but to avoid depending on an implicit preference, `dotnet.yml`'s `build` and `integration` jobs now request `/p:Platform="Any CPU"` explicitly. `Release|Any CPU` mirrors `Release|x64` exactly, including *not* building `Development`, whose missing `Release|x64.Build.0` is what keeps it out of Release builds. `TestAndRelease` has no `Any CPU` counterpart, so selecting that platform drops the configuration; harmless, and noted here because `.sln` files cannot carry comments
- builds and tests the release with `/p:Platform="Any CPU"`
- repoints all 63 `nuspec` file entries from `bin\x64\Release` to `bin\Release`

Any CPU assemblies still run 64-bit on a 64-bit runtime, so x64 consumers see no behavioural change.

## Verification

- Any CPU solution build succeeds, and **all 64 nuspec source paths resolve** against the resulting output. Note a wrong path is *not* a silent failure, contrary to an earlier version of this description: `dotnet pack`/`nuget pack` hard-fails with `error NU5019: File not found` and produces no package, which means the `integration` job's `nuget pack` already guards all 63 paths on every PR
- packing the nuspec produces 68 entries with `lib/net8.0` (31) and `lib/net8.0-windows7.0` (32) both complete; of the DLLs inside the package, 39 are architecture-neutral and the only x64 ones are the ten third-party/native binaries
- every packaged managed assembly is `PE32 ... Mono/.Net assembly` (MSIL) rather than `PE32+ ... x86-64`
- the existing `Release|x64` solution build still succeeds
- **end-to-end:** dropping these assemblies into a MetaMorpheus build with its own `PlatformTarget` removed (smith-chem-wisc/MetaMorpheus#2701) lets `CMD.dll --test` complete the full Calibration → GPTMD → Search pipeline on a native arm64 .NET 8 runtime, exit code 0, with results **identical to the x64 baseline**: 89 target PSMs, 60 peptides, 47 protein groups at 1% FDR, 482 precursors

## Scope: managed code only

The bundled native libraries — `timsdata`, `baf2sql_c`, `isodeclib`, `isogenmass`, `svml_dispmd`, `libmmd`, `vcomp110`, `msvcp110`, `msvcr110` — and the third-party `ThermoFisher.CommonCore.MassPrecisionEstimator` are x64-only, and they are copied into `lib/` rather than `runtimes/<rid>/native/`.

So after this change **mzML and mgf work on arm64, while Bruker, Thermo .raw precision estimation and IsoDec do not.** Supporting those per-architecture would mean moving to `runtimes/<rid>/native/` assets and obtaining arm64 builds of each dependency; that's a separate piece of work and I've kept it out of scope. Worth deciding how much arm64 support to advertise on the strength of this alone.

Paired with smith-chem-wisc/MetaMorpheus#2701 — both are required; neither is sufficient alone.


trishorts (changes requested):
Reviewed this in a clean worktree and re-ran every verification claim in the description rather than taking them on trust. **All of them hold** — nice diagnosis, and the two-lever shape (the sln has to *offer* the platform, the build has to *request* it) is right; the account of why the two smaller fixes fail matches what I see in the tree.

I did find one blocking issue that the diff doesn't reveal on its own — the new `Release|Any CPU` isn't quite the mirror of `Release|x64` it looks like, and it breaks the release workflow's test step. Details in finding 0. Everything else is small.

## What I reproduced

| Claim | Result |
|---|---|
| `Release\|Any CPU` solution build succeeds | ✅ 0 errors (977 warnings, all pre-existing CS8632/CS0169) |
| `Release\|x64` solution build still succeeds | ✅ 0 errors — no regression for existing consumers |
| All 64 nuspec `<file src>` paths resolve after an Any CPU build | ✅ 64 / 64, zero missing |
| Packaged managed assemblies are MSIL, not x86-64 | ✅ all 20 now `PE32 / I386 / managed` (were `PE32+ / AMD64`) |
| `nuget pack mzLib.nuspec` end-to-end | ✅ succeeds, 68 entries, `lib/net8.0` and `lib/net8.0-windows7.0` both complete |
| Release workflow's `dotnet test … /p:Platform="Any CPU" --no-build` step | ✅ discovers, loads and passes against `Test/bin/Release/Test.dll` |
| All 63 `bin\x64\Release` refs converted, no stragglers | ✅ zero `bin\x64` left in the nuspec |
| Any project missing an `Any CPU` mapping | ✅ none — 19 projects, 38 `ActiveCfg` entries, complete |

`Omics.dll` goes from `PE32+ AMD64` to `PE32 I386`. That's the Apple Silicon `FileLoadException`, fixed.

**One correction to the description**, and it's in your favour:

> checked programmatically, since a wrong path would silently drop a file from the package

It wouldn't. I injected a bogus `<file src>` to check, and `nuget pack` fails hard:

```
Error NU5019: File not found: 'Chemistry\bin\Release\net8.0\DOES_NOT_EXIST.dll'.
```

Explicit (non-wildcard) entries are a hard error in NuGet. So the failure mode is loud, not silent — and it means the `integration` job in `dotnet.yml`, which runs `nuget pack` on every PR, already guards all 63 paths.

---

## Findings

### 0. ⛔ `Release|Any CPU` isn't a faithful mirror of `Release|x64`, and it breaks the release workflow

This is the one blocking item, and it's the reason I'd hold the merge. It's also invisible from the diff alone, which is why I'd have missed it too.

`mzLib.sln:127` gives the `Development` project a `Release|x64.ActiveCfg` with **deliberately no `.Build.0`**:

```
{04615EBF-20FC-437C-9CF7-47DA6260FAE3}.Debug|x64.ActiveCfg   = Debug|x64
{04615EBF-20FC-437C-9CF7-47DA6260FAE3}.Debug|x64.Build.0     = Debug|x64
{04615EBF-20FC-437C-9CF7-47DA6260FAE3}.Release|x64.ActiveCfg = Release|x64      <-- no .Build.0
```

That omission is what keeps `Development` out of Release builds. The new block adds both lines (`mzLib.sln:216-217`):

```
{04615EBF-20FC-437C-9CF7-47DA6260FAE3}.Release|Any CPU.ActiveCfg = Release|AnyCPU
{04615EBF-20FC-437C-9CF7-47DA6260FAE3}.Release|Any CPU.Build.0   = Release|AnyCPU   <-- opts it back in
```

The solution metaproject says so in as many words — `The project "Development" is not selected for building in solution configuration "Release|x64"` — with `BuildProjectInSolution="False"` for `Release|x64` and `"True"` for `Release|AnyCPU`. On disk after each build: `Development/bin/Release/net8.0/` exists, `Development/bin/x64/Release/net8.0/` does not.

**Why it matters:** `release.yml:46` runs `dotnet test` **solution-wide** (cwd `mzLib`, no project argument). So `Development.dll` is now enumerated. I ran exactly that command:

```
$ dotnet test --configuration Release /p:Platform="Any CPU" --no-build

Test run for ...\Test\bin\Release\Test.dll
Test run for ...\Development\bin\Release\net8.0\Development.dll     <-- new
  Failed Benchmark_ParallelFragmentation [696 ms]
  Failed BaselineMetrics_NoNormalization_UPS [41 ms]
  Failed CombinedQuantHelaUPS [1 ms]
  Failed EvaluateFoldChanges_UPSProteins [< 1 ms]
  Failed LoadAndRunSpikeInData_BasicPipeline [< 1 ms]
  Failed StrategyEvaluation_AllCombinations_UPS [1 ms]
Failed!  - Failed: 6, Passed: 0, Skipped: 124, Total: 130
```

And the counterfactual on the same tree, confirming this is new rather than pre-existing:

```
$ dotnet test --configuration Release /p:Platform=x64 --no-build

Test run for ...\Test\bin\x64\Release\Test.dll        <-- Development.dll absent entirely
```

`Development` has two live fixtures that were never meant to leave a dev box — `Fragmentation/BenchmarkFragmentation.cs:43` hardcodes `D:\Kelly_ALS_motor_nueron_dataset\...`, and `QuantificationDevelopment/TmtSpikeInDevelopmentTests.cs` carries the comment *"These tests reference local files (TMT_Spike-In_Info/) and should NOT be pushed to CI"* while its five `[Test]`s are not `[Explicit]`. They fail on any machine but the one that wrote them, so they'll fail on the runner.

`dotnet test` failing means **the release workflow aborts at the Test step, before Pack and Push** — so the first tag after this merges wouldn't publish. And **PR CI cannot catch it**: `dotnet.yml:30/:54` always name `./Test/Test.csproj` explicitly, and `release.yml` only fires on tag push (`release.yml:4-6`). Green checks here are not evidence.

**Fix — either one:**
- drop `mzLib.sln:217` (`Release|Any CPU.Build.0`) so `Release|Any CPU` mirrors `Release|x64` exactly; or
- scope `release.yml`'s test step to `./Test/Test.csproj`, matching what `dotnet.yml` already does.

I'd take the first — it keeps "Any CPU is Release with a different platform" true, which is what the rest of the PR assumes. Worth a glance at `Development`'s live fixtures separately either way; they're one `dotnet test` away from biting someone regardless of this PR.

### 1. The PR changes the default platform for every unqualified build, including CI — and doesn't say so

> leaving the existing `x64` and `TestAndRelease` configurations untouched so local and developer builds are unaffected

The configurations are untouched, but *which one you get by default* is not. MSBuild's solution metaproject picks its default platform as `Mixed Platforms` → `Any CPU` → first listed. Before this PR the only option was `x64`; now `Any CPU` wins. Measured, not inferred:

```
BEFORE (master sln):  Platform = x64
AFTER  (this sln):    Platform = Any CPU
```

and the output really does move:

```
after a bare `dotnet build`:
  Chemistry\bin\Debug\net8.0\Chemistry.dll      -> True
  Chemistry\bin\x64\Debug\net8.0\Chemistry.dll  -> False
```

That catches three CI invocations that pass no `/p:Platform` — `dotnet.yml` build (L20-22), external-service (L31), and `integration` (L57) — plus every local build and VS's default dropdown.

**The consequences are actually fine, and I checked each one.** The `integration` job's build-then-`nuget pack` lands on `bin\Release`, exactly where the nuspec now points; I reproduced that combination locally and it packs cleanly. Ironically the nuspec paths end up *better* covered than before, because that job now exercises them on every PR.

But it's load-bearing behaviour arrived at by accident. It means the `x64` configuration developers use in VS is now built by no CI job at all, and the release pipeline's correctness depends on the ordering of a list in a `.sln` with nothing recording why.

Suggestion: make it deliberate rather than emergent — pass `/p:Platform="Any CPU"` explicitly in `dotnet.yml`'s build and integration jobs the same way `release.yml` now does, and add a line to the description acknowledging the flip. Small change, converts an invisible coupling into a stated one.

### 2. All 19 csprojs still declare `<Platforms>x64</Platforms>`

The solution now offers an `Any CPU` platform that no project advertises support for — BayesianEstimation:5, Chemistry:5, Chromatography:7, Development:5, FlashLFQ:5, MassSpectrometry:5, MzIdentML:5, MzLibUtil:5, mzPlot:5, Omics:5, pepXML:5, PredictionClients:7, Proteomics:5, Quantification:5, Readers:5, SpectralAveraging:4, Test:6, Transcriptomics:5, UsefulProteomicsDatabases:5.

It builds, because `Platform` defaults to `AnyCPU` regardless of what `Platforms` lists — which is why this didn't surface in testing. But `Platforms` isn't decorative: VS's Configuration Manager reads it, and so does `SetPlatform` negotiation across `ProjectReference` (`EnableDynamicPlatformResolution`). Leaving the sln and the csprojs disagreeing is the kind of thing that bites later. `<Platforms>AnyCPU;x64</Platforms>` — ideally not 19 more times, see follow-up A.

### 3. `TestAndRelease` has no `Any CPU` counterpart — minor

`Debug|Any CPU` and `Release|Any CPU` were added; `TestAndRelease|Any CPU` wasn't, so selecting the `Any CPU` platform drops that configuration entirely. Probably intentional and harmless — worth a comment either way.

### 4. Loadable on arm64, but not functional, and with no clean failure

Agreed with the scoping, and I think keeping the natives out of this PR is right. Just confirming what actually ships, from the PE headers of the packed nupkg:

```
isodeclib.dll     PE32+  AMD64  native      msvcp110.dll   PE32+  AMD64  native
isogenmass.dll    PE32+  AMD64  native      msvcr110.dll   PE32+  AMD64  native
libmmd.dll        PE32+  AMD64  native      vcomp110.dll   PE32+  AMD64  native
svml_dispmd.dll   PE32+  AMD64  native      timsdata.dll   PE32+  AMD64  native
baf2sql_c.dll     PE32+  AMD64  native
ThermoFisher.CommonCore.MassPrecisionEstimator.dll   PE32+  AMD64  MANAGED   <-- note
```

The last one is worth calling out separately from the rest. It's a **mixed-mode C++/CLI assembly** consumed as a managed `<Reference>` (Readers.csproj L42-44), not as a P/Invoke target — so on arm64 it fails at *assembly load*, the moment any code path touches one of its types. That's a `BadImageFormatException` out of the loader, not a tidy "unsupported on this platform". Same class of problem for the `[DllImport("isodeclib.dll")]` at `MassSpectrometry/Deconvolution/Algorithms/IsoDecAlgorithm.cs:67` and the Bruker/timsTOF imports.

I searched for any architecture guard — `RuntimeInformation.ProcessArchitecture`, `OSArchitecture`, `Is64BitProcess` — and there is none anywhere in the repo. So an arm64 user who reaches `ThermoRawFileReader` or IsoDec gets a loader crash rather than a diagnosable message.

That the full Calibration → GPTMD → Search run came back clean on arm64 shows the mzML path doesn't transitively touch any of these, which is genuinely good news. It's the reason I'd still want one guard before advertising anything (follow-up C).

### 5. Stale comment — nit

`Test/DatabaseTests/TestDatabaseLoaders.cs:429`

```csharp
public void FilesLoading() //delete mzLib\Test\bin\x64\Debug to update your local unimod list
```

Now `mzLib\Test\bin\Debug` for a default build. Ignore if you'd rather keep the diff tight.

---

## Is this the right home for the change?

I checked for existing machinery this should have reused instead, and there is none. Confirmed absent from the repo: no `Directory.Build.props` / `.targets` / `Directory.Packages.props` anywhere; no `RuntimeIdentifier(s)` and no `runtimes\<rid>\native\` layout; no nuspec generation or templating and no `GeneratePackageOnBuild`; no build or pack helper scripts of any kind; no `PlatformTarget` element anywhere; and no test or CI check asserting anything about the packaged output.

So hand-editing the sln and the nuspec is the only lever available, and this PR pulls it correctly.

The real observation isn't that the PR reused the wrong thing — it's that **there's nothing to reuse, and that absence is now the durability risk.** After this merges, platform knowledge lives in three unlinked places: the sln platform list, the `/p:Platform` flags in `release.yml`, and 19 stale `<Platforms>` elements. Nothing ties them together and nothing tests the result. Hence:

## Follow-ups (separate PRs, not this one)

**A. Add `mzLib/Directory.Build.props`.** MSBuild auto-imports it into all 19 projects, giving `<Platforms>AnyCPU;x64</Platforms>` a single home and letting the duplicated lines go in one commit. Natural home for future packaging/versioning properties too. Doesn't replace the sln mapping block — solution configuration mappings are sln-only.

**B. A CI check on the packaged output.** Highest-value of these. Nothing currently asserts the nupkg's file list or PE machine type, so a regression back to `bin\x64` — or a new project added without `Any CPU` mappings — would surface only as a broken *published* package. A dozen-line step that packs and asserts every managed assembly is MSIL would lock in exactly what this PR just achieved, and is what stops it silently rotting.

**C. An architecture guard on the x64-only readers.** `RuntimeInformation.ProcessArchitecture != X64` → `PlatformNotSupportedException` naming the feature, at the entry points to IsoDec, Bruker/timsTOF and Thermo raw. Turns three loader crashes into three legible messages.

**D. On "how much arm64 support to advertise"** — you're right to ask. My read: "mzML and MGF work on arm64; Bruker, timsTOF, Thermo .raw and IsoDec remain x64-only" is honest and shippable *once C exists*. Without C the failure is a crash, and I'd advertise nothing yet. Moving the natives to `runtimes/<rid>/native/` is the real fix and is correctly out of scope here.

---

Happy for this to go in once finding 0 is addressed — it's a one-line fix and the rest of the change is sound. Same merge order as smith-chem-wisc/MetaMorpheus#2701 — **both required, neither sufficient**: this one merges and cuts a release first, then MetaMorpheus consumes it, otherwise MM builds architecture-neutral against an x64-stamped mzLib and Apple Silicon still fails one layer down. Which is also why finding 0 matters more than it might look: the release has to actually publish for #2701 to have anything to consume.

Requesting changes only for finding 0 — drop `mzLib.sln:217` (or scope the release test step) and I'm happy to approve. Everything else on the list is a suggestion, not a gate.

acesnik:
🤖 Thank you for reviewing in a clean worktree and re-running the claims rather than taking them on trust — finding 0 is a real bug in this PR and I would not have caught it from the diff either.

**Finding 0 — fixed in 7697d2a9.** You diagnosed it exactly: `Development`'s missing `Release|x64.Build.0` is load-bearing, and my generated block added `ActiveCfg` and `Build.0` uniformly for all 19 projects, which opted it back in. Rather than deleting `mzLib.sln:217` directly I made the generation faithful — mirror each project's existing x64 `Build.0` presence — which removes exactly that one line and keeps "Any CPU is Release with a different platform" true, as you preferred. Verified both directions on the solution: `Release|Any CPU` no longer produces `Development/bin/Release`, matching `Release|x64`, while `Debug|Any CPU` still builds it, matching `Debug|x64`.

**Your correction is right, and it also killed a worse alternative.** I had independently hit finding 1 (the default-platform flip) and, before reading your review, rewritten this PR to avoid the `.sln` entirely — building `Test.csproj` and `pepXML.csproj` at project level, since project-level builds default to Any CPU. Your review is what showed me that would be worse: the `integration` job builds the **solution** as Release and then packs, so with the nuspec pointing at `bin\Release` and no Any CPU solution platform, every PR would fail `NU5019`. I've discarded that approach and kept yours. It's a good illustration of your closing point — the coupling between the sln platform list, the `/p:Platform` flags and the nuspec paths is real and currently recorded nowhere.

**Finding 1 — made deliberate.** `dotnet.yml`'s `build` and `integration` jobs now pass `/p:Platform="Any CPU"` explicitly, so the packaging path rests on a stated intent rather than on MSBuild's implicit preference and the ordering of a list. Description updated to acknowledge the flip; my "leaving the existing configurations untouched so local and developer builds are unaffected" was wrong as written, since which configuration you get by default *did* change.

**Correcting my own description, and I've now checked it myself.** You're right that a wrong nuspec path is not a silent failure. I couldn't originally verify this — there's no `nuget` CLI available to me and `NuGet.CommandLine` won't install as a dotnet tool — but `dotnet pack` accepts `NuspecFile`, which routes through the same `NuGet.Build.Tasks.Pack` target. Injecting a bogus `<file src>` gives:

```
NuGet.Build.Tasks.Pack.targets(221,5): error NU5019: File not found: 'Chemistry\bin\Release\net8.0\DOES_NOT_EXIST.dll'
```

and no package is produced. So the failure is loud, the `integration` job's `nuget pack` already guards all 63 paths on every PR, and the sentence claiming otherwise is gone. That's better coverage than I credited.

I also packed the nuspec **after** the finding-0 fix, since your pack verification predates it: 68 entries, `lib/net8.0` (31) and `lib/net8.0-windows7.0` (32) both complete, and of the DLLs inside the package 39 are architecture-neutral with 0 unexpected x64 — the only x64 ones are the ten third-party/native binaries. All 64 `<file src>` paths still resolve.

**Finding 5** — stale `bin\x64\Debug` comment updated.

**Findings 2 and 3** left as you scoped them. Agreed `<Platforms>AnyCPU;x64</Platforms>` belongs in a `Directory.Build.props` (follow-up A) rather than 19 more times, and that `TestAndRelease` having no Any CPU counterpart is harmless — `.sln` files don't take comments, so it's noted in the description instead.

**Follow-up B** — I had this written and verified (reads the COFF machine field of every `.dll` the nuspec ships, allow-listing the ten genuinely x64-only third-party binaries; exits 0 as built, exits 1 naming each offender when an mzLib project is rebuilt with `PlatformTarget=x64`). Keeping it out of this PR per your scoping and opening it separately once this merges, since it asserts on the output this PR produces.

**Finding 4 / D** — agreed, and the `MassPrecisionEstimator` distinction is worth calling out separately: it's a mixed-mode C++/CLI assembly consumed as a managed `<Reference>`, so it fails at *assembly load* rather than at P/Invoke. Tracked with the rest in smith-chem-wisc/MetaMorpheus#2702, and I'll advertise nothing about arm64 until a guard exists.

Agreed on merge order — this cuts a release first, then MetaMorpheus consumes it.

trishorts (approved):
Approving. Finding 0 is fixed in 7697d2a9, and I re-verified it the same way I found it — in a clean worktree, from a build rather than from the diff.

You took the option I'd have picked, and improved on it: rather than deleting `mzLib.sln:217` by hand you made the generation faithful, mirroring each project's existing x64 `Build.0` presence. That fixes the cause instead of the symptom, and it means a future project added to the solution inherits the correct behaviour instead of re-introducing this. Net effect on the sln is exactly the one line.

## What I re-ran

I deleted `Development/bin` before building so a stale assembly couldn't produce a false pass.

| Check | Result |
|---|---|
| `Release\|Any CPU` mirrors `Release\|x64` for all 19 projects | ✅ only `TestAndRelease` differs — finding 3, deliberate |
| `Release\|Any CPU` build omits `Development` | ✅ directory deleted first, did not reappear |
| `Debug\|Any CPU` still builds `Development` | ✅ mirrors `Debug\|x64` — the right line, and only that line |
| `release.yml`'s solution-wide test step | ✅ 5385 passed, 0 failed, `Development.dll` absent |
| All 64 nuspec `<file src>` paths resolve | ✅ 64 / 64, zero `bin\x64` stragglers |
| Packaged assemblies still MSIL | ✅ incl. `Omics.dll` → `I386/MSIL` |

The release test step, which is what would have aborted before Pack and Push:

```
$ dotnet test --configuration Release /p:Platform="Any CPU" --no-build

Test run for ...\Test\bin\Release\Test.dll (.NETCoreApp,Version=v8.0)
Passed!  - Failed: 0, Passed: 5385, Skipped: 29, Total: 5414, Duration: 3 m 12 s
```

against what the same command produced before the fix:

```
Test run for ...\Test\bin\Release\Test.dll
Test run for ...\Development\bin\Release\net8.0\Development.dll     <-- was here
Failed!  - Failed: 6, Passed: 0, Skipped: 124, Total: 130
```

Confirming separately that the finding-0 fix didn't disturb what the PR is actually for: the packaged managed assemblies are still architecture-neutral, `Omics.dll` included. That's the Apple Silicon `FileLoadException`, still fixed.

## The rest

Finding 1 is now stated rather than emergent — `dotnet.yml`'s `build` and `integration` jobs request `/p:Platform="Any CPU"` explicitly, and the comments explain *why*, which is the part that will still be legible in a year. Finding 5's stale comment is updated. Findings 2 and 3 and follow-ups A–D are correctly left out; they were suggestions, not gates.

Good catch discarding the project-level-build rewrite. You're right that it would have failed `NU5019` on every PR once the nuspec points at `bin\Release` with no Any CPU solution platform — the `integration` job builds the solution, not the projects. That the two approaches diverge only at that one job is a decent illustration of why follow-up B matters: nothing currently asserts anything about the packaged output, so the sln platform list, the `/p:Platform` flags and the 63 nuspec paths stay coupled with nothing tying them together. Worth opening as soon as this lands.

Merge order unchanged, and it matters here — **both required, neither sufficient**: this merges and cuts a release first, then smith-chem-wisc/MetaMorpheus#2701 consumes it. Otherwise MetaMorpheus builds architecture-neutral against an x64-stamped mzLib and Apple Silicon still fails one layer down.

Nice work on this one — the diagnosis was the hard part, and it was right the first time.

acesnik:
🤖 Thanks for the approval, and particularly for deleting `Development/bin` before rebuilding so a stale assembly couldn't produce a false pass. Agreed on follow-up B — I'll open it as soon as this lands.

One note on the red `integration` check, since it isn't from this PR: **mzLib master is red on its own.** The push run for `f6b0f0d1` ("Remove y ions from the ETD and ECD product sets", #1114) fails the identical four tests — `NandO_GlycoSearch_onlyO_Run2`, `OGlycoTest_FragmentIons2`, `OGlycoTest_FragmentIonsHash`, `TestMixedMs2Ms2` — and every PR run since that push has failed in `integration` while `build` and `external-service-tests` stayed green. MetaMorpheus master still asserts the old `c, y, zDot` counts, so the job fails when it builds MetaMorpheus against the packed 9.9.9.

Worth flagging for step 2 of the merge order rather than for this PR: `release.yml` runs only mzLib's own tests, so this doesn't block cutting the release — but it will surface in MetaMorpheus once smith-chem-wisc/MetaMorpheus#2701 consumes the new package, and looks like it needs a MetaMorpheus-side update to the expected ETD fragment sets.

Alexander-Sol (approved):

PR #1130, opened 2026-08-03
fix(flashlfq): bound each peak window to one charge state

trishorts:
@Alexander-Sol — this targets `PeakEnvelopeOutput`, so it lands inside #1122 rather than alongside it. It's the one blocking item from my review there; the other four findings are design calls I've left to you.

### The problem

`ChromatographicPeak.IsotopicEnvelopes` holds one envelope per scan **per charge state** — that is what `NumChargeStatesObserved` counts. `PeakWindow.Create` took a single `Min`/`Max` of `IndexedPeak.M` across all of them, so a peptide traced at more than one charge got one window spanning the *gap between* the charge states.

Measured on `PeakEnvelopeOutput` before this change, with EGFQVADGPLYR (1350.66 Da) at z=2 (676.3 m/z) and z=3 (451.2 m/z), over a synthetic MS1 with one peak every 0.5 m/z:

```
NumChargeStatesObserved = 2
Window    = 450.700 - 676.800 m/z   (width 226.1 Da)
PeakCount = 1356 over 3 scans       (452 peaks/scan)
TSV rows for this ONE peak = 1356

[same peptide, single charge] width 1.0 Da, PeakCount = 6, rows = 6
```

226x the rows, and essentially none of that 226 Da is co-elution around the precursor — it's the rest of the spectrum. On real Orbitrap MS1 data (10-50k peaks over 350-1500 m/z) the output is effectively unbounded. I suspect the 200 Da window you benchmarked in the PR description is exactly this rather than a pathological case.

### The change

- `PeakWindow.Create` now takes the `chargeState` whose envelopes bound the window, and only those envelopes set `MinMz`/`MaxMz`, the scan span, and the peakfinding flags.
- `PeakWindow.CreateForEachChargeState` yields one window per traced charge state, ascending.
- `PeakWindow.ChargeState` is exposed, and the `Peak Charge` column now carries the **window's** charge state rather than `Apex?.ChargeState ?? 0`.
- `WritePeakWindows` writes every charge state of a peak under that peak's single `peakId`, distinguished by `Peak Charge`. The correspondence with QuantifiedPeaks that the comment there promises is unchanged.
- `ApexRetentionTime` is now the apex of the window's own charge state. The peak's overall apex can belong to a different charge state whose scan lies outside the window entirely, which would have put an out-of-range RT in the `RT Apex` column.
- `Peak RT Start` / `Peak RT Apex` / `Peak RT End` renamed to `Window RT *`, since they describe the window's scan span and that span is now per charge state.

### Tests

Your 13 tests still pass, updated for the new signature. Two added:

- `TestPeakWindowIsBoundedPerChargeState` — the regression above. Builds the two-charge peak over a spectrum with a peak every 0.5 m/z, asserts two windows in ascending charge order, each exactly `2 * mzExpansion` wide and centred on its own m/z, and that the total peak count stays under 50 where the old bounding box swept in ~1356.
- `TestPeakWindowApexIsPerChargeState` — z=2 apexes in the first scan and z=3 in the last; asserts each window reports its own apex rather than the peak's.

`TestWritePeakWindows` additionally asserts every row's `Peak Charge` is a charge the peak was actually traced at, and that every traced charge state contributed rows.

Also added `BuildMultiChargePeak`, so a test can build the peak FlashLFQ actually produces for a multiply-charged peptide; `BuildPeak` now delegates to it and is unchanged for callers.

### Verification

- Solution builds clean, `x64|Debug`, 0 errors.
- All 15 tests in `TestPeakWindow` pass.
- Patch coverage over the whole of #1122 plus this change, Coverlet/Release against `upstream/master`: **98.99%** (196/198 changed executable lines), target 90. No file below the bar.

Take it, adapt it, or close it — whichever is least friction for you.

### DIFF
--- a/mzLib/FlashLFQ/FlashLFQResults.cs
+++ b/mzLib/FlashLFQ/FlashLFQResults.cs
@@ -715,9 +715,10 @@ namespace FlashLFQ
 
         /// <summary>
         /// Writes one row per MS1 peak observed around each quantified chromatographic peak: every peak between the
-        /// chromatographic peak's first and last retention time, from mzExpansion below the lowest m/z observed for
-        /// that peak to mzExpansion above the highest. Peaks belonging to other species are included, so the output
-        /// shows the interference and co-elution surrounding each quantified precursor.
+        /// first and last retention time the peak was traced at a given charge state, from mzExpansion below the
+        /// lowest m/z observed at that charge state to mzExpansion above the highest. Peaks belonging to other
+        /// species are included, so the output shows the interference and co-elution surrounding each quantified
+        /// precursor. A peak traced at several charge states contributes one such window per charge state.
         ///
         /// This is not part of WriteResults because it re-reads each spectra file: FlashLfqEngine discards the raw
         /// data once quantification finishes, and holding the extracted peaks in memory instead would cost far more
@@ -770,15 +771,17 @@ namespace FlashLFQ
                     foreach (ChromatographicPeak peak in peaksInFile.OrderByDescending(p => p.Intensity))
                     {
                         peakId++;
-                        PeakWindow window = PeakWindow.Create(peak, dataFile, ms1ScanNumberMap, mzExpansion);
-                        if (window == null)
-                        {
-                            continue;
-                        }
 
-                        foreach (string row in window.ToTsvRows(peakId))
+                        // One window per charge state the peak was traced at. They share the peak's id and are told
+                        // apart by the Peak Charge column, because a peptide's m/z at one charge is unrelated to its
+                        // m/z at another and a single window bounding both would span the gap between them.
+                        foreach (PeakWindow window in PeakWindow.CreateForEachChargeState(peak, dataFile,
+                            ms1ScanNumberMap, mzExpansion))
                         {
-                            output.WriteLine(row);
+                            foreach (string row in window.ToTsvRows(peakId))
+                            {
+                                output.WriteLine(row);
+                            }
                         }
                     }
 
--- a/mzLib/FlashLFQ/PeakWindow.cs
+++ b/mzLib/FlashLFQ/PeakWindow.cs
@@ -9,11 +9,16 @@ using System.Text;
 namespace FlashLFQ
 {
     /// <summary>
-    /// A rectangular slice of the MS1 data surrounding a single ChromatographicPeak. The slice spans the scans the
-    /// peak was observed in, and runs from mzExpansion below the lowest m/z observed for the peak to mzExpansion
-    /// above the highest. Every MS1 peak in that region is retained, including peaks that belong to species other
-    /// than the quantified precursor, which is the point of the type: it captures the co-eluting and interfering
-    /// signal that quantification discards.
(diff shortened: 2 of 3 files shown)


trishorts:
Flagging a collision before this goes any further: **this PR and #1122 add the same class under two names.**

`PeakWindow` here and `PeakWindowData` in #1122 are both new, both in namespace `FlashLFQ`, both documented as a rectangular MS1 slice around **one charge state** of a single `ChromatographicPeak`, both carry `DefaultMzExpansion` / `ChargeState` / `MinMz` / `MaxMz` / `MinRetentionTime` / `MaxRetentionTime` / `ApexRetentionTime` / `PeakCount`, and both enforce the same one-charge-state invariant.

The two PRs also collide on files: `FlashLFQResults.cs` (+81 here, +218 there) and the indexing engine (`IndexingEngine.cs` here, `PeakIndexingEngine.cs` there).

#1122's test file calls `PeakWindow.Create(...)` and `PeakWindow.BuildMs1ScanNumberMap(dataFile)` returning `int[]` — this PR's API, not its own — which suggests #1122 is a later redesign of this same work whose test never caught up. If that reading is right, this PR is the older generation and the two should be reconciled into one rather than both landing.

Not touching either until that is decided. Details and the full API drift table are on #1122.

trishorts:
Correcting my comment above — I had the relationship backwards, and the conclusion flips with it.

These are **not** two PRs adding the same class in parallel. This branch is **stacked on #1122**: `git merge-base` of the two is `20262b92` (Aug 3), and this PR carries four of #1122's commits plus exactly one of its own.

```
933463d5 2026-08-03 trishorts      fix(flashlfq): bound each peak window to one charge state   <- this PR's only own commit
20262b92 2026-08-03 Alexander-Sol  Merge branch 'master' into PeakEnvelopeOutput
4c1011f4 2026-08-03 Alex           test(flashlfq): cover the peak window branches codecov flagged
44994f92 2026-08-03 trishorts      Merge branch 'master' into PeakEnvelopeOutput
e8683273 2026-07-31 Alex           feat(flashlfq): export the raw MS1 peaks around each quantified peak
```

So `PeakWindow` here and `PeakWindowData` there were never two designs. They are the same class at two points in its own history — this branch froze it on Aug 3, and #1122 kept going, renaming it in `53ac091c` on Aug 5.

And the one fix this PR does add is already in #1122: `Ms1FeatureData.Create` groups `peak.IsotopicEnvelopes.GroupBy(e => e.ChargeState)` and emits one window per charge state (`Ms1FeatureData.cs:93,150`), and `PeakWindowData.Create` throws on mixed charge states. There is nothing left here that #1122 does not have.

Closing this as superseded by #1122. Nothing is lost: everything in this branch is either #1122's own commits or a fix #1122 has since absorbed.

#1122's red CI is unrelated to any of this — `53ac091c` renamed the test file with zero content changes while renaming both classes under it, so `TestMs1FeatureData.cs` is one generation behind. Details on #1122.

trishorts:
Superseded by #1122 — this branch is stacked on it, and its one distinct fix (one peak window per charge state) is already in #1122's `Ms1FeatureData.Create`. See the comment above for the branch topology.

PR #1131, opened 2026-08-04
Refactor to separate out chronologer

Alexander-Sol:
Make Chromatography a high-level project and split Chronologer into an opt-in package

## Problem

`Chromatography` sat at level 4 of the dependency graph, just above `Omics`, and it is the
only project carrying `TorchSharp` + `libtorch-cpu`. Because `Proteomics` referenced it, and
`UsefulProteomicsDatabases` -> `Readers` -> `FlashLFQ`/`SpectralAveraging`/`TopDownSimulator`
all sit above `Proteomics`, **11 of 21 projects transitively bound libtorch** - including
`Readers` and `FlashLFQ`, the two most widely consumed.

The nuspec made it worse: `mzLib` shipped as a single flat package declaring `TorchSharp` and
`libtorch-cpu` as hard dependencies, so every consumer restored the lot. `libtorch-cpu` is a
pure metapackage that unconditionally pulls all three RIDs:

    libtorch-cpu-win-x64      318 MB   (torch_cpu.dll alone is 237.8 MB)
    libtorch-cpu-linux-x64    563 MB
    libtorch-cpu-osx-arm64    236 MB
    -------------------------------
    total                   1,117 MB

The actual coupling turned out to be two files and about 13 lines.

## What changed

**1. `IRetentionPredictable` moved from `Chromatography` into `Omics`**
(`Omics/RetentionTimePrediction/`). Every real implementer is an Omics type, so the interface
was simply in the wrong project - its own doc comment claimed "no dependencies on higher
layers (e.g. Omics)" while living below the types that implement it.

Note: I deliberately did *not* make `IBioPolymerWithSetMods` extend it. SSRCalc3, CZE and
Chronologer are all peptide-specific models, so making `OligoWithSetMods` retention-predictable
would be scientifically wrong, and re-declaring `BaseSequence`/`FullSequence` across the two
interfaces trips CS0108 hiding. Straight relocation only.

**2. The `Proteomics -> Chromatography` edge is cut.** The back-compat shim
(`ChromatographyExtensions.cs`, whose own comment says it exists "for backwards compatibility
with the old structure prior to the existence of the Chromatography namespace") moved into
`Chromatography/Compat/`, keeping its `Proteomics.RetentionTimePrediction` namespace.
`ScoreSequence`'s parameter widened from `PeptideWithSetModifications` to
`IBioPolymerWithSetMods`, so the shim depends only on `Omics` and `Chromatography` never has to
reference `Proteomics`. Both shims are now `[Obsolete]`.

Also removed a dead `using Chromatography.RetentionTimePrediction;` at
`Readers/SpectralLibrary/MslLibrary.cs:1` that had no usages and only compiled via the
transitive path.

**3. New `Chromatography.Chronologer` project** holding `Chronologer.cs`,
`ChronologerRetentionTimePredictor.cs`, the embedded weights, and
`RetentionTimePredictorFactory` (the one type that must reference all three concrete
predictors). `Chromatography` now carries only the abstractions plus the pure-C# SSRCalc3 and
CZE predictors, with no dependencies beyond `Omics` and `MzLibUtil`.

This finishes a refactor `RetentionTimePredictorFactory`'s own docstring already started: it
decoupled the interface from the concrete predictors, but left the factory in the same
assembly, so referencing the assembly still dragged in Torch. Splitting the assembly was the
missing half.

**4. The NuGet package is split.** `mzLib` no longer declares any Torch dependency. New
`mzLib.Chronologer` ships `Chromatography.Chronologer.dll` with `TorchSharp` +
`libtorch-cpu-win-x64` (win-x64 only rather than the 1,117 MB metapackage - consistent with the
package already shipping Windows-only Thermo/Bruker natives) and an exact version pin back to
`mzLib`, since the two are built from the same release tag.

`release.yml` packs and pushes both. The integration job in `dotnet.yml` stamps both nuspecs,
packs them explicitly (a bare `nuget pack` fails once two nuspecs share a folder), and adds
`mzLib.Chronologer` to MetaMorpheus's `EngineLayer` and `Test`.

## Result

Projects transitively binding TorchSharp/libtorch: **11 -> 2**
(`Chromatography.Chronologer` by design, and `Test` because it references it).
Verified with `dotnet list package --include-transitive` across all 21 projects.

## Verification

- Debug and Release builds: 0 errors.
- Full suite (`Category!=ExternalService`): **5320 passed, 0 failed, 29 skipped**.
- Both packages pack cleanly; unzipped and inspected to confirm `mzLib` carries no Torch
  dependency and `mzLib.Chronologer` ships only its own DLL.
- The embedded-weights lookup builds its resource name from the assembly name, so moving the
  assembly could have broken weight loading silently at runtime. 94 Chronologer tests pass, and
  an explicit `<LogicalName>` is now pinned so a future folder rename fails at build time
  instead.

## Downstream impact (MetaMorpheus)

**No source changes required.** Namespaces were preserved throughout:

- `GuiFunctions/MetaDraw/PlotModelStat.cs` (3 sites) constructs
  `Proteomics.RetentionTimePrediction.SSRCalc3` - still resolves via the moved shim, now with
  an obsolescence warning. Neither repo sets `TreatWarningsAsErrors`.
- `EngineLayer/FdrAnalysis/FdrAnalysisEngine.cs` uses
  `Chromatography.RetentionTimePrediction.Chronologer` - namespace unchanged, now supplied by
  the new assembly.
- `EngineLayer/FdrAnalysis/PEPAnalysisEngine.cs:11`'s `using Proteomics.RetentionTimePrediction;`
  is already dead (it uses `SSRCalc3RetentionTimePredictor` from the Chromatography namespace)
  and survives either way.

## Follow-ups (not in this PR)

1. MetaMorpheus master should carry `<PackageReference Include="mzLib.Chronologer" />` in
   `EngineLayer.csproj` itself; the lines added to the integration job are a stopgap so CI
   passes in the meantime.
2. Worth guarding the unguarded `Lazy<ChronologerRetentionTimePredictor>` at
   `FdrAnalysisEngine.cs:15-18`. Chronologer is the *default* RT predictor
   (`CommonParameters.cs:67`), and now that libtorch is opt-in, a consumer who takes `mzLib`
   without the extra package gets a `DllNotFoundException` wrapped in a
   `TypeInitializationException` on first use rather than a graceful skip.
3. Delete the `[Obsolete]` shims in the next major version.


trishorts (changes requested):
Requesting changes - detail is in the inline comments on this diff, one thread per issue.

The two blockers are the `mzLib.sln` `Any CPU` rows and the `mzLib.Chronologer.nuspec` `bin\x64\Release` paths, both consequences of this branch predating #1127. The rest are design and contract questions worth a round of discussion.

The `Omics` relocation, the `Compat/` folder, the pinned `<LogicalName>`, and the explicit two-nuspec packs are all good calls.

mzLib/mzLib.sln
```diff
@@ -161,6 +163,12 @@ Global
 		{9DD8E48A-8A6F-D546-3B4D-A20D488457A4}.Release|x64.Build.0 = Release|x64
 		{9DD8E48A-8A6F-D546-3B4D-A20D488457A4}.TestAndRelease|x64.ActiveCfg = Release|x64
 		{9DD8E48A-8A6F-D546-3B4D-A20D488457A4}.TestAndRelease|x64.Build.0 = Release|x64
+		{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Debug|x64.ActiveCfg = Debug|x64
+		{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Debug|x64.Build.0 = Debug|x64
+		{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Release|x64.ActiveCfg = Release|x64
+		{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Release|x64.Build.0 = Release|x64
+		{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.TestAndRelease|x64.ActiveCfg = Release|x64
+		{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.TestAndRelease|x64.Build.0 = Release|x64
```
trishorts, line 171: **Blocker — this branch predates #1127.**

#1127 ("Package architecture-neutral assemblies so mzLib can be used on arm64") merged at `b92fc65d`, roughly twelve minutes before this branch's head commit; `git merge-base --is-ancestor b92fc65d <head>` returns false. It added `Debug|Any CPU` / `Release|Any CPU` to `SolutionConfigurationPlatforms` **and a row pair for every one of the 21 project GUIDs**.

`{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}` gets only the three `|x64` variants here, so `dotnet build /p:Platform="Any CPU"` — now what both `dotnet.yml` and `release.yml` run — will skip `Chromatography.Chronologer`. `Test.csproj` `ProjectReference`s it, so the build and integration jobs both break after the rebase.

```
{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Debug|Any CPU.ActiveCfg = Debug|AnyCPU
{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Debug|Any CPU.Build.0 = Debug|AnyCPU
{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Release|Any CPU.ActiveCfg = Release|AnyCPU
{C5C7359A-88EF-4D6D-87B3-0CE85ADD8E1F}.Release|Any CPU.Build.0 = Release|AnyCPU
```

mzLib/mzLib.Chronologer.nuspec
```diff
@@ -0,0 +1,42 @@
...
+      <group targetFramework="net8.0-windows7.0">
+        <dependency id="mzLib" version="[$mzlibversion$]" />
+        <dependency id="TorchSharp" version="0.105.1" />
+        <dependency id="libtorch-cpu-win-x64" version="2.5.1" />
+      </group>
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="mzLib.png" target="images\" />
+    <file src="Chromatography.Chronologer\bin\x64\Release\net8.0\Chromatography.Chronologer.dll" target="lib\net8.0" />
```
trishorts, line 39: **Blocker — these paths no longer exist after rebasing on #1127.**

#1127 repointed every `<file src>` in `mzLib.nuspec` from `bin\x64\Release\…` to `bin\Release\…`, because the release build now runs as Any CPU. Both lines here still say `bin\x64\Release\net8.0\`.

That either fails the pack outright, or — if a stale `bin\x64` tree survives on the runner — quietly ships an AMD64-stamped assembly, which is exactly the bug #1127 was written to fix.

mzLib/Chromatography.Chronologer/Chromatography.Chronologer.csproj
```diff
@@ -0,0 +1,45 @@
...
+      loading silently at runtime rather than at build time.
+    -->
+    <EmbeddedResource Include="RetentionTimePrediction\Chronologer\Chronologer_20220601193755_TorchSharp.dat">
+      <LogicalName>Chromatography.Chronologer.RetentionTimePrediction.Chronologer.Chronologer_20220601193755_TorchSharp.dat</LogicalName>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </EmbeddedResource>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="libtorch-cpu-win-x64" Version="2.5.1" />
```
trishorts, line 35: **This drops Apple Silicon and Linux, right as MetaMorpheus gains them.**

Narrowing from the `libtorch-cpu` metapackage to the win-x64 RID package does fix the 1,117 MB restore, but it also removes `libtorch-cpu-osx-arm64` (236 MB) and `libtorch-cpu-linux-x64` (563 MB) from the graph entirely. Chronologer is MetaMorpheus's **default** RT predictor (`CommonParameters.cs:67`), so on the Apple Silicon build that #1127 + MetaMorpheus #2701 just enabled, every FDR analysis hits a `DllNotFoundException` on first use.

The comparison to the Windows-only Thermo/Bruker natives isn't quite parallel — those gate a file format, this gates the default scoring path. Referencing all three RID packages costs the same as the metapackage on any single machine and preserves current behaviour. If Windows-only really is the intent, it should be a stated scope decision with the arm64 consequence spelled out, not a side effect of the size fix.

mzLib/Chromatography/RetentionTimePrediction/IRetentionTimePredictor.cs
```diff
@@ -14,22 +15,22 @@ namespace Chromatography.RetentionTimePrediction;
...
+/// concrete predictors. This keeps the abstraction free of references to concrete types —
+/// most importantly the TorchSharp-backed <c>ChronologerRetentionTimePredictor</c>, which
+/// would otherwise force every consumer of this interface to transitively depend on
+/// TorchSharp and ship its large native binaries.
 /// </para>
 /// <para>
-/// <b>Migration:</b> the previously-nested <c>Create</c> method and
-/// <c>PredictorType</c> enum on this interface have been moved to the top-level
-/// <see cref="RetentionTimePredictorFactory"/> and <see cref="PredictorType"/>.
+/// <b>Assembly layout:</b> the factory, the <c>PredictorType</c> enum and the Chronologer
```
trishorts, line 30: **Moving the factory into the Torch assembly gates the *free* predictors behind libtorch.**

(Commenting here since `RetentionTimePredictorFactory.cs` is a 100%-similarity rename with no diff hunk to anchor on.)

Two of the three branches — `SSRCalc3` and `CZE` — are pure C# with, per the enum's own docs, "no external model files or native dependencies". After this PR a consumer taking `mzLib` alone can no longer reference the factory *or* even name `PredictorType.SSRCalc3`, so the one documented construction entry point for the lightweight predictors requires the 318 MB package. This interface's remarks still say construction goes through the factory. That inverts the PR's goal, and it removes public API from `Chromatography.dll` without a major version bump.

"The factory must reference all three concrete predictors" is true of *this* shape, not of the split. An additive alternative that keeps the dependency cut:

```csharp
// Chromatography\RetentionTimePrediction\RetentionTimePredictorFactory.cs — stays put
public static IRetentionTimePredictor Create(PredictorType type) => type switch
{
    PredictorType.SSRCalc3    => new SSRCalc3RetentionTimePredictor(),
    PredictorType.CZE         => new CZERetentionTimePredictor(),
    PredictorType.Chronologer => CreateChronologer(),
    _ => throw new ArgumentOutOfRangeException(nameof(type), type, null)
};

// Resolved by name, so Chromatography never references Chromatography.Chronologer.
private static IRetentionTimePredictor CreateChronologer()
{
    var t = Type.GetType("Chromatography.RetentionTimePrediction.Chronologer." +
                         "ChronologerRetentionTimePredictor, Chromatography.Chronologer");
    if (t is null)
        throw new InvalidOperationException(
            "Chronologer retention time prediction requires the mzLib.Chronologer package. " +
            "Add <PackageReference Include=\"mzLib.Chronologer\" /> or use PredictorType.SSRCalc3.");
    return (IRetentionTimePredictor)Activator.CreateInstance(t)!;
}
```

A registration hook works equally well. It would also turn follow-up #2 — the unguarded `Lazy<ChronologerRetentionTimePredictor>` in `FdrAnalysisEngine` — into an actionable message rather than a `DllNotFoundException` wrapped in a `TypeInitializationException`. I'm fine with the current placement if the trade-off is stated in the factory's remarks; mainly I want to know it was chosen rather than fallen into.

mzLib/Chromatography/Compat/ChromatographyExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// <summary>
+    /// Scores a biopolymer's base sequence with SSRCalc3.
+    /// </summary>
+    /// <remarks>
+    /// The parameter was widened from <c>PeptideWithSetModifications</c> to
+    /// <see cref="IBioPolymerWithSetMods"/> so this shim depends only on Omics. Existing call sites
+    /// that pass a <c>PeptideWithSetModifications</c> continue to compile unchanged.
+    /// </remarks>
+    public static double ScoreSequence(this Chromatography.RetentionTimePrediction.SSRCalc.SSRCalc3 predictor,
+        IBioPolymerWithSetMods peptide) => predictor.ScoreSequence(peptide.BaseSequence);
```
trishorts, line 31: **The widening admits RNA to a peptide-only model.**

`OligoWithSetMods : NucleolyticOligo, IBioPolymerWithSetMods, INucleicAcid` — so an RNA oligo now satisfies this signature and gets forwarded into a Krokhin reversed-phase model built on the 20 proteinogenic residues. Because A, C, G and U are all also valid amino-acid one-letter codes, SSRCalc3 won't throw; it returns a plausible-looking `double`.

That's the exact outcome the PR body argues against when it declines to make `IBioPolymerWithSetMods` extend `IRetentionPredictable` — arriving through the back door, since the `PeptideWithSetModifications` parameter *was* the type-level barrier.

The widening doesn't need to go this far. `IRetentionPredictable` now lives in `Omics` too, exposes `BaseSequence`, and is implemented by `PeptideWithSetModifications` but deliberately not by `OligoWithSetMods`:

```csharp
public static double ScoreSequence(this SSRCalc3 predictor, IRetentionPredictable peptide)
    => predictor.ScoreSequence(peptide.BaseSequence);
```

Same Omics-only dependency, same source compatibility for every existing peptide call site, and the unsupported analyte is excluded structurally rather than by convention.

mzLib/Omics/RetentionTimePrediction/IRetentionPredictable.cs
```diff
@@ -1,9 +1,16 @@
-namespace Chromatography.RetentionTimePrediction;
+namespace Omics.RetentionTimePrediction;
```
trishorts, line 1: **This namespace change is a source break, contrary to "namespaces were preserved throughout".**

`Chromatography.RetentionTimePrediction.IRetentionPredictable` → `Omics.RetentionTimePrediction.IRetentionPredictable`, with no shim. Every other relocation in this PR preserved its namespace precisely to avoid that — `ChromatographyExtensions` changed assemblies while keeping `Proteomics.RetentionTimePrediction`, and Chronologer kept `Chromatography.RetentionTimePrediction`.

Anyone outside this repo with a custom `IRetentionTimePredictor` implementation, or a `using Chromatography.RetentionTimePrediction;` that names this interface, gets CS0246 with no deprecation cycle.

Either leave an `[Obsolete]` interface in the old namespace extending the relocated one, or state in the PR body that this one is a deliberate break — defensible, it's a young interface — so it isn't covered by the blanket claim.

.github/workflows/release.yml
```diff
@@ -39,11 +39,16 @@ jobs:
...
+        # Two packages ship from this tag and must stay in lockstep: mzLib, and mzLib.Chronologer,
+        # which carries the TorchSharp/libtorch payload as an opt-in extra. mzLib.Chronologer pins an
+        # exact dependency on mzLib, so mzLib must be pushed first.
         run: |
           $version= &git describe --tags
           cd mzLib
           nuget pack mzLib.nuspec -Version ${version} -Properties Configuration=Release
+          nuget pack mzLib.Chronologer.nuspec -Version ${version} -Properties "Configuration=Release;mzlibversion=${version}"
           dotnet nuget push mzLib.${version}.nupkg --source https://api.nuget.org/v3/index.json --api-key ${env:GITHUB_TOKEN}
+          dotnet nuget push mzLib.Chronologer.${version}.nupkg --source https://api.nuget.org/v3/index.json --api-key ${env:GITHUB_TOKEN}
```
trishorts, line 52: **Nothing enforces the push ordering the comment above describes.**

The step's comment states the invariant — "mzLib.Chronologer pins an exact dependency on mzLib, so mzLib must be pushed first" — but this is a single `run:` block, `$ErrorActionPreference` doesn't abort on native-executable failure, and the step's exit code comes from the last command. Before this PR the mzLib push *was* last, so its failure failed the step; now it isn't.

Failure mode: the mzLib push fails, this one succeeds, the job reports green, and nuget.org permanently holds an `mzLib.Chronologer` pinned to `[version]` of an `mzLib` that was never published — unrestorable, and NuGet.org versions are immutable.

Worth a `$LASTEXITCODE` check after each `nuget pack` and each `dotnet nuget push`, or splitting them into separate `run:` steps so Actions enforces the ordering.

.github/workflows/dotnet.yml
```diff
@@ -95,6 +101,12 @@ jobs:
         dotnet add EngineLayer/EngineLayer.csproj package mzLib -v 9.9.9;
         dotnet remove Test/Test.csproj package mzLib;
         dotnet add Test/Test.csproj package mzLib -v 9.9.9;
+        # Chronologer moved out of the main mzLib package into mzLib.Chronologer. EngineLayer is the
+        # only direct consumer (FdrAnalysisEngine constructs ChronologerRetentionTimePredictor); the
+        # remaining projects pick it up transitively. Added to Test as well for parity with mzLib.
+        # Once MetaMorpheus master carries these PackageReferences itself, these lines can go.
+        dotnet add EngineLayer/EngineLayer.csproj package mzLib.Chronologer -v 9.9.9;
+        dotnet add Test/Test.csproj package mzLib.Chronologer -v 9.9.9;
```
trishorts, line 109: **The MetaMorpheus counterpart needs to merge ahead of the mzLib release, not follow it.**

These two lines keep *CI* green, and the comment is honest that they're a stopgap — but the first real release cut from this breaks MetaMorpheus master, since `FdrAnalysisEngine` constructs `ChronologerRetentionTimePredictor` and `EngineLayer.csproj` carries no reference to the new package.

Follow-up #1 should be a companion MetaMorpheus PR, named in this description and merged first, rather than a follow-up. Worth noting `Test/RetentionTimeTest.cs` also uses `Proteomics.RetentionTimePrediction`, on top of the `PlotModelStat.cs` sites listed in the body.

Alexander-Sol:
The MetaMorpheus installer sits at 80mb with the new dependencies. While this is a ~150% increase in size, the absolute size of the installer remains small and as such this work is unnecessary.

PR #1132, opened 2026-08-05
Add Reviewed.all so a complete proteome can be downloaded

trishorts:
> **Stacked on #1126.** Until that merges, this PR's diff also contains #1126's two commits, because `master` does not have them yet. **The change under review here is the single commit `25b73c9e`** — 2 files, +189/−27. The diff collapses to just that once #1126 merges. Merge #1126 first.

## The problem

`ProteinDbRetriever.RetrieveProteome` cannot download a whole organism's proteome. `Reviewed.yes` and `Reviewed.no` are mutually exclusive halves — Swiss-Prot only, or TrEMBL only — so between them there is no way to ask for all of it:

| Homo sapiens (UP000005640) | entries |
|---|---|
| `Reviewed.yes` | 20,416 |
| `Reviewed.no` | 127,090 |
| what a caller means by "the human database" | **147,506** |

## The change

One enum member and the four places that consume it.

```csharp
public enum Reviewed { yes, no, all }
```

- **Query** — `all` omits the `reviewed:` clause rather than asking for both values, because the absence of that clause is how UniProt spells "every entry". `reviewed:true OR reviewed:false` is the obvious-looking mistake.
- **File name** — `_all`, so a complete proteome is not mistaken on disk for either half.
- **Empty-result message** — no longer says "no *reviewed* proteins" when review status was not what was asked about.
- **Validation** — the review status is rejected if it is not one of the three defined values, beside the existing `format` check, so an undefined value cannot fall through to `all`.

## Tests

- Both URLs pinned for all six enum combinations, including that `all` sends no `reviewed:` clause at all.
- The four `_all` file names.
- A live canary asserting the arithmetic that makes `all` mean complete — `all == reviewed + unreviewed` — rather than a fixed count, which UniProt changes whenever it curates. Plasmodium falciparum (UP000001450) is the smallest proteome with a substantial amount of both, so neither half can pass for the whole by accident.
- A live canary for the complete proteome in xml, the format that carries the annotations.

112 tests green; the live fixture runs in 6 s. New live tests carry `[Category("ExternalService")]` and route through `ExternalServiceTestHelper.RunAsync`, so they land in the non-blocking external-service job rather than the required one.

## Scope and compatibility

Touches `RetrieveProteome` and the `Reviewed` enum only — additive, with no existing member body or signature changed. Adding a member to a public nested enum is source- and binary-compatible.

**No MetaMorpheus counterpart needed.** MM's only use of this class is `ProteinDbRetriever.UniprotProteomesList` (via `GlobalVariables.AvailableUniProtProteomes`), whose signature and behaviour are untouched. Nothing in MM references `Reviewed`, `ProteomeFormat`, or `RetrieveProteome`.

## Deliberately not here

Each is a real fix, and each is its own change — kept out so this stays reviewable on its own:

- **The isoform parameter.** UniProt's parameter is `includeIsoform=`; this class sends `&includeIsoforms:` — plural, and joined with a colon instead of `=`, so it is a parameter name with no value and has always been ignored. Isoforms have never been included, in any release, while the file is still named `_isoform`. Streaming P02768 returns one sequence with the current spelling and three with the correct one. Fixing it changes which sequences arrive, which is why #1126 left it alone and why it does not ride along here.
- **Field-qualifying the query** as `(proteome:UP...)` rather than sending the bare ID as free text. Checked against four proteomes that both forms return identical counts today, so this is hardening, not a fix.
- **The proteome catalogue.** `DownloadAvailableUniProtProteomes` uses the paged `search` endpoint, so it returns the first 25 of roughly 1,086,000 proteomes and reports no problem — which is why looking an organism up in it fails. That needs its own PR, together with a way to find one organism's proteome without downloading all of them.

https://claude.ai/code/session_01Phj6bqnFkScdVJN7y3k3C7


Alexander-Sol (dismissed):

acesnik (approved):
🤖 Reviewed this one properly rather than skimming it, since it changes what a user ends up searching against. The change itself I'd take as-is — one documentation request before it goes in, and two stale bits in the description.

## The request: the new docs recommend a lookup this PR says is broken

The body is explicit that the catalogue path does not work:

> `DownloadAvailableUniProtProteomes` uses the paged `search` endpoint, so it returns the first 25 of roughly 1,086,000 proteomes and reports no problem — which is why looking an organism up in it fails.

Confirmed — `:401` is `proteomes/search?query=*&format=tsv&compressed=true` with no `size=`, so 25 rows.

But the same PR adds remarks pointing callers at exactly that path, in two places:

> The usual sequence for "give me the complete database for this organism" is to find the proteome ID in the catalogue and then download it with `RetrieveProteome` asking for `Reviewed.all` […]

> This is how a caller who knows an organism but not its proteome ID finds one: download the catalogue, read it with `UniprotProteomesList`, and pass the ID to `RetrieveProteome` […]

Someone following the new documentation looks up their organism, doesn't find it among the 25, and concludes UniProt has no proteome for it. Leaving the fix out is the right call — it needs paging plus a single-organism lookup, and that is its own change. It is advertising the route without the caveat that I'd rather not ship: the deferral is recorded in the PR description, which nobody reads from an IDE tooltip.

A sentence in the `DownloadAvailableUniProtProteomes` remarks saying it currently returns only the first page, and that finding an arbitrary organism this way is not yet reliable, would settle it. Happy either way on whether the class-level paragraph needs the same note or just a pointer.

## A nit, genuinely take it or leave it

```csharp
Reviewed.yes => "_reviewed",
Reviewed.no  => "_unreviewed",
_            => "_all",
```

The `Enum.IsDefined` check upstream does guarantee what the body claims, that an *undefined* value cannot reach `_all`. The gap is a newly *defined* member: add a fourth status later and it silently files under `_all`, which is the one outcome the validation exists to prevent. `Reviewed.all => "_all"` with a throwing discard makes the compiler carry that instead of a reviewer.

## Two things in the description that have gone stale

- #1126 merged on 5 August, so the header — "Stacked on #1126 […] Merge #1126 first" — no longer applies, and the diff has already collapsed to the 2 files / +189 −27 it describes.
- The commit under review is now `983f1a91`, not `25b73c9e`, after the two merges from master.

Worth tidying so the description matches what lands, since this one will be read later as the record of why `all` omits the clause.

## What I checked

- `all` omitting the `reviewed:` clause is right: absence is how UniProt spells "everything", and `reviewed:true OR reviewed:false` is the plausible-looking wrong answer. Good that the comment says so at the call site.
- `yes` and `no` are byte-identical to before — same suffix, same query, same exception text. This really is additive.
- Adding a member to the public nested enum is source- and binary-compatible.
- The no-MetaMorpheus-counterpart claim holds. MetaMorpheus touches this class only through `ProteinDbRetriever.UniprotProteomesList`, via `GlobalVariables.AvailableUniProtProteomes`, and nothing there references `Reviewed`, `ProteomeFormat` or `RetrieveProteome`.
- The gzip caveat you added is real and worth having: `ProteinDbLoader` decompresses to a hardcoded `temp.xml` / `temp.fasta` beside the source (`ProteinDbLoader.cs:110,248`), so two gzipped databases in one directory do collide.
- Asserting `all == reviewed + unreviewed` rather than a fixed count is the right shape for a live test, and `[Category("ExternalService")]` keeps a UniProt outage out of the required job.
- All five checks green.

Approving on the documentation note above; nothing here needs another round from my side once that is in.

PR #709, opened 2023-06-09
Raw file connection not closing 

elaboy:
Thermo Raw files were not closing once MM finished tasks., users had to close MM in order to not get the error indicating that the file was still being used. Tackled this issue including any method accessing the file inside a using statement.

Tested reading a file and moving it from the directory. Test is passing after the fix.

PR #1133, opened 2026-08-05
Fixed bug in .raw file reader where dynamic connections weren't re-used

Alexander-Sol:
## Fix: `GetOneBasedScanFromDynamicConnection` ignored the existing dynamic connection

`GetOneBasedScanFromDynamicConnection` opened a new `IRawDataPlus` connection on every call via
`RawFileReaderAdapter.FileFactory` instead of using the `dynamicConnection` field established by
`InitiateDynamicConnection`. This had two consequences: each dynamic scan retrieval paid the full
cost of opening the raw file, and the locally-created connection was never disposed, leaking a
file handle per call.

There was also a dead null-check — `dymConnection` was tested for null *after* being assigned
from `FileFactory`, so the intended guard for "connection not initialized" was unreachable.

**Changes:**
- Removed the local `dymConnection` variable
- Moved the null-check on `dynamicConnection` to the top of the method, before any use
- All reads now go through the already-open `dynamicConnection`, matching the intended
  `InitiateDynamicConnection` → `GetOneBasedScanFromDynamicConnection` → `CloseDynamicConnection`
  lifecycle


mzLib/Readers/Thermo/ThermoRawFileReader.cs
```diff
@@ -164,21 +164,18 @@ public override MsDataScan GetOneBasedScanFromDynamicConnection(int oneBasedScan
...
-            if (oneBasedScanNumber > dymConnection.RunHeaderEx.LastSpectrum ||
-                oneBasedScanNumber < dymConnection.RunHeaderEx.FirstSpectrum)
+            if (oneBasedScanNumber > dynamicConnection.RunHeaderEx.LastSpectrum ||
+                oneBasedScanNumber < dynamicConnection.RunHeaderEx.FirstSpectrum)
             {
                 return null;
             }
 
-            return ThermoRawFileReader.GetOneBasedScan(dymConnection, filterParams, oneBasedScanNumber);
+            return ThermoRawFileReader.GetOneBasedScan(dynamicConnection, filterParams, oneBasedScanNumber);
```
trishorts: This is giving back a deliberate fix. `git blame` puts the two lines removed here on **bee61eb8 - "Made Dynamic reading thread safe (#936)"**, which changed this exact method in the opposite direction:

```diff
-            dynamicConnection = RawFileReaderAdapter.FileFactory(FilePath);
+            var dymConnection = RawFileReaderAdapter.FileFactory(FilePath);
```

The per-call connection was not an oversight - it is what makes concurrent dynamic reads safe on Thermo. `IRawDataPlus` is stateful and not thread-safe, and this same file already treats it that way: `LoadAllStaticData` uses `RawFileReaderFactory.CreateThreadManager()` with one `CreateThreadAccessor()` per worker. Sharing one handle lets two threads interleave `GetFilterForScanNumber` / `GetCentroidStream` / `GetTrailerExtraInformation`, so you can get a scan whose header comes from one spectrum and whose peaks come from another - silently, with no exception.

#936 also added `protected readonly object DynamicReadingLock` on `MsDataFile` (`MassSpectrometry/MsDataFile.cs:31`), and Mzml, MsAlign, Mgf and Bruker all take it in this method. `MsAlign.GetOneBasedScanFromDynamicConnection` is the shape I think you want: static-load early return -> `lock (DynamicReadingLock)` -> null guard -> read through the field. This PR does 1, 3 and 4 and skips the lock.

CI will not catch it either. `TestDynamicConnection_IsThreadSafe` (`Test/FileReadingTests/ReadersInfrastructure/TestReaders.cs:108`) does cover `small.RAW` and `sliced_ethcd.raw`, but it calls `LoadAllStaticData()` first - so ever since #948 added the `CheckIfScansLoaded()` early return on line 164, every call in that test short-circuits before it touches the connection. That guard test has been passing vacuously for both Thermo cases.

This is not hypothetical downstream: in MetaMorpheus, `MetaDrawDataLoader` initiates one dynamic connection per file and keeps it open, then fires `ProcessDeconTab` and `ProcessChimeraTab` as concurrent `Task.Run`s over the same `MsDataFile` instances while the UI thread can also be reading scans.

The leak you are fixing is real and worth fixing. Taking `DynamicReadingLock` around the read gets you the reuse and the leak fix without giving back #936.

mzLib/Readers/Thermo/ThermoRawFileReader.cs
```diff
@@ -164,21 +164,18 @@ public override MsDataScan GetOneBasedScanFromDynamicConnection(int oneBasedScan
             if (CheckIfScansLoaded() && oneBasedScanNumber <= Scans.Length)
                 return GetOneBasedScan(oneBasedScanNumber);
 
-            var dymConnection = RawFileReaderAdapter.FileFactory(FilePath);
-            dymConnection.SelectInstrument(Device.MS, 1);
-
-            if (dymConnection == null)
+            if (dynamicConnection == null)
```
trishorts: The guard still cannot see a *closed* connection. `CloseDynamicConnection()` (line 187) calls `Dispose()` but never nulls the field, so `dynamicConnection` stays a non-null reference to a disposed `IRawDataPlus`. `Initiate -> Get -> Close -> Get` therefore falls straight through this check and dies on `RunHeaderEx` with an `ObjectDisposedException` from the vendor layer, where before this PR it silently worked.

The description says the moved guard makes a call "after `CloseDynamicConnection`" properly detected - that half is not delivered yet. Setting `dynamicConnection = null` after the `Dispose()` in `CloseDynamicConnection` closes the hole and makes a double close harmless at the same time.

mzLib/Readers/Thermo/ThermoRawFileReader.cs
```diff
@@ -164,21 +164,18 @@ public override MsDataScan GetOneBasedScanFromDynamicConnection(int oneBasedScan
             if (CheckIfScansLoaded() && oneBasedScanNumber <= Scans.Length)
```
trishorts: This early return needs to handle the out-of-range case too, or the PR changes documented behaviour.

`CheckIfScansLoaded()` is just `Scans != null && Scans.Length > 0`. So when scans came from `LoadAllStaticData` and `oneBasedScanNumber > Scans.Length`, this condition is false, we fall through, and the new guard throws `MzLibException("The dynamic connection has not been created yet!")`. Before this PR the same call opened a throwaway connection, ran the `FirstSpectrum`/`LastSpectrum` range check and returned `null` - which is what the XML doc directly above still promises: "Returns null if the raw file does not contain the scan number specified." The message is misleading as well, since the caller's actual problem is an out-of-range scan number, not an uninitialized connection.

Answering out-of-range from `Scans.Length` when the scans are already loaded keeps the documented contract and leaves the new guard for the genuinely dynamic path:

```csharp
if (CheckIfScansLoaded())
    return oneBasedScanNumber > 0 && oneBasedScanNumber <= Scans.Length
        ? GetOneBasedScan(oneBasedScanNumber)
        : null;
```

trishorts (commented):
Reviewed this with `/oracle` and `/review-pr`. The diagnosis in the description is right - the old code did reopen the raw file on every call, did leak the handle, and the null check was dead - so the direction is good. Three things to sort out first; the one on line 178 is the one I really care about.

1. **line 178** - the per-call connection was #936's thread-safety fix. Reusing the field without taking `DynamicReadingLock` hands that back, and the existing `TestDynamicConnection_IsThreadSafe` can no longer catch it (the `CheckIfScansLoaded()` early return short-circuits it).
2. **line 167** - the guard still cannot see a *closed* connection, because `CloseDynamicConnection` disposes without nulling the field.
3. **line 164** - an out-of-range scan number on a statically-loaded file now throws where the XML doc promises `null`.

A lifecycle test would be worth adding as well (`Initiate -> Get -> Close -> Get`, plus get-without-initiate). None of the PR's claims - reuse, no leaked handle, reachable guard - are asserted anywhere right now.

mzLib/Readers/Thermo/ThermoRawFileReader.cs
Alexander-Sol: Pull #936 doesn't change the behavior of the RawFileReader, it just renamed some fields

This PR is intended to revert a bug that was introduced way back in PR #709 . The other PRs that are referenced here all revolve around MsDataFile, not the implementation of the ThermoRawFileReader.

trishorts (approved):

RayMSMS (approved):

PR #1135, opened 2026-08-05
Speed up Thermo `LoadAllStaticData` by overlapping the file hash with scan reading, reading in file sequentially

Alexander-Sol:
### Problem
Loading a Thermo `.raw` file via `LoadAllStaticData` was slower than it needed to be.
The parallel file loop that was used to read in scans from the .raw file turned out to be slower simply reading in scans sequentially, without any parallelization. So now we do that. 

### Results
Full `LoadAllStaticData` on the 978 MB file (Release, warm cache, best of 3):

| Configuration | Load time |
| --- | --: |
| Original, single-threaded | ~6.9 s |
| Original, 4 threads | ~5.4 s |
| **This PR (any `maxThreads`)** | **~4.2 s** |

~22–39% faster, and no longer sensitive to how many threads the caller passes.

However, one major cost in raw file reading turns out to be calling  `GetSourceFile()`, which computes a **SHA-1 over the entire file** (~3.0 s, roughly half the total load time) and ran serially *after* the scan loop. Now, that runs as a task while scans are being read in. So hashing occurs in parallel with file loading

#### Change
- Run `GetSourceFile()`'s hash on a background task so it **overlaps** the scan reading:
  one thread reads the scans, one thread hashes the file.
- The hash runs on a **dedicated long-running thread** (`Task.Factory.StartNew` with
  `TaskCreationOptions.LongRunning`), not a pooled one. `LoadAllStaticData` is a synchronous
  entry point that blocks on the hash when it joins; if it is itself called from a parallel
  loop over files, a pooled hash task would compete for the same thread pool the join is
  blocking, which can cause pool starvation. A dedicated thread keeps the overlap independent
  of pool capacity.
- The hash is started **after** the `IsError` / `IsOpen` / `InAcquisition` guards, not before
  them, so an unreadable or still-acquiring file is refused before it is hashed (hashing an
  in-acquisition file would checksum a target that is still being written).
- The scan loop is wrapped in `try` / `finally`. On any failure — a guard, or an
  `MzLibException` out of `GetOneBasedScan` — the `finally` joins the hash task (releasing its
  read handle so the caller can retry/delete/move the file, and observing any exception it
  raised) and disposes the `FileFactory` handle. Previously a mid-load failure could leave the
  hash running with an open handle and swallow its exception.
- With the hash overlapped, the parallel scan loop only adds contention, so scan reading
  is now a **plain sequential loop over a single accessor**. The `Partitioner` /
  `Parallel.ForEach` machinery (and the `System.Collections.Concurrent` import) are removed.
- `maxThreads` is retained on the signature for `MsDataFile` API compatibility but no
  longer affects reading.

No behavioural change on the success path — `Scans` and `SourceFile` are identical to before.

### Concurrency
`LoadAllStaticData` is called once per spectra file. Callers may invoke it concurrently across
files (e.g. an outer parallel loop). The hash is deliberately placed on a dedicated thread
rather than the thread pool for exactly this case, so the internal overlap does not degrade
when many files load at once. The benchmark numbers below were measured loading a single file
at a time; per-file speedup under a saturated outer loop will be smaller, since the machine's
cores are already busy, but the change does not regress relative to the previous sequential
hash.

### Testing
- All 44 existing raw-file reader/source-file tests pass.
- Adds `Development/FileReading/BenchmarkFileReading.cs`: an `[Explicit]` NUnit benchmark
  that sweeps `LoadAllStaticData` across thread counts for a local file (any supported
  format), reporting best/median and speedup vs a sequential baseline, and asserting every
  configuration reads an identical peak count. Point it at a file with the
  `MZLIB_BENCHMARK_FILE` environment variable or by editing `DefaultBenchmarkFile`.


mzLib/Readers/Thermo/ThermoRawFileReader.cs
```diff
@@ -46,6 +45,11 @@ public override MsDataFile LoadAllStaticData(FilteringParams filteringParams = n
             // I don't know why this line needs to be here, but it does...
             var temp = RawFileReaderAdapter.FileFactory(FilePath);
 
+            // GetSourceFile computes a SHA-1 over the entire raw file, which is roughly
+            // half of the total load time and does not touch the Thermo reader. Run it on a background task so
+            // it overlaps the scan reading below: one thread reads the scans, one thread hashes the file.
+            var sourceFileTask = Task.Run(GetSourceFile);
```
trishorts: **Major - the background hash task is never observed on the failure paths.**

`sourceFileTask` starts here, but the only place it is ever touched is `sourceFileTask.GetAwaiter().GetResult()` on line 88, *after* the `using (var threadManager ...)` block. Every throw site inside that block - the three `MzLibException` guards just below, plus any `MzLibException` out of `GetOneBasedScan` ("unknown MS order", "Could not centroid data from scan", "Could not get precursor for scan #") - exits the method without touching the task.

Before this PR, `GetSourceFile()` ran inside the block after the scan loop, so a scan failure meant the hash simply never started. Now:

- a background thread keeps hashing up to ~1 GB after the caller already has its exception, holding a read handle that can block the caller's retry / delete / move of that file;
- if the hash itself throws (`IOException` from a sharing violation, or the file being deleted mid-load), it is swallowed as an unobserved task exception;
- `temp` is left undisposed on those same paths - pre-existing, but the new task compounds it by adding a second live handle nobody owns.

Wrapping the accessor block in `try`/`finally` and observing (or joining) `sourceFileTask` in the `finally`, alongside `temp.Dispose()`, closes both.

**Related, and it is the same edit:** the `Task.Run` sits *before* the `IsError` / `!IsOpen` / `InAcquisition` guards, so a corrupt or in-acquisition file gets hashed anyway before being refused. `InAcquisition` is the awkward one - hashing a file that is actively being written produces a checksum of a moving target, while the guard exists precisely to say "don't touch this file yet." Moving the `Task.Run` to just after the guards and `SelectInstrument` preserves the overlap in full (the scan loop is the long part) while letting invalid files fail fast.

mzLib/Readers/Thermo/ThermoRawFileReader.cs
```diff
@@ -68,27 +72,20 @@ public override MsDataFile LoadAllStaticData(FilteringParams filteringParams = n
...
+                    msDataScans[s] = GetOneBasedScan(rawFileAccessor, filteringParams, s + 1);
+                }
 
                 rawFileAccessor.Dispose();
                 Scans = msDataScans;
-                SourceFile = GetSourceFile();
             }
-            
+
+            SourceFile = sourceFileTask.GetAwaiter().GetResult();
```
trishorts: **Major - blocking join on a pooled work item, inside a synchronous public entry point.**

`LoadAllStaticData` is synchronous, public, and called once per spectra file from the MetaMorpheus task layer. Blocking a thread-pool thread here while another pool thread runs the hash is the classic starvation shape: a caller that wraps this in a `Parallel.ForEach` over files parks every outer worker on an inner pool work item that cannot be scheduled until the pool hill-climbs in new threads (roughly one or two per second). That is exactly the regime where the overlap this PR adds can invert into a regression.

`Task.Factory.StartNew(GetSourceFile, TaskCreationOptions.LongRunning)` puts the hash on a dedicated thread rather than a pooled one, so the join below cannot starve the pool. Same overlap, no coupling to pool capacity.

Secondary: an exception from `GetSourceFile()` (`UnauthorizedAccessException` or `IOException` out of `File.OpenRead`) used to propagate directly with its original stack; coming back through the awaiter rewrites it.

Could you also note in the PR description whether this reader is expected to be called concurrently from an outer parallel loop? That is the case where the benchmark numbers may not carry over.

trishorts (commented):
Two blocking comments on the concurrency change, both on `ThermoRawFileReader.LoadAllStaticData`. The idea is good and the speedup is worth having - these are about what happens on the paths the benchmark does not exercise.

Nothing to say about `BenchmarkFileReading.cs`; the `[Explicit]` guard and the missing-file skip look right.

trishorts:
@Alexander-Sol I put up a PR against your branch to clear the `codecov/patch` failure: **Alexander-Sol/mzLib#1** (`trishorts:pr1135-coverage` → `RawFilePerformance`). Merge it into your branch and the gate here should go green.

Short version of what it does and why it isn't tests-only:

The patch is 32 lines — 22 hits, 7 misses, 3 partials. `>92%` means ≥30 of 32, so 8 of the 10 gaps have to close, and all ten are the three accessor guards (`IsError` / `!IsOpen` / `InAcquisition`) plus the `catch` in your new `finally`.

The guards have **no test-only route**. `RawFileReaderFactory.CreateThreadManager` is a hard-coded static on Thermo's sealed API — nothing can hand `LoadAllStaticData` an accessor that's already in a failing state. A hand-corrupted `.raw` doesn't work either, because `RawFileReaderAdapter.FileFactory` runs before the guards and throws there first (and with `OpenMcdf` pinned to 2.x for the known CFB DoS, a crafted file could hang instead of throwing). There's no mocking library anywhere in mzLib.

So the three guards move into an `internal static ThrowIfNotReadable(bool, bool, bool)` — same order, same messages — and your 9-line block becomes one call that every existing raw test already covers. Primitives rather than `IRawDataPlus` because `Test.csproj` deliberately carries no ThermoFisher reference; `internal` because `Readers.csproj` already declares `InternalsVisibleTo("Test")`, so no project file changes. Same shape as `ProteinDbRetriever.ThrowIfServiceUnavailable` from #1126.

The `catch` needed no seam — `GetSourceFile` is a non-sealed `public override`, so a test subclass throws from the hash task.

Two things worth your eyes:

- I left `InitiateDynamicConnection`'s copy of the same three checks alone. It evaluates them in a different order (`IsOpen` → `IsError` → `InAcquisition`), so sharing the helper would change which message a doubly-bad file reports — a behaviour change in code this PR doesn't touch. Flagged as a separate follow-up.
- The test for the `catch` also asserts the file is fully released once `LoadAllStaticData` unwinds, which is the behaviour your new `finally` exists to provide. Heads up that my first version failed against the shared `DataFiles\small.RAW` — other tests in the suite hold handles on it — so it runs against a private copy in a scratch dir.

823/823 `FileReadingTests` pass locally, and a cobertura run confirms every previously-uncovered line is hit with no partial branches left in the file.

mzLib/Readers/Thermo/ThermoRawFileReader.cs
trishorts: Verified fixed at `9092ec5c`. The whole body is now wrapped in `try`/`finally`, and the `finally` joins `sourceFileTask` on every exit path -- the three guards, an `MzLibException` out of `GetOneBasedScan`, or the success path -- so the hash can no longer outlive the call holding a read handle. The join is inside its own `try`/`catch` that swallows the hash's exception with a comment saying why, so it is observed without masking the original failure, and `temp.Dispose()` moved into the same `finally`. Resolving.

mzLib/Readers/Thermo/ThermoRawFileReader.cs
trishorts: Verified fixed at `9092ec5c`. Now `Task.Factory.StartNew(GetSourceFile, CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default)`, so the hash runs on a dedicated thread and the blocking join cannot starve the pool when `LoadAllStaticData` is called from an outer parallel loop over files.

You also took the secondary point: the task now starts *after* `ThrowIfNotReadable` rather than before, so an unreadable or in-acquisition file is refused without being hashed, and the comment above it explains both the dedicated-thread choice and the ordering. Resolving.

trishorts (approved):
Approving at `9092ec5c`. Both Major findings from my 2026-08-05 pass are fixed, and I have verified each against the current code rather than the commit message:

- **Hash task observed on every path.** `try`/`finally` around the whole body, the `finally` joins `sourceFileTask` and swallows its exception inside its own `catch` so it cannot mask the original, and `temp.Dispose()` moved in with it. No path now leaves a ~1 GB SHA-1 running against an open handle after the caller already has its exception.
- **Dedicated thread instead of a pooled work item.** `Task.Factory.StartNew(..., TaskCreationOptions.LongRunning, TaskScheduler.Default)`, so the synchronous join cannot starve the pool when this is invoked from a parallel loop over files -- which is the regime where the overlap would otherwise have inverted into a regression.

You also picked up the secondary point I raised, starting the hash after the readability guards so an in-acquisition file is not checksummed while it is still being written.

Two things noted while re-reading, neither blocking and neither needing action:

- This is a net improvement on disposal, not just a wash. Before this PR the outer `rawFileAccessor` was never disposed at all -- only the per-thread accessors inside the `Parallel.ForEach` were -- and it is now disposed on the success path.
- `ThrowIfNotReadable` taking the three flags as primitives is what made the guards reachable from the Test project at all, given it carries no ThermoFisher reference. The `<remarks>` block explaining that is worth keeping if this method is ever refactored.

Coverage is settled: the patch-coverage gap is closed and `codecov/patch` / `codecov/project` are both green.

Two mechanical notes before this can go in:

1. The branch is one commit behind master (#1132 merged at 18:06). `dismiss_stale_reviews` is on for this repo, so updating the branch will dismiss this approval and it will need re-giving -- worth updating first and then collecting both approvals, rather than the other way round.
2. `integration` is red for reasons unrelated to this PR: MetaMorpheus master's `Product.wxs` still lists five EF6/SqlClient files that #2710 stopped producing, so every open mzLib PR is failing that job identically right now. smith-chem-wisc/MetaMorpheus#2720 fixes it. `integration` is not a required check here -- only `build` is -- so it does not block the merge.

RayMSMS (approved):

PR #1136, opened 2026-08-08
Keep the NCBI taxonomy id when loading a UniProt FASTA

trishorts:
The same protein loaded from XML and from FASTA disagreed about whether it had a taxonomy identifier.

`LoadProteinXML` keeps it, though not through any deliberate taxonomy handling: `ProteinXmlEntry` stores every `dbReference` generically by whatever type the file supplies, so the one inside the `<organism>` block comes along for free. A protein from `disulfidetests.xml` already carried `DatabaseReference("NCBI Taxonomy", "10090")`.

`LoadProteinFasta` did not. `OX=` already appeared in two header regexes — `UniprotFullNameRegex` and `UniprotOrganismRegex` — but only ever as a **terminator** for the preceding field. Nothing captured it, so `OS=Homo sapiens OX=9606` yielded the organism name and dropped the identifier.

The organism *name* is free text and ambiguous across databases; the taxonomy id is the stable machine identifier, and any consumer that wants to resolve an organism to a controlled-vocabulary term needs it. Requiring callers to know which format a protein came from in order to find it was the defect.

## What changed

- `UniprotOrganismIdRegex` captures `OX=`, and a FASTA-loaded protein gets the same `DatabaseReference("NCBI Taxonomy", …)` the XML path already produced.
- `Protein.NcbiTaxonomyId` and `Protein.NcbiTaxonomyDatabaseReferenceType`, so anything holding a `Protein` can find the taxonomy without depending on how it was read. `ProteinDbLoader` forwards the constant.

## Three defects found by audit before this shipped

Each left the taxonomy present in some proteins and absent in others — the exact failure the change exists to prevent.

**Decoys did not inherit it.** `DecoyProteinGenerator` calls the `Protein` constructor positionally and passed `null` for `databaseReferences`, so a FASTA target carried the taxonomy and its reverse decoy carried nothing. It now carries the taxonomy across, and **only** the taxonomy — a decoy is a reversed sequence from this organism's database so its organism is genuinely the same, but it is not annotated in EMBL or GO, and copying every reference would assert things about it that are false.

**The FASTA writer dropped it.** `Protein.GetUniProtFastaHeader` emitted `OS=… GN=…` with no `OX=`, so `ProteinDbWriter.WriteFastaDatabase` discarded the id `LoadProteinFasta` had just been taught to keep, and a FASTA→FASTA round trip lost it again. `OX=` is now written between `OS=` and `GN=`, matching real UniProt headers and matching `UniprotOrganismRegex`, which already treats `OX=` as a terminator. Omitted entirely when unknown, so a header without taxonomy is unchanged.

**The regex was unanchored.** `@"OX=(\d+)"` matched a description token ending in `OX=` followed by digits, so `… SOX=5 … OS=Homo sapiens OX=9606` yielded taxon `5`. A taxonomy id is a join key, where a silently wrong value is worse than a missing one. Now `(?:^|\s)OX=`.

## Compatibility

Additive. `organismIdRegex` is an optional trailing parameter defaulted for UniProt headers, and `databaseReferences` was already an optional `Protein` constructor argument. A header without `OX=` produces no taxonomy reference rather than an empty or zero one. `DatabaseReferences` does not participate in `Protein` equality or hashing, and `MergeProteins` dedupes correctly.

## Tests

9 in `TestUniProtTaxonId`: both load paths, absence, cross-format agreement, decoy inheritance, decoys *not* inheriting unrelated references, the FASTA round trip, header omission when unknown, and the `SOX=5` anchor case.

Full mzLib suite green — 5,415 passed / 0 failed.


Alexander-Sol (dismissed):
The core change is solid — the `(?:^|\s)OX=(\d+)` anchor correctly rejects `SOX=5`, the OS= terminator regex is untouched, the round-trip writes `OX=` with correct spacing and omits it cleanly when unknown, decoy aliasing is harmless (`DatabaseReference` is immutable), and `Merge` dedupes the new reference.

One consistency gap worth fixing before merge: **slide decoys still drop the taxonomy.** `GenerateReverseDecoys` now carries the reference across, but `GenerateSlideDecoys` (`DecoyProteinGenerator.cs:374`) still passes `null` for `databaseReferences`. So `LoadProteinFasta(path, true, DecoyType.Slide, …)` on an `OX=9606` header gives the targets a taxonomy and their decoys `null` — the exact "half the database answers, half doesn't" split this PR exists to close. The reverse-decoy justification in your comment at :166 applies verbatim to slide. Tests only exercise `DecoyType.Reverse`, so CI stays green over it; a slide-decoy case would catch it.

Minor: the OX regex takes the first match, so a space-delimited `OX=<digits>` token appearing in free text before `OS=` would win over the real one. Real UniProt headers never do this, but given the id is framed as a join key where silently-wrong beats missing, it's slightly at odds with that framing.

zhuoxinshi (dismissed):

PR #1137, opened 2026-08-08
Surface and preserve the instrument model that produced a data file

trishorts:
Nothing in mzLib exposed which mass spectrometer produced a data file, and every mzML mzLib wrote silently discarded it. Both ends are fixed here.

## Read

`SourceFile` gains an init-only `InstrumentModel` (`MzLibUtil.CvParam`), so none of the 8 existing construction sites change. `Mzml` already walked the element the term lives on — `GetMsDataOneBasedScanFromConnection` kept `componentList.analyzer[0]` for the mass analyzer and discarded the rest.

Two things about mzML make this less obvious than it looks:

- **The model is usually not a direct child of `instrumentConfiguration`.** ProteoWizard, which converted essentially every vendor file anyone will read, hoists it into a `referenceableParamGroup` and points at it with `referenceableParamGroupRef`. Reading only the element's own cvParams finds nothing on any real converted file.
- **There is no closed list to match.** Instrument models are the several hundred children of `MS:1000031`. They are identified negatively instead — an `MS:` term that is named, carries no value, and is not one of the abstract parents. A model is written as a bare presence flag; the serial number beside it carries a value.

The abstract parents must be excluded or the result is worse than no answer. `MS:1000031` is the generic parent, and PSI-MS gives every vendor a branch node named `<Vendor> instrument model` — `MS:1000483` Thermo Fisher Scientific, `MS:1000122` Bruker, `MS:1000126` Waters, `MS:1000121` SCIEX, `MS:1000490` Agilent, `MS:1000489` Shimadzu, `MS:1000495` Applied Biosystems. All are valueless, so a value test alone admits every one, and **`badScan7192.mzML` in this repo's own test data carries `MS:1000483`**. Matching the naming convention rather than listing accessions covers vendors added later.

Thermo RAW records only free text (`"Orbitrap Fusion Lumos"`), so that path sets `Name` with an **empty `Accession`**. Resolving a name to an accession needs a controlled vocabulary, which is a separate concern from reading a file and would put an ontology dependency in the raw reader. Callers must check `Accession` for emptiness rather than assume a populated model is accessioned; this is documented on the property.

## Write

`MzmlMethods` carried a standing `// ToDo: read instrumentConfigurationList from mzML file` and stamped the bare parent `MS:1000031` on every file it produced, so **calibration and spectral averaging discarded the instrument as a matter of course** — invisibly, because `MS:1000031` looks like a declaration rather than an absence.

It now carries an accessioned model through, falling back to `MS:1000031` exactly as before when there is none. The accession is required, not optional: mzML types `cvParam/@accession` as mandatory, so emitting a name-only term would produce a file that does not validate — worse than the loss being fixed. A model read from a Thermo RAW therefore still does not survive a write.

`MsDataFileExtensions` rebuilds `SourceFile` field-by-field for the mzML snip path. Because `InstrumentModel` is init-only and was not copied, snipped files still lost the instrument; it is now carried explicitly. (Worth noting because "none of the 8 construction sites change" was true but misleading — that site is a *copy* constructor.)

## Known gaps, deliberate

- Bruker and timsTOF do not populate `InstrumentModel`.
- One configuration is still emitted per mass analyzer, so a file genuinely declaring two different instruments is rewritten with the default configuration's model on both.

## Tests

9 in `TestInstrumentModel` and `TestInstrumentModelRoundTrip`, over three real ProteoWizard-converted files (Orbitrap Fusion Lumos, Orbitrap Fusion, LTQ Orbitrap Velos), the serial-number confusion, the bare parent term, the Thermo vendor branch node, and survival of a write.

Full mzLib suite green.


Alexander-Sol (dismissed):
Read/write/round-trip is correct and genuinely exercised — I confirmed the test files actually contain the `referenceableParamGroup` model, the vendor-node case (`badScan7192.mzML` → MS:1000483), and the bare-parent fallback, and the copy path covers the one SourceFile-rebuild site.

The design point worth acting on: vendor branch-node rejection is done by **name** (`cv.name.EndsWith("instrument model")`), not by accession. `NonInstrumentModelAccessions` doesn't contain the seven vendor parents (MS:1000483, MS:1000122, …) the description lists — they're excluded only by the name check. That contradicts the "match on Accession, never on Name" rule the PR documents on `CvParam` and `SourceFile.InstrumentModel`, and the doc comment on the set ("caught by the value check") is wrong for exactly these terms, since they're valueless. Those seven are a closed, stable set — adding them to `NonInstrumentModelAccessions` is strictly more robust and makes the code match its own stated rule.

Relatedly, "first qualifying term wins" is only safe because non-model terms carry values; a valueless named non-model MS term on `instrumentConfiguration` or in the referenced group would be misread. Unlikely on real files (ionization/analyzer terms live in `componentList`, which `GetInstrumentModel` doesn't scan), so this is a fragility note, not a bug.

Test gaps: the Thermo-RAW empty-`Accession` path and the writer's `haveAccessionedModel == false` fallback are untested — `UnknownInstrument_StillWritesTheParentTerm` uses a *null* model, not a name-only one, so the Thermo-origin fallback-to-MS:1000031 that the comments emphasize is unverified. The snip copy-constructor fix (`MsDataFileExtensions.cs:176`) also has no round-trip test.

Heads-up for whoever merges this against the lazy-deserialize work: `GetInstrumentModel` assumes `_mzMLConnection` is fully materialized, which is true on this base but not on that branch. Nits: `haveAccessionedModel` is recomputed each loop iteration though it's loop-invariant; `SourceFile.cs` has no trailing newline.

zhuoxinshi (approved):

PR #1138, opened 2026-08-08
Read, write, validate and pool SDRF-Proteomics experimental-design files

trishorts:
Adds `Readers\Sdrf\` — a document model, reader/writer, structural validator, a collection type that pools documents, and a lint that reports concepts annotated inconsistently **across** documents. Registered as `SupportedFileType.Sdrf` (`.sdrf.tsv`).

SDRF-Proteomics is the HUPO-PSI standard for the sample-to-data-file relationship of an experiment (v1.1.0). mzLib exposes `IExperimentalDesign` with no production implementation; this is the format that will supply one in follow-up work.

> **Note on scope:** nothing in mzLib consumes this yet. The builder that populates it from search parameters, and the MetaMorpheus call site that invokes it, are the next PR. Reviewing this as a standalone format layer is intended — it is self-contained, and the corpus tests below exercise it far harder than a caller would.

## Why the reader is hand-rolled

This departs from every other reader here, and the reason is that the house pattern cannot express the format. `[Name(...)]` attributes map **fixed** columns onto a record; SDRF's column set is open-ended, its names are data, and names repeat within one file — 649 of the 1,236 curated corpus files repeat `comment[modification parameters]`, up to 8 times. CsvHelper's dynamic mode is out for the same reason: its record is keyed by header name, so 8 identical keys collapse to one. (A raw `CsvParser` in `CsvMode.NoEscape` would give positional `string[]` and is a fair alternative; ~50 dependency-free lines was judged simpler.) The nearest existing precedent for dynamic headers is `SpectrumMatchFromTsv`.

## Four properties measured, not assumed

These were measured across the 1,236-file curated corpus at `bigbio/sdrf-annotated-datasets` rather than taken from the specification, **because the two disagree**:

| Property | What the corpus shows |
|---|---|
| **No quoting or escaping** | Two files carry literal double quotes *inside* cell values; a CSV parser with default quoting strips them and the file no longer round-trips. Splitting on tab with no escape handling is both simpler and more faithful. |
| **Line endings are preserved, not chosen** | The document remembers what it read and reproduces it; in-memory documents use CRLF. Pinning either ending makes the writer normalise the one thing this class exists not to normalise. |
| **Rows may be short** | `PXD059974` has a 46-column header with 17 of its 22 data rows carrying 42 cells. Preserved, not padded. |
| **Header case is not normalised** | The spec says lowercase and case-sensitive; the corpus contains `comment[MS min charge]` and `Material Type`. |

Writing rejects a cell **or column name** containing a tab or newline, which the format cannot represent, and validates everything before opening the stream so a refusal cannot truncate the file it declined to write.

## Cells are stored raw

The `NT=`/`AC=` grammar is decoded only on demand, via `SdrfCell`. That is a correctness requirement, not laziness: `comment[file uri]` cells carry pre-signed URLs whose query strings contain `Signature=` and `Expires=`, so a parser triggered by cell *shape* decodes them as CV descriptors. Detection keys off a leading known key instead. It also makes lossless round-tripping true by construction rather than by careful re-serialisation.

## Validator

Kept out of `LoadResults` so a malformed-but-readable file still parses — a reader that repaired what it read could not round-trip, and silently rewriting a curated annotation is worse than reporting it.

Findings carry a severity, and **severities are calibrated against the corpus, not the spec**, because a rule that condemns most of the community reference set is a wrong rule. Required columns are the **thirteen present in every one of the 1,236 files**, so requiring them condemns nothing: files with at least one error stay at exactly 1 (`PXD059974`, genuinely ragged) whether Required holds three entries or thirteen. The three columns that really are sometimes absent — disease, cell type, modification parameters — are Recommended warnings. `CorpusCalibration` pins the rates so a miscategorised rule fails there rather than quietly condemning the reference set.

## Collection and drift lint

These answer the question no single-file validator can: **every document can be individually valid and the pooled table still be unusable**, because validity is a property of one file and comparability is a property of the relationship between files.

Over the curated corpus the lint reports **1,491 inconsistencies among documents that all pass validation** — including `MS:1001251` written `Trypsin` in 642 documents and `trypsin` in 382, and HCD spanning three vocabularies. Its majority report *describes* what a corpus does; it is not advice about what to write.

`SdrfCoverage` measures the opposite axis — how much a corpus actually *says*. Reserved words count as unfilled, because `SdrfValidator` correctly treats `not available` as a valid absence and `SdrfDriftLint` correctly skips it, so a corpus that is 90% empty passes both and reads as flawless.

## Tests

60 across `TestSdrf`, `TestSdrfValidator` and `TestSdrfCollection`, each pinning a measured corpus property or a specific failure mode — including regressions for a zero-row in-place write, a lone carriage return, a tab in a header, and a write guard that must not truncate the file it refuses to write.

Three corpus-wide harnesses (round-trip, validator calibration, drift report) are `[Explicit]` and read `MZLIB_SDRF_CORPUS`, so **CI never needs the 1,236-file corpus**. Locally: 1,236/1,236 round-trip byte-identical.

`.gitattributes` gains `*.sdrf.tsv -text` so the committed fixtures stay byte-exact — without it, `* text=auto` stores them LF and hands them back CRLF on a Windows checkout, and the round-trip test would be validating git's conversion rather than the reader.

Full mzLib suite green.

---

# Update — review round (2026-08-14)

Responding to @acesnik's five inline comments and @Alexander-Sol's request to try an SDRF the reader
has never seen. Two commits: `0cc05c88` and `bcca3241`. Numbers in the sections above were measured
before this round; where one has moved, it is corrected here rather than edited above, so the figures
the review was conducted against stay legible.

## One real defect: `"N"` was not a key

`SdrfCell.KnownLeadingKeys` contained `"N"`, which the specification's key grammar does not define —
`NT` is the only name key. Measured across the corpus: **45 cells in 2 files** (27 in PXD039582,
18 in PXD039585), every one of them `N=Orbitrap` in `comment[ms2 analyzer type]`. One submitter's
typo for `NT=`, accepted as grammar.

The comment defending it cited corpus frequency, which is the failure mode this project has already
had to reverse once at the column level. Accepting `N` also **hid the drift from `SdrfDriftLint`**,
which is the one component whose job is to report exactly that. `TryParseTerm` carried a matching
`"N"` fallback for the name key; that is gone too.

**Corpus drift findings move `1,491` → `1,492`** — the figure quoted above is now the *before*. The
extra finding is those cells arriving where they belonged all along. Both ends measured, not inferred.

`SN` was raised in the same comment and is **not** the same case: it is specified for
`characteristics[pooled sample]` (`SN=sample1;SN=sample2`, `sdrf-proteomics/README.adoc` lines 352
and 362, `TERMS.tsv`), and all 383 leading `SN` cells in the corpus match that shape. It stays.

## `KnownKeys` is now derived, and pinned as a superset

It was dead code that had drifted from the grammar it appeared to describe. It is now built from
`SpecKeys` (the 16 the specification defines) plus `ObservedNonSpecKeys` (`MM` 26,828 occurrences,
`CS` 3,810, `CF` 2,602 — widespread, unspecified, tolerated on **read only**), with a test pinning
`KnownLeadingKeys ⊆ KnownKeys`. That test would have caught `"N"` unaided. `PMID` removed (the spec
uses it as a value format, never as a key); `PS` added (specified, unused in the corpus so far).

## Validator: the message, not the severity

A mis-cased required column reported as flatly *missing* while `ColumnNameCase` separately warned
about its casing — the validator knew and did not say. Severity is unchanged, because the spec makes
lowercase a MUST, but the message now reads:

> Required column 'comment[data file]' is missing, but 'Comment[data file]' is present and differs
> only in casing; column names are case-sensitive and must be lowercase.

The case policy itself is now written on the class. It is deliberately **not** uniform: existence is
ordinal (spec-faithful), ordering is `OrdinalIgnoreCase` because an ordinal `BlockRank` cannot rank
`Factor Value[...]` at all — so the file most likely to be disordered would be the one the ordering
rule silently stops checking.

## Public surface: eight types down to four

`SdrfCell`, `SdrfValidator`, `SdrfValidation*` and `SdrfDriftLint` are now `internal`
(`Readers.csproj` already grants `InternalsVisibleTo("Test")`). `SdrfDocument`, `SdrfHeader`,
`SdrfRow` and `SdrfCollection` stay public because MetaMorpheus calls all four on the builder branch
and `SdrfBuilder.Build` returns `SdrfDocument` — `InternalsVisibleTo` cannot reach another product,
so making those internal would block the consumer rather than defer the commitment.

## Verification against a corpus that has moved

@acesnik noted, correctly, that no corpus-derived number in this PR had been independently checked.
`CorpusCalibration` was re-run against a current clone; it passes. Every figure is exact except one:

| Claim | Measured | |
|---|---|---|
| 1,236 files | 1,236 | exact |
| thirteen universal columns | 13, and exactly `RequiredColumns` | exact |
| `AC` leads 190,369 cells | 190,369, in 323 files | exact |
| `CT` leads 389 / `SN` leads 383 | 389 in 6 files / 383 in 4 files | exact |
| `NT` leads 1,459,772 | **1,459,508**, in all 1,236 files | corrected in code |

The upstream corpus has since moved **31 commits** past the revision those numbers were taken at —
**46 files added, 996 modified in place**. Re-run against that current state (`ba3793a`) at true LF
bytes: **1,282/1,282 round-trip byte-identically, 0 threw**, and validation still holds the
calibration thresholds (88.1% entirely clean, 1 file with an error — `PXD059974`, the known ragged
one). The thresholds were not overfit to the snapshot they were measured on.

Separately, an SDRF from outside the corpus entirely — `examples/PXD019515/PXD019515Hela.sdrf.tsv`
in `bigbio/proteomics-sample-metadata` — round-trips byte-identically as found, re-encoded CRLF,
with the trailing newline stripped, and through the `FileReader` factory, validating clean.

**One input shape does not round-trip**, recorded rather than buried: a file ending in a **bare CR**
has that CR completed to CRLF. It affects no well-formed file and nothing in the corpus.

## `RoundTrip_EntireCorpus` now reports what it measured (`bcca3241`)

Both ways a green run there can mean nothing have now actually happened, so the summary line carries
the corpus path, file count, line-ending distribution and BOM count, and the assertion message
repeats them:

- Pointing `MZLIB_SDRF_CORPUS` at the clone **root** rather than `datasets/` also picks up
  `sandbox/` — ~1,531 files against ~1,282 — so every corpus number quoted here comes out different
  and reads as invented.
- The corpus is **LF upstream with no `.gitattributes`**, so a clone made with `core.autocrlf=true`
  has an entirely CRLF working tree. The round trip still passes, because the writer reproduces
  whatever it read — but it did not test the upstream bytes, and reporting it as though it had is
  precisely the mistake that once made this reader pin CRLF on a bad survey. Nothing is asserted
  about endings; the reader of the result can simply see which bytes were measured.

Full suite green: **5,541 passed / 0 failed / 29 skipped**.


Alexander-Sol (dismissed):
LGTM, would like to test a SCP SDRF file to verify that works (ex here: https://github.com/bigbio/proteomics-sample-metadata/tree/dev/examples/PXD019515), but I don't think SCP is different enough that it would break where the other three tests pass

mzLib/Readers/Sdrf/SdrfCell.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// CT (389), SN (383), and N. Everything else in <see cref="KnownKeys"/> only ever appears
+        /// after a semicolon.
+        ///
+        /// "A" was removed from the set entirely. It is a single character, it never leads a real
+        /// cell, and as a leading key it can only ever produce false positives -- any free-text
+        /// value beginning "A=" would have been decoded as a controlled-vocabulary term.
+        /// </summary>
+        private static readonly HashSet<string> KnownLeadingKeys = new(StringComparer.OrdinalIgnoreCase)
+        {
+            "NT", "AC", "CT", "SN", "N", "MT", "TA", "PP", "CS", "CF", "CL", "CV", "MM", "VV"
```
acesnik: **`N` is not a key in the SDRF grammar.**

The specification's key list is `NT`, `AC`, `TA`, `MT`, `PP`, `CL`, `MH`/`ML`, `CV`, `CT`, `QY`, `PS`, `CN`, `SP`, `VV`. There is no `N`, and `NT` is the only name-related key.

So a cell of free text beginning `N=` is decoded as a controlled-vocabulary term on the strength of a key the format does not define — which is verbatim the argument the comment just above uses to remove `"A"`:

> It is a single character, it never leads a real cell, and as a leading key it can only ever produce false positives.

Every word of that applies to `N`, and the spec settles it. What made me look was the counts: NT 1,459,772, AC 190,369, CT 389, SN 383, and then "and N" with no number. If `N` genuinely does lead cells in the corpus, that is drift being decoded as valid grammar, and `SdrfDriftLint` is the right place to notice it rather than `SdrfCell` quietly accepting it.

Worth checking `SN` at the same time — it leads 383 corpus cells but I do not see it in the spec's key list either, so it may be in the same category.

mzLib/Readers/Sdrf/SdrfCell.cs
```diff
@@ -0,0 +1,178 @@
...
+    public static class SdrfCell
+    {
+        /// <summary>
+        /// Keys observed across the whole 1,236-file curated corpus, most frequent first:
+        /// NT name, AC accession, MT modification type, TA target amino acid, VV version value,
+        /// PP position, MM monoisotopic mass, SN specificity name, CS cleavage site,
+        /// CF chemical formula, CT compound type, QY quantity, SP species, CN common name,
+        /// ML/MH mass low/high, A generic attribute.
+        /// </summary>
+        private static readonly HashSet<string> KnownKeys = new(StringComparer.OrdinalIgnoreCase)
```
acesnik: **Dead, and already drifting from the grammar it appears to describe.**

Nothing reads `KnownKeys` — `IsTerm` uses `KnownLeadingKeys`, and the only other mention is the `<see cref="KnownKeys"/>` in the comment below. It also does not match the spec: it carries `PMID`, which is not in the key list, and omits `PS`.

So it reads like a whitelist that validates keys, validates nothing, and has drifted. Either delete it and reword that cross-reference, or make it the single source of truth that `KnownLeadingKeys` is a subset of — the second would have caught the `N` question on its own.

mzLib/Readers/Sdrf/SdrfValidator.cs
```diff
@@ -0,0 +1,368 @@
...
+            }
+        }
+
+        /// <summary>
+        /// The specification's block order. Compared case-INSENSITIVELY on purpose: the corpus
+        /// contains "Factor Value[organism part]", and an ordinal test silently disarmed the whole
+        /// rule on those files while a mixed-case pair elsewhere produced a false positive. Casing
+        /// is reported separately by ColumnNameCase, which is where it belongs.
+        /// </summary>
+        private static int? BlockRank(string columnName)
```
acesnik, line 273: **Strictness is not uniform across header comparisons.**

`BlockRank` compares `OrdinalIgnoreCase`, so it ranks `Comment[data file]` correctly. But the existence checks in `ValidateHeader` go through `SdrfHeader.Contains`, which is ordinal — eight `OrdinalIgnoreCase` comparisons in this file sitting alongside ordinal ones on the same header strings.

Both are defensible on their own: the spec makes column names case-sensitive and lowercase, so ordinal is faithful, and being lenient about *ordering* is a reasonable kindness. Worth a sentence saying which is deliberate, because the next reader will assume one policy applies throughout.

mzLib/Readers/Sdrf/SdrfValidator.cs
```diff
@@ -0,0 +1,368 @@
...
+                messages.Add(new SdrfValidationMessage(
+                    SdrfValidationSeverity.Error, "EmptyHeader", "The document has no columns."));
+                return;
+            }
+
+            foreach (var required in RequiredColumns)
+            {
+                if (!header.Contains(required))
+                    messages.Add(new SdrfValidationMessage(
+                        SdrfValidationSeverity.Error, "RequiredColumn",
```
acesnik, line 167: **A mis-cased required column reports as missing.**

`Comment[data file]` fails `header.Contains("comment[data file]")` because that comparison is ordinal, so this fires as an Error saying the column is missing — while `ColumnNameCase` separately warns about its casing, which is to say it knows the column is there.

The severity is right: the spec makes lowercase a MUST and names `Characteristics[organism]` as a failure, so the file is non-conformant. It is the message that misleads. A curator reading "missing comment[data file]" on a file that visibly contains it will not go looking for capitalisation.

Naming the case explicitly when the column is present under a different casing would make it self-diagnosing.

mzLib/Readers/Sdrf/SdrfHeader.cs
```diff
@@ -0,0 +1,67 @@
...
+                if (string.Equals(_names[i], columnName, StringComparison.Ordinal))
+                    found.Add(i);
+            }
+            return found;
+        }
+
+        /// <summary>
+        /// The index of the first column with this name, or -1 if absent.
+        /// </summary>
+        public int IndexOf(string columnName) => _names.FindIndex(n => string.Equals(n, columnName, StringComparison.Ordinal));
```
acesnik, line 57: Checked this against the spec, since it is the assumption the validator's existence rules ride on, and it holds: column names are case-sensitive **and** must be lowercase, with `Source Name` and `Characteristics[organism]` given as explicit failures. Ordinal is faithful and the comment is accurate.

Recording it because I had this written up as the headline problem — that the reader deliberately preserves `Material Type` and `comment[MS min charge]` while the lookup will not find them — and was going to ask for a case-insensitive lookup. That would have pushed this away from the spec. Leaving it as is.

What remains is downstream rather than here: `null` from `SdrfRow`'s indexer conflates "absent" from "present but non-conformant", which the builder in the next PR will want to tell apart to say anything useful to a curator.

acesnik (changes requested):
🤖 Careful work, and the measured-not-assumed approach is the best thing about it — calibrating severities against the reference set rather than the specification, and catching that occurrence counts are not file counts, are the parts I would have gotten wrong.

I reviewed this in three passes: the format layer, then the validator and drift lint, then the four spec-shaped claims against the SDRF-Proteomics specification itself rather than taking them from the description. **That last pass moved my emphasis**, so the inline comments are lighter on what I expected to raise and heavier on one thing I nearly waved through.

**One item I would fix before merge:** `KnownLeadingKeys` contains `"N"`, which the specification's key grammar does not define. Details inline, along with the dead `KnownKeys` set, the non-uniform header strictness in the validator, and a misleading `RequiredColumn` message.

**What I checked and found correct:** the four reserved words are the complete set; `NT` before `AC` is a spec MUST rather than a convention, which is a stronger argument than the comment makes; `BlockRank`'s ordering matches the spec's block sequence; and ordinal header comparison is spec-faithful — the spec makes column names case-sensitive *and* lowercase, giving `Source Name` and `Characteristics[organism]` as explicit failures.

**One decision worth making deliberately rather than by default.** Around 1,600 lines and eight public types with nothing in mzLib calling them. Staging it this way is genuinely easier to review, and the corpus harnesses exercise it harder than a first caller would. But mzLib ships as NuGet, so these become a compatibility commitment on the next release, before the builder that motivates them has put any pressure on their shape. `internal` plus `InternalsVisibleTo("Test")` until the consumer lands would keep the freedom to change them. No strong feeling either way.

**Also good:** validating every cell and column name before `File.Create` truncates, with the comment recording that the guard used to destroy the file it refused to write. `UTF8Encoding(false)` for the BOM and `.gitattributes -text` so the round-trip test measures the reader rather than git's line-ending conversion. Ragged rows bounds-checked in both `SdrfRow` accessors, which is what I most expected to be wrong. `LineNumber => RowIndex + 2` with the reasoning written down. `SdrfDriftLint` staying ordinal throughout is correct and should stay — it exists to find case variants, so a case-insensitive comparison would delete its purpose.

The branch is behind master and will need it merged.

**Limit of this review:** I verified the spec claims, but no corpus-derived number — the 1,236 files, the thirteen universal columns, the 1,491 drift findings, the per-key counts. Those are load-bearing for the severity calibration, so someone running `CorpusCalibration` against a current clone is worth more than my reading of it, particularly for the leading-key counts where the spec and the corpus appear to disagree.

mzLib/Readers/Sdrf/SdrfCell.cs
trishorts: Fixed, and the missing number was the right thing to pull on.

Measured against the corpus: `N` leads **45 cells in 2 files** — 27 in PXD039582 and 18 in PXD039585 — and every one of them is `N=Orbitrap` in `comment[ms2 analyzer type]`. One submitter's typo for `NT=`, in two files, decoded as grammar.

`N` is gone from both sets, and so is the other half of it that isn't visible from this hunk: `TryParseTerm` carried an explicit `pairs.TryGetValue("N", out name)` fallback treating it as an alternative spelling of the name key. Removing it from the key sets alone would have left that special case sitting there.

Your point about where it belongs is now measurable rather than arguable. Drift findings over the corpus go **1,491 → 1,492**: the `N=Orbitrap` cells leave the term index, land in free text, and `SdrfDriftLint` reports them as the `MixedTermAndFreeText` they always were. Which is exactly what you said should happen to it.

**On `SN` — I checked, and it is specified, so it stays.** It is absent from the modification-style key list, which is what makes it look like the same case, but the spec defines it for `characteristics[pooled sample]`: `SN=sample1;SN=sample2`, "Lists source names of pooled samples when known" (`sdrf-proteomics/README.adoc` line 352, worked example at 362, and `TERMS.tsv` "pooled sample"). All 383 leading `SN` cells in the corpus are that exact shape in that exact column. `N` had no such defence; `SN` does.

mzLib/Readers/Sdrf/SdrfCell.cs
trishorts: Took your second option, since you're right that it would have caught `N` on its own — and it is now the test that does:

```csharp
Assert.That(SdrfCell.KnownLeadingKeys, Is.SubsetOf(SdrfCell.KnownKeys));
```

`KnownKeys` is no longer a hand-maintained list. It is derived from two named sets:

- `SpecKeys` — the 16 the specification defines: NT AC MT TA PP VV CT QY PS SP CN CV CL MH ML SN.
- `ObservedNonSpecKeys` — MM (26,828 occurrences), CS (3,810), CF (2,602). Widespread, unambiguous in context, and *not* in the spec.

Kept as two sets rather than one so that "the format says so" and "the community writes it anyway" stay distinguishable. Anything in the second set is a read-side tolerance and must never reach authored output — which is the same line D13 draws between the writer and the pooling layer.

`PMID` removed: the spec uses PMID only as a **value** format for reference columns ("URL, DOI, or PMID"), never as a key, and the 3 corpus occurrences are of that kind. `PS` added — specified for spiked-compound cells (`CT=peptide;PS=PEPTIDESEQ;QY=10 fmol`), though no corpus file uses it yet, which is precisely the sort of gap a spec-derived set should carry and an observation-derived one cannot.

`SdrfCell` as a whole is `internal` now — see the summary comment on the staging question.

mzLib/Readers/Sdrf/SdrfValidator.cs
trishorts: Agreed that it needed saying out loud rather than being inferred. Added a CASE POLICY block to the class doc recording that the split is deliberate and what each side is for.

The part that only becomes obvious when you try to make it uniform: an ordinal `BlockRank` cannot rank `Factor Value[organism part]` at all, so it returns null, the column is skipped — and the file most likely to have disordered columns becomes the one file the ordering rule stops checking. Leniency there isn't a kindness to the author, it's what keeps the rule alive on exactly the files it exists for.

Existence stays ordinal for the reason you give: the spec makes names case-sensitive *and* lowercase, so a mis-cased column genuinely is not that column, for this validator and for anyone joining a corpus on column names.

mzLib/Readers/Sdrf/SdrfValidator.cs
trishorts: Fixed, severity unchanged. `CasingHint` appends, when the column turns out to be present under another casing:

> Required column 'comment[data file]' is missing, but 'Comment[data file]' is present and differs only in casing; column names are case-sensitive and must be lowercase.

Applied to `RecommendedColumn` as well — it is the same helper and the same defect, and it would have read oddly for one to self-diagnose while the other kept saying "absent" about a column you can see.

Two tests: one that the hint fires and names the column that *is* there, one that a genuinely missing column gets no hint and the message stays clean.

mzLib/Readers/Sdrf/SdrfHeader.cs
trishorts: Noted, and thank you for recording the reasoning rather than deleting the comment once it checked out. "I had this written up as the headline problem and the spec talked me out of it" is worth more to whoever reads this next than a silent approval would have been — it stops the same case-insensitive lookup being proposed again in six months.

No change here.

Your downstream point is the one that survives, and I've carried it into the builder PR: `null` from `SdrfRow`'s indexer conflates "column absent" with "column present but non-conformant", and the builder is exactly the caller that has to tell a curator which of the two happened. It needs a distinguishable result there, not a nullable string.

trishorts:
@acesnik — thanks. Checking the spec claims against the spec rather than against my description is what made this useful, and it moved the outcome: the item you nearly waved through was the only real defect in the four.

All four inline items are addressed in `0cc05c8`; details in the threads. One correction, one decision, and your stated limit closed.

### The correction: `SN` stays

You flagged it alongside `N`, and it looks like the same case for the same reason — absent from the modification-style key list. But the spec does define it, for `characteristics[pooled sample]`: `SN=sample1;SN=sample2`, "Lists source names of pooled samples when known" (`README.adoc:352`, example at `:362`, `TERMS.tsv` "pooled sample"). All 383 leading `SN` cells in the corpus are that exact shape in that exact column.

`N` had no such defence and is gone — along with the `TryParseTerm` fallback that also treated it as a name key, which the hunk you commented on didn't show.

### The decision: `internal`, as far as it actually goes

Four of the eight are now `internal` — `SdrfCell`, `SdrfValidator`, `SdrfValidation*`, `SdrfDriftLint`. Nothing outside the assembly names them, so they cost nothing to hold back, and you're right that shipping them commits us before the builder has put any pressure on their shape.

Four cannot be, and it is worth recording why rather than looking like half-hearted compliance. `SdrfDocument`, `SdrfHeader`, `SdrfRow` and `SdrfCollection` are already called from **MetaMorpheus** on the builder branch — a separate assembly consuming mzLib via NuGet, which `InternalsVisibleTo("Test")` cannot reach. `SdrfBuilder.Build` returns `SdrfDocument`, so it and its public surface have to stay reachable. Making those internal wouldn't defer the compatibility question, it would block the consumer that motivates it.

Net: the NuGet surface this PR commits to goes from eight types to four, and the four remaining each have a named consumer already written against them.

One honest cost, since it bit me while checking this: the round-trip harness I ran for @Alexander-Sol's file called `SdrfValidator.Validate` from an outside assembly, and no longer compiles. Validating an arbitrary file is now only reproducible from inside this repo's test project — which is where I re-ran it, so the numbers in that comment are against this commit and not the pre-review code.

### Your stated limit: I ran `CorpusCalibration`, and re-measured every number you flagged

It passes against a current clone. Results at the pinned corpus revision, `datasets/` only:

| Claim in the PR | Measured | |
|---|---|---|
| 1,236 files | 1,236 | exact |
| thirteen universal columns | 13, and exactly the `RequiredColumns` list | exact |
| AC leads 190,369 cells | 190,369, in 323 files | exact |
| CT leads 389 | 389, in 6 files | exact |
| SN leads 383 | 383, in 4 files | exact |
| NT leads 1,459,772 | **1,459,508**, in all 1,236 files | off by 264 — comment corrected |
| 1,491 drift findings | 1,491 before this change, **1,492** after | expected: the `N` fix hands one to the lint |

You were right to single out the leading-key counts. That is where the one wrong number was, and where the one absent number was hiding a real defect.

One trap for anyone reproducing this: the corpus repo carries a `sandbox/` directory as well as `datasets/`, and recursing the repo root yields **1,531** `.sdrf.tsv` files rather than 1,236. `CorpusCalibration` uses `SearchOption.AllDirectories`, so `MZLIB_SDRF_CORPUS` must point at `datasets/`. Point it at the root and every count above comes out different, which would reasonably read as the numbers having been invented.

Full suite green (5,541 passed / 0 failed / 29 skipped). The branch is not behind master — the merge at `94ca30d4` brought it current; 6 ahead, 0 behind as of now.

trishorts:
@Alexander-Sol — ran the SCP file you linked, since it's a fair thing to want before merge.

`examples/PXD019515/PXD019515Hela.sdrf.tsv` from bigbio/proteomics-sample-metadata @ dev — 4,606 bytes, 29 columns, 6 rows — **round-trips byte-identically**, SHA-256 in equal to SHA-256 out, through both `new SdrfDocument(path)` and the `FileReader.ReadFile<SdrfDocument>` factory. Validator: 0 errors, 0 warnings.

Re-encoded it three more ways to check the identity isn't luck: as CRLF, and with the trailing newline stripped — both byte-identical — and written twice, identical each time.

PXD019515 turns out to carry three different annotations under that one accession: yours, plus two 12-row files in the curated corpus (43 and 46 columns, differing from each other as well as from yours). All three round-trip on all four variants, 12 for 12. Yours is the more meaningful test of the set, since it is the one that isn't in the corpus this reader was developed against.

Two things it surfaced that are worth having on the record:

- It is a harder test than the three fixtures, mainly because `comment[modification parameters]` repeats three times and a single row mixes `AC=Unimod:35` with `AC=UNIMOD:4`. Both preserved exactly, which is the whole point — a reader that tidied either could not round-trip.
- Every row carries `AC=MS:1001313;NT=Trypsin`, but `MS:1001313` is *Trypsin/P* in psi-ms.obo; plain Trypsin is `MS:1001251`. The name and the accession in that cell disagree. That is a defect in the annotation rather than in this PR, and it passes validation cleanly because structural validation does not check whether an accession means what its name claims — that arrives with the CV resolver. Same defect class as the `proteases.tsv` trypsin row being fixed separately.

One edge case found while building the harness, reported because it is the only input shape I could find that changes under a round trip: a file ending in a **bare CR** has that CR completed to CRLF (4,612 bytes in, 4,613 out). It affects no well-formed file and nothing in the corpus — my first harness manufactured it by truncating a CRLF file by one byte, and produced a false failure that way.

acesnik:
👨🏼‍🚀  I woke up with the question, should this be somewhere more in HUPO-PSI's domain like TopDownProteomics so can be reused across labs that are implementing SDRF?

trishorts:
> 👨🏼‍🚀 I woke up with the question, should this be somewhere more in HUPO-PSI's domain like TopDownProteomics so can be reused across labs that are implementing SDRF?

bro. give me a check mark

trishorts:
> 👨🏼‍🚀 I woke up with the question, should this be somewhere more in HUPO-PSI's domain like TopDownProteomics so can be reused across labs that are implementing SDRF?

based loosely on https://github.com/NCEMS/HAMLET

Please note that all of this will be available as individual modules in python, rust and R. So anyone can use it anywhere dependency free. for example: https://github.com/smith-chem-wisc/pyMzLib/

trishorts:
> 👨🏼‍🚀 I woke up with the question, should this be somewhere more in HUPO-PSI's domain like TopDownProteomics so can be reused across labs that are implementing SDRF?

also i can make a pr of this to the topdownproteomics SDK since that is in C# and they can have it if they like.

acesnik (approved):
🤖 Approving. Every item is addressed, and I checked them in the code rather than in the replies.

| requested | state at `bcca3241` |
|---|---|
| `"N"` is not in the SDRF key grammar | gone from both sets, and the `TryParseTerm` fallback that also treated it as a name key is gone with it. `"N"` now survives only in comments explaining why it left |
| dead, drifting `KnownKeys` | derived rather than hand-maintained: `SpecKeys` (16, matching the specification's list) ∪ `ObservedNonSpecKeys` (`MM`, `CS`, `CF`). `PMID` dropped, `PS` added |
| the invariant that would have caught it | `LeadingKeysAreASubsetOfKnownKeys`, `TestSdrfCollection.cs:312`. I checked all thirteen leading keys against the union by hand as well |
| non-uniform header strictness | `CASE POLICY` block on the class doc, recording which comparison answers which question and why not to tidy one into the other |
| misleading `RequiredColumn` message | `CasingHint` at `SdrfValidator.cs:140`, applied at both 168 and 178 |
| corpus numbers unverified (my stated limit) | re-measured; `NT` corrected 1,459,772 → 1,459,508 and the comment updated to match |

**You were right about `SN` and I was wrong.** I flagged it beside `N` on the same reasoning — absent from the modification-style key list — and the specification does define it for `characteristics[pooled sample]`. The distinction you drew is the correct one: `N` had no defence, `SN` has one.

The `N` finding is also better closed than I asked for. Removing it from the key sets was the visible half; the `TryParseTerm` special case was the half that would have kept the behaviour alive after the sets looked clean. And 1,491 → 1,492 drift findings is the right kind of evidence — the cells did not vanish, they moved to the layer that should have been reporting them all along.

Two numbers earned their keep: the absent count next to `N` marked the one real defect, and the one wrong count was in the same family. Documenting the `sandbox/` versus `datasets/` trap matters too — 1,531 files against 1,236 would read as invented numbers to the next person reproducing it.

## One note, non-blocking, and it is mine

Validation no longer has a public entry point. `SdrfValidator`, `SdrfValidationSeverity`, `SdrfValidationMessage` and `SdrfValidationResult` are all `internal`, and neither `SdrfDocument` nor `SdrfCollection` exposes a `Validate` method — so a change titled "read, write, validate and pool" ships validation that only this repo's test project can call. The harness that stopped compiling was the symptom; it reads as a testing inconvenience but it is the feature having no reachable door.

That is a consequence of my suggestion, applied faithfully, and I do not think it should block: `internal` → `public` is additive while the reverse breaks consumers, so holding it back is the direction that keeps the choice open. The thing to carry into the builder PR is a public `Validate()` on `SdrfDocument`/`SdrfCollection` returning public result types, rather than exposing the static validator — which also gives the natural home for distinguishing "column absent" from "column present but non-conformant", the point that survived the `SdrfHeader` thread.

Splitting the surface eight types → four, with a named consumer already written against each of the four, is a better answer than either extreme.

## Also

@Alexander-Sol's approval was dismissed by the new commits, though the SCP file he asked for now round-trips byte-identically across four encodings — worth a re-approve from him rather than my check mark standing in for it.

The `MS:1001313` versus `MS:1001251` disagreement in that file is a good catch and correctly scoped out: structural validation cannot know whether an accession means what its name claims, and the CV resolver is where that belongs.

On where this should live — HUPO-PSI, TopDownProteomics, or here: that question is orthogonal to merging. Donating it upstream later is a copy either way, and nothing about landing it here forecloses that.

CI is green on all five checks, and the branch is current with master.

nbollis (approved):

PR #1139, opened 2026-08-08
Pinned PSI-MS and PRIDE controlled vocabularies

trishorts:
Adds `ControlledVocabulary` — embedded, version-pinned PSI-MS (4,114 terms) and PRIDE (981 terms) with lookup by accession and by name — and `DissociationTypeCvTerms`, which maps mzLib's `DissociationType` onto PSI-MS terms and back.

> **Note on scope:** nothing in mzLib consumes this yet; it is the vocabulary layer for the SDRF work in #1138. Reviewing it standalone is intended.

## Why pinned and embedded

Resolving a term against something that moves means the same experiment annotated a year apart gets different accessions, and a corpus built that way silently stops joining to itself. `Version` is exposed so the snapshot can be recorded alongside whatever the terms end up in, and refreshing a vocabulary becomes a reviewable change to this repository rather than something that happens to users overnight.

It reuses what is already here: TopDownProteomics' `OboParser` (already a dependency, already used by `Loaders.ReadPsiModFile`), `ProteaseDictionary`'s lazy embedded-resource pattern, and `MzLibUtil.CvParam` as the term type. **No new dependency.**

The large sample ontologies are deliberately absent — NCBITaxon is ~600 MB and unnecessary, because an organism identifier comes from the search database that was already loaded (see #1136).

## Package impact

+1.4 MB uncompressed inside `UsefulProteomicsDatabases.dll`, ~226 KB compressed in the nupkg. `Omics` already ships 6.7 MB of ontologies the same way.

`.gitattributes` gains `*.obo -text`, added **before** these files were first tracked: `* text=auto` would store them LF and hand back CRLF on a Windows checkout, so what CI embeds would not be what was committed or reviewed.

## Licensing

Both vocabularies are **CC BY 4.0**, which requires attribution to reach downstream recipients — and NuGet consumers get assemblies, not this repository. Adds `THIRD-PARTY-NOTICES.md` (shipped via `mzLib.nuspec`) naming both, their publishers, sources, pinned versions and licence, and covering the pre-existing `unimod.xml` / `PSI-MOD.obo.xml` / `ptmlist.txt` embedded in `Omics`. `Resources/PROVENANCE.md` records fetch URLs, dates and the refresh procedure.

## DissociationType mapping

This table did not exist. The accessions lived only in XML doc comments on the enum, where no code could reach them. 17 values are mapped; four (`aEPD`, `Custom`, `Autodetect`, `AnyActivationType`) have no PSI-MS term and return `false` rather than a plausible neighbour.

HCD writes `MS:1000422` — the plurality in the curated corpus at 273 files against 47 and 23 — and reads all three spellings, so other people's documents still join to ours.

## Five wrong accessions were found by audit before this shipped

Recorded because the reason they survived matters more than the fix:

| Wrote | Actually is |
|---|---|
| `SID → MS:1000406` | "surface ionization" — an **ionization** type |
| `PD → MS:1000435` | mzLib's `PD` is **plasma desorption**, `MS:1000134` — as `DissociationType.cs:33` and `Mzml.cs:99` both already said |
| `MS:1000045 → CID` | "collision energy", a numeric parameter with `has_units eV` that accompanies nearly every HCD scan |
| `PRIDE:0000598 → ETD` | "Alkylation reagent". PRIDE has no ETD term at all |
| "UVPD/NETD have no term" | `MS:1003246` and `MS:1003247`; UVPD is used in the curated corpus, so refusing to write it was a joinability gap |

**Every one of those resolved cleanly against the vocabulary.** The test meant to catch them pinned each accession to the name the OBO gives it, which is circular — the ontology guarantees that link, so the assertion could only fail on a typo. It could not detect a real accession naming the wrong *concept*.

The replacement, `AgreesWithTheMzmlReadersOwnAccessionTable`, checks this table against `Readers.Mzml.DissociationDictionary` — public, in the same assembly, and right all along. Disagreeing with it means an mzML this library reads cannot be described by the terms this library writes. Read-aliases now get the same name-pinning discipline as the primary table; they had none, which is how two wrong ones survived.

The obsolete test had the same defect: it asserted things about "Q Exactive", which is not obsolete, so deleting the entire obsolete filter left it green. It now uses `minute`, where an obsolete `MS:1000038` precedes the live `UO:0000031` in file order and would win without the filter.

## Other fixes from the same audit

`CvLabel` now derives from the **accession** rather than from which file the term came out of — `psi-ms.obo` imports 128 foreign terms (`UO`, `PEFF`, `NCIT`) and `pride_cv.obo` 19, all of which were being stamped with the wrong `cvRef`.

## Known follow-up

This is the **fourth** in-repo source of PSI-MS accessions, alongside `Readers/MzML/Mzml.cs` (~25 consts + 3 dictionaries), `Readers/MzML/MzmlMethods.cs`, and `UsefulProteomicsDatabases/PrideArchiveExtensions.cs` (17 consts). Consolidating them is now *unblocked* — `PrideArchiveExtensions.cs:230` already complains that `Mzml.AnalyzerDictionary` "is not reachable from this project", and moving those dictionaries into `UsefulProteomicsDatabases` (which `Readers` already references) is exactly what resolves it. Not attempted here to keep the diff reviewable.

## Tests

53 in `TestControlledVocabulary`. Full mzLib suite green at 5,459 passed / 0 failed.


trishorts:
## Why these aren't split into an opt-in package, à la #1131

Raised before review does: #1131 splits Chronologer into an opt-in `mzLib.Chronologer` package, and "1.4 MB of vendored ontologies in the core package" looks like the same shape of problem. It isn't, and the reasoning is worth having on record so it doesn't get re-litigated.

**#1131's problem was a dependency, not a payload.** `Chromatography` carried `TorchSharp` + `libtorch-cpu`, and because 11 of 21 projects transitively depended on it, every consumer restored **1,117 MB** — including `Readers` and `FlashLFQ`, the two most widely used. The split removed a hard third-party dependency from the core package.

Neither half of that applies here:

| | #1131 Chronologer | this PR |
|---|---|---|
| Payload | 1,117 MB of libtorch | **1.4 MB** (~226 KB compressed in the nupkg) |
| New dependencies | `TorchSharp` + `libtorch-cpu` | **none** — `OboParser` (TopDownProteomics) was already a dependency of this project and is already used by `Loaders.ReadPsiModFile` |
| Projects affected | 11 of 21 bound libtorch | 1; nothing new is transitively bound |
| Genuinely optional? | Yes — one of three RT predictors | **No** — see below |

**The optionality row is the decisive one.** Chronologer is one retention-time predictor among three, so opt-in degrades gracefully: don't take the package, don't use that predictor. The vocabularies are not optional for the feature they exist to serve — the SDRF work in #1138 authors controlled-vocabulary terms unconditionally, so every write needs them. Making them a separate package means the core feature fails by default, which is precisely the failure mode #1131 flags in its own **follow-up #2**: a consumer who takes `mzLib` without the extra gets a `DllNotFoundException` wrapped in a `TypeInitializationException` on first use rather than a graceful skip. Inheriting that accidentally is a bug; designing it in deliberately would be worse.

**The in-repo precedent points the other way.** `Omics` already embeds 6.4 MB of ontology data — `unimod.xml` (2.3 MB) and `PSI-MOD.obo.xml` (4.4 MB) — as `EmbeddedResource`, shipped in the same nupkg, with no split. The 1.4 MB added here is smaller than either of those files individually.

**And splitting has real cost.** A new `.csproj` re-enters the #1127 territory — solution platform rows, `PlatformTarget` left unset rather than `AnyCPU`, nuspec paths — plus a second nuspec and changes to both `dotnet.yml` and `release.yml`. That is a lot of moving parts to save 226 KB on restore.

### Where the pattern does become right

When the **sample** ontologies arrive. Organism is currently resolved from the search database (#1136) rather than from NCBITaxon, which is why that ~600 MB file is deliberately absent here. But tissue and disease should also be CV terms, and those need UBERON and MONDO. At that point this is squarely Chronologer-shaped — large payload, genuinely separable, plausibly optional — and an `mzLib.Ontologies` opt-in package following #1131's structure is the right answer.

Proposed threshold, so the decision has a trigger rather than a vibe: **split when the vocabularies stop being small enough to embed comfortably, or when they would add a dependency the core does not already have.** Neither is true today; both would be true of NCBITaxon.

### One practical note

#1131 and this PR both touch `mzLib.nuspec` — #1131 removes four `<dependency>` lines, this adds a `<file>` entry for `THIRD-PARTY-NOTICES.md` in a different section. No semantic conflict, but whichever merges second will want a trivial rebase.

trishorts:
## This reduces accession duplication rather than adding to it

Worth stating concretely, because "another file full of accessions" is the natural first reaction.

**mzLib does not currently have these vocabularies.** What it has is two different things that can both sound like it:

1. **Per-project data from the PRIDE API.** `PrideProject.Organisms`, `.Instruments`, `.OrganismParts`, `.Diseases` are `List<CvParam>` — but they are the terms PRIDE supplied *for one fetched dataset*. Network-dependent, only for projects you have retrieved, and only the handful of terms that project happens to use. You cannot author with them, cannot resolve a name to an accession, cannot validate an accession you were handed, and cannot do any of it offline.

2. **Twenty hardcoded constants** in `PrideArchiveExtensions.cs:19-25, 161-214` — `MsLevelAccession = "MS:1000511"`, `SelectedIonMzAccession = "MS:1000744"`, `FtpLocationAccession = "PRIDE:0000469"`, and so on. A hand-picked subset for PROXI spectrum parsing.

**All twenty of those resolve from the vocabularies added here — zero misses.** Verified against the embedded `psi-ms.obo` and `pride_cv.obo`:

```
MS:1000016  scan start time          MS:1000041  charge state
MS:1000042  peak intensity           MS:1000127  centroid spectrum
MS:1000129  negative scan            MS:1000130  positive scan
... 20/20 resolvable, 0 not
```

So this PR is the first thing in the repo that could actually retire them, and the same is true of the larger tables in `Readers/MzML/Mzml.cs` (~25 consts plus `PolarityDictionary`, `AnalyzerDictionary`, `DissociationDictionary`) and `Readers/MzML/MzmlMethods.cs`.

**Not done here, deliberately.** Folding a cross-file cleanup into a PR that adds a vocabulary is unrelated scope, and touching the mzML reader's dictionaries is a change to a hot path that deserves its own review. Tracked as a follow-up to run once this and the sibling PRs have merged.

It is also the consolidation the codebase already asked for: `PrideArchiveExtensions.cs:230` carries a standing comment that `Readers.Mzml.AnalyzerDictionary` "is not reachable from this project". Moving those tables into `UsefulProteomicsDatabases` — which `Readers` already references — is exactly what resolves it, and this PR is what makes it possible.

Alexander-Sol (dismissed):
Verified the audit fixes independently against the pinned OBO bytes — all 17 forward accessions and all 5 read-aliases resolve to the right concepts, including SID→MS:1000136 where the old MS:1000406 is a *live* `surface ionization` term. Duplicate-name handling (first-writer-wins with obsolete terms kept by accession but out of the name index), the reverse map, lazy init thread-safety, and the `AgreesWithTheMzmlReadersOwnAccessionTable` oracle all check out. `Mzml.DissociationDictionary` is a genuine independent oracle, not circular.

The one coverage gap that could regress silently: **`ExactSynonyms` is effectively untested.** Every by-name test hits a primary `name:`, never a `synonym:` line, so the custom quote/scope parsing — extract quoted text, require `EXACT`, reject `RELATED`/`BROAD` — has no assertion behind it. A resolve-via-EXACT-synonym case plus a RELATED-does-not-resolve case would cover the only piece of hand-written parsing here.

Also: the oracle doesn't reach UVPD, NETD, or LowCID — the three newly-added mappings most at risk of a wrong accession are covered only by name-pinning against the same OBO, which the class doc itself calls circular. I checked them by hand and they're right; the suite doesn't. And `DissociationType.Unknown` belongs in `UnmappedDissociationTypes_ReturnFalse` (it does return false; it's just not asserted).

Nit: `<CopyToOutputDirectory>Never</CopyToOutputDirectory>` on the two `EmbeddedResource` items is a no-op — that metadata applies to `None`/`Content`, not embedded resources — so it reads as if it's doing something it isn't. Worth dropping.

pcruzparri (dismissed):

PR #1141, opened 2026-08-12
Move from .NET 8 to .NET 10

### LINKED ISSUE #1146: GetProxiSpectrumAsync has no cancellation test coverage
🤖 Filed at @trishorts's request from the review of #1141, but **narrower than the item we discussed** — I checked the code before filing and most of it is already done.

Already honoured, no action needed:

- Koina `PredictionClients/Koina/Client/HTTP.cs` threads the token through every await — `PostAsync` (:27), `ReadAsStringAsync` (:31), `ReadAsStreamAsync` (:37), `ReadToEndAsync` (:39) and `SendAsync` (:45).
- `PrideArchiveClient.DownloadFileAsync` (`UsefulProteomicsDatabases/PrideArchiveClient.cs:518`) takes a `CancellationToken`, and commit `f7d5167a` on the #1141 branch added the already-cancelled guard. It is covered by `DownloadFileAsync_PreCancelledToken_Throws_LeavesNoFile` (`Test/DatabaseTests/PrideArchiveDownloadTests.cs:432`), which also asserts no partial file is left behind.

What remains is one test gap:

**`GetProxiSpectrumAsync` (`PrideArchiveClient.cs:627`) accepts a `CancellationToken` but has no cancellation coverage.** The only `CancellationToken` in `Test/DatabaseTests/PrideProxiSpectrumTests.cs` is the mock handler's own `SendAsync` override signature, so the parameter is never exercised. Its sibling download path has the test above; the PROXI path does not.

Worth doing because it is free to run: a pre-cancelled token short-circuits before any network I/O, so the test needs no live service and belongs in the offline job rather than `Category=ExternalService`. Mirroring the existing `DownloadFileAsync_PreCancelledToken_Throws` shape should be enough.

One caveat for whoever picks this up: a pre-cancelled-token test can pass for the wrong reason. `HttpClient` throws `TaskCanceledException` for an already-cancelled token by itself, so such a test passes whether or not the *library* checks the token — it only fails if the token stops being passed through at all. If the intent is to pin the library's own guard, the test needs to observe that no request was attempted (the mock handler's invocation count) rather than only that an `OperationCanceledException` came out.

acesnik:
🤖

## Move from .NET 8 to .NET 10

.NET 8 leaves support in **November 2026**. .NET 10 is the current LTS, supported to November 2028.

All 19 projects move to `net10.0` / `net10.0-windows`. No package versions change, and nothing needed
a code change to compile.

### What changed

- The 19 project files
- `mzLib.nuspec` — both dependency groups and all 64 `<file src=…>` paths, which point into
  `bin/Release/<framework>/`
- The workflows now declare `DOTNET_VERSION` and `DOTNET_FRAMEWORK` once each, instead of repeating
  the version in four places

### Verification

On macOS/arm64 with SDK 10.0.400 the solution builds with 0 errors
(`-p:EnableWindowsTargeting=true`, since mzPlot and Test are `net10.0-windows`).

After a Release build, **every one of the 64 nuspec file paths resolves**. That is worth checking
explicitly rather than eyeballing: the nuspec is hand-maintained and a stale path there does not fail
the build, it just publishes a package quietly missing an assembly.

Not verified here: the test suite. `Test` and `mzPlot` are `net10.0-windows` because mzPlot is
WPF, so they cannot run on macOS. CI covers them.

### One real behaviour change, fixed here

`GetProjectAsync_Cancelled_ThrowsOperationCanceledException` fails on .NET 10. The PRIDE client never
checked the cancellation token itself — it left that to `HttpClient`. On .NET 8, `GetAsync` threw for
an already-cancelled token before reaching the handler, so the test passed on behaviour the client
did not implement. On .NET 10 it does not, the test's stub handler ignores the token, and the call
returns a project instead of cancelling.

That is a genuine gap rather than a test artefact: cancellation is in the documented contract
(`GetSpectrumAsync` carries an `<exception cref="OperationCanceledException">` tag), and *any* handler
that ignores the token would have behaved this way. Three methods in that file already opened with
`ThrowIfCancellationRequested`; three did not. `TryGetProjectAsync` and `GetProxiSpectrumAsync` now do
too, covering `GetProjectAsync` and `GetSpectrumAsync` since they delegate. This restores the .NET 8
behaviour rather than changing it.

Verified by compiling `PrideProjectTests` into a net10.0 harness, since the Test project is
net10.0-windows and cannot run on macOS: **34 of 34 pass with the fix, and the cancellation test fails
without it.**

### Two things for maintainers to decide

**1. This is consumer-breaking.** A `net10.0`-only package cannot be referenced by a `net8.0`
consumer, so everyone downstream has to move to .NET 10 to take any future mzLib release. Given
.NET 8 has about three months of support left that seems reasonable, and it keeps the package simple.

The alternative is multi-targeting `net8.0;net10.0`, which breaks nobody and lets consumers move at
their own pace. The cost is that the hand-maintained nuspec roughly doubles — 64 file entries become
about 128, across two `lib/` folders — and every build produces both. Happy to switch to that if you
would rather not force the move; it is a mechanical change to this same branch.

**2. Merge order matters.** mzLib's CI clones MetaMorpheus master and builds and tests it against the
freshly built package. A `net8.0` MetaMorpheus cannot reference a `net10.0` mzLib, so **the
integration job here will fail until smith-chem-wisc/MetaMorpheus#2710 merges.** That PR does not
depend on this one — a `net10.0` app consumes the existing `net8.0` mzLib 1.0.584 package fine — so
the order is simply #2710 first, then this.

If multi-targeting is chosen instead, that ordering constraint disappears entirely, since MetaMorpheus
on `net8.0` would still resolve a `net8.0` mzLib.

Closes #1146.


trishorts:
Reviewed against a worktree at `39800571`, including a full Release build, the whole test suite, and an actual `nuget pack`. **The migration is clean and mzLib is functionally green on .NET 10 — 5435/5435 tests pass.** The nuspec, which is the file I most expected to be broken, is correct in both directions. Findings below are mostly small, plus one thing outside this PR that should land first.

---

## The `integration` failure is not the TFM — it's `global.json`, and it's about to get worse

Worth being precise about, because the fix is not what it looks like. The job never reaches `dotnet add package`; it dies at SDK resolution:

```
Requested SDK version: 8.0.204
global.json file: D:\a\mzLib\mzLib\MetaMorpheus\global.json
Installed SDKs:
Install the [8.0.204] .NET SDK or update [...] to match an installed SDK.
10.0.400 [C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk]
```

`setup-dotnet` installs into a user-local dotnet root and the resolver then sees **only what that step installed** — note `Installed SDKs:` lists nothing but `10.0.400`. This branch asks for `10.0.x`; MetaMorpheus master's `global.json` asks for `8.0.204`; nothing satisfies it.

The mirror image is coming. Once MetaMorpheus#2710 raises its floor, mzLib **master** and **every open mzLib PR** still asking `8.0.x` go red the same way:

| mzLib branch asks | MM master floor | integration |
|---|---|---|
| `8.0.x` | `8.0.204` (today) | green |
| `10.0.x` | `8.0.204` | **red — this PR now** |
| `8.0.x` | `10.0.400` (after #2710) | **red — master + all open PRs** |
| `10.0.x` | `10.0.400` | green |

Both off-diagonal cells are red, so no merge ordering avoids it. I've opened #1143 — three lines giving that job both bands. Suggested sequence: **#1143 → MetaMorpheus#2710 → rebase this → merge → tag the release.**

**One consequence for the multi-targeting question in your PR body:** temporarily multi-targeting `net8.0;net10.0` would **not** fix CI, because the break is SDK resolution rather than TFM. Its only real benefit is letting ProteaseGuru and other consumers take the next release without retargeting, and it costs doubling the 64 hand-maintained `<file src>` entries to 128 — a format you rightly flag as failure-prone. I'd recommend against it, and ship a ProteaseGuru retarget PR instead (see below).

Also worth noting: `integration` is **not** a required check on `master` (required contexts are just `["build"]`), so the red is noise rather than a merge gate.

---

## Verified good

- **5435/5435 tests pass**, 29 skipped (Ignore/Explicit, not the filter). Build clean, 0 errors.
- **The nuspec is right.** All 64 `<file src>` paths resolve against a real Release build; `nuget pack` succeeds and produces exactly 64 archive entries (31 `lib/net10.0`, 32 `lib/net10.0-windows7.0`, 1 icon). Zero `x64` anywhere in it, so it's consistent with the post-#1127 `bin\Release` layout. The reverse direction is clean too — the only unshipped build output is `.lib` import libraries, duplicate copies, and the `Omics` `.txt` files that are `EmbeddedResource`-compiled into the DLL anyway. The `src`/`target` asymmetry (`net10.0-windows` vs `net10.0-windows7.0`) is correct, not a typo.
- **No conditional code to miss**: zero `#if` anywhere in the repo, zero `NET*_0`/`NETCOREAPP`, zero `Environment.Version`, zero TFM-conditioned MSBuild, zero conditional `PackageReference`, zero `LangVersion`. All 8 `RuntimeInformation` uses are OS probes in `SystemInfo.cs`.
- **The scientific output is unchanged.** I ran a real Classic bottom-up search through MetaMorpheus on both runtimes with identical inputs (four runs, two per runtime as a determinism control). `AllPSMs.psmtsv`, `AllPeptides.psmtsv` and `AllProteinGroups.tsv` are **byte-identical** between .NET 8 and .NET 10 — same scores, same row order, same q-values, same FDR boundary.

## NU1510 — two references this now invites you to delete

A controlled experiment (two throwaway projects identical but for TFM) established that the NU1902/1903/1904 transitive-audit warnings are **not** new — they fire identically on `net8.0` with a modern SDK. The only genuinely new category is **NU1510**, and it names exactly two now-redundant references:

- `Readers.csproj` — `System.Runtime.InteropServices`
- `MzLibUtil.csproj` — `Microsoft.Win32.Registry`

Note the second is **also** declared as a `<dependency>` in `mzLib.nuspec`, so removing it touches both files.

## The one TFM miss: two benchmark monikers

`Development/MSL/MslBenchmarks.cs:26` and `MslVsMspBenchmarks.cs:30` still carry `[SimpleJob(RuntimeMoniker.Net80)]`. BenchmarkDotNet generates and builds a child project at that moniker, so these now request a `net8.0` build out of a `net10.0` project. It **compiles**, so CI stays green and nothing catches it — it only fails when someone actually runs the benchmarks, and only if a net8 targeting pack isn't installed. Should be `RuntimeMoniker.Net10_0`.

For what it's worth, an exhaustive scan for legacy TFM strings across every file type returned exactly three hits: those two plus one cosmetic comment. The migration is otherwise complete.

## The cancellation fix is right, but incomplete — and one gap is a trap

The two added `ThrowIfCancellationRequested()` calls are correct, well-placed (after argument validation, matching `GetProjectFilesAsync`'s existing shape), and they close a doc-vs-behaviour gap since both methods already documented `OperationCanceledException`. The evidence for your .NET 8→10 rationale is actually stronger than you state it: the stub handlers split cleanly — the ones that self-guard (`PrideArchiveDownloadTests.cs:31`, `PrideFtpListingTests.cs:33`) belong to the tests that didn't break, and the ones that don't (`PrideProjectTests.cs:30`, `PrideProxiSpectrumTests.cs:32`) cover exactly the methods that did. That's precisely what you'd predict.

Three gaps:

**1. Koina has the same bug, and a test that asserts the invariant the library doesn't implement.**
`PredictionClients/Koina/Client/HTTP.cs:27` (`InferenceRequest`) and `:45` (`TestConnectionAsync`) take a token and go straight to `PostAsync`/`SendAsync` with no pre-check. And this is genuinely reached: `FragmentIntensityModel.cs:335` creates a **session-scoped** `CancellationTokenSource(timeout)` and loops chunks at `:336-345`, so once the session times out mid-loop every later iteration hands `PostAsync` an already-cancelled token. Same pattern at `CollisionalCrossSectionModel.cs:97`, `CrosslinkFragmentIntensityModel.cs:158`, `DetectabilityModel.cs:170`, `RetentionTimeModel.cs:157`. Under .NET 8 those failed instantly with no network call; under .NET 10 a burst of doomed requests goes to `koina.wilhelmlab.org`. Not a result change — a latency/politeness failure.

The trap: `Test/KoinaTests/HttpClientUnitTests.cs:23-31` (`InferenceRequest_AlreadyCanceledToken_Throws`) carries the comment *"A pre-canceled token short-circuits before any network I/O, so this stays offline."* It is **not** `[Category("ExternalService")]`, so it runs in the required, supposedly-offline `build` job — and it passes only because the real `SocketsHttpHandler` honours the token, not because the library does. That's exactly the assumption this PR disproves elsewhere.

**2. `DownloadFileAsync` is the third unguarded PRIDE method.** The body says three lacked the guard and two now have it — `PrideArchiveClient.cs:519-551` is the third. With a pre-cancelled token it still runs `Directory.CreateDirectory` and issues the HTTP GET before `CopyToAsync` throws. Mitigating: `DownloadProjectFilesAsync` guards in its own loop, so the multi-file path is covered, and this method's XML doc is the one that *doesn't* declare `OperationCanceledException`.

**3. Half the fix ships untested.** `PrideProxiSpectrumTests.cs` has no cancellation test — no `CancellationTokenSource`, no `.Cancel()` — and its stub ignores the token too, so **nothing would fail without the second added line**. The body's "the cancellation test fails without it" is true only of the `TryGetProjectAsync` half. A mirror of `PrideProjectTests.cs:603` costs six lines.

Suggestion: keep the two lines here since they unblock CI, and open a follow-up that adds the missing test, guards `DownloadFileAsync`, and guards both Koina methods.

## The nuspec has no validation, and the repo already has the pattern

You verified all 64 paths by hand and said so — that's exactly the verification that shouldn't have to be redone next time. `nuget pack` warns on a missing `src` but nothing fails the job, and nothing checks the reverse direction at all.

The repo already has the shape to copy: `.github/scripts/Check-TestNameHygiene.ps1`, wired into `dotnet.yml:32-34` as a `pwsh` step placed deliberately *before* the expensive step so a structural regression fails in seconds. A sibling `Check-NuspecPaths.ps1` is about five lines:

```powershell
$xml = [xml](Get-Content mzLib.nuspec)
$missing = $xml.package.files.file | Where-Object { -not (Test-Path $_.src) }
if ($missing) { $missing | ForEach-Object { Write-Error "nuspec src not found: $($_.src)" }; exit 1 }
```

Worth extending it to assert every `target` folder matches a declared `<dependencies><group targetFramework>`, which would catch a half-migrated nuspec — the exact failure mode this PR had to avoid by hand.

## `DOTNET_FRAMEWORK` conflates two repos' target frameworks

`dotnet.yml:9` defines it as `net10.0`, and its **only** use is `:135` — a path into **MetaMorpheus's** GUI output, not mzLib's. MM master is `net8.0-windows` today, so that path is wrong *right now* and the GUI artifact upload silently produces nothing (`upload-artifact` defaults to `if-no-files-found: warn`, so the step passes unnoticed). It becomes correct after #2710, then breaks again silently the next time the two repos diverge. Either rename it `METAMORPHEUS_FRAMEWORK` or glob `bin/Debug/net*-windows/`. The `DOTNET_VERSION` hoist itself is a genuine improvement.

## Two .NET 9+ behaviour changes with live call sites here

**Float→int conversions are now IEEE-saturating.** Measured: `(int)double.NaN` was `-2147483648`, is now `0`; `(int)1e30` was `-2147483648`, is now `2147483647`.

- Observed and benign: MetaMorpheus's written pruned database changed `mass="-2147483648"` → `mass="0"` for 436 of 9,159 sequences — all containing an ambiguous residue (X/B/Z/J/U) whose mass is `NaN`. I confirmed it doesn't propagate. But it will change the bytes of any stored regression baseline.
- **Live and worth a guard:** `MassSpectrometry/Deconvolution/Algorithms/ClassicDeconvolutionAlgorithm.cs:77` and `:81`, and `MassSpectrometry/MzSpectra/MzSpectrum.cs:264` — `charge = (int)Math.Floor(±1 / deltaMass)`. A duplicated m/z (vendor artefact, or post-processing) makes `deltaMass` exactly `0.0`, so the expression is `±∞`. Old runtime gave `charge = -2147483648`; new gives `+2147483647`. Both are nonsense that gets filtered — but they fail *different* bounds checks. A `deltaMass != 0` guard at those three sites is correct on any runtime.

**Unstable sorts** (not a .NET 10 regression — latent fragility a runtime bump is unusually likely to expose). `Array.Sort`/`List<T>.Sort` are contractually unstable and their introsort cutoffs have changed between runtimes before without a breaking-change note. Two sites sit upstream of every MetaMorpheus search and have no tie-breaker:
- `MassSpectrometry/WindowModeHelper.cs:15` — `Array.Sort(intensities, mArray)` feeding the top-N peak filter on **every spectrum MetaMorpheus loads**. Tied intensities are routine (integer detector counts, values thresholded to 0). A tie-order change keeps a *different set of peaks*.
- `Readers/SpectralLibrary/MslProteoformScorer.cs:202` — sorts by `CompositeScore` with no secondary key; changes the reported top proteoform.

My end-to-end diff shows neither fired for this runtime pair, so this is a hardening suggestion rather than a defect. Adding a total-order tie-breaker makes the output deterministic across *any* runtime.

## Consumers — one that breaks and one non-obvious trap

Since this single-targets `net10.0`:

- **ProteaseGuru breaks.** It's `net8.0`/`net8.0-windows` across five projects, pins mzLib **1.0.580**, and is actively released (0.2.1 on 2026-08-03) to NuGet.org. Its `global.json` is `8.0.204`, its two workflows say `8.0.x`, and its own nuspec carries 28 `net8.0` strings. It also publishes `--self-contained false`, so its users need the .NET 8 Desktop Runtime installed. Worth filing its retarget PR now so it's never stranded between the mzLib release and its own update. (Its nuspec already says mzLib 1.0.579 while the csprojs say 1.0.580 — drifted independently of this.)
- **pyMzLib has a trap a find-and-replace will miss.** Beyond the obvious `net8.0` TFMs and `DOTNET_VERSION: 8.0.x`, `wheels.yml:230-241` derives the wheel platform tags (`manylinux_2_28_x86_64`, `macosx_12_0_*`) from **.NET 8's** documented glibc and macOS floors, with the reasoning written into comments at `:232` and `:234`. .NET 10 raises those floors. The failure mode is a wheel that installs cleanly on an older cluster and then dies at runtime.
- **mzLibR** pins the bridge by version plus four hardcoded SHA-256 digests (`R/install-bridge.R:39-56`) and refuses to install on mismatch — all five values need re-recording after a new bridge release.
- **mzLibRust** needs no source change, but `ci.yml:92` tracks pyMzLib's newest release **unpinned**, so it'll consume a net10 bridge automatically with zero lead time. Good canary, no warning.

## Minor

- The body says "All **20** projects move" and names a **`TestFlashLFQ`** project. There are **19**, and no such project exists (`git ls-files` shows only `mzLib/FlashLFQ/*`). Harmless, but a PR whose value is completeness shouldn't misstate its own scope.
- `actions/setup-dotnet@v1` (3× here, 1× in `release.yml`), `actions/checkout@v2` (4×) and `nuget/setup-nuget@v1` are all deprecated — `setup-dotnet@v1` is archived and runs on a deprecated Node runtime. #1143 bumps the two in `dotnet.yml`; `release.yml` still wants doing.
- No `Directory.Build.props` exists, so all 19 csprojs declare `TargetFramework` and `Platforms` independently — which is what made this a 19-file edit instead of a 1-line one. Worth its own PR *before* the next migration, definitely not bundled into this one.

## On splitting the cancellation fix

The two-commit split (TFM / behaviour) is the right instinct and makes this reviewable. I'd still lean toward the behavioural change being its own PR: it's a public behaviour change (a call that previously returned data now throws) inside a PR titled "Move from .NET 8 to .NET 10", it's incomplete in ways a focused review would have caught, and half of it ships untested. Soft call — not a blocker, and keeping the two lines here does unblock CI.

acesnik:
🤖 Thank you for this review; the level of attention in it is obvious and it made the PR materially better. The NU1510 catch is a fair hit — two warnings were sitting in my own build output and I read straight past them.

What I most appreciate is the deconvolution comment, because that required knowing where the risk in this PR actually lives. Retargeting 8 → 10 means jumping *over* .NET 9, and the saturating float-to-integer conversion landed in 9 — so it's a behaviour change that never appears in either the source or the target version's notes and is invisible in the diff. Spotting that a `1/deltaMass` two files deep in a charge-state loop is exposed to it, from a PR whose diff is almost entirely `<TargetFramework>` lines, is not the kind of thing a retarget review usually surfaces. Same for the two `RuntimeMoniker.Net80` attributes and the `DOTNET_FRAMEWORK` variable that quietly described a different repository — both real, both things I'd have shipped.

All five applied in `7b5def3c`:

**1. Pruned package references.** Removed `System.Runtime.InteropServices` from `Readers` and `Microsoft.Win32.Registry` from `MzLibUtil`. You were right that the nuspec matters too — `Microsoft.Win32.Registry` was declared in **both** dependency groups, so that's two removals there, not one. Build goes from 39 warnings to 37, and NU1510 is now zero.

**2. Benchmark monikers.** `MslBenchmarks.cs:26` and `MslVsMspBenchmarks.cs:30` were still `RuntimeMoniker.Net80`, so the benchmarks would have measured a runtime the library no longer targets. Both now `Net10_0`.

**3. `DOTNET_FRAMEWORK`.** Agreed the name was wrong — it described MetaMorpheus's framework, and its only use was the MetaMorpheus GUI artifact path in the `integration` job. Rather than rename it I dropped the variable and globbed the path (`bin/Debug/net*-windows/`), so the two repos can retarget independently without this workflow needing an edit.

**4. Zero `deltaMass`.** Guarded at all three sites.

> **Edited to correct this point.** It originally claimed the change was result-neutral. That is true of `ClassicDeconvolutionAlgorithm`, where the guard sits inside the branch, but **false of the `MzSpectrum` copy as I first wrote it**: that method's outer condition has an `else` that is a `break`, so folding `deltaMass > 0` into the condition ended charge-state discovery for the candidate rather than skipping one neighbour. I verified the bounds check rejected the saturated `int.MaxValue` and never read the `else`. @trishorts measured 888/4000 synthetic cases diverging from master. Fixed in `a364c161`; details in [that thread](https://github.com/smith-chem-wisc/mzLib/pull/1141#discussion_r3769775990).

For the version now on the branch: a duplicated m/z gives `1/0 = +Infinity`, the cast used to wrap to `int.MinValue` and now saturates to `int.MaxValue`, and the bounds check below rejects both — so with the guard placed inside the branch the peak is skipped exactly as before. Worth doing because the code shouldn't depend on that, but it isn't a bug fix.

**5. PR body.** Corrected: 19 projects, not 20, and I invented `TestFlashLFQ` — the Windows-targeted pair is `Test` and `mzPlot`. (`TestFlashLFQ` is one of the stale artifact directories in my checkout with no tracked files; it isn't a project.)

On the deferred items, to confirm I've read them rather than dropped them:

- **Cancellation gaps** (Koina `HTTP.cs:27`/`:45`, `DownloadFileAsync` at `PrideArchiveClient.cs:519-551`, the missing `PrideProxiSpectrumTests` coverage) — out of scope for a retarget, and there's a trap worth flagging for whoever takes it: `HttpClientUnitTests.cs:23-31` asserts an invariant the library doesn't actually implement, and it sits in the offline job, so it passes today for the wrong reason.
- **`Check-NuspecPaths.ps1`**, the unstable-sort tie-breakers at `WindowModeHelper.cs:15` and `MslProteoformScorer.cs:202`, and the consumer PRs (ProteaseGuru retarget, pyMzLib wheel tags) — all separate changes.

Happy to file issues for any of those if you'd rather they not live only in this thread.

One limitation I should state plainly: `Test` is `net10.0-windows`, so I can't run the suite on macOS — the deconvolution changes are covered by CI, not by anything I ran locally.

mzLib/MassSpectrometry/MzSpectra/MzSpectrum.cs
```diff
@@ -258,7 +258,10 @@ public IEnumerable<IsotopicEnvelope> Deconvolute(MzRange theRange, int minAssume
                     for (int i = candidateForMostIntensePeakIndex + 1; i < XArray.Length; i++) //look at peaks of higher m/z
                     {
                         double deltaMass = XArray[i] - candidateForMostIntensePeakMz;
-                        if (deltaMass < 1.1) //if we're past a Th spacing, we're no longer looking at the closest isotope
+                        // A duplicated m/z makes deltaMass exactly zero, so 1/deltaMass is infinite. The int
+                        // cast of that changed with .NET 9 - it used to wrap to int.MinValue and now saturates
+                        // to int.MaxValue - so exclude it rather than rely on either bound being rejected.
+                        if (deltaMass > 0 && deltaMass < 1.1) //if we're past a Th spacing, we're no longer looking at the closest isotope
```
trishorts: The two halves of this one fix aren't the same fix, and this is the half that changes behaviour.

`ClassicDeconvolutionAlgorithm` puts the guard *inside* the branch — `if (deltaMass == 0) continue;` — so a duplicated m/z skips one neighbour and the scan carries on. Here `deltaMass > 0` is folded into the outer condition, whose `else` is a **`break`** (:279-282). So a duplicated m/z doesn't skip a neighbour, it **ends charge-state discovery for that candidate**, leaving `allPossibleChargeStates` empty or truncated where master kept walking.

Measured rather than argued — 4000 synthetic isotopic ladders, one duplicated m/z injected into each, through this overload, comparing the whole envelope set (charge, monoisotopic mass, peak count, total intensity) against master:

| variant | cases differing from master |
|---|---|
| this branch (`deltaMass > 0 &&` … `else break`) | **888 / 4000** |
| `continue`-shaped, as in `ClassicDeconvolutionAlgorithm` | **0 / 4000** |

Envelopes vanish, and where the truncated scan changes which candidate wins, the reported mass and peak count move:

```
master : case 25  nEnv=1  z=1, mono=1671.77825, n=4, ti=2069.129
this PR: case 25  nEnv=1  z=1, mono=1672.78083, n=3, ti=1223.217
master : case 21  nEnv=1  z=1, mono=298.14448,  n=2, ti=952.067
this PR: case 21  nEnv=0
```

**Calibrating that honestly: nothing ships broken.** This method body is unreachable. Its only caller is `MsDataScan.GetIsolatedMassesAndCharges(MzSpectrum, int, int, double, double)` (`MsDataScan.cs:232`), which is itself `[Obsolete]` and has no callers; the one test that touches that overload (`TestMsDataFile.cs:151`) returns on a null `IsolationRange` before reaching here, as its own comment says; and all four MetaMorpheus call sites — including the main search path at `TaskLayer\MetaMorpheusTask.cs:274` — bind the `DeconvolutionParameters` overloads, which route to `Deconvoluter.Deconvolute`. So the live charge-discovery loop is the one you got right.

It also catches a case wider than the one you meant: because `deltaMass > 0` sits in the outer condition, a **negative** `deltaMass` now breaks too, where before it computed a negative charge, failed the bounds test, and kept scanning. Nothing enforces sorted m/z — the constructor at `:98` neither sorts nor validates (and with `shouldCopy: false` it aliases the caller's array outright), and `ReplaceXbyApplyingFunction` (`:635`) rewrites `XArray` in place through an arbitrary `Func` without re-sorting, which is what `MsDataScan.TransformMzs` drives. So the library can manufacture the input itself. Guarding inside the branch handles both cases at once; note that moving `deltaMass != 0` into the outer condition instead would fix the negative case but leave the zero case still hitting `break`.

Worth the one-line change anyway, because this file and `ClassicDeconvolutionAlgorithm.cs:25-165` are a near-verbatim clone pair, and they already disagree on master in two ways — this copy is positive-polarity only (`:264` vs `ClassicDeconvolutionAlgorithm.cs:75-82`), and it reads the hard-wired `mostIntenseMasses` table where the live one uses the pluggable `AverageResidueModel` (`:289` vs `:109`). Adding a third divergence, on precisely the case the commit message says it's guarding, sets a trap for whoever eventually reconciles the two — they'd reasonably read the older copy as the reference:

```csharp
double deltaMass = XArray[i] - candidateForMostIntensePeakMz;
if (deltaMass < 1.1) //if we're past a Th spacing, we're no longer looking at the closest isotope
{
    // A duplicated m/z makes deltaMass exactly zero, so 1/deltaMass is infinite. The int cast of
    // that changed with .NET 9 - it used to wrap to int.MinValue and now saturates to int.MaxValue
    // - so guard rather than rely on either.
    if (deltaMass == 0)
    {
        continue;
    }
    //get the lower bound charge state
    int charge = (int)Math.Floor(1 / deltaMass);
```

Related: patch coverage on this PR is **33%** (2 hits / 3 misses / 1 partial over the six changed lines), and the partial is this condition — the new `deltaMass > 0` half is never evaluated false, in either copy. Neither guard has a test.

.github/workflows/release.yml
```diff
@@ -5,6 +5,10 @@ on:
     tags:
     - "[0-9]+.[0-9]+.[0-9]+"
 
+env:
+  DOTNET_VERSION: 10.0.x
+  DOTNET_FRAMEWORK: net10.0
```
trishorts: `DOTNET_FRAMEWORK` survived here, and in this file it is dead — nothing in `release.yml` references it (only `DOTNET_VERSION`, at :21).

Since the reason you dropped it from `dotnet.yml` was that the name described the wrong repository, worth deleting this copy in the same pass rather than leaving a variable that reads as authoritative and means nothing. Otherwise the next person to need a framework string in a workflow finds this one and reintroduces the ambiguity with a third meaning.

.github/workflows/dotnet.yml
```diff
@@ -6,6 +6,11 @@ on:
   pull_request:
     branches: [ master ]
 
+env:
+  DOTNET_VERSION: 10.0.x
+  # MetaMorpheus's target framework, not mzLib's - its only use is the GUI artifact path in
```
trishorts: Small thing: this comment is now orphaned. The variable it explains is gone from this block, so as written it reads as annotating `DOTNET_VERSION: 10.0.x` on the line above — which would tell a reader that `10.0.x` is MetaMorpheus's framework.

The explanation is worth keeping; it belongs at the artifact step where the glob actually is (:146).

trishorts (dismissed):
Re-reviewed at `7b5def3c` — fresh Release build, a forced `dotnet restore`, a full nuspec re-verification, and a measured probe of the deconvolution guard. **All five items you applied are correct, and the two you deferred were the right ones to defer.** Closing each of this morning's findings below; three small things are inline, none of them blocking.

## Closed

**1. NU1510 / pruned references — verified.** `dotnet restore --force` at this head reports **zero NU1510**. You were also right that I undercounted the nuspec: `Microsoft.Win32.Registry` sat in **both** dependency groups. I re-verified the part that could have broken silently — after the removals, all **64** `<file src>` paths still resolve against a real `Release` build, the two dependency groups are intact at 16 and 17 entries, and no `Registry` or `InteropServices` string remains in the file. Build clean, 0 errors.

**2. Benchmark monikers — verified.** Both now `RuntimeMoniker.Net10_0`, and BenchmarkDotNet 0.15.8 does have that enum member, so `Development` still compiles.

**3. `DOTNET_FRAMEWORK` — verified, and dropping the variable beats the rename I suggested.** Globbing `bin/Debug/net*-windows/` means neither repo's retarget has to touch this workflow again. Two leftovers, inline.

**4. `deltaMass` guards — and your correction is right; I'll take it.** The bounds check rejects `int.MinValue` and `int.MaxValue` alike, and `charge++` wraps rather than throwing — nothing in the repo sets `CheckForOverflowUnderflow`, which I checked rather than assumed. So the peak was skipped before and is skipped now, calling it a bug fix would have been wrong, and *"the code shouldn't depend on that coincidence"* is the accurate framing. I'd rather have that corrected than flattered.

The `ClassicDeconvolutionAlgorithm` guard — the one on the **live** path — is correct. The `MzSpectrum` one is a different shape and does change behaviour; measured, inline. It's unreachable code, so nothing ships broken.

**5. PR body — verified.** 19 projects, no `TestFlashLFQ`, and `Test`/`mzPlot` named correctly.

## Deferred, correctly

The cancellation gaps (Koina `HTTP.cs:27`/`:45`, `DownloadFileAsync`, the absent `PrideProxiSpectrumTests` coverage), `Check-NuspecPaths.ps1`, the sort tie-breakers and the consumer PRs are all genuinely separate changes — agreed on every one, and thank you for restating the `HttpClientUnitTests.cs:23-31` trap rather than letting it drop with the rest.

**Yes please, file issues** — at least for the cancellation set and `Check-NuspecPaths.ps1`. Those two are the ones with a real chance of being lost, and the first carries a written-down trap that is much cheaper to record now than to rediscover. The ProteaseGuru retarget and the pyMzLib wheel-tag re-derivation are on my side of the fence.

## CI, since it moved

`integration` is still red, but **the failure changed, which is the whole point**: it now dies at `NU1202` — *"Package mzLib 9.9.9 is not compatible with net8.0"* — during MetaMorpheus's restore, rather than at SDK resolution. #1143 merged at 17:29 and did its job; this job now installs both bands and gets all the way to the MetaMorpheus restore before hitting the genuine, expected TFM incompatibility. That clears the moment MetaMorpheus#2710 merges. `build` is green.

## One that can't go inline (unchanged file)

`mzLib/Directory.Build.props` arrived from master with #1142 and sets `NuGetAuditMode=all`, commented as *"the .NET 9+ default, but the CI setup-dotnet pin means it would otherwise stay off until that moves."* **This PR is that move** — the pin goes to `10.0.x` — so by the comment's own reasoning the property is now redundant and the comment describes a state this PR ends. Either drop the property or reword it. Not load-bearing; it's just the kind of stale conditional comment that outlives everyone's memory of the condition.

## Verified unchanged elsewhere

Full suite green locally at this head: **5333 passed, 0 failed, 29 skipped** (`Category!=ExternalService`; CI's `external-service-tests` leg is green separately). All 19 csprojs on `net10.0`/`net10.0-windows`, and a sweep of every tracked source, project and workflow file for legacy TFM strings returns **nothing** outside the deliberate `8.0.x` band in the `integration` job and the vendored Thermo binaries. Also confirmed the guard inventory is complete: `Math.Floor(±1 / deltaMass)` occurs at exactly three places repo-wide, and this PR reaches all three. The migration is complete.

---

Approving. Everything from this morning is closed, and the three inline items are small: a one-line guard shape in unreachable code, plus two dead workflow leftovers. I'd like the `MzSpectrum` guard flipped to `continue` before this merges so the clone pair doesn't drift further, but it isn't worth holding the PR for — and merge is gated on MetaMorpheus#2710 either way.

mzLib/MassSpectrometry/MzSpectra/MzSpectrum.cs
acesnik: 🤖 You're right, and my "changes no results" claim was wrong. I traced the saturated `int.MaxValue` into the bounds check, saw it rejected, and stopped there without reading the outer `else`. That `break` is the whole difference.

Fixed in `a364c161` using the shape you posted: the guard is inside the branch, so a duplicated m/z skips one neighbour. That also leaves negative `deltaMass` behaving as it does on master, which the outer-condition version didn't.

Taking the clone-pair point seriously too — this copy and `ClassicDeconvolutionAlgorithm.cs:25-165` already diverge on polarity and on `mostIntenseMasses` vs `AverageResidueModel`, and adding a third divergence on precisely the case the commit message says it's guarding would have set a trap for whoever reconciles them. That the method is unreachable doesn't excuse it.

On coverage: correct that neither guard has a test. The live one is reachable through `Deconvoluter`, so a duplicated-m/z case is writeable — happy to add it, with the caveat that `Test` is `net10.0-windows`, so CI would be its first run rather than my machine.

.github/workflows/release.yml
acesnik: 🤖 Removed in `a364c161`. You're right that it was never referenced in this file at all, so I left behind a variable that read as authoritative and meant nothing — same reasoning as dropping it from `dotnet.yml`, and it should have gone in the same pass.

.github/workflows/dotnet.yml
acesnik: 🤖 Moved to the artifact step at :146 in `a364c161`. Orphaning it under `DOTNET_VERSION: 10.0.x` was worse than deleting it would have been, since it implied `10.0.x` was MetaMorpheus's framework.

acesnik:
🤖 **Why `integration` is red, and why it isn't fixable from inside this PR.**

The job packs mzLib as 9.9.9, clones MetaMorpheus **master**, and builds it against that package. MetaMorpheus master is still .NET 8, so:

```
error: NU1202: Package mzLib 9.9.9 is not compatible with net8.0 (.NETCoreApp,Version=v8.0)
```

The 743 `CS0246` errors two steps later are just fallout — nothing got referenced, so every mzLib type disappeared. Framework compatibility is one-directional: a `net10.0` project can consume a `net8.0` package, never the reverse.

**So these two PRs are order-dependent, and neither description said so.** MetaMorpheus [#2710](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2710) is green on all checks and mergeable today — it passes because a `net10.0` MetaMorpheus consumes the current `net8.0` mzLib 1.0.583 quite happily. Merge #2710 first and this job goes green with no change here. The reverse order cannot work.

Two related changes in `705d3954`:

**Dropped the `8.0.x` SDK band from this job.** Nothing needs an 8.0 SDK once MetaMorpheus is on .NET 10.

**Corrected the rationale comment from #1143**, which said installing both bands *"keeps it green through a .NET migration in either repo, in either order."* This PR is the counterexample: the dual band covers SDK **resolution** only, and the constraint that actually binds is target-framework **compatibility**, which no number of installed SDKs affects. The comment now records the real invariant — mzLib retargets after MetaMorpheus, not before. Flagging it because the old wording would have led the next person to expect this job to tolerate either order.

Two notes on the other checks:

- `build` passes at 9m38s, so the full Windows suite is green on the retarget — including the deconvolution change.
- `codecov/patch` fails only on the new guard lines, which is fair; see the deconvolution thread.

I'd rather not disable or delete this job to get a green tick. It's the only cross-repo signal mzLib has, and it is correctly telling us something true about merge order. Happy to make it non-blocking temporarily if you'd prefer that over sequencing the merges.

trishorts (dismissed):
Re-approving at `705d3954` — my earlier approval was dismissed by the push, and all three inline threads are fixed and now resolved.

**`MzSpectrum` guard (`a364c161`)** — this is the shape I measured at 0/4000 divergence from master, so the clone pair now agrees on the zero case and negative `deltaMass` behaves as it did before. **`release.yml`** and the **orphaned comment** both done, and moving the explanation down to the artifact step reads better than the rename I originally suggested.

## On `705d3954` — you're right that my #1143 comment overclaimed, and the real mechanism is a third thing

Your correction is correct: dual SDK bands only ever addressed **SDK resolution**, and the binding constraint is **target framework** — a net10.0-only package is NU1202-incompatible with a net8.0 consumer no matter what SDKs are installed. "Green through a migration in either repo in either order" was wrong, and your replacement comment is the accurate version.

Worth adding what I found checking that the band removal was safe, because it's neither of the reasons stated. MetaMorpheus master is still `net8.0` with `global.json: 8.0.204` and #2710 is unmerged — so with only `10.0.x` requested, this job should have died at SDK resolution again. It didn't; it reached `NU1202`, exactly as before. The reason is in the step log:

```
install-dotnet.ps1 -SkipNonVersionedFiles -Runtime dotnet -Channel LTS
```

`setup-dotnet@v4` installs into the **standard** location, so the runner's preinstalled 8.0.x SDKs stay visible to the resolver. `@v1` installed into a user-local root, which is why the original failure printed `Installed SDKs:` with only `10.0.400` in it. **So the half of #1143 that actually fixed SDK resolution was the `@v1 → @v4` bump, not the dual band** — I credited the wrong line in my own PR. Your removal is safe today for that reason, and safe after #2710 for the reason you give.

Net effect: `integration` is red solely on `NU1202` against a net8.0 MetaMorpheus master, which clears when #2710 merges. `build` green.

## Remaining, none blocking

- A duplicated-m/z test through `Deconvoluter` would be welcome whenever you get to it — CI being its first run is fine, that's true of most of this repo's Windows-only tests.
- `Directory.Build.props`'s `NuGetAuditMode` comment still describes the SDK pin this PR moves.

Merge order is unchanged: #2710, then this, then tag.

mzLib/MassSpectrometry/MzSpectra/MzSpectrum.cs
acesnik: 🤖 Tests added in `08b8f399` — `Test/Deconvolution/TestDeconvolutionDuplicateMz.cs`, four cases per loop, duplicating each peak of a charge-1 ladder in turn and asserting the envelope set is unchanged versus the clean spectrum.

Two things worth reporting, because my first attempt at this test was worthless.

**The obvious version of the test passes against the broken code.** I started by duplicating the most intense peak of a 5-peak ladder at m/z 1000 and comparing envelopes. Identical output either way — because when the `break` kills discovery for that candidate, the *next* candidate (the duplicate itself) finds the same charge and reconstructs the same envelope. So the bug is invisible unless the duplicate lands somewhere the recovery doesn't happen.

Scanning 216 configurations (4 masses × 3 charges × 3 ladder lengths × every duplicate position) against both variants: **72 diverge**, i.e. 33%, in the same ballpark as your 888/4000. Every divergence was in the `MzSpectrum` copy; the `Deconvoluter` path was byte-identical across all 216, which is independent confirmation of your reachability analysis.

**Verified failing before it was verified passing.** With the guard folded into the outer condition, `duplicateAt` 0 and 2 fail:

```
Expected: < "z=1 mono=700.0002 peaks=2 intensity=1500.00", ... "intensity=583.33" >
But was:  < "z=1 mono=700.0002 peaks=2 intensity=1400.00", ... "intensity=550.00" >
```

The 1500 → 1400 case is the clearest statement of the defect: the truncated scan drops the *most intense* peak from the envelope and reconstructs it from the duplicate instead.

The four live-path cases pass against both variants. I kept them anyway rather than testing only the broken loop, since they pin the invariant on the path that actually runs and would catch anyone re-introducing the `break` shape there.

One caveat on how this was run: `Test` is `net10.0-windows` and cannot execute on macOS, so I ran the file through a throwaway `net10.0` NUnit project referencing `MassSpectrometry` directly. It compiles clean inside the real `Test` project (0 errors, no warnings from the new file), but this PR's CI is the first time it runs in situ.

acesnik:
🤖 Six commits since your first review, and a lot of dismissed approvals — but each closed something this PR would otherwise have shipped wrong, so I think we came out ahead.

**The duplicated-m/z tests are in** (`08b8f399`, four cases per loop). Instructively, my first attempt was worthless: duplicating the most intense peak of a ladder at m/z 1000 gives identical output either way, because the next candidate — the duplicate itself — quietly reconstructs the same envelope. Scanning 216 configurations to find where that recovery *doesn't* happen, 72 diverge, every one in the `MzSpectrum` copy and none in the `Deconvoluter` path — your reachability finding from the other direction. The committed version fails at `duplicateAt` 0 and 2 against the unfixed code, `intensity=1500.00` → `1400.00`.

**`Directory.Build.props`** (`961e15c6`): worse than stale, since the SDK pin was never the mechanism — the audit default follows the target framework. Lifting the OpenMcdf suppressions so there was something to count gave 12 NU1902 with the property, 12 without, and 2 with `direct` forced, so it's redundant exactly as you said. I kept it with an honest comment so a future downward retarget can't quietly reduce auditing, but happy to drop it instead.

**#1146 dissolved on contact** and is done in `4b7be54b` — the guard at `PrideArchiveClient.cs:632` existed but nothing exercised it. It also caught me wrong in my own issue body, where I claimed such a test passes regardless because `HttpClient` throws by itself. Deleting line 632 gives no exception at all and the request reaches the handler: `StubHandler` completes synchronously, so `HttpClient` never hits a suspension point where it would notice. Useful generalisation — cancellation tests against a synchronous stub exercise the library's guards, not the transport's. The rest of that set was already done (`HTTP.cs` threads the token through all five awaits; `DownloadFileAsync` has `f7d5167a` plus its own test), so my deferred list was stale and #1146 is scoped to the one real gap.

**#1145 I left open deliberately**, and my own attempt is why: a quick version confidently reported 32 problems in a healthy nuspec, thanks to a regex spanning newlines. Done properly it has to handle two `lib` targets where the csproj says `net10.0-windows` and the nuspec says `net10.0-windows7.0`, plus `mzLib.png`, native sub-folders, and non-DLL payloads. A naive version reddens CI on good code, which beats having no check only in appearance.

Your `setup-dotnet` finding cleared something up here too: I dropped the `8.0.x` band expecting NU1202, and got it — but via your `@v1 → @v4` install location, not my reasoning.

Everything outstanding is closed. `integration` is red solely on NU1202 against a net8.0 MetaMorpheus master, clearing when #2710 merges, and I'll leave the branch alone until then so your approval stands. Order unchanged: #2710, then this, then tag.

trishorts (approved):
Re-approving at `098be601`. I've also resolved the merge conflict and pushed it to this branch — details at the bottom.

## The three new commits

**`08b8f399` — duplicated-m/z regression tests.** This is better than what I asked for. Two things stand out:

- Every case carries `Assert.That(clean, Is.Not.Empty, "the ladder itself must deconvolute, or the test proves nothing")`. That negative control is the difference between a regression test and a test that passes because it exercises nothing — and it's exactly the failure mode of `HttpClientUnitTests.cs:23-31`, which we discussed as passing for the wrong reason.
- You verified it against the *unfixed* code and reported which cases fail and by how much (`duplicateAt` 0 and 2, intensity 1400 vs 1500 and 550 vs 583.33). That's the part I'd have had to redo otherwise.

Covering the obsolete overload too, behind `#pragma warning disable CS0618` with a justification, matches how `DeconvolutionScorer.cs:600` already reaches into that surface. And `Test/Deconvolution/` is the right home — seven sibling deconvolution files already live there.

Noting for the record that the four live-path cases pass either way. That is the correct result and worth keeping: it pins the behaviour of the loop that actually runs, so a future edit to `ClassicDeconvolutionAlgorithm` can't quietly acquire the bug the obsolete copy had.

**`4b7be54b` — the `GetProxiSpectrumAsync` cancellation test.** You also corrected my framing, and I'll take it: I'd said this half of the fix shipped untested and that a mirror test would cost six lines. You found the sharper version — deleting the guard shows the call doesn't throw *at all* and the request reaches the handler, because a stub handler completes synchronously so `HttpClient` never observes the token. So `Is.Empty` on `RequestedUris` is load-bearing, not decoration. That closes the "half the fix is untested" objection properly rather than nominally.

**`961e15c6` — the `NuGetAuditMode` comment.** Right correction, and you found the real mechanism: the default follows the target framework, not the SDK, so my "the CI pin is what's holding it off" reading was wrong too. Keeping the property with the inverted rationale — so retargeting a project downwards can't quietly reduce auditing to direct dependencies — is better than deleting it. The measurement (12 / 12 / 2 with the suppressions lifted) is the kind of thing that makes the comment trustworthy.

## The merge conflict — resolved and pushed

`#1139` merged and added a `THIRD-PARTY-NOTICES.md` entry to `mzLib.nuspec`'s `<files>` block; this branch had rewritten every path in that same block net8.0 → net10.0, so the whole block conflicted despite the two changes being unrelated.

Resolution keeps both: this branch's net10.0 paths plus master's three lines in master's position, right after the icon entry. The only diff against your previous nuspec is those three added lines — no reordering, no line-ending churn.

I **merged rather than rebased**, deliberately. Your history here is rebase-only, so a rebase would have matched it better — but it needs a force-push, and you'd pushed twenty minutes earlier, so a merge that lands as a fast-forward seemed the wrong thing to get clever about on someone else's branch. Squash it out or rebase over it if you'd rather; nothing depends on the merge commit existing.

Verified after the merge: build 0 errors, **5395 passed / 0 failed / 29 skipped**, all **65** nuspec `<file src>` paths resolve against a real Release build, dependency groups intact at 16 / 17 with `Microsoft.Win32.Registry` still removed, and zero `net8.0` strings left in the nuspec.

`integration` stays red on `NU1202` until MetaMorpheus#2710 lands — which is approved and one check from mergeable.

nbollis (approved):

PR #1142, opened 2026-08-12
Drop the SQLite EF6 adapter, clearing a critical and a high severity advisory

acesnik:
🤖

## Drop the SQLite EF6 adapter, clearing a critical and a high severity advisory

`Readers` referenced the **`System.Data.SQLite` metapackage**, which is `.Core` plus the EF6 and
LINQ-to-Entities providers. Nothing in mzLib uses either — `TimsTofFileReader` talks raw ADO.NET
(`SQLiteConnection` / `SQLiteCommand` / `SQLiteDataReader`, all in `.Core`) — but they dragged in:

```
System.Data.SQLite → .EF6 → EntityFramework → System.Data.SqlClient 4.8.1        HIGH
                                            → …ConfigurationManager
                                               → …Permissions
                                                  → System.Windows.Extensions
                                                     → System.Drawing.Common 4.7.0   CRITICAL
```

Referencing `System.Data.SQLite.Core` directly removes **17 packages** (54 → 37, measured on the
full `dotnet list package --include-transitive` graph) and adds none.

### This cannot change what gets read

The provider assembly and the native SQLite engine both come from `System.Data.SQLite.Core`, which
stays at the same **1.0.118** it already resolved to. Only the unused ORM adapters go away.

- **removed:** `EntityFramework`, `System.Data.SqlClient` + its SNI natives,
  `…ConfigurationManager`, `…Permissions`, `System.Windows.Extensions`, `System.Drawing.Common`,
  the metapackage, `…SQLite.EF6`, and their transitive tails
- **retained:** `System.Data.SQLite.Core 1.0.118`
- **added:** nothing

### Auditing, suppressions, and one deliberate piece of noise

`NuGetAuditMode` and the OpenMcdf suppressions live in a new **`Directory.Build.props`**, not in
`Readers.csproj`. OpenMcdf reaches six projects transitively, so a suppression in one project left
the advisories firing from `Development`, `FlashLFQ`, `PredictionClients`, `SpectralAveraging` and
`Test` under `NuGetAuditMode=all` — which is the .NET 9+ default, and would have become the actual
behaviour the moment #1141 landed. One place for it also gives #1141 somewhere to put the TFM.

The OpenMcdf advisories themselves are unfixable here: Thermo's CommonCore binaries are compiled
against OpenMcdf 2.x and 3.x is an API-incompatible rewrite, so taking the fix breaks `.raw`
reading. Filed with Thermo and forwarded to their CommonCore developers:
[thermofisherlsms/RawFileReader#22](https://github.com/thermofisherlsms/RawFileReader/issues/22).
There are **two** such advisories, and NuGet reports both: `GHSA-jxpf-xq2m-q525` (CVE-2026-41511,
fixed 3.1.3) and `GHSA-5qwm-7pvp-w988` (CVE-2026-45785, fixed 3.1.4). Both are suppressed by URL, so
any *new* OpenMcdf advisory still surfaces.

**Turning transitive auditing on repo-wide makes two genuinely new high advisories visible**, and
they are deliberately left unsuppressed: `System.Net.Http 4.3.0` and
`System.Text.RegularExpressions 4.3.0`, reaching `mzPlot` and `Test` via
`NETStandard.Library 1.6.1` ← `OxyPlot.Wpf 2.0.0`. They are real, they are high, and hiding them
would defeat the point of enabling the audit. The fix is the OxyPlot 2.0.0 → 2.2.0 upgrade, an
API-breaking change to WPF plotting code, so it belongs in its own PR.

So: `Readers` goes from four flagged packages to none, and the repo as a whole from unaudited to
audited with two known, attributed, high findings left standing on purpose.

### A watchdog for the suppression

Most of the added lines are a new scheduled workflow,
`.github/workflows/check-thermo-rawfilereader.yml`, and it deserves describing rather than sneaking
in alongside the dependency change. A suppression with nothing watching it becomes permanent by
accident, so this asks Thermo's repo monthly for the RawFileReader they publish, downloads the
nupkg, and reads the OpenMcdf dependency out of the published nuspec — testing the actual condition
for removing the suppression rather than a proxy for it.

It fails if the published version has moved past what we vendor, or if OpenMcdf has left the 2.x
line. Three outcomes are kept deliberately distinct, because the one thing this check must never do
is advise dropping a security control when it has simply lost track of the dependency:

| what it reads | what it says |
|---|---|
| OpenMcdf 2.x | all clear, suppression still justified |
| OpenMcdf 3.x+ | the pin and suppressions can probably go |
| nothing, or unparseable | this proves nothing — do **not** drop the suppression; fix the parsing |

The nuspec is parsed as XML with a namespace-agnostic lookup, and versions are compared on the major
number, so attribute reordering, line breaks inside the tag and exact-version ranges like `[2.3.0]`
cannot be mistaken for "the pin is gone". Verified against seven cases offline, and against the real
published nupkg, which reports `OpenMcdf 2.3.0` and passes.

### Why this surfaced now

.NET 9+ audits transitive packages by default when targeting net9.0+, so the .NET 10 work made these
visible. The vulnerabilities were always in the tree. This PR is independent of #1141.

### Verification

The solution builds and the graph diff is exactly the 17 removals with nothing added.

**Not verified here:** reading an actual Bruker file. `System.Data.SQLite` ships no `osx-arm64`
native (`SQLite.Interop.dll`), so that path cannot run on macOS either before or after this change.
CI covers `TestBruker` and the timsTOF reader tests, and those are the checks that matter for merge.


.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
...
+          deps = dict(re.findall(r'<dependency id="([^"]+)" version="([^"]+)"', text))
+          openmcdf = deps.get("OpenMcdf", "absent")
+          print(f"published RawFileReader {published} depends on OpenMcdf {openmcdf}")
+
+          problems = []
+          if published != vendored:
+              problems.append(
+                  f"Thermo publishes {published}, we vendor {vendored}. "
+                  f"Consider updating Readers/Thermo and VENDORED_VERSION in this workflow.")
+          if openmcdf == "absent" or not openmcdf.startswith("2."):
```
trishorts: **This is the one I''d fix before merge.** A parse failure and a genuine all-clear produce the same message, and that message tells the reader to remove a security control.

```python
if openmcdf == "absent" or not openmcdf.startswith("2."):
    problems.append(
        f"RawFileReader no longer pins OpenMcdf 2.x (now {openmcdf}). "
        f"The OpenMcdf pin and the NuGetAuditSuppress entries in Readers.csproj can probably go.")
```

`"absent"` is what `deps.get()` returns when the regex on line 56 matched nothing â€” which is a *parsing* outcome, not a fact about Thermo''s dependencies. Any change on their side that the regex doesn''t anticipate (attribute reordering, single quotes, a line break inside the tag, an exact-version range like `version="[2.3.0]"` which also fails `startswith("2.")`) lands in this branch and reports "no longer pins OpenMcdf 2.x â€¦ the suppressions can probably go."

So the failure mode points the wrong way: the watchdog''s advice on losing track of the dependency is *drop the suppression*.

**To be clear, it works today** â€” I downloaded `thermofisher.commoncore.rawfilereader.8.0.37.nupkg` and ran your exact regex against the real nuspec: 5 dependencies matched, `OpenMcdf â†’ 2.3.0`, `startswith("2.")` is True, no problem reported. This is latent, not broken.

Separating the two cases is a few lines and makes the watchdog fail safe:

```python
if openmcdf == "absent":
    problems.append(
        f"Could not find an OpenMcdf dependency in {spec}. The nuspec format may have changed - "
        f"this check needs updating; do NOT treat it as the pin being gone.")
elif not openmcdf.startswith("2."):
    problems.append(
        f"RawFileReader no longer pins OpenMcdf 2.x (now {openmcdf}). "
        f"The OpenMcdf pin and the NuGetAuditSuppress entries in Readers.csproj can probably go.")
```

Worth considering parsing the nuspec as XML rather than by regex, since you already have it decoded â€” `ElementTree` plus a namespace-agnostic `.//{*}dependency` lookup removes the whole class of formatting sensitivity.

.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
...
+          match = next((n for n in names
+                        if n.startswith("thermofisher.commoncore.rawfilereader.") and n.endswith(".nupkg")), None)
+          if not match:
+              print(f"::error::No rawfilereader nupkg under {DIR}. Layout changed: {names}")
+              sys.exit(1)
+
+          published = match[len("thermofisher.commoncore.rawfilereader."):-len(".nupkg")]
+          print(f"vendored={vendored} published={published}")
+
+          url = f"https://raw.githubusercontent.com/{REPO}/main/{DIR}/{match}"
```
trishorts: Minor: the listing you already fetched at line 35 carries a `download_url` for every entry, so this URL doesn''t need constructing.

I checked, and `main` is correct today â€” `repos/thermofisherlsms/RawFileReader` reports `default_branch: main`, so this works. But the API response gives you the answer for free:

```python
entry = next((f for f in listing
              if f["name"].startswith("thermofisher.commoncore.rawfilereader.")
              and f["name"].endswith(".nupkg")), None)
...
with urllib.request.urlopen(entry["download_url"], timeout=180) as r:
```

That also survives a default-branch rename, which would currently surface as a bare `HTTPError` traceback rather than a `::error::` line.

.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python3 - <<'PY'
+          import json, os, re, subprocess, sys, urllib.request, zipfile, io
+
+          REPO = "thermofisherlsms/RawFileReader"
+          DIR = "Libs/NetCore/Net8"
+          vendored = os.environ["VENDORED_VERSION"]
+
+          listing = json.loads(subprocess.run(
```
trishorts: `capture_output=True` with `check=True` means a `gh api` failure raises `CalledProcessError` with the stderr captured into the exception and never printed â€” so a rate limit, an auth problem, or a 404 from a repo rename all look like the same bare traceback in the log.

Since the whole point of this job is to be read once a month by someone who has forgotten it exists, it''s worth spending three lines to say what went wrong:

```python
proc = subprocess.run(["gh", "api", f"repos/{REPO}/contents/{DIR}"],
                      capture_output=True, text=True)
if proc.returncode != 0:
    print(f"::error::gh api failed for {REPO}/{DIR}: {proc.stderr.strip()}")
    sys.exit(1)
listing = json.loads(proc.stdout)
```

.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
...
+  # The RawFileReader build vendored under Readers/Thermo. Bump when new binaries land.
+  VENDORED_VERSION: "8.0.37"
+
+jobs:
+  check:
+    name: Compare against thermofisherlsms/RawFileReader
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
```
trishorts: Two small things about this step:

1. `actions/checkout@v4` isn''t needed â€” nothing in the job reads the repository. The Python block only talks to `gh api` and `raw.githubusercontent.com`. Dropping it saves the clone.

2. Related, and more interesting: `VENDORED_VERSION` restates a fact that is already on disk. The four DLLs under `mzLib/Readers/Thermo/` all report `FileVersion 8.0.37.0`, so the vendored version is discoverable rather than declarative. As written, someone who updates the binaries without remembering to edit this env var gets a *false* "Thermo publishes a newer version" alert â€” or, worse, a false all-clear.

If you keep the checkout, reading the version off the DLL would make the two impossible to desync. If you drop it, consider a comment noting that this constant must be bumped in the same commit as the binaries.

.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
...
+          if openmcdf == "absent" or not openmcdf.startswith("2."):
+              problems.append(
+                  f"RawFileReader no longer pins OpenMcdf 2.x (now {openmcdf}). "
+                  f"The OpenMcdf pin and the NuGetAuditSuppress entries in Readers.csproj can probably go.")
+
+          if problems:
+              for p in problems:
+                  print(f"::error::{p}")
+              sys.exit(1)
+          print("No change: still 8.0.37-equivalent and still on OpenMcdf 2.x. Suppression still justified.")
```
trishorts: Nit: this hardcodes `8.0.37` where the rest of the script uses `vendored`, so after a version bump the success message will claim the wrong number.

```python
print(f"No change: still {vendored} and still on OpenMcdf {openmcdf}. Suppression still justified.")
```

.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
+name: Check whether Thermo has released a new RawFileReader
+
+# The OpenMcdf advisories are suppressed in Readers.csproj because Thermo's CommonCore
+# binaries link OpenMcdf 2.x and 3.x is an API-incompatible rewrite. A suppression with
+# nothing watching it becomes permanent by accident, so this asks Thermo's repo monthly
+# whether either half of that has changed.
+on:
+  schedule:
+    - cron: "37 7 8 * *"
```
trishorts, line 9: Worth thinking about who actually sees a failure here.

A `schedule`-triggered workflow runs as the repository, and GitHub emails failed-run notifications to the *workflow file''s author* â€” not to the maintainers generally, and not to anyone watching the repo. A monthly watchdog whose entire value is being noticed once a year is exactly the kind of thing that fails silently for eighteen months.

Options, roughly in order of effort: add a step on failure that opens (or updates) an issue via `gh issue create`, so the signal lands somewhere durable and visible; or accept it and note in the header comment who is expected to watch it.

Also, GitHub disables scheduled workflows in repositories with no activity for 60 days â€” not a concern for mzLib, but worth knowing the mechanism exists.

.github/workflows/check-thermo-rawfilereader.yml
```diff
@@ -0,0 +1,75 @@
+name: Check whether Thermo has released a new RawFileReader
```
trishorts, line 1: Process note rather than a code comment: this file is 75 of the PR''s 86 added lines, and the PR description doesn''t mention it exists. The description covers the SQLite/EF6 swap and the OpenMcdf suppressions thoroughly â€” this watchdog only appears in commit `71cd9e3d`.

It''s a good idea and it clearly belongs with the suppression it guards, so I''m not suggesting it be split out. But a paragraph in the description would help whoever reads this in a year, and it would have made the review easier â€” a reviewer skimming the body would not expect the majority of the diff to be a scheduled GitHub Action.

mzLib/Readers/Readers.csproj
```diff
@@ -8,18 +8,19 @@
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 	</PropertyGroup>
 
+	<ItemGroup>
+		<!-- Two moderate infinite-loop DoS advisories in OpenMcdf 2.x, unfixable while Thermo
+		     CommonCore links 2.x: thermofisherlsms/RawFileReader#22 -->
+		<NuGetAuditSuppress Include="https://github.com/advisories/GHSA-jxpf-xq2m-q525" />
```
trishorts: The suppression is correct and I verified it works: at master `Readers` emits two `NU1902` warnings for OpenMcdf; at this head it emits zero. Both advisory URLs resolve (HTTP 200) and they are genuinely distinct.

**One gap, and it''s in exactly the scenario the description uses as motivation.** The body''s rationale is that .NET 9+ flipped `NuGetAuditMode` from `direct` to `all`, making transitive packages visible. Building with `-p:NuGetAuditMode=all`, the OpenMcdf advisories still fire from **four other projects** that reach OpenMcdf through `Readers`:

- `Test.csproj`
- `FlashLFQ.csproj`
- `SpectralAveraging.csproj`
- `PredictionClients.csproj`

They''re quiet today only because CI pins `setup-dotnet` to `8.0.x`, whose default audit mode is `direct`. So this is sufficient right now and becomes insufficient the moment #1141 lands â€” which is the PR that makes the `all` default real for this repo.

Since mzLib has no `Directory.Build.props`, there''s no single home for a repo-wide suppression today. Creating one for these two items would fix it in one place and would also give #1141 somewhere to put the TFM. Alternatively, duplicating the `ItemGroup` into the four projects works but ages badly.

Not a blocker for this PR â€” but worth deciding now rather than rediscovering it as five new warnings after the .NET 10 merge.

mzLib/Readers/Readers.csproj
```diff
@@ -8,18 +8,19 @@
...
 	<ItemGroup>
 		<PackageReference Include="CsvHelper" Version="32.0.3" />
-		<!-- OpenMcdf is intentionally pinned to the 2.x line. DO NOT upgrade to 3.x.
-		     The Thermo CommonCore RawFileReader binary is compiled against OpenMcdf 2.x
-		     (v8.0.37 targets OpenMcdf 2.3.0); OpenMcdf 3.x is an API-incompatible rewrite
-		     that breaks Thermo .raw reading. Thermo's latest release still ships against
-		     OpenMcdf 2.x, so there is no 3.x-based option today. GHSA-5qwm-7pvp-w988
-		     (fixed in 3.1.4) is a DoS on crafted/untrusted CFB files and does not apply -
-		     we only read trusted local instrument files. -->
+		<!-- Pinned to 2.x: Thermo CommonCore links 2.x, and 3.x is an incompatible rewrite. -->
```
trishorts, line 13: Flagging for your judgement rather than asking for a change: this condenses eight lines of rationale to one, and the two lines it documents are unchanged in this diff.

What the old comment carried that this doesn''t: that v8.0.37 specifically targets OpenMcdf 2.3.0, that 3.x breaks Thermo `.raw` reading rather than merely being incompatible, that Thermo ships no 3.x-based option, and â€” the part I''d most want kept â€” **the threat-model argument that the DoS applies to crafted/untrusted CFB files while we only read trusted local instrument files.** That last one is the actual justification for suppressing rather than fixing, and it''s the first question a future reader (or an auditor) will ask.

Some of it now lives in the `NuGetAuditSuppress` comment above and in the new workflow header, so the information isn''t lost repo-wide â€” but it''s no longer next to the pin it explains. The surrounding file leans the other way: the Bruker native block below still carries a multi-line "why these four DLLs must sit together" comment, and `dotnet.yml` / `release.yml` both carry long rationale comments.

Your call â€” just noting that the threat-model sentence is the one I''d restore.

mzLib/Readers/Readers.csproj
```diff
@@ -8,18 +8,19 @@
...
 		<PackageReference Include="OpenMcdf" Version="2.3.1" />
 		<PackageReference Include="OpenMcdf.Extensions" Version="2.3.1" />
-		<PackageReference Include="System.Data.SQLite" Version="1.0.118" />
+		<PackageReference Include="System.Data.SQLite.Core" Version="1.0.118" />
```
trishorts, line 16: The core change, and I''ve verified the claim it rests on: the EF6 / LINQ-to-Entities adapters really are unused.

A repo-wide search across 959 files for `System.Data.Entity`, `DbContext`, `DbSet`, `EntityFramework`, `SQLiteProviderFactory`, `DbProviderFactories`, `EntityConnection`, `ObjectContext`, `IQueryable`, `SQLiteFactory` and `System.Data.SQLite.EF6` returns **zero hits**. I also closed the config-file loophole, which is the way EF6 usually gets wired up without a code reference: **there is no `*.config` file anywhere in the repo**, so no `<entityFramework>` or `<DbProviderFactories>` registration can exist. `Readers` and `Test` both build with 0 errors against `.Core` alone, and `SQLite.Interop.dll` still lands in the output for all four RIDs (`linux-x64`, `osx-x64`, `win-x64`, `win-x86`) via the package''s own targets.

Three corrections to the description, all in the same direction â€” the change is better than advertised, but the numbers shouldn''t be stated as verified when they don''t match:

1. **It removes 17 packages, not 11.** Measured with `dotnet list package --include-transitive` before and after on this csproj: 54 â†’ 37, 17 removed, 0 added. Beyond the eleven you list, it also drops `Microsoft.CSharp`, `Microsoft.Win32.SystemEvents`, `System.CodeDom`, `System.ComponentModel.Annotations` and `System.Security.Cryptography.ProtectedData`. "The verified graph diff is exactly the eleven removals" is the one phrase I''d change, since the whole argument of the PR is that the graph diff was checked.

2. **The advisory correction is itself slightly off.** The description says the old comment cited `GHSA-5qwm-7pvp-w988` but "the one NuGet actually reports is `GHSA-jxpf-xq2m-q525`." Building at master, NuGet reports **both**. So the old comment was *incomplete* (one of two), not *wrong*. Naming both is still the right fix.

3. **"Audit output goes from four flagged packages to none" is `Readers`-only.** Solution-wide, `mzPlot` and `Test` still flag `System.Net.Http 4.3.0` and `System.Text.RegularExpressions 4.3.0`, both high. Pre-existing and out of scope here â€” just not "none".

Minor: the description credits `TimsTofFileReader` as the raw-ADO.NET consumer, but `BrukerFileReader` is the larger one (13 `SQLiteCommand`/`SQLiteConnection` sites plus the `SQLiteException` handling, vs 12). Same conclusion either way.

mzLib/mzLib.nuspec
```diff
@@ -26,7 +26,7 @@
     <dependency id="SharpLearning.Optimization" version="0.28.0" />
     <dependency id="SharpZipLib" version="1.4.2" />
     <dependency id="SkiaSharp" version="2.88.9" />
-    <dependency id="System.Data.SQLite" version="1.0.118" />
+    <dependency id="System.Data.SQLite.Core" version="1.0.118" />
```
trishorts, line 29: Both dependency groups are correctly updated â€” this one and the `net8.0-windows7.0` group at line 50 â€” and the version stays `1.0.118`, so what''s declared matches what the package was built against. No half-update here.

One consequence worth a sentence in the description: this changes the *published* contract, not just mzLib''s internal graph. A downstream consumer that was implicitly picking up `System.Data.SQLite.EF6` or `System.Data.SQLite.Linq` through mzLib''s transitive closure will stop resolving them after upgrading. The fix on their side is one explicit `PackageReference`, so it''s a note rather than a blocker â€” but "this cannot change what gets read" is argued entirely from mzLib''s own readers, and the shipped nuspec reaches further than that.

Confirming MetaMorpheus doesn''t reference the EF6 provider (and saying so in the body) would close it out.

trishorts (commented):
Reviewed with a local pipeline plus a dependency-graph pass, and I verified the load-bearing claims rather than taking the description on trust. **The core change is correct and safe â€” I''d merge it once the watchdog''s failure mode is fixed.** Commenting rather than requesting changes, because nothing here affects the dependency swap itself.

## The main claim holds up

The EF6 / LINQ-to-Entities adapters are genuinely unused: zero hits across 959 files for twelve different EF surfaces, **and no `*.config` file exists anywhere in the repo**, which closes the `<DbProviderFactories>` / `<entityFramework>` route by which EF6 normally gets wired up without a code reference. `Readers` and `Test` build clean against `.Core` alone, `SQLite.Interop.dll` still reaches the output for all four RIDs the package ships, and both nuspec dependency groups are correctly updated. The suppression works â€” two `NU1902`s at master, zero at this head â€” and both advisory URLs are live and distinct.

Measured, it removes **17 packages, not 11** (54 â†’ 37, nothing added). Better than advertised.

## What I''d change before merge

**One real defect, in the new workflow** (inline at line 65): a regex parse failure and a genuine all-clear produce the same message, and that message says *"the OpenMcdf pin and the NuGetAuditSuppress entries can probably go."* So if Thermo ever reformats their nuspec, the watchdog''s advice is to drop a security suppression. It works today â€” I downloaded the real nupkg and ran your regex against it, `OpenMcdf â†’ 2.3.0`, passes â€” so this is latent, not broken. Separating the two branches is a few lines.

**Three numbers in the description** don''t match what I measured (detail inline on `Readers.csproj`): the "exactly eleven removals" count, the claim that NuGet reports only `GHSA-jxpf-xq2m-q525` (it reports both, so the old comment was incomplete rather than wrong), and "four flagged packages to none" (that''s `Readers`-only; `mzPlot` and `Test` still flag two others). All harmless in themselves, but this PR''s value is that the graph diff was verified, so the stated numbers carry more weight than usual.

**One gap worth deciding now, not later** (inline on `Readers.csproj`): under `NuGetAuditMode=all` â€” the .NET 9+ default your description cites as the reason this surfaced â€” the OpenMcdf advisories still fire from `Test`, `FlashLFQ`, `SpectralAveraging` and `PredictionClients`. They''re quiet today only because CI pins `setup-dotnet` to `8.0.x`, which defaults to `direct`. This becomes incomplete the moment #1141 lands. A `Directory.Build.props` would fix it in one place and give #1141 somewhere to put the TFM too.

## Smaller

Inline: `download_url` is already in the listing you fetch (line 49); `gh api` stderr is captured and discarded (line 35); `actions/checkout` isn''t needed and `VENDORED_VERSION` duplicates a version that''s already in the DLL metadata (line 22); the success message hardcodes `8.0.37` (line 74); and it''s worth deciding who actually receives a failed scheduled run (line 9).

Also â€” and this is the reason the review is mostly about a file the title doesn''t mention â€” the watchdog is **75 of the 86 added lines** and doesn''t appear in the description at all. It''s a good idea and it belongs with the suppression it guards; it just deserves a paragraph.

## Checked and explicitly not raising

The `main` branch hardcoding at line 49 flagged as a possible break â€” I checked, `default_branch` really is `main`, so it works; I''ve left it as a simplification suggestion only. A case-sensitivity concern about the nupkg filename match turned out to be wrong: the published files really are lowercase (`thermofisher.commoncore.rawfilereader.8.0.37.nupkg`), so the prefix match is correct as written. And per your framing I''ve stayed off the OpenMcdf 2.x pin itself, the vendored Thermo binaries, the missing `osx-arm64` native, and the absence of a new test â€” all correctly scoped out.

acesnik:
🤖

Thank you for the care that went into this. Verifying the claims rather than taking the description on trust, downloading the real nupkg and running the regex against it, checking twelve EF surfaces across 959 files — and confirming there's no `*.config` anywhere to wire EF6 in without a code reference. That last check is the one I should have thought of and didn't; it closes the only remaining route by which the dependency could have mattered. Two of your measurements also correct mine.

## The watchdog defect

Good catch, and I've fixed it. You're right that "no OpenMcdf dependency found" and "OpenMcdf is no longer 2.x" were collapsed into one branch, and that the branch's advice was to drop a suppression — so a reformatted nuspec would have pointed someone at exactly the wrong action.

There are three distinct branches now, and reading nothing says so plainly: *"this run proves nothing either way. Do NOT drop the suppression on the strength of it; fix the parsing first."* Tested offline against four cases — today's 2.3.0, a hypothetical 3.1.4, a reformatted nuspec, and a new Thermo release.

## Your numbers are better than mine

**17 packages removed, not 11** — confirmed, 54 → 37. My figure came from a grep-filtered view rather than the full graph. Corrected in the description, and thank you for measuring it rather than taking the number on trust.

**Both advisories are reported by NuGet** — corrected.

**"Four flagged to none" was Readers-only** — corrected; see below.

## The audit gap, now fixed repo-wide

Confirmed, and your instinct to raise it as "decide now, not later" was the right call — it would have become wrong the moment #1141 landed. One small addition: it fires from **five** projects, not four; `Development` is also affected.

Both the audit mode and the suppressions now live in a new `Directory.Build.props`, so there's a single place for them, and #1141 has somewhere natural to put the TFM.

## What that surfaced, and a question for you

Enabling transitive auditing repo-wide makes two genuinely new **high** advisories visible: `System.Net.Http 4.3.0` and `System.Text.RegularExpressions 4.3.0`, reaching mzPlot and Test via `NETStandard.Library 1.6.1` ← **`OxyPlot.Wpf 2.0.0`**.

I've left them visible rather than suppressed, on the grounds that they're real and hiding them would defeat the purpose of turning auditing on. The fix is the OxyPlot 2.0.0 → 2.2.0 upgrade, which is an API-breaking change to WPF plotting code I can't exercise beyond compiling on macOS, so it wants its own PR. If you'd rather CI stay quiet in the meantime, I'm happy to suppress them with a pointer to that work — your call.

## Smaller points

All taken: `download_url` reused from the listing rather than reconstructed; `gh` stderr no longer swallowed, so an auth or rate-limit failure is visible; success message reports the version it actually checked instead of a hardcoded `8.0.37`.

Two I'd push back on gently. `actions/checkout` is still needed for `VENDORED_VERSION` — reading the version out of the DLL metadata instead would mean checking out and parsing a binary, which seems like more moving parts than the duplication costs, though I'm open to being wrong. And **who receives a failed scheduled run is genuinely a maintainer decision** — I can add issue-creation via `peter-evans/create-issue-from-file` if you'd like it to nag someone rather than sit in the Actions tab.

## On the description

Fair point that the watchdog is most of the diff and wasn't described. It has its own section now.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Fixed, and your XML suggestion caught a hole my first attempt left open.

I initially just split the branches, keeping the regex. But you listed `version="[2.3.0]"` as a case that fails `startswith("2.")` — and that is still 2.x. So my "fix" would have said *"no longer pins OpenMcdf 2.x, the suppressions can probably go"* for an exact-version range. Same wrong-direction advice, relocated rather than removed.

Now: `ElementTree` with a namespace-agnostic `.//{*}dependency` lookup as you suggested, comparison on the major version number so ranges normalise, and a `ParseError` guard so malformed XML lands in the fail-safe branch instead of crashing with a traceback.

Tested offline against seven cases: `2.3.0`, `[2.3.0]`, reordered attributes, a line break inside the tag, `3.1.4`, the dependency absent, and malformed XML. Only `3.1.4` now says the suppression can go.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Done — reusing `download_url` from the listing. It also takes the default-branch assumption out of the picture entirely, which was the better part of the suggestion.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Done, and worth flagging that my first pass did not actually do it: I added a comment saying stderr was visible while leaving `capture_output=True` with `check=True`, which still swallowed it. It now checks the return code and prints stderr into an `::error::` line, as you wrote it.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Checkout dropped — you are right that nothing in the job reads the repo. I had argued in my summary that it was needed for `VENDORED_VERSION`, which was wrong: that is an env var in this file, not something read from disk.

On the desync: I have kept the constant with a comment that it must be bumped in the same commit as the binaries, rather than reading `FileVersion` off the DLL. Reasoning — the dangerous direction is a *false all-clear*, which needs someone to bump the env var without the binaries; the likelier mistake (new binaries, stale constant) produces a false drift alert, which fails loudly and gets looked at. Happy to read it from the DLL if you would rather make them impossible to desync, though that means restoring the checkout.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Fixed, using `vendored`.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Good point, and implemented.

A failed scheduled run emails the workflow file author and nobody else, so a monthly check can fail unnoticed for a very long time — and the author here is me, which makes it worse.

Done with `gh issue create` as you suggested, rather than pulling in an extra action: on failure it opens an issue, or comments on the existing one if it is already open. The body warns that a parsing failure is *not* a reason to drop the suppression, since that is the reading someone skimming an alert is most likely to get wrong.

Noted too on the 60-day inactivity disabling — not a concern for mzLib, but useful to know the mechanism exists.

.github/workflows/check-thermo-rawfilereader.yml
acesnik: 🤖 Entirely fair, and the description now has a section on it — what it checks, the three distinct outcomes in a table, and what was tested. You were right that the majority of the diff being a scheduled Action is not something a reader should discover from the commit list.

For what it is worth, this was also the item I claimed to have fixed before I actually had; the description still said "eleven packages" and did not mention the watchdog when I first replied. It does now.

mzLib/Readers/Readers.csproj
acesnik: 🤖 Thank you for verifying it rather than taking it on trust — two `NU1902` at master and none at head is exactly the check I would want someone to run.

One thing has moved since: the suppressions now live in `Directory.Build.props` rather than here, because OpenMcdf reaches six projects and a `Readers`-only suppression left the advisories firing from five others under `NuGetAuditMode=all`. The verification still holds, just at a different location.

mzLib/Readers/Readers.csproj
acesnik: 🤖 Thank you for testing the load-bearing claim directly. The `*.config` check is the one I should have thought of and did not: a code-reference sweep alone would not have ruled out EF6 being wired up through `<DbProviderFactories>` or `<entityFramework>`, so without it my "genuinely unused" claim rested on less than I implied.

Your package count also corrected mine — 17 removed, not 11. I had counted from a filtered view rather than the full graph, and the description now says 17 with the measurement noted.

mzLib/mzLib.nuspec
acesnik: 🤖 Thanks for checking both groups. The `net8.0-windows7.0` one is easy to miss, and it would have shipped a package whose Windows-targeted consumers still pulled the EF6 chain while the `net8.0` ones did not — a difference nobody would have noticed until it mattered.

mzLib/Readers/Readers.csproj
acesnik: 🤖 You picked the right sentence, and it is restored.

Context on why it was cut: the terseness was deliberate — the maintainer asked me to condense those comments, on the principle that rationale belongs in the commit and PR rather than the project file. I took that further than it should have gone. Documenting *what* a pin is for is one thing; documenting *why suppressing beats fixing* is the justification itself, and losing it makes the suppression look unexamined.

So the threat-model line is back as one line rather than the eight it replaced:

```xml
<!-- ... Suppressed rather than fixed
     because the DoS needs a crafted CFB file and we read trusted local instrument files. ... -->
```

It now sits in `Directory.Build.props` next to the `NuGetAuditSuppress` entries, which is where the suppression moved, rather than beside the `OpenMcdf` pin in this file. The pin keeps the short "2.x, Thermo links 2.x, 3.x is an incompatible rewrite" note.

On the three other facts you listed — that 8.0.37 targets 2.3.0 specifically, that 3.x breaks `.raw` reading rather than merely being incompatible, and that Thermo ships no 3.x option — I have left those out of the file on purpose. All three are now verified rather than asserted, in the watchdog and in the PR body, and the watchdog will say so if any of them stops being true. Happy to restore them inline if you would rather they sit next to the pin.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified at `37d018c3` â€” resolved, and the XML rewrite is a better fix than what I suggested.

Confirmed in the new file: three genuinely distinct outcomes (`openmcdf is None` -> parse failure, `major(openmcdf) != 2` -> real all-clear, plus an `ET.ParseError` handler), and both failure branches now say *"do NOT drop the suppression"* in as many words. The `major()` helper handles `[2.3.0]` and `(2.0,3.0]`, which closes the range case.

I re-ran your parse path against Thermo''s live nuspec to be sure the XML route works on the real input, not just in principle: `ThermoFisher.CommonCore.RawFileReader.nuspec` parses, `OpenMcdf -> 2.3.0`, no problem reported. Worth noting it survives a UTF-8 BOM and an `encoding="utf-8"` declaration in the decoded string, which is the one thing I thought might bite `ET.fromstring` â€” it does not, on stdlib ElementTree.

One residual edge, purely for the record and not worth changing: a lower-bound-exclusive range like `(2.0,3.0]` reads as major 2 but would permit 3.0. Vanishingly unlikely in a published nuspec, and the docstring already states that reading deliberately.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified â€” resolved. Line 61 is now `url = by_name[match]["download_url"]`, reusing the listing entry rather than reconstructing the URL, so the default-branch assumption is gone entirely.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified â€” resolved. `check=True` is gone; the return code is tested and `proc.stderr` goes into the `::error::` line.

Appreciate you flagging that the first pass added the comment without the behaviour â€” that is the failure mode that would have been hardest to catch on re-review, since the comment reads as if it were done.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified â€” resolved. `actions/checkout` is gone, with a comment at line 27 saying why.

Your pushback on `VENDORED_VERSION` is right and I withdraw the suggestion. Reading `FileVersion` off the DLL would mean checking out the repo again and parsing a binary to avoid duplicating one string â€” more moving parts guarding a smaller risk. The comment at lines 13-14 tying the bump to the binaries is the proportionate fix.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified â€” resolved. Line 112 reports `{vendored}` and `{openmcdf}` rather than a literal.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified â€” resolved, and implemented better than I asked. The `if: failure()` step opens an issue or comments on the existing one, so a repeated monthly failure does not accumulate duplicates, and the issue body repeats the parse-failure warning where whoever acts on it will actually read it. Doing it with `gh issue create` rather than a third-party action is the right call for a job that already has `GH_TOKEN`.

The `permissions:` block (`contents: read`, `issues: write`) is a good addition â€” it was implicit before.

.github/workflows/check-thermo-rawfilereader.yml
trishorts: Verified â€” resolved. The description now has a section covering what the watchdog checks, its three outcomes, and what was tested.

mzLib/Readers/Readers.csproj
trishorts: Verified at `37d018c3` â€” resolved, and I re-ran the check against the new head rather than assuming.

`dotnet restore` across the solution with the new `Directory.Build.props`: **zero OpenMcdf advisories anywhere**, where a `Readers`-only suppression previously left them firing from the other projects. `NuGetAuditMode=all` is demonstrably live too â€” the two transitive advisories below only appear at all because it is on.

Moving them to `Directory.Build.props` is the better fix, and it gives #1141 somewhere natural to put the TFM.

mzLib/Readers/Readers.csproj
trishorts: Verified â€” resolved. The threat-model sentence is back, in `Directory.Build.props` lines 11-12: *"Suppressed rather than fixed because the DoS needs a crafted CFB file and we read trusted local instrument files."*

That is the one I wanted kept, and it is now next to the suppression it justifies rather than next to the pin, which is arguably the better home. The distinction you drew â€” documenting *what* a pin is for versus documenting *why suppressing beats fixing* â€” is the right line, and worth remembering the next time comments get condensed.

mzLib/Readers/Readers.csproj
trishorts: Verified â€” resolved. The description now says 17 packages, drops "eleven", and states that NuGet reports both advisories.

(The `BrukerFileReader`-is-the-larger-consumer nit was never worth a description change; ignore it.)

mzLib/mzLib.nuspec
trishorts: Resolving this, because I went and checked the downstream question myself rather than leaving it as an ask.

**MetaMorpheus is unaffected by this change.** It declares the `System.Data.SQLite` metapackage *directly* in four projects â€” `CMD.csproj:31`, `EngineLayer.csproj:31`, `GUI.csproj:75`, `TaskLayer.csproj:22` â€” alongside `.Core`, so it never relied on mzLib to supply the EF6 chain. Its two `App.config` files contain only binding redirects and a vestigial `supportedRuntime`; no `<entityFramework>` or `<DbProviderFactories>` section. Nothing to break.

**Two things that fell out of checking, both for MetaMorpheus rather than this PR:**

1. Those four direct metapackage references mean **MetaMorpheus still carries the exact advisories this PR just cleared from mzLib** â€” `System.Drawing.Common 4.7.0` (critical) and `System.Data.SqlClient 4.8.1` (high). The same one-word swap would work there.

2. It would very nearly be a clean swap. The only EF surface in the whole repo is `using System.Data.Entity.Core;` at `Test/DIATests/TestPfPairAndPfGroup.cs:10` â€” and it is a **stray unused using**: no `ObjectContext`, `EntityKey`, `EntityException` or any other EF type appears in the file. Deleting that one line removes the last thing standing between MetaMorpheus and the same fix.

Worth its own PR over there. Happy to open it if you would rather not carry it.

trishorts (approved):
Approving — all eleven review threads are resolved, and I verified each fix against `37d018c3` rather than taking the replies on trust.

Spot-checks worth recording: the new XML parse path runs correctly against Thermo's **live** nuspec (`OpenMcdf -> 2.3.0`, no problem reported), and a full `dotnet restore` at this head produces **zero OpenMcdf advisories anywhere in the solution**, with `NuGetAuditMode=all` demonstrably active — the two remaining transitive advisories only appear at all because it is on. Moving the suppressions to `Directory.Build.props` was the better fix and it gives #1141 somewhere natural to put the TFM.

## On the red checks — none of them are yours

I looked into all three before approving:

- **`external-service-tests` (failure):** `FilesEqualHash` and `TestUpdateUniprot`, both at exactly 1m40s — the `HttpClient` default timeout against UniProt. This job passed on your own earlier head `cea54305` and failed on **master** at `c7917c1a` two days ago.
- **`build` (cancelled):** it hit `timeout-minutes: 15` at 15m12s. Every step before `Test` succeeded; `Restore` actually got *faster* (55s -> 46s), so `NuGetAuditMode=all` costs nothing measurable. The `Test` step went 433s -> 759s, and the middle commit shows why: `Failed FilesLoading [1 m 41 s] -> TaskCanceledException`. That is a network timeout **inside the required job**, so the required suite still contains network-touching tests that are not `[Category("ExternalService")]`. A handful of 100-second hangs is the whole regression.
- **master is red too, for a third reason:** the required build on `70894e59` fails `TestUniSpecAcceptsValidPeptides` and `TestUniSpecChargeStateBoundaries` with `400 Bad Request` and `RuntimeError: cuDNN error: CUDNN_STATUS_INTERNAL_ERROR` — a GPU fault on Koina's model server, entirely outside this repo.

Three distinct third-party outages, all leaking into a required check. This is a good advertisement for #1140 ("keep the required CI job off the network, and make outages report as skips") — with that in, none of the above would have blocked anything.

**Please wait for a green `build` before merging** — approval does not bypass branch protection, and re-running once UniProt and Koina recover should be enough.

One loose end I have **not** cleared: `TestVariantCombinationOrderingPreventsOutOfBounds` failed in 1 ms on `0468a627` — a real assertion failure, not a timeout, and unrelated to anything here. It did not recur on the later runs, so it may be flaky, but it is the one failure I cannot attribute to the network. Worth a glance on the re-run.

## Two follow-ups this surfaced, both elsewhere

1. **MetaMorpheus carries the same advisories.** It declares the `System.Data.SQLite` metapackage directly in `CMD`, `EngineLayer`, `GUI` and `TaskLayer`, so it never relied on mzLib for the EF6 chain — but it still gets `System.Drawing.Common 4.7.0` (critical) and `System.Data.SqlClient 4.8.1` (high). Better still, `System.Data.SQLite.Core` is *already* declared alongside in all four, so the fix is a pure deletion rather than a swap. The only EF surface in that entire repo is a stray unused `using System.Data.Entity.Core;` in `Test/DIATests/TestPfPairAndPfGroup.cs`. I have that change prepared and am validating it now.

2. **`OxyPlot.Wpf 2.0.0 -> NETStandard.Library 1.6.1`** is the sole source of the two remaining high advisories, declared in one place (`mzPlot.csproj:10`). Your call to leave them visible rather than suppress them is the right one — suppressing them would defeat the point of turning auditing on, and they are genuinely fixable, unlike OpenMcdf. The 2.2.0 upgrade also happens to be what the Avalonia GUI work needs, so it may find an owner sooner than expected.

Nice work on the watchdog. Three distinct outcomes with the parse failure saying "this run proves nothing either way" is materially better than what I suggested, and catching the exact-version-range case yourself closed a hole my review had only half-seen.

nbollis (approved):

PR #1144, opened 2026-08-12
Point ptmlist.txt at UniProt's current location (master is red because of this)

acesnik:
🤖

**Draft, because I cannot verify it** — see the last section. Opening it now because this looks like
why **master has been red since 2026-08-10**, which is currently masking the real signal on every
open PR including #1141, #1142 and #1143.

## Diagnosis

`Loaders.cs:279` requests `http://uniprot.org/docs/ptmlist.txt`. That URL now answers **301** and
redirects twice, ending at:

```
https://ftp.uniprot.org/pub/databases/uniprot/knowledgebase/complete/docs/ptmlist.txt
```

The evidence that this is what is breaking the build:

- The failing tests are exactly the ones that load modifications at fixture setup —
  `FastaTest`, `BadFastaTest`, `DisulfideXmlTest`, `FilesLoading`, `MergeACoupleProteins`,
  `CompareOxidationWithAndWithoutCf`, and others — **plus `TestUpdateUniprot` itself**.
- They fail with `TaskCanceledException`, i.e. a timeout, not a parse error or a contract change.
- I reproduced the same failure locally against **both** mzLib 1.0.540 and 1.0.584, so it is not
  related to any recent change in this repo.
- master's last two runs failed independently of any PR.

The other four download URLs in `Loaders.cs` are fine: unimod answers 200 with no redirect, NIST
redirects http→https and resolves, and both PSI-MOD URLs resolve through GitHub.

## The change

One line: request the redirect destination directly, over HTTPS, instead of relying on two hops from
an `http://` URL.

## Why this is a draft

**`ftp.uniprot.org` does not respond from where I am** — the file times out at 40s and 180s, and so
does the host root. So I could not confirm the new URL actually fetches. It may be genuinely down, it
may be rate-limiting, or my network may be blocked. Someone who can reach that host should run
`TestUpdateUniprot` before this is trusted.

## If UniProt's FTP turns out to be unreliable

The durable fix is already the pattern in this file. `DownloadPsiMod` fetches from
`github.com/smith-chem-wisc/psi-mod-CV` — a lab-owned mirror — rather than from HUPO directly.
Mirroring `ptmlist.txt` the same way would take UniProt's availability off the critical path for
every test run and every MetaMorpheus search that builds a database.

I could not do that here because I cannot download the file to mirror it.

## Related

If this is the cause, it is worth deciding whether `Loaders` should fail loudly on a download problem
rather than surfacing as a dozen unrelated-looking test failures — that is what made this look like a
migration artifact at first glance.


trishorts:
i think it is because uniprot has been intermitantly down. i merged a pr this morning that should eliminate that redness. but we'll look at this anyway.

PR #1147, opened 2026-08-13
Carry retention times and protein groups through MergeResultsWith

acesnik:
🤖 `FlashLfqResults.MergeResultsWith` copies two of the three per-file values a peptide carries, and drops the third.

A peptide row in `AllQuantifiedPeptides.tsv` is written as three parallel per-file column blocks — intensity, retention time, detection type (`Peptide.cs:221-237`). For a peptide that both merged runs observed, the merge sets intensity and detection type but never retention time, and `GetRetentionTime` returns 0 for a missing key. So merged peptides report `0` in the retention time column for every merged-in file while their intensity and detection type are correct.

To be clear about what is and is not being moved here: this is not match-between-runs. There, the donor's retention time is deliberately *not* copied — it only feeds `PredictRetentionTime` to decide where to look, and the resulting peak's retention time is the acceptor file's own apex. `MergeResultsWith` combines two result sets covering **disjoint** file sets, so the retention time being carried over is a measurement that already belongs to the merged-in file. Nothing crosses a file boundary.

Same shape for `Peptide.ProteinGroups`: it is a `HashSet` that was not unioned, so protein associations found only in the second run were dropped. The constructor already unions protein groups when two identifications share a sequence (`FlashLFQResults.cs:50-53`), so this makes the merge path agree with it.

**Why this went unnoticed:** `MergeResultsWith` has no callers in production code, and its one test (`TestFlashLfqMergeResults`) runs with match-between-runs and normalization both off and asserts only collection counts.

**Test:** `MergeCarriesRetentionTimeAndProteinGroupsForASharedPeptide` quantifies the same peptide in two single-file runs under different protein group names, merges them, and asserts the retention time measured in the merged-in file survives along with both protein groups. It captures the pre-merge retention time and asserts it is greater than zero first, so the test cannot pass vacuously. Confirmed failing before the change on the retention time assertion.

Synthetic mzML only, so it runs anywhere — unlike the existing merge test, which reads a Thermo `.raw`.

---

### Merge order with #1151

Re-verified against `master` @ `5ba13155` (2026-08-18), after updating this branch: the two PRs **auto-merge cleanly in both directions**, `dotnet build` 0 errors, and the FlashLFQ + IsoTracker subset is **118/118** with both applied. No rebase conflict either way.

#1151 adds an XML `<remarks>` block to `MergeResultsWith` enumerating what the method carries over. This PR adds retention times and peptide `ProteinGroups` to that list, so **if this merges second, it must add those two bullets to that block** — otherwise the documentation ships describing behaviour that has moved. Suggested order: #1151 first, then this one picks up the doc edit in the same rebase.

### On the red `integration` job

Not caused by this PR. The run failed on 5x `WIX0103` from MetaMorpheus's `Product.wxs` (`EntityFramework.dll`, `EntityFramework.SqlServer.dll`, `System.Data.SqlClient.dll`, `System.Data.SQLite.EF6.dll`, `sni.dll`) — the .NET 10 migration fallout that hit every open mzLib PR at once, since the `integration` job clones MetaMorpheus `master` and builds it. Those components are no longer in MetaMorpheus `master`, so the branch update above should bring this job back green with no code change.


mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -89,6 +89,12 @@ public void MergeResultsWith(FlashLfqResults mergeFrom)
                     {
                         mergeToPep.SetIntensity(file, mergeFromPep.GetIntensity(file));
                         mergeToPep.SetDetectionType(file, mergeFromPep.GetDetectionType(file));
+                        mergeToPep.SetRetentionTime(file, mergeFromPep.GetRetentionTime(file));
+                    }
+
+                    foreach (ProteinGroup proteinGroup in mergeFromPep.ProteinGroups)
+                    {
+                        mergeToPep.ProteinGroups.Add(proteinGroup);
```
trishorts, line 126: **The protein-group union crosses the file boundary the retention-time half is careful not to cross.**

`Peptide.ProteinGroups` isn't owned by the `Peptide`. `FlashLFQResults.cs:46` constructs it as `new Peptide(..., id.ProteinGroups)` and `Peptide.cs:28` stores that reference verbatim (`this.ProteinGroups = proteinGroups;`). So `mergeToPep.ProteinGroups` **is** the `HashSet` belonging to the first `Identification` seen for that modified sequence in the receiving run — and that `Identification` is still attached to run A's `ChromatographicPeak`s.

I ran your fixture with the assertion extended down to the peaks:

```
file A peak identifications BEFORE merge: ProteinA
file A peak identifications AFTER  merge: ProteinA;ProteinB
```

`ChromatographicPeak.ToString()` builds the Protein Group and Organism columns from `Identifications.SelectMany(p => p.ProteinGroups)` (`ChromatographicPeak.cs:197,201`), so after a merge every **file A** peak row in `QuantifiedPeaks.tsv` reports `ProteinA;ProteinB`. The comment block immediately above that line documents the semicolon as asserting a specific parsimony relationship — which is exactly the claim we don't have here, since ProteinB was only ever observed in file B.

That's the one thing the description promises doesn't happen ("Nothing crosses a file boundary"). The retention-time half honours it; this half doesn't.

Smallest fix is to give the peptide ownership of its set, in `Peptide`'s constructor:

```csharp
this.ProteinGroups = new HashSet<ProteinGroup>(proteinGroups);
```

That also closes the identical aliasing on the existing `FlashLFQResults.cs:50-53` path you cite as precedent, and changes no public signature. With the set owned, this loop can become the one-liner mzLib already uses for set unions in `BioPolymerGroup.MergeWith` (`BioPolymerGroup.cs:719`):

```csharp
mergeToPep.ProteinGroups.UnionWith(mergeFromPep.ProteinGroups);
```

Worth a regression assertion as well: after the merge, `resultsA.Peaks[fileA].SelectMany(p => p.Identifications).SelectMany(i => i.ProteinGroups)` should still contain only `ProteinA`.

One second-order effect to decide on either way: unioning raises `Peptide.ProteinGroups.Count` for a peptide the two runs assigned different group names, and both protein-quant paths skip peptides with more than one group unless `useSharedPeptides` is set — `.Where(p => p.ProteinGroups.Count == 1 || useSharedPeptides)` at `FlashLFQResults.cs:358` and the `continue` at `:389`. So a merged peptide can quietly stop contributing to protein quant. That may well be the intended semantic (it matches the constructor), but it's worth stating.

trishorts (commented):
Reviewed locally against the merge base. The bug is real and the fix is the right shape — I checked out master, added only your test file, and it fails at the retention-time assertion with `But was: 0.0d`, then passes on this branch. So it's a genuine, non-vacuous regression guard, and the existing `TestFlashLfqMergeResults` still passes (its `pgA`/`pgB` are both `"MyProtein"`, so the new union dedupes to 1 and its count assertions hold).

One blocking issue, inline: the protein-group union writes through into run A's `Identification` objects and leaks the merged-in run's protein group into file A's peak rows. The retention-time change itself is clean and correctly keyed — `SpectraFileInfo` has full value equality (`SpectraFileInfo.cs:99-160`), so `mergeFrom`'s instances resolve properly against `this`'s dictionaries.

One clarification on the description rather than the code: `Peptide.cs:221-237` is inside the `if (IsoTracker)` branch. The default `TabSeparatedHeader` (`Peptide.cs:48-65`) and `ToString` (`:243-275`) emit only Intensity and Detection Type — there is no retention-time column in a non-IsoTracker `AllQuantifiedPeptides.tsv`, and `GetRetentionTime` has exactly one consumer in mzLib (`Peptide.cs:230`, inside that branch). Doesn't change the fix; the honest framing is "the merge drops one of the three per-file dictionaries `Peptide` maintains, regardless of which the current writers happen to emit."

I'm holding back a few smaller test notes to keep this focused — happy to add them if useful.

Separately, while tracing this I hit a pre-existing bug in the same method that is **not** yours: `MergeResultsWith` never merges `_peptideModifiedSequencesToQuantify`, so a `CalculatePeptideResults`/`ReNormalizeResults` after a merge silently zeroes every peptide that came in only from the merged-in run — including the retention time this PR carries. Reproduced and filed as #1149.

Alexander-Sol (dismissed):
Code changes look good and make sense. 

Currently, the MergeResultsWith function isn't used in MetaMorpheus or the FlashLFQ standalone. However, it's possible there could be a use case within pyMzLib, so this fix was a good catch.

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -118,6 +118,12 @@ public void MergeResultsWith(FlashLfqResults mergeFrom)
...
```
nbollis, line 126: [synthesis] Protein-group union mutates an aliased HashSet from existing identifications

`mergeToPep.ProteinGroups` aliases the `Identification.ProteinGroups` set already attached to chromatographic peaks from the receiving run, so unioning in `mergeFromPep.ProteinGroups` writes back into pre-existing file-A peak-identification rows and makes them report protein groups observed only in file B.

Suggestion: Give `Peptide` ownership of its protein-group set before merging, then use `UnionWith` on that owned copy.

nbollis (commented):
## Review Summary

- 1 major issue identified in `mzLib/FlashLFQ/FlashLFQResults.cs`.

All findings in this review resolved to inline anchors; no summary-only comments were needed.

mzLib/FlashLFQ/FlashLFQResults.cs
trishorts: Fixed in `46d0add9`. Confirmed, and it reaches further than the merge path.

`Peptide` held the caller's `HashSet` rather than a copy, and `FlashLfqResults` builds one from the **first** `Identification` carrying a given modified sequence:

```csharp
new Peptide(id.ModifiedSequence, id.BaseSequence, id.UseForProteinQuant, id.ProteinGroups)
```

Every later identification for that sequence then does `peptide.ProteinGroups.Add`, which writes into that first Identification's set — so the aliasing already corrupted results **within a single run**, before this PR. `MergeResultsWith` extends it across runs, which is the case you caught.

Fixed at the source rather than in the merge: `Peptide` copies, exactly as `Identification` already does with its own input. Nothing depended on the aliasing — every mutation site intends to change the Peptide.

The test walks `resultsA.Peaks[fileA]` to the identifications the rows are written from and asserts none gains `ProteinB`, while the peptide-level union still happens. Verified it fails against the previous behaviour with that exact assertion.

nbollis (approved):

mzLib/FlashLFQ/FlashLFQResults.cs
trishorts: Closing this out — status against head (`46d0add9`), point by point:

**1. Peptide ownership — done.** `Peptide.cs:35` copies rather than storing the caller's set, which is the fix as prescribed here. Worth noting nbollis arrived at the same finding independently on 26 Aug from the `MergeResultsWith` side, and the fix was written from that thread — so this comment sat unanswered for two weeks while the bug it identified was rediscovered. My fault for not reading back before fixing.

**2. `UnionWith` one-liner — not done.** Still `foreach` + `Add` at `:126`. Behaviourally identical now the set is owned, so it is cosmetic; folded into #1223 rather than asked for here.

**3. Regression assertion — done.** `TestFlashLfqMergeResultsFields.cs:85` walks `resultsA.Peaks[fileA]` to the identifications the rows are written from and asserts none gains `ProteinB`, while the peptide-level union still happens. Verified it fails against the previous behaviour.

**4. The protein-quant eligibility effect — real, deferred to #1223.** Both guards are unchanged (`:394` and `:424`), so the behaviour you described still holds: a peptide both runs assigned to different groups gains a second group on merge and stops contributing to protein quant, silently.

Deferring rather than fixing here, because the normal workflow runs inference once over all files — both runs assign the same group and the union is a no-op. Reaching it needs two results from separate inference, which nobody is known to be doing. And the exclusion itself is arguably correct: two inferences disagreeing means "we do not know", which is what `Count > 1` already encodes. What is missing is that it is invisible, so #1223 asks for a report rather than a guard.

Approving on that basis.

trishorts (approved):
Approving.

The aliasing bug this raised is fixed at the source — `Peptide` copies its protein-group set rather than holding the caller's, so the merge no longer writes back into an `Identification` still attached to the receiving run's peaks. That was corrupting results within a single run too, before this PR; `MergeResultsWith` only extended it across runs.

The retention-time half does what the title says and the test pins the file-boundary property directly, by walking `Peaks[fileA]` to the identifications the output rows are built from.

Two things deliberately left out rather than forgotten:

- **#1223** — a peptide both runs assigned to different protein groups gains a second group on merge and silently stops contributing to protein quant. Real, but the normal workflow runs inference once over all files, so both runs agree and the union is a no-op. It needs separate inference to fire, which nobody is doing. The exclusion is also arguably correct; what is missing is a report, not a guard.
- The `UnionWith` tidy at `:126`, folded into the same issue. Cosmetic now that the set is owned.

Note for the record: commit `46d0add9` on this branch is mine, pushed while working through the review findings — so this approval covers a change I wrote as well as your own.

PR #1148, opened 2026-08-13
Keep match-between-runs within a digestion agent

acesnik:
🤖 Match-between-runs treats every file as a possible acceptor for every donor peptide. When a dataset mixes digestion agents — one protease per file, which is how MetaMorpheus's multiprotease search works — a tryptic peptide can be "found" in a Glu-C file. It isn't in that digest, so the transferred peak is either noise or a coeluting species at the same mass.

This is the mzLib half of **smith-chem-wisc/MetaMorpheus#1034**, filed by @rmillikin in March 2018 and noted by @zrolfs in 2019 as the last issue left in the MetaMorpheus 1.0 milestone. The MetaMorpheus side is one argument at the `Identification` call site and closes that issue once this lands and ships.

### The principle, and the precedent already here

MBR's premise is that the peptide is present in the acceptor run but was never selected for MS2. That splits file metadata into two axes wanting opposite treatment:

- **Analyte pool** — digestion agent, fraction. A hard constraint. If the agent cannot produce the peptide, no amount of retention time calibration makes the transfer real.
- **Experimental design** — condition, biological and technical replicate. Must stay crossable, since filling in missing values across conditions is the entire point of MBR.

FlashLFQ already implements exactly this split — just not for digestion. On the experimental axis it deliberately crosses conditions, and `CalculateFoldChangeBetweenFiles` exists to score cross-condition donors differently; `RequireMsmsIdInCondition` is an opt-in knob (default `false`), not a physical rule. On the analyte axis it already blocks fraction mismatches: `PredictRetentionTime` refuses a transfer when both samples are fractionated and `|acceptorFraction - donorFraction| > 1`.

So this isn't a new concept, it's the missing case. MetaMorpheus#2718 has the experimental designs where this does and does not bite.

### What changed

1. `Identification` gains an optional `DigestionAgentName`. Null means unknown, so existing callers are unaffected.
2. `FindPeptideDonorFiles` chooses a donor per **(sequence, agent)** rather than one per sequence globally.
3. `QuantifyMatchBetweenRunsPeaks` skips donor files that don't share the acceptor's agent, via `SharesAnalytePool`.

Point 2 matters as much as point 3, and is the less obvious failure. Choosing one global donor per sequence lets a file from the wrong agent win the slot for a peptide both agents produce. Retention time calibration against a different-agent acceptor then has few shared anchor peptides, and `if (!scorer.IsValid(donorFile)) continue` drops that donor file's transfers for the acceptor entirely — so the legitimate same-agent transfer never happens. Spurious peaks are at least visible in the output; that loss is silent.

### Tradeoff to weigh

Agent equality also blocks transfers of peptides that **both** agents could produce. With trypsin and Lys-C, everything ending in K is a plausible product of either; in a synthetic run of mine, 15 of 27 transferred peptides fell in that category. So this trades false positives for some false negatives on shared peptides.

The stricter-but-better rule is to ask whether the donor peptide is a valid product of the *acceptor's* agent, using `DigestionAgent.DigestionMotifs`. It's a drop-in replacement for the body of `SharesAnalytePool`, so it can layer on later, but it is genuinely more subtle: the check has to permit missed cleavages (Woessmann et al. 2022 found miscleaved peptides quantify about as well as fully cleaved ones), and it has to decide what "valid product" means under semi-specific and non-specific search. It would also want an explicit `Omics` reference, which FlashLFQ currently only has transitively. Happy to go that way instead if reviewers prefer it.

### Decoys stay symmetric

Worth stating since it's the first thing I checked: MBR q-values are not biased by this. `randomDonor` comes from `donorPeaksMassOrdered`, which `GetRtCalSpline` builds from `_results.Peaks[donorFile]` — already past the guard. Target and decoy MBR peaks are restricted identically.

### Tests

Five tests in `Test/FlashLFQ/TestMbrDigestionAgent.cs`. Every file is written with the same MS1 peaks, so a refused transfer is a decision rather than an absence of signal.

| Test | Fails without |
|---|---|
| `TransferCrossesReplicatesButNotDigestionAgents` | the `SharesAnalytePool` guard |
| `DonorFromAnotherAgentDoesNotDisplaceTheSameAgentDonor` | per-(sequence, agent) donor selection |
| `UnknownAgentOnOneSideOnlyStillTransfers` | treating an unknown agent as unrestricted |
| `FileWithConflictingAgentsIsTreatedAsUnknown` | treating an unknown agent as unrestricted |
| `UnknownDigestionAgentLeavesTransfersUnrestricted` | nothing — pins backward compatibility |

Each of the first four was confirmed failing by reverting the relevant half of the change on its own, so none of them is padding. Synthetic mzML only, so they run anywhere.

The third and fourth cover the fallbacks that leave transfers unrestricted when no single agent can be attributed to a file — the two partial branches codecov flagged. `DigestionAgentOf` now uses `GetValueOrDefault`, dropping a lookup-miss branch that was unreachable, since the map is built from the same identifications that produce `SpectraFileInfoList`.

### Notes

- Needs a release before MetaMorpheus can consume it; the MetaMorpheus side is one argument at the `Identification` call site.
- Not measured on real data. Quantifying how much this distorts a real multiprotease experiment wants something like MassIVE MSV000083304 (Miller et al., JPR 2019 — Jurkat lysate, six proteases), which has no replicates per protease, so it can show spurious transfers but not the donor-displacement case the unit test covers.


mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -91,6 +96,16 @@ public FlashLfqEngine(FlashLfqParameters flashLfqParameters, List<Identification
                 .ThenBy(p => p.TechnicalReplicate).ToList();
 
             _allIdentifications = allIdentifications;
+            _fileToDigestionAgent = allIdentifications
+                .GroupBy(id => id.FileInfo)
+                .ToDictionary(
+                    group => group.Key,
+                    group =>
+                    {
+                        var agents = group.Select(id => id.DigestionAgentName).Distinct().ToList();
```
trishorts: **`Distinct()` counts `null` as a value, so one unannotated identification turns the whole file unknown.**

```csharp
var agents = group.Select(id => id.DigestionAgentName).Distinct().ToList();
return agents.Count == 1 ? agents[0] : null;
```

Everywhere else in this PR `null` means *unknown* — `SharesAnalytePool` reads `donorAgent == null || acceptorAgent == null`, and the `DigestionAgentName` doc says "or null if unknown". Only here does `null` become a third value that can *conflict* with a real agent. `{"trypsin", null}` gives `agents.Count == 2`, so the file resolves to unknown and is left unrestricted. But an unannotated identification doesn't contradict "trypsin" — it says nothing about it.

I ran it with one file Glu-C for every shared peptide except one, which carries no agent:

```
file3 agents = { null, Glu-C x5 }
tryptic peptide transferred into the Glu-C file? True
```

So a single unannotated identification anywhere in a file silently restores exactly the cross-agent transfer this PR exists to block. There's no warning and no output difference to notice. It's reachable in practice: `MzLibExtensions.MakeIdentifications` constructs identifications with no `digestionAgentName`, so any dataset mixing that path with an annotated caller lands in this state — as would an ambiguous PSM or an untagged decoy on the MetaMorpheus side.

Filtering nulls before `Distinct()` fixes it:

```csharp
var agents = group.Select(id => id.DigestionAgentName).Where(name => name != null).Distinct().ToList();
```

I applied exactly that and ran your fixture plus the probe above: 6/6. Both documented fallbacks survive — a file with no named agent still resolves to unknown (zero entries), and genuinely disagreeing agents still resolve to unknown (two or more entries). Might be worth a sixth test pinning the mixed case, since it is the one a real multiprotease run is most likely to hit.

trishorts (commented):
Verified this one locally rather than by reading, since the change is subtle. Everything in the description holds up:

- All five new tests pass on the branch, and the full `Test.FlashLFQ` suite is **118/118** — no regression to the existing MBR tests.
- Non-vacuity confirmed independently, one half at a time. Removing the `SharesAnalytePool` guard fails `TransferCrossesReplicatesButNotDigestionAgents` and `UnknownAgentOnOneSideOnlyStillTransfers`; reverting the donor keying to sequence-only fails `DonorFromAnotherAgentDoesNotDisplaceTheSameAgentDonor` and nothing else. The tests pin the two halves separately, as claimed.
- The decoy-symmetry argument checks out: `randomDonor` comes from `donorPeaksMassOrdered`, produced by `GetRtCalSpline` at line 990, which is downstream of the guard at 956. Targets and decoys are restricted identically.
- Backward compatibility is safe — the long-parameter constructor chains via `: this(...)`, so `_fileToDigestionAgent` is never null, and `SpectraFileInfoList` is built from the same `allIdentifications`, so the lookup-miss branch you dropped really is unreachable.
- The fraction precedent you cite is real (`FlashLfqEngine.cs:726-732`), though it guards per donor *peak* inside `PredictRetentionTime` while this guards per donor *file* — coarser and cheaper, and I think correct at that granularity.

One blocking issue inline: the per-file agent aggregation treats `null` as a conflicting value rather than as absence, which silently disables the guard for any partly-annotated file. One-line fix, verified.

Two smaller notes I'm leaving out of the diff to keep this focused, both pointing the same way. `DigestionAgentName` is a free-form string validated against nothing, so a typo silently becomes a distinct agent and blocks all MBR between two files that actually share a protease — a failure with no diagnostic. `Omics.Digestion.DigestionAgent` defines `Equals` as `agent.Name == Name`, so the typed form compares identically to what you have, but can't be typo'd and carries `DigestionMotifs` for the stricter rule you describe as the follow-up. `FlashLfqParameters.cs:3` already imports `Proteomics.ProteolyticDigestion` and the reference is transitively available, so it would cost one `using` and no csproj change. Genuinely a separate question from the motif rule you deferred, so take it or leave it.

trishorts (commented):
One follow-up note on the IsoTracker path, verified the same way. Non-blocking - the default path is unaffected since `isoTracker` defaults to false.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -627,23 +642,44 @@ private RetentionTimeCalibDataPoint[] GetRtCalSpline(SpectraFileInfo donor, Spec
...
+        /// <summary>
+        /// Whether an identification in the donor file could be present in the acceptor file at all.
+        /// Digestion determines which analytes exist in a run, so an identification is only transferable
+        /// between files digested with the same agent. This is the same kind of constraint as the fraction
+        /// check in <see cref="PredictRetentionTime"/>, and is deliberately separate from the experimental
+        /// design: transfers still cross conditions, biological replicates and technical replicates,
+        /// because filling in missing values across those is the point of match-between-runs.
+        /// Unknown agent on either side means unrestricted, so callers that supply no agent are unaffected.
+        /// </summary>
+        private bool SharesAnalytePool(SpectraFileInfo donorFile, SpectraFileInfo acceptorFile)
```
trishorts, line 678: **IsoTracker does its own cross-file identification borrowing, and it never reaches this guard.**

Two sites transfer an identification into a file that did not identify it, both outside `QuantifyMatchBetweenRunsPeaks`:

- `CollectChromPeakInRuns` — when no identification from the XIC's own file falls in the peak window, `case 0` re-queries `xICGroups.IdList` with the `p.FileInfo.Equals(xic.SpectraFile)` predicate dropped, and stamps the peak `DetectionType.IsoTrack_MBR`.
- `QuantifyIsobaricPeaks` — `if (xicIds.Count == 0) xicIds = mostCommonChargeIdGroup.Where(p => !p.IsDecoy).ToList();`, which pulls from every file to build the XIC.

I checked whether this actually matters by replicating `TestIsoSequence_Mixture` (a scenario that reliably produces `IsoTrack_MBR`) and varying only the agents on its two files:

```
[baseline - no agents]   agents = (null, null)       IsoTrack_MBR=1  Ambiguous=2  peaksWithIdsBorrowedFromAnotherFile=2
[SAME agent]             agents = (trypsin, trypsin) IsoTrack_MBR=1  Ambiguous=2  peaksWithIdsBorrowedFromAnotherFile=2
[DIFFERENT agents]       agents = (trypsin, Glu-C)   IsoTrack_MBR=1  Ambiguous=2  peaksWithIdsBorrowedFromAnotherFile=2
```

Identical across all three — the agent assignment has no effect on that path. So under `isoTracker: true` a tryptic identification is still borrowed into a Glu-C file and emitted as an `IsoTrack_MBR` peak.

Not blocking: `isoTracker` defaults to `false`, so the default MBR path is genuinely fixed by this PR. But the doc comment above reads as a property of the engine rather than of one method, so as written the next reader will assume the constraint holds everywhere. Two ways out, either fine:

- add the same donor/acceptor agent check where IsoTracker borrows, or
- say here that the constraint covers the `QuantifyMatchBetweenRunsPeaks` path only and that IsoTracker transfers are deliberate follow-up.

The second is a one-line doc change and records the gap rather than leaving it to be rediscovered.

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Correction to my comment above, from reviewing the MetaMorpheus counterpart (smith-chem-wisc/MetaMorpheus#2718).

I wrote that the null-agent case was reachable "as would an ambiguous PSM or an untagged decoy on the MetaMorpheus side." That part is wrong. `CommonParameters` coalesces the digestion params (`CommonParameters.cs:93`, `DigestionParams = digestionParams ?? new DigestionParams();`), `SpectralMatch` copies that reference (`SpectralMatch.cs:42`), and `DigestionParams` resolves its protease from `ProteaseDictionary` (`DigestionParams.cs:20`, with `DigestionAgent => Protease`). So every MetaMorpheus identification carries a non-null agent name, decoys and ambiguous PSMs included, and the `?.` chain in #2718 never actually falls through.

What stands: `MzLibExtensions.MakeIdentifications` still constructs identifications with no agent, `Identification` is public, and any caller mixing an annotated source with an unannotated one lands in the `{"trypsin", null}` state where the file resolves to unknown. So the bug is real and the one-line filter is still worth doing — but it is a robustness fix for the public API, not something MetaMorpheus will trip over. Lower priority than I implied.

Separately, and more relevant to the pair: #2718 reads `DigestionAgent.Name`, which for a non-specific search is `singleN`/`singleC` rather than the protease, so two files digested with different proteases can report the same agent. Details on that PR. It does not change anything in this one - the mzLib side is agnostic about what the string means - but it does mean the end-to-end behaviour needs that fixed too before the pair delivers what it says.

trishorts:
Pushed the null-agent fix to this branch directly (`c704bff6`) rather than leaving you to apply it — shout if you'd rather I had opened a PR against your branch instead.

Two files:

- `FlashLfqEngine.cs` — filters nulls before `Distinct()`, so an identification that names no agent no longer counts as a value that can disagree with a real one. Both documented fallbacks are unchanged: no named agent at all still resolves to unknown, and two or more genuinely different names still resolve to unknown.
- `TestMbrDigestionAgent.cs` — `UnannotatedIdentificationDoesNotMakeAFileUnknown`, built on the same three-file shape as `TransferCrossesReplicatesButNotDigestionAgents` so it carries its own positive control. File 3 is Glu-C except for one unannotated identification; the test asserts the trypsin-to-trypsin transfer still happens **and** the Glu-C one is still refused, so it can't pass vacuously.

Confirmed failing before the change (1 failed, your 5 passed) and `Test.FlashLFQ` is 119/119 after. Your five tests are untouched in both states.

Still outstanding on your side: the `singleN`/`singleC` issue on smith-chem-wisc/MetaMorpheus#2718, which is the other half of getting the agent string right — worth reading the two together, since they're the same failure in opposite directions (this one made a file unknown when it shouldn't be, that one makes two different proteases look identical).

acesnik:
🤖 Thanks for verifying this by running it rather than by reading — the non-vacuity check one half at a time is exactly the thing I would have wanted a reviewer to redo independently.

Taking the null-agent fix as pushed (`c704bff6`). I confirmed it the same way before accepting it: 6/6 on the branch, and reverting just the `.Where(name => name != null)` filter fails `UnannotatedIdentificationDoesNotMakeAFileUnknown` and nothing else, so the new test pins that half on its own. You are right that `null` was the odd one out — every other site in the PR reads it as *unknown*, and only the aggregation let it act as a third value that could disagree. Thanks for adding the test rather than just the fix.

**IsoTracker.** Verified, and you are right on both sites: `CollectChromPeakInRuns` `case 0` re-queries `xICGroups.IdList` with the `p.FileInfo.Equals(xic.SpectraFile)` predicate dropped, and `QuantifyIsobaricPeaks` falls back to `mostCommonChargeIdGroup.Where(p => !p.IsDecoy)` across every file. Neither routes through `QuantifyMatchBetweenRunsPeaks`, so `SharesAnalytePool` never runs.

I took your second option, since it records the gap instead of widening this PR: the doc comment on `SharesAnalytePool` now says the constraint covers the `QuantifyMatchBetweenRunsPeaks` path only and names the two IsoTracker sites. The code fix is filed as #1152, with your measured numbers, so it does not have to be rediscovered from the doc comment.

**Typed `DigestionAgent` instead of the string.** Checked your claim and it holds — `FlashLfqParameters.cs:3` already imports `Proteomics.ProteolyticDigestion`, and `DigestionAgent.Equals` is `agent.Name == Name`, so the typed form compares identically to what is here. I am inclined to leave the string anyway, for a reason your own follow-up comment surfaced: the typed form would not have caught the `singleN`/`singleC` problem either, since that is a genuinely wrong agent object rather than a mistyped name. The typo risk is real but it is a risk to a caller writing a literal, and the value crossing this boundary is a label rather than something FlashLFQ digests with. Happy to be talked round if you would rather have it typed before this ships — it is a small change and better made now than after release.

**`singleN`/`singleC` on smith-chem-wisc/MetaMorpheus#2718.** Confirmed, and thank you for catching it — `DigestionParams.RecordSpecificProtease` moves the real protease into `SpecificProtease` and overwrites `Protease` with `singleN`/`singleC` when `SearchModeType == CleavageSpecificity.None`, and `DigestionAgent => Protease`. So a non-specific search reports the same agent name for two files digested with different proteases, and the guard silently does nothing. It fails safe — that is current master's behaviour, not something worse — but the pair does not deliver for non-specific searches until #2718 reads the specific protease instead. Fixing it there.

Nothing outstanding on this one from my side, so this is good to merge whenever you are happy with it.

Alexander-Sol (dismissed):

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -627,23 +647,48 @@ private RetentionTimeCalibDataPoint[] GetRtCalSpline(SpectraFileInfo donor, Spec
...
         /// </summary>
         private void FindPeptideDonorFiles()
         {
             DonorFileToPeakDict = new Dictionary<SpectraFileInfo, List<ChromatographicPeak>>();
 
-            Dictionary<string, List<ChromatographicPeak>> seqPeakDict = _results.Peaks
+            Dictionary<(string Sequence, string DigestionAgent), List<ChromatographicPeak>> seqPeakDict = _results.Peaks
                     .SelectMany(kvp => kvp.Value)
                     .Where(peak => peak.NumIdentificationsByFullSeq == 1
                         && peak.IsotopicEnvelopes.Any()
```
nbollis, line 703: [synthesis] Donor grouping uses file-level agent instead of the identification's own digestion agent

The donor key uses `DigestionAgentOf(peak.SpectraFileInfo)`, i.e. the file-level summary. A file whose identifications mostly agree on one agent but contain a minority identification naming another agent can still group that minority peak under the majority agent and transfer it into a same-agent acceptor.

Suggestion: Group donor peaks by the identification-level digestion agent, or explicitly document that the restriction is file-granularity only.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -91,6 +96,21 @@ public FlashLfqEngine(FlashLfqParameters flashLfqParameters, List<Identification
...
             _allIdentifications = allIdentifications;
+            _fileToDigestionAgent = allIdentifications
+                .GroupBy(id => id.FileInfo)
+                .ToDictionary(
+                    group => group.Key,
+                    group =>
+                    {
+                        // an identification that names no agent is unknown, not a conflicting agent -
+                        // it must not turn a file whose other identifications agree into an unknown one
+                        var agents = group.Select(id => id.DigestionAgentName)
```
nbollis, line 113: [synthesis] Agent-name comparison is case-sensitive and whitespace-sensitive

`Distinct()` compares raw digestion-agent strings, so labels like `Trypsin` vs `trypsin` or trailing whitespace resolve the file to unknown and silently disable the intended transfer restriction.

Suggestion: Normalize agent names before comparison, e.g. trim and compare case-insensitively.

nbollis (commented):
## Review Summary

- 2 minor issues identified in `mzLib/FlashLFQ/FlashLfqEngine.cs`.
- No summary-only findings.

These comments focus on remaining edge-case behavior after the earlier null-agent and IsoTracker review threads were addressed.

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Fixed in `43834fde`. Names are trimmed and compared case-insensitively now, both when summarising a file and when comparing donor to acceptor; blank counts as no agent, as null already did.

Worth naming the failure shape, because it is the bad one: two spellings made a file look mixed, a mixed file resolves to unknown, and unknown means *unrestricted* — so the rule would have switched itself off silently rather than erring toward blocking.

Two tests, both failing against raw comparison: two spellings across files still transfer while Glu-C stays blocked, and one file spelling its own agent two ways stays known rather than turning unknown and re-permitting everything.

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Documented in `43834fde` rather than changed, because the premise does not hold — worth checking me on this.

`_fileToDigestionAgent` assigns an agent only when every identification in the file that names one agrees:

```csharp
var agents = group.Select(id => id.DigestionAgentName).Where(...).Distinct().ToList();
return agents.Count == 1 ? agents[0] : null;
```

So a file whose identifications mostly agree but contain a minority naming another agent does **not** group that peak under the majority agent — `agents.Count` is 2 and the file resolves to `null`.

What is true is that such a file is then unrestricted rather than partly restricted, and that was implicit. Now stated on `IsTransferable`, with the reasoning: a genuinely mixed file gets today's behaviour instead of having some of its identifications declared untransferable on the strength of a majority vote. Happy to take the identification-level key instead if you would rather have the stricter default — it is a design call rather than a defect, so I did not make it unilaterally on someone else's PR.

nbollis (approved):

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Addressed at head: the summary now filters `IsNullOrWhiteSpace` before `Distinct`, so an identification that names no agent no longer conflicts with one that does, and the comparison is trimmed and case-insensitive on top. Resolving.

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Addressed at head via the doc route (the second option): `SharesAnalytePool` now states that the restriction is scoped to the `QuantifyMatchBetweenRunsPeaks` path and that IsoTracker's own transfers in `CollectChromPeakInRuns` and `QuantifyIsobaricPeaks` remain unrestricted. The gap is recorded rather than left to be rediscovered, which is what I was after. Resolving.

trishorts (commented):
Closing an issue open since 2018 with a mechanism this small is a good outcome, and restricting donor selection rather than partitioning the engine is the right call — it keeps normalization and median polish spanning every file, which is a separate decision that should not fall out of where an engine boundary lands.

Things I verified rather than assumed:

- **`_fileToDigestionAgent` is built once, in the constructor (`:99`).** That matters more than it looks: MBR creates identifications as it goes, and if the map were computed lazily a transferred identification could reclassify its own acceptor file part-way through a run.
- **Backward compatible.** `digestionAgentName` defaults to null, null on either side means unrestricted, and `UnknownDigestionAgentLeavesTransfersUnrestricted` pins it. Nothing in mzLib populates `DigestionAgentName` yet — MetaMorpheus #2718 is what will — so merging this is behaviourally inert here until that lands.
- **The restriction is applied in one place** (`:980`), on the donor loop, which is the narrowest place it could go.

One design question inline. I do not think it blocks, but I would like your answer on it before I approve.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -627,23 +654,56 @@ private RetentionTimeCalibDataPoint[] GetRtCalSpline(SpectraFileInfo donor, Spec
...
+        /// identification in it that names one agrees; a file whose identifications disagree resolves to
+        /// unknown and is therefore unrestricted rather than being assigned its majority agent. That is
+        /// the permissive choice: a genuinely mixed file gets today's behaviour instead of having some of
+        /// its identifications quietly declared untransferable on the strength of a majority vote.
+        ///
+        /// Scoped to the <see cref="QuantifyMatchBetweenRunsPeaks"/> path. IsoTracker borrows identifications
+        /// across files of its own accord, in CollectChromPeakInRuns and QuantifyIsobaricPeaks, and does not
+        /// come through here, so those transfers are still unrestricted by digestion agent.
+        /// </summary>
+        private bool SharesAnalytePool(SpectraFileInfo donorFile, SpectraFileInfo acceptorFile)
```
trishorts, line 678: **This fails open, and the direction it fails in is the one the feature exists to prevent.**

A file whose identifications disagree on an agent resolves to `null`, and `null` here means unrestricted on *either* side. So a single stray identification naming a different agent switches the protection off for that file entirely — not just for that identification, and not just in one direction.

Concretely: file A is cleanly tryptic, file B is tryptic except for one identification that names Glu-C. B resolves to null, so transfers **into B from A** and **into A from B** are both permitted again. The second is the bad one — a Glu-C-only peptide can be transferred into a tryptic file, which is exactly the nonsense the issue describes.

The permissive default is defensible where the agent is simply unknown — that is the backward-compatibility case and it is clearly right. It is less obviously right when the agent is *known and conflicting*, which is a different situation wearing the same `null`.

**One asymmetry worth naming, because it constrains the fix.** The acceptor side genuinely cannot be identification-level: MBR transfers a peak into a file that has no identification for it yet — that is the whole point — so the acceptor's agent can only ever be a property of the file. But the **donor** side can be: the donating peak has identifications, and `peak.Identifications.First().DigestionAgentName` is right there. Keying donors by their own identification's agent rather than their file's would tighten the half that can be tightened, and leave the acceptor as-is.

That also happens to be the shape nbollis was reaching for in his second comment. His stated mechanism did not hold — a mixed file resolves to `null`, not to its majority agent — but the underlying worry, that file granularity is too coarse, lands here instead.

Happy either way, and happy for it to be a follow-up: the case needs a genuinely mixed file to arise, which is unusual. But `null` currently means two different things — "not told" and "told, inconsistently" — and only the first deserves to be unrestricted.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -627,23 +654,56 @@ private RetentionTimeCalibDataPoint[] GetRtCalSpline(SpectraFileInfo donor, Spec
...
+        /// Unknown agent on either side means unrestricted, so callers that supply no agent are unaffected.
+        ///
+        /// The restriction is file-granularity, deliberately. A file is assigned an agent only when every
+        /// identification in it that names one agrees; a file whose identifications disagree resolves to
+        /// unknown and is therefore unrestricted rather than being assigned its majority agent. That is
+        /// the permissive choice: a genuinely mixed file gets today's behaviour instead of having some of
+        /// its identifications quietly declared untransferable on the strength of a majority vote.
+        ///
+        /// Scoped to the <see cref="QuantifyMatchBetweenRunsPeaks"/> path. IsoTracker borrows identifications
+        /// across files of its own accord, in CollectChromPeakInRuns and QuantifyIsobaricPeaks, and does not
```
trishorts, line 675: Not a change request — flagging so the gap is on the record with the code rather than only in this thread.

The scoping note is accurate: `QuantifyIsobaricPeaks` (`:1899`) and `CollectChromPeakInRuns` (`:1962`) borrow identifications across files without going through `SharesAnalytePool`, so IsoTracker runs unrestricted. That is fine as a starting point — IsoTracker defaults off, and widening the rule into a second code path in the same PR would make this harder to reason about.

Worth an issue though, rather than living as a remark. If someone runs IsoTracker on a multi-protease set they get the pre-#1034 behaviour with nothing to tell them, and the comment explaining that is inside a method they have no reason to read.

trishorts:
@acesnik — pulling together what is waiting on you across the batch, so it is in one place rather than spread over six PRs.

**Needs a decision from you**

- **#1148** (this one) — reviewed above, 1/2. One design question: the agent check fails open on a *mixed* file, in the direction that matters. `null` currently means both "not told" and "told, inconsistently", and only the first deserves to be unrestricted. The donor side could be identification-level; the acceptor side genuinely cannot. Answer either way and I will approve — I do not think it blocks.
- **#1180** — changes requested, and only on one point: `TimePointsOnApexCharge` is `trace.Count`, but its doc and the PR's own argument for FWHM ("points across the peak means across FWHM, not across the full base") will send a consumer to it for a number it does not hold. It is public API in a shipped package, so it is a line now and a break after release. Two smaller non-blocking notes there too.

**Needs your approval, and only yours**

- **#1212** — Top3. You skipped it earlier, I think because I told you it was stacked on #1211 and to take that first. #1211 has since merged, so its diff has collapsed to the two Top3 files. Smallest review in the batch.

**For information**

- **#1147 merged** (`9b46fd07`), with #1223 opened for the deferred piece — a peptide the two runs assigned to different protein groups silently stops contributing to protein quant after a merge. Rare, and the exclusion is arguably right; the issue asks for a report rather than a guard.
- Five of the nine are in: #1202, #1208, #1209, #1211, #1147.
- Three commits on your branches are mine, pushed while working through nbollis's findings — `43834fde` here, `81fc5c0f` on #1180, and `46d0add9` on the merged #1147. Flagging plainly rather than leaving you to find them in the log.

Also worth your eye when you get to it: **#1213** records that the isobaric channel table cannot be built until we decide where mzLib gets its modifications from, which is the same question **#1185** is answering from the packaging side.

trishorts (commented):
Second pass, this time through the test file rather than the mechanism. It is a genuinely good suite — in particular writing every file with the same MS1 peaks, so that a refused transfer is a decision rather than an absence of signal. Negative assertions in MBR tests are usually worth very little and these ones aren't.

One gap, and it happens to sit exactly on the design question I raised earlier, which changes my read of it. Inline.

mzLib/Test/FlashLFQ/TestMbrDigestionAgent.cs
```diff
@@ -0,0 +1,278 @@
...
+            {
+                identifications.Add(Id(file1, shared, "trypsin"));
+                identifications.Add(Id(file2, shared, shared == PepSequences[0] ? "Glu-C" : "trypsin"));
+            }
+
+            identifications.Add(Id(file1, TransferredSequence, "trypsin"));
+
+            var results = new FlashLfqEngine(identifications, matchBetweenRuns: true, maxThreads: 1).Run();
+
+            Assert.That(WasTransferredInto(results, "mbr_conflict_2"), Is.True);
```
trishorts, line 183: **This asserts the harmless direction only, and the name claims the general property.**

The fixture gives `TransferredSequence` to file 1 alone (`:179`), so file 1 is the only possible donor and the mixed file 2 can only ever be an acceptor. The assertion therefore checks that a transfer is allowed *into* a file with conflicting agents — which is fine, nobody minds that.

The direction that matters is the other one: a mixed file as **donor** into a clean file. `SharesAnalytePool` returns true when either side is null, so a file that is mostly Glu-C with one tryptic identification will happily donate into a cleanly tryptic file. That is the transfer the whole feature exists to refuse, and no test covers it.

Worth contrasting with `UnannotatedIdentificationDoesNotMakeAFileUnknown` twenty lines below, which does assert both directions — `Is.True` for the trypsin pair and `Is.False` for the Glu-C file, with a message on each saying why. That is the shape this one wants: give the mixed file its own `TransferredSequence`, and assert what happens when it tries to donate.

I raised the fails-open behaviour as a design question in my earlier review and said I was happy either way. Having read the tests I would put it slightly more strongly: the benign half of the property is pinned and the risky half is unverified, so whichever way you decide, that assertion should exist. If you keep the current semantics it documents them honestly; if you tighten the donor side it is the test that proves it.

trishorts (approved):
Approving.

On the open design question — the permissive reading of `null` stands. A file whose identifications disagree resolves to unknown and is unrestricted in both directions, which means the protection can switch off for a whole file on one stray identification. Taking that deliberately rather than by omission, for three reasons: nothing in mzLib populates `DigestionAgentName` yet, so merging this is behaviourally inert until MetaMorpheus#2718 lands; reaching the case needs a genuinely mixed file; and this closes an issue open since March 2018, which should not wait on a refinement to a case nobody has hit.

Tracked as #1224 — tightening the donor side to the identification's own agent, which is the half that *can* be tightened, since MBR transfers into a file that has no identification yet and the acceptor can therefore only ever be classified at file level. The missing test assertion goes with it: `FileWithConflictingAgentsIsTreatedAsUnknown` currently asserts only a transfer *into* the mixed file, so the risky direction is unverified whichever semantics win.

What tipped it for me on the rest: the map is built once in the constructor, so an identification created during MBR cannot reclassify its own acceptor file part-way through a run. That is the subtle failure this design could have had and does not. And every file in the test fixture carries the same MS1 peaks, so the negative assertions mean something — a refused transfer is a decision rather than an absence of signal.

For the record: commit `43834fde` on this branch is mine, pushed while working through nbollis's findings, so this approval covers a change I wrote as well as yours.

Issue #1150, opened 2026-08-13
MergeResultsWith drops every field added to FlashLfqResults since 2018 (IsoTracker, IsobaricPeptideDict, MbrQValueThreshold, PepResultString)


trishorts:
`FlashLfqResults.MergeResultsWith` was written in 2018 and has never been updated for state added to the class since. Every field added after it is dropped by it. #1149 is one instance of that pattern — the quantification allow-list — and is being fixed on its own because it causes silent data loss. This issue tracks the rest, so the list is not closed along with #1149.

`git blame` on `mzLib/FlashLFQ/FlashLFQResults.cs`:

| Field | Line | Introduced | Merge behaviour |
|---|---|---|---|
| `_peptideModifiedSequencesToQuantify` | 20 | 2024-05 | **fixed by #1149** |
| `IsoTracker` | 15 | 2025-03 | dropped |
| `IsobaricPeptideDict` | 21 | 2025-03 | dropped |
| `PepResultString` | 22 | 2024-11 | dropped |
| `MbrQValueThreshold` | 23 | 2024-11 | `this`'s value silently wins |

Every line of `MergeResultsWith` itself dates to `a04db61b` (2018-08-28) or `3800df56c` (2018-07-11).

### The individual gaps

**`MbrQValueThreshold`** is a hard filter, read at `:149`, `:187` and `:260`. Merging results from two runs that used different MBR thresholds silently applies the receiving object's threshold to the merged-in peaks. `Math.Max` and `Math.Min` are both defensible and there is no precedent in the class, so this needs a deliberate decision rather than a default. An equality guard that throws on a mismatch is a third option.

**`IsoTracker` and `IsobaricPeptideDict` have to move as a pair.** `RevisedModifiedPeptides()` at `:226` guards on `IsoTracker && IsobaricPeptideDict != null`, so OR-ing the flag without unioning the dictionary is inert, and unioning the dictionary without the flag is equally inert. Merging an IsoTracker run into a non-IsoTracker run currently drops the isobaric dictionary entirely. Note the field is declared `IDictionary` but the engine assigns a `ConcurrentDictionary` (`FlashLfqEngine.cs:250`), so an implementation cannot assume a plain `Dictionary`.

[...]

PR #1151, opened 2026-08-13
Carry the quantification allow-list through MergeResultsWith

### LINKED ISSUE #1149: FlashLfqResults.MergeResultsWith drops _peptideModifiedSequencesToQuantify, silently zeroing merged-in peptides on requantification
`FlashLfqResults.MergeResultsWith` copies `SpectraFiles`, `PeptideModifiedSequences`, `ProteinGroups` and `Peaks`, but not the private `_peptideModifiedSequencesToQuantify` set (`FlashLFQResults.cs:20`). That set is the allow-list `CalculatePeptideResults` uses to decide which peptides may be repopulated, so after a merge every peptide that came in only from `mergeFrom` is silently zeroed the next time peptide results are calculated.

### Mechanism

`CalculatePeptideResults` opens by blanking every peptide in every file:

```csharp
// FlashLFQResults.cs:132-140
foreach (var sequence in PeptideModifiedSequences)
{
    foreach (SpectraFileInfo file in SpectraFiles)
    {
        sequence.Value.SetDetectionType(file, DetectionType.NotDetected);
        sequence.Value.SetIntensity(file, 0);
        sequence.Value.SetRetentionTime(file, 0);
    }
}
```

and then only repopulates sequences that pass the allow-list:

```csharp
// FlashLFQResults.cs:151
.Where(group => _peptideModifiedSequencesToQuantify.Contains(group.Key))
// FlashLFQResults.cs:193
if (!_peptideModifiedSequencesToQuantify.Contains(id.ModifiedSequence)) continue;
```

Because the merge never adds `mergeFrom`'s sequences to `this._peptideModifiedSequencesToQuantify`, the blanking at :132-140 applies to them but the repopulation at :151/:193 refuses them. Their `ChromatographicPeak`s were merged correctly at :118-128, so the underlying data is present and simply discarded.

This is reachable through public API: `ReNormalizeResults` (`FlashLFQResults.cs:66-75`) calls `CalculatePeptideResults` directly, and `IntensityNormalizationEngine` calls it at four sites.

### Reproduction

Quantify two different peptides in two single-file runs, merge, then requantify:

```csharp
var resultsA = QuantifyOneFile("probe_a", "PEPTIDE",  out SpectraFileInfo fileA);
var resultsB = QuantifyOneFile("probe_b", "PEPTIDEK", out SpectraFileInfo fileB);

resultsA.MergeResultsWith(resultsB);
resultsA.CalculatePeptideResults(quantifyAmbiguousPeptides: true);
```

Observed:

```
=== AFTER MERGE ===
  PEPTIDE : intensity(A)=982333.97  intensity(B)=0            rt(B)=0     detection(B)=NotDetected
  PEPTIDEK: intensity(A)=0          intensity(B)=975854.76    rt(B)=1.02  detection(B)=MSMS

=== AFTER CalculatePeptideResults ===
  PEPTIDE : intensity(A)=982333.97  intensity(B)=0            rt(B)=0     detection(B)=NotDetected
  PEPTIDEK: intensity(A)=0          intensity(B)=0            rt(B)=0     detection(B)=NotDetected
```

`PEPTIDEK` — present only in the merged-in run — goes from a real MSMS quantification to `0` / `NotDetected`. The receiving run's own peptide is unaffected, which is what makes this easy to miss: the failure is silent, produces no exception, and only affects the merged-in half of the data.

### Why it has gone unnoticed

`MergeResultsWith` has no production callers — its only caller anywhere is `TestFlashLFQ.TestFlashLfqMergeResults` (`Test/FlashLFQ/TestFlashLFQ.cs:488`), which asserts collection counts only and never requantifies after merging. MetaMorpheus does not call `MergeResultsWith` at all. So there is no current wrong-numbers-in-output impact in the Smith-lab stack; this is a latent trap for any external consumer of `FlashLfqResults` that merges and then normalizes.

### Suggested fix

Union the allow-list in `MergeResultsWith` alongside the other collections, e.g.:

```csharp
this._peptideModifiedSequencesToQuantify.UnionWith(mergeFrom._peptideModifiedSequencesToQuantify);
```

While in there, the following `FlashLfqResults` state is also dropped by the merge and may deserve the same treatment or an explicit decision: `IsobaricPeptideDict` and the `IsoTracker` flag (these interact — RT columns are only emitted in IsoTracker mode, so a merge that loses the flag also loses the output that would show the RT), `MbrQValueThreshold`, `PepResultString`, and `ProteinGroup.ConditionToQuantificationResults` on a name collision.

A merge test that calls `CalculatePeptideResults`/`ReNormalizeResults` afterwards and asserts the merged-in peptide survives would pin this down.

Found while reviewing #1147, which fixes two adjacent gaps in the same method (retention times and peptide protein groups). Not caused by that PR, but it does undo the retention time that PR carries.

trishorts:
Closes #1149.

`MergeResultsWith` copies `SpectraFiles`, `PeptideModifiedSequences`, `ProteinGroups` and `Peaks`, but not the private `_peptideModifiedSequencesToQuantify` set. That set is the allow-list `CalculatePeptideResults` uses to decide which peptides it may repopulate, so a peptide that arrived only from `mergeFrom` was blanked by that method's opening pass and then rejected by the allow-list filter — a real MSMS quantification silently became `0` / `NotDetected`. The peaks had merged correctly, so the data was present and simply discarded.

It is reachable through public API: `ReNormalizeResults` calls `CalculatePeptideResults` directly, and `IntensityNormalizationEngine` calls it at four sites.

### The fix

```csharp
this._peptideModifiedSequencesToQuantify.UnionWith(mergeFrom._peptideModifiedSequencesToQuantify);
```

Union is the semantically right direction — merging two runs should quantify the union of both allow-lists — and it matches the method's existing contract that nothing throws on a duplicate. The field is `readonly`, so in-place mutation is also the only option that compiles without changing the declaration.

### Why the constructor changed too

`FlashLfqEngine.cs:191` passes the engine's own `HashSet` in, and the constructor stored it **by reference**. Unioning into the results object would therefore have reached back and silently widened the engine's allow-list, which is read at eight sites inside the engine. The constructor now copies:

```csharp
_peptideModifiedSequencesToQuantify = peptideModifiedSequencesToQuantify != null
    ? new HashSet<string>(peptideModifiedSequencesToQuantify)
    : identifications.Where(id => !id.IsDecoy).Select(id => id.ModifiedSequence).ToHashSet();
```

`FlashLfqEngine` already copies its own input the same way (`FlashLfqEngine.cs:94-96`), so this brings the two sides into line. Behaviour is otherwise unchanged: the `null` path is untouched, and nothing depended on the two sets being reference-identical.

I also added an XML `<remarks>` block documenting what each member's state is after a merge, following `Omics.BioPolymerGroup.MergeWith`, which is the existing house pattern for this kind of method.

### Test

`TestFlashLfqMergeResultsKeepsMergedInPeptideQuantifiable` builds two single-file results with **different** modified sequences, merges, then calls `CalculatePeptideResults` and asserts the merged-in peptide still has a non-zero intensity and `DetectionType.MSMS`.

Both details matter. The existing `TestFlashLfqMergeResults` cannot catch this bug: both halves quantify the same sequence `EGFQVADGPLYR`, which makes the union a no-op, and it never requantifies after merging. Verified by removing the fix and re-running — the new test reports `Expected: greater than 0 / But was: 0.0d`, while `TestFlashLfqMergeResults` still passes.

The helper builds its inputs in memory rather than from fixture files, so it needs no `Test.csproj` entry. It passes real `Identification`s to the constructor deliberately: an empty allow-list means "quantify nothing", not "quantify everything", so a test built on `new FlashLfqResults(files, new List<Identification>())` would prove nothing.

### Verification

- Full suite `Category!=ExternalService`: **5387 passed, 0 failed, 29 skipped** (net8.0)
- FlashLFQ + IsoTracker subset: 117/117
- Test-merged #1147 locally: **auto-merges cleanly**, no conflict. That PR fixes the shared-peptide branch of this same method (retention times and protein groups); this one fixes the merged-in-only branch. Independent, either order.

### Scope

Four other pieces of `FlashLfqResults` state are still dropped by this merge — `IsoTracker`, `IsobaricPeptideDict`, `MbrQValueThreshold` and `PepResultString`. All were added after the method was written in 2018, and each needs its own decision (`MbrQValueThreshold` in particular has no obviously correct reconciliation). Filed as #1150 rather than folded in here.

MetaMorpheus is unaffected — it never calls `MergeResultsWith`, verified against MetaMorpheus `master` @ `dac54917f`. No companion PR needed.

*Files touched: `mzLib/FlashLFQ/FlashLFQResults.cs`, `mzLib/Test/FlashLFQ/TestFlashLFQ.cs`.*

---

### Merge order with #1147

Re-verified against `master` @ `5ba13155` (2026-08-18), after updating this branch: the two PRs **auto-merge cleanly in both directions**, `dotnet build` 0 errors, and the FlashLFQ + IsoTracker subset is **118/118** with both applied.

The only coupling left is documentation. The `<remarks>` block added here enumerates what `MergeResultsWith` carries over, and #1147 adds retention times and peptide `ProteinGroups` to exactly that list. **Whichever of the two merges second has to add those two bullets**, or the block ships describing behaviour that has moved. Cheapest order is this PR first, then #1147 folds the doc edit into its rebase.


Alexander-Sol (approved):
Code changes look good and make sense. However, I want to note that the `MergeResultsWith ` function isn't used in MetaMorpheus or the FlashLFQ standalone. However, it's possible there could be a use case within pyMzLib, so this fix was a good catch

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -74,10 +78,35 @@ public void ReNormalizeResults(bool integrate = false, int maxThreads = 10, bool
...
+        /// <remarks>
+        /// After merging:
+        /// <list type="bullet">
+        ///   <item><description>The set of peptides eligible for quantification contains the union of both
+        ///   results' sets, so a peptide identified only in <paramref name="mergeFrom"/> survives a later
+        ///   call to <see cref="CalculatePeptideResults"/></description></item>
+        ///   <item><description><see cref="SpectraFiles"/> contains both results' files, without deduplication</description></item>
+        ///   <item><description><see cref="PeptideModifiedSequences"/> and <see cref="ProteinGroups"/> contain
+        ///   both results' entries; where a key is present in both, <paramref name="mergeFrom"/>'s per-file
+        ///   intensities win</description></item>
```
acesnik, line 95: This holds for `mergeFrom.SpectraFiles` only — the receiving object's own files keep their values, so "wins" is stronger than what the loop does. Worth noting the asymmetry just below too: peptides get intensity and detection type copied, protein groups get intensity alone.

mzLib/FlashLFQ/FlashLFQResults.cs
```diff
@@ -74,10 +78,35 @@ public void ReNormalizeResults(bool integrate = false, int maxThreads = 10, bool
...
+        /// </list>
+        /// </remarks>
         public void MergeResultsWith(FlashLfqResults mergeFrom)
         {
             this.SpectraFiles.AddRange(mergeFrom.SpectraFiles);
 
+            // Peptides arriving from mergeFrom are quantifiable in their own right. Without this union
+            // CalculatePeptideResults would blank them along with everything else and then refuse to
+            // repopulate them, silently zeroing the merged-in half of the data.
+            this._peptideModifiedSequencesToQuantify.UnionWith(mergeFrom._peptideModifiedSequencesToQuantify);
```
acesnik, line 108: Unconditional, so a deliberately restrictive allow-list on the receiving side gets widened by the merge. Right trade against silent zeroing, and in the merge case both sets are legitimate — but if a caller ever passes a q-value-filtered list here and a looser one to `mergeFrom`, this relaxes the filter. Might be worth a clause in the remarks.

acesnik (approved):
🤖 Approving. The bug is real and precisely described: `CalculatePeptideResults` blanks intensity, detection type and RT for every peptide/file pair and then gates repopulation on `_peptideModifiedSequencesToQuantify`, so a peptide that arrived via `mergeFrom` is zeroed and never restored.

Worth recording that no current output is affected: `MergeResultsWith` has no caller in mzLib or MetaMorpheus, so nothing shipped reaches this path today. That argues for merging rather than against it — it is public API, and the next caller would have hit this silently.

The copy-instead-of-alias in the constructor is the part I expected to break something, and it doesn't. `FlashLfqEngine.PeptideModifiedSequencesToQuantify` is `{ get; init; }` and every other use of it is a `.Contains` read, so nothing depends on sharing the reference — and without the copy the `UnionWith` would mutate the engine's own filter set.

The test holds up where this kind of test usually doesn't: `Condition` participates in `SpectraFileInfo.Equals`, so the two empty-path files really are distinct dictionary keys. The note about the two runs needing different sequences is the right thing to have written down.

One thing outside the diff, so I could not put it in place: the aliasing this PR fixes is still present twice further down the same method — `Peaks.Add(fromPeaks.Key, fromPeaks.Value)` aliases mergeFrom's list, and `PeptideModifiedSequences.Add(pep.Key, pep.Value)` shares the `Peptide` instance, so after a merge mutating one result mutates the other. Pre-existing, and unreachable for the same reason as above; flagging it because you were already in there.

PR #1153, opened 2026-08-13
Stop a slow Windows build job reading as a failed one

acesnik:
🤖 The `build` job allowed 15 minutes for work whose test step alone varies between roughly 6 and 13 minutes on `windows-latest`. The spread straddled the limit, so the job was cancelled mid-run and reported a **red required check on PRs whose tests had actually passed**.

The log of a cancelled run ends:

```
Build succeeded.  0 Warning(s)  0 Error(s)
Test Run Successful.  Total tests: 5421  Passed: 5392  Skipped: 29
Total time: 12.7309 Minutes
##[error]The operation was canceled.
```

Everything passed. The step then spent its remaining seconds in teardown — the log shows three orphan `dotnet` processes being terminated — and the job hit the wall.

### It is not rare, and it is not confined to PRs

Three of the last eight `build` jobs died this way, all within five seconds of the same wall, two of them on `master`:

| Run | Branch | Result | Build job |
|---|---|---|---|
| [31627167078](https://github.com/smith-chem-wisc/mzLib/actions/runs/31627167078) | master | cancelled | 15.08m |
| [31621174182](https://github.com/smith-chem-wisc/mzLib/actions/runs/31621174182) | master | cancelled | 15.13m |
| [31712557499](https://github.com/smith-chem-wisc/mzLib/actions/runs/31712557499) | #1148 | cancelled | 15.17m |

For contrast, the five most recent successful `build` jobs came in at **8.8, 9.7, 10.2, 12.6 and 14.2 minutes**. So the fastest had six minutes of headroom and the slowest had 48 seconds. Nothing distinguishes the cancelled runs from the successful ones except runner speed.

Raising the budget to 30 is not masking a hang. The `Test` step's own duration ranges 6.3–12.7 minutes across those runs — ordinary `windows-latest` variance of about 2×, on a suite of 5,421 tests — and the job completes normally whenever the runner is not on the slow end.

### `release.yml` had the same budget, and worse exposure

Its `Test` step runs the whole suite with **no `Category` filter**, unlike the PR job's `Category!=ExternalService`, making it strictly slower than the job that already overran. Because the timeout aborts before `Pack and Push`, a tag push would silently fail to publish rather than publishing something broken — but it would still mean a release that looks initiated and never lands. Raised to 30 as well.

That one is worth noticing now rather than later: the next mzLib release is what smith-chem-wisc/MetaMorpheus#2718 is waiting on.

### Why 30

It is a little over twice the slowest observed success, and sits sensibly against the neighbouring budgets in the same workflow — `external-service-tests` at 20 and `integration` at 45, both untouched here since neither has overrun. It leaves room for the suite to keep growing without anyone having to revisit this.

The alternative would be to chase the teardown hang itself, most likely the coverlet data collector not releasing its test hosts. That is worth doing if the orphan processes turn out to cost real minutes, but it is a different change, and it would not have saved the two `master` runs above — those had the same 15-minute ceiling regardless.

### Risk

A cancelled job currently produces no coverage upload, so Codecov silently reports nothing for the affected PR. Beyond that, the only cost of a larger timeout is that a genuinely hung job burns 30 runner-minutes instead of 15 before being killed.


trishorts (approved):
Approve. I verified the diff is complete: master carries exactly two `timeout-minutes: 15` (`dotnet.yml:15`, `release.yml:14`) and this changes both. The untouched neighbours are right to leave alone — `integration` measured 16.0–16.7 min across five recent runs, so 45 has plenty of headroom.

I also checked the premise independently. Five recent `build` jobs came in at 7.7, 8.9, 8.9, 9.0 and 10.2 minutes, with the Test step alone at 6.87. That matches the 8.8–14.2 range here and confirms the spread genuinely straddles a 15-minute ceiling rather than there being a hang to chase.

Three comments, none blocking.

**The `release.yml` half is undersold.** It is the more consequential of the two and currently reads as a footnote. That job runs the suite with no `Category` filter, so it includes the network-dependent external-service tests — strictly slower and more variable than the PR job that already overran — and its failure mode is worse: a silent non-publish on a tag push.

**One correction on the coverlet aside.** The PR suggests the teardown hang is "most likely the coverlet data collector not releasing its test hosts". I measured collection directly on this suite: **214.7s with `--collect:"XPlat Code Coverage"` versus 212.4s without** — a 2.3s difference, within noise. That does not disprove the teardown theory (orphan processes at shutdown are a different thing from per-test slowdown), but anyone chasing coverlet expecting to reclaim *minutes of test time* will come back empty. Worth knowing before someone spends a day on it.

**The Codecov note belongs under benefits, not risks.** A cancelled job uploading no coverage is a reporting hole this change closes.

Separately, and not for this PR: `check-thermo-rawfilereader.yml` has no `timeout-minutes` at all, so it inherits the 360-minute default. Same class of bug, worth its own issue.

nbollis (approved):

PR #968, opened 2025-10-25
Update Handling of Sequence Variants (Part 5): VariantApplication Improved Specificity

trishorts:
This pull request introduces improvements and clarifications to variant expansion and protein XML parsing logic in the `mzLib` library. The most significant changes include a new, more flexible variant expansion API, enhanced XML parsing documentation, and improved handling of sequence variants and modifications. The changes are grouped below by theme.

### Variant expansion API improvements

* Added a new method `GetConsensusAndVariantBioPolymers<TBioPolymerType>` to `VariantApplication`, providing a more flexible and clearly documented API for expanding biopolymers into consensus and variant isoforms. The legacy `GetVariantBioPolymers` method is now retained for backward compatibility and delegates to the new method. [[1]](diffhunk://#diff-cb8ea208f2df0e2d15577156901b251725f9384283f56ab3b732a5affc451a4fL18-R23) [[2]](diffhunk://#diff-cb8ea208f2df0e2d15577156901b251725f9384283f56ab3b732a5affc451a4fL42-R51)
* Updated parameter names and documentation to clarify the meaning and usage of isoform limits and allele depth in variant expansion. [[1]](diffhunk://#diff-cb8ea208f2df0e2d15577156901b251725f9384283f56ab3b732a5affc451a4fL18-R23) [[2]](diffhunk://#diff-cb8ea208f2df0e2d15577156901b251725f9384283f56ab3b732a5affc451a4fL42-R51)

### Sequence variant handling

* Improved the logic for creating post-application `SequenceVariation` objects in `ApplySingleVariant`, ensuring correct handling of VCF data and descriptions.
* Enhanced pruning of out-of-range sequence variants by adding logging to warn when variants are removed, improving traceability and debugging.

### Protein XML parsing and documentation

* Rewrote and clarified XML parsing method documentation in `ProteinXmlEntry`, making it easier to understand the parsing flow and the construction of `Protein` objects from XML entries. [[1]](diffhunk://#diff-b8b6bd68c05002361ff53fda9096e99b1c9965225b3508e2a72f4e2d32a3309aR58-R76) [[2]](diffhunk://#diff-b8b6bd68c05002361ff53fda9096e99b1c9965225b3508e2a72f4e2d32a3309aL293-R314) [[3]](diffhunk://#diff-b8b6bd68c05002361ff53fda9096e99b1c9965225b3508e2a72f4e2d32a3309aL312-R324)
[...]

PR #1154, opened 2026-08-14
A variant database was not one decoy per target, each a reversal of it

acesnik:
🤖 ## The problem

Target-decoy FDR rests on one assumption: that the decoy database is a stand-in for the target database, matched entry for entry. Every target has exactly one decoy, and that decoy is a reversal of it — same length, same residue composition, cleavage sites preserved. A decoy hit is then a proxy for a false target hit at the same score, and the ratio estimates the false discovery rate.

For a database carrying sequence variants, that did not hold. On a U2OS Spritz database of 65,873 canonical entries:

- **337 more decoys than targets** — 96,762 against 96,425.
- **459 decoys were not a reversal of any target.** The shortest was **2 residues**. The longest was 3,876.
- **122 targets had no decoy at all**, so nothing calibrated them.
- **229 decoys shared an accession with a different decoy sequence**, so the name did not identify the entry. Spritz's accession-keyed summary tables had 232 duplicate keys, and its variant-description table silently omitted variations that were present in the sequence.

The failure is not in one direction. A 2-residue decoy is never identified, so it pads the decoy count without contributing hits. A 3,876-residue one that mirrors nothing can contribute hits that reflect no target. And 122 targets had no counterpart at all. So the decoy space stopped matching the target space in a way that is not uniformly conservative or anti-conservative — which is worse than a bias with a known sign, because it cannot be reasoned about.

To my knowledge no published result rests on these databases yet, which is an argument for fixing it now rather than later.

## Why it happened

Decoys were built by reversing a protein's sequence variations and then expanding the decoy separately, so the decoy half came off the same pipeline as the target half. That instinct is sound — a null should reflect the process that produced the targets, which is why shuffling preserves cleavage sites instead of being uniform-random. It does not survive the coordinate change.

Reversal maps a variation at `[b,e]` onto `[L-e+2, L-b+2]`. Two variations sharing an **end** on the target therefore share a **begin** on the decoy, and overlapping variations compose differently under those two geometries. `Q9UHF7`, length 1281, two variations both ending at the C terminus:

- **target, both applied:** the higher-begin edit goes first, so the second indexes past the shortened sequence, its tail is dropped, and the result is byte-identical to that edit alone. `ApplyVariants` ends with `GroupBy(x => x.BaseSequence)`, so the duplicate is deduplicated away — **3 entries**.
- **decoy, both applied:** reversal put them at the same begin, so the two edits compose into a new, shorter sequence of 42 residues that mirrors no target — **4 entries**.

Ordering cannot repair this. Forcing the wider edit last (`ThenBy(OneBasedEndPosition)`) recovered 21 of the surplus and left `Q9UHF7` untouched: a variation ending at the target's last residue has no tail, while its mirror begins at position 2 and has a large one, and tails are what make composition order-dependent. That is geometry, not sequencing.

## The fix

Expand the targets first, then mirror each expanded entry. One decoy per target, same length, same composition — by construction, rather than by two derivations agreeing.

This keeps the variant character of the null, which was the reason for running decoys through the same pipeline: the reversal of a variant protein still contains that variant's junction, so variant peptides still have decoys to be scored against. 19,197 variant targets, 19,170 variant decoys.

Measured end to end through Spritz's own `SpritzModifications`, against a control built from this same branch point **without** these commits, so the numbers isolate this change rather than everything since the last release:

| `combined.spritz.snpeff.protein.withdecoys.fasta` | before | after |
|---|---|---|
| targets | 96,425 | 96,425 |
| decoys | 96,762 | **96,425** |
| surplus decoys | 337 | **0** |
| decoys mirroring no target | 459 | **0** |
| targets with no decoy | 122 | **0** |
| duplicate decoy accessions | 229 | **0** |

Spritz's summary now doubles exactly: 65,873 → 131,746 canonical entries, 79,296 → 158,592 total.

Four things this needed beyond the reordering, one per commit:

1. **A variation that cannot be reversed no longer empties the decoy.** `GetVariantBioPolymers` only takes the combinatorial path when every variation is valid, otherwise falling through to genotype-aware application, which drops variations without genotype data — so a single unreversible variation discarded *every* variation on that decoy. `O43653.xml` gave four targets and one decoy. Also fixes an insertion after the initiator methionine reversing to a variation that begins one past the end of the sequence, and an `ArgumentOutOfRangeException` when reversing an applied stop gain.
2. **Every applied variation is recorded, so an accession names one sequence.** Overlapping edits are spliced together rather than replacing one another — `OverlappingVariants_NullVcf_ProducesCorrectCombinedSequence` pins that deliberately — but the record dropped any earlier variation whose span the current edit "included", and a deletion's post-application span collapses to the single residue it leaves behind. An edit therefore stayed in the sequence and vanished from the name.
3. **A decoy is named for the sequence it holds**, rebuilt from the reversed variations: `DECOY_P04406_A315G`, not `DECOY_P04406_A22G`. `TestThatProteinVariantsAreGeneratedDuringRead` already asked for this and was failing.
4. **A mirrored decoy carries a mirrored consensus.** `Protein` sets `NonVariantProtein` to itself when none is given, which made every variant decoy its own consensus — and Spritz reads `NonVariantProtein` to separate canonical from variant entries and to group variants by gene. Each distinct consensus is mirrored once and passed in, for which the main `Protein` constructor takes a new optional `nonVariantProtein` argument, appended and defaulted so existing callers are unaffected.

A stop gain also keeps an annotation. An applied stop gain truncates the sequence while its variation still records the position it cut at, so the original residue is no longer there to reverse; the decoy records the stop against its own last residue, which keeps the entry identifiable as a stop gain rather than quietly becoming an ordinary decoy.

This is the substantive half of smith-chem-wisc/Spritz#220; the Spritz-side PR closes that issue, and Spritz users see the change once this ships and its `PackageReference` is bumped.

## Target entries are renamed, and their sequences are not

The one part of this that touches the target half, stated plainly:

- **Target sequences are unchanged** — all 79,292 compare identical as a multiset against the control, zero differences either way.
- **419 of 79,292 target names change (0.53%)** — the entries from point 2, whose accession and description now list a variation that was already in the sequence and missing from the label.

Anyone holding search results keyed on those 419 accessions will see the key move while the sequence behind it stays the same. On the decoy side Spritz's accession-keyed table went from 38,676 rows over 38,444 distinct keys to 38,367 over 38,367, and its variant-description table gained roughly 1,100 rows — the applied variations previously left out.

## Known limitations

- 19,197 variant targets against 19,170 variant decoys. The 27 are single-residue start losses, `M` to a non-`M` residue: dropping the initiator methionine and reversing the rest changes both ends of the sequence at once, which no single contiguous variation can express. Those decoys mirror their target's sequence correctly and are simply not labelled as variants.
- `VariantApplication.RestoreModificationIndex` now sums a retained deletion's length change. That is the correct consensus index rather than the previous one, but it moves GPTMD's reported position in that case (MetaMorpheus `TaskLayer/MetaMorpheusTask.cs:899`).

## Tests

166 green across the variant and decoy suites, including the deliberate overlapping-variants test. The new tests state the property as an oracle — reversal must commute with variant application — with five ordinary cases (substitutions, indels, multi-residue substitutions) as positive controls that pass before and after, plus accession uniqueness across a target-decoy pair.

Three existing tests selected entries by list position or relied on the order of `AppliedSequenceVariations`, both of which mirroring changes. They now select by content and read variations in position order; the variations and their mirrored coordinates are asserted exactly as before. I verified the mirroring independently first — `M1646V`→`M62V`, `N1488D`→`N220D`, `V409L`→`V1299L`, `K781EEK`→`K922KEE` — so those edits are not covering a regression.


trishorts (commented):
Reviewed against `smith/master @ 5ba13155`, with the findings below reproduced on a worktree of this branch rather than read off the diff.

The direction is right and the diagnosis in the description is careful. Expand-then-mirror is the correct shape; `ReverseOne`'s coordinate math is master's own formula carried through unchanged; and the new optional `nonVariantProtein` parameter is a line-for-line copy of the idiom already on the clone-with-overrides constructor (`Protein.cs:149/173`), so it introduces no overload ambiguity and no positional call site in the repo reaches past `isEntrapment`.

Three things to settle inline, two of them with a reproduction attached. Plus two coordination notes:

- **#968 touches the same code.** *Update Handling of Sequence Variants (Part 5)* rewrites the same `ApplySingleVariant` region and renames `GetVariantBioPolymers` to `GetConsensusAndVariantBioPolymers` behind a legacy shim. It is already `CONFLICTING` with master. Whichever lands first, the other needs a rebase that will not be mechanical, so the order is worth agreeing now.
- **The RNA path keeps the ordering this PR argues against.** `RnaDbLoader.cs:386` is the same `GenerateDecoys(targets, ...)` then `SelectMany(GetVariantBioPolymers)` shape, and `RnaDecoyGenerator.cs:82-83` assigns `reverseBegin`/`reverseEnd` unswapped where the truncation loop three lines below swaps correctly. Out of scope for this PR - flagging it so it can get its own issue rather than quietly diverging from the protein side.

mzLib/Omics/BioPolymer/VariantApplication.cs
```diff
@@ -261,10 +261,15 @@ private static TBioPolymerType ApplySingleVariant<TBioPolymerType>(SequenceVaria
...
+            // Keep every previously applied variation. Overlapping edits are spliced into one sequence rather
+            // than replacing one another, so an earlier edit stays in the sequence and has to stay in the
+            // record - the accession is built from this list. Includes() compares post-application spans, and
+            // a deletion's span collapses to a single residue, so it read as overwritten by any later edit
+            // beginning at the same position: two entries then differed by a deletion while carrying the same
+            // accession. Reversal makes that common, because it maps variations sharing an end onto
+            // variations sharing a begin.
             appliedVariations = appliedVariations
-                .Concat((protein.AppliedSequenceVariations ?? Enumerable.Empty<SequenceVariation>()).Where(x => !variantGettingApplied.Includes(x)))
+                .Concat(protein.AppliedSequenceVariations ?? Enumerable.Empty<SequenceVariation>())
```
trishorts, line 272: Removing the filter outright also retains variations that a later edit genuinely *did* overwrite, and what it leaves behind is not benign.

A/B on this one line, nothing else on the branch changed:

```csharp
var p = new Protein("ACDEFGHIKL", "D2", sequenceVariations: new List<SequenceVariation> {
    new SequenceVariation(5, 5, "F", "W", "narrow", (Dictionary<int, List<Modification>>)null),
    new SequenceVariation(3, 8, "DEFGHI", "QQ", "wide", (Dictionary<int, List<Modification>>)null) });
p.GetVariantBioPolymers(4, 1);
```

with master's `.Where(x => !variantGettingApplied.Includes(x))`:

```
D2                seq=ACDEFGHIKL  applied=0
D2_F5W            seq=ACDEWGHIKL  applied=1
D2_DEFGHI3QQ      seq=ACQQKL      applied=1
D2_DEFGHI3QQ      seq=ACQQKL      applied=1   <- identical pair, Merge collapses it
```

with the line as written here:

```
D2                seq=ACDEFGHIKL  applied=0
D2_F5W            seq=ACDEWGHIKL  applied=1
D2_DEFGHI3QQ      seq=ACQQKL      applied=1
D2_F0W_DEFGHI3QQ  seq=ACQQKL      applied=2   <- new accession, same sequence
```

The retained variation is `[0,0] 'F'->'W'`. Three things follow from it:

- `OneBasedBeginPosition == 0`, which `SequenceVariation.AreValid()` (`begin > 0`) rejects. Commit 1 of this PR exists precisely because a single invalid variation poisons a whole entry; this line manufactures them.
- `ACQQKL` contains no `W`, so the accession names a variation the entry does not carry — commit 2's stated goal, inverted.
- `Merge`'s key is `(Accession, BaseSequence, IsContaminant, IsDecoy, IsEntrapment)`, so the two `ACQQKL` rows no longer collapse. That is a surplus target entry that did not exist before.

`RestoreModificationIndex` sums `VariantSequence.Length - OriginalSequence.Length` over exactly this list, and its only production caller is MetaMorpheus's GPTMD pruned-database writer (`TaskLayer/MetaMorpheusTask.cs:899`), where the returned index is the residue a PTM gets written at. A retained-but-overwritten variation with a non-zero length delta moves it.

The defect you describe is real — `Includes()` compares a pre-application span against post-application spans, and a deletion's post-application span collapses to one residue. But that argues for making the comparison like-for-like, or exempting deletions, rather than dropping the filter: as written it trades "an edit that is in the sequence goes missing from the name" for "an edit that is not in the sequence appears in the name."

mzLib/UsefulProteomicsDatabases/DecoyGeneration/DecoyProteinGenerator.cs
```diff
@@ -37,155 +37,193 @@ public static List<Protein> GenerateDecoys(List<Protein> proteins, DecoyType dec
         /// <returns></returns>
         private static List<Protein> GenerateReverseDecoys(List<Protein> proteins, int maxThreads = -1, string decoyIdentifier = "DECOY")
         {
+            // Mirror each distinct consensus once, so that a decoy built from a variant-applied protein points
+            // at a mirrored consensus instead of at itself. Consumers read ConsensusVariant / NonVariantProtein
+            // to tell a canonical entry from a variant one and to group an entry's variants by gene, so a decoy
+            // left as its own consensus makes every variant decoy look canonical.
+            // Keyed by reference: Protein overrides Equals, and this map is about object identity.
+            var consensusDecoys = new Dictionary<object, Protein>(ReferenceEqualityComparer.Instance);
+            foreach (Protein target in proteins)
```
trishorts, line 46: This pre-pass fires for **every** entry of **every** XML database, including ones with no sequence variations at all.

`GenerateDecoys` is now handed `expandedTargets`, and `GetVariantBioPolymers` always routes through `ApplyVariants` → `CreateVariant` → `new Protein(...)` even when `SequenceVariations` is empty (`uniqueEffectsToApply.Count == 0` returns `{ proteinCopy }`). So an entry is always a *copy* whose `ConsensusVariant` is the original loaded object, `!ReferenceEquals(consensus, target)` is never false, and the map ends up with one entry per protein in the file.

Measured on this branch, 20,000 proteins x 405 aa, **zero** sequence variations, `maxThreads = -1`:

```
raw targets      (ConsensusVariant == self, pre-pass skipped): 106 ms,     0 / 20000 decoys with a foreign consensus
expanded targets (pre-pass fires for every entry)            : 239 ms, 20000 / 20000
```

2.25x, and the added half is a serial `foreach` in front of the `Parallel.ForEach`. It also retains a duplicate `Protein` per decoy — for the 96,425-entry database in the description that is tens of thousands of extra full protein objects whose only job is to be pointed at by a decoy that is byte-identical to them.

A canonical entry does not need a mirrored consensus; only one carrying applied variations does. Gating on that skips the pre-pass entirely for a plain UniProt XML while keeping it for the Spritz case:

```csharp
if (target.AppliedSequenceVariations is { Count: > 0 }
    && target.ConsensusVariant is Protein consensus
    && !ReferenceEquals(consensus, target)
    && !consensusDecoys.ContainsKey(consensus))
```

Where it does still fire, this loop is the one part of decoy generation that is not parallel — worth a `Parallel.ForEach` over the distinct consensuses with a concurrent map, given the box sizes these databases are built on.

mzLib/UsefulProteomicsDatabases/DecoyGeneration/DecoyProteinGenerator.cs
```diff
@@ -241,6 +279,22 @@ private static List<SequenceVariation> ReverseSequenceVariations(IEnumerable<Seq
...
+                    // An applied stop gain truncates the sequence at the stop, but the variation it records
+                    // still points at its position in the untruncated sequence, which is then past the end - so
+                    // the original residue it replaced is no longer there to be read and reversed. Record the
+                    // stop against the last residue the decoy does have instead, keeping the position, so the
+                    // entry is still identifiable as a stop gain rather than losing its annotation entirely.
+                    if (sv.OneBasedEndPosition > reversedSequence.Length)
+                    {
+                        if (reversedSequence.Length > 0)
+                        {
+                            decoyVariations.Add(new SequenceVariation(sv.OneBasedBeginPosition, sv.OneBasedBeginPosition,
```
trishorts, line 291: The position and the residue come from different places here: the position is `sv.OneBasedBeginPosition`, carried over unchanged, while the residue is `reversedSequence[^1]` — the last one. Those coincide only when the stop sits exactly one past the end, which is the single-residue case.

With a multi-residue stop-gain variant sequence they part company:

```csharp
var consensus = new Protein("MCDEAKKKPPPP", "S2");
var sv = new SequenceVariation(5, 8, "AKKK", "EKR*", "sg", (Dictionary<int, List<Modification>>)null);
var variant = new Protein("MCDEEKR", consensus, new[] { sv }, null, null, null);
DecoyProteinGenerator.GenerateDecoys(new List<Protein> { variant }, DecoyType.Reverse, 1, "DECOY");
```

```
decoy.BaseSequence = MRKEEDC
decoy sv  begin=5 end=5  orig='C'  var='*'      <- position 5 holds 'E'; 'C' is at position 7
```

The entry stays identifiable as a stop gain, which is the goal, but it now records a residue that is not at the position it records. Anchoring both on the same place — `reversedSequence.Length` for the position too, as the insertion branch below already does — would keep them consistent. Is the original position deliberately preserved here for something downstream?

nbollis (approved):

Alexander-Sol (approved):

PR #1155, opened 2026-08-14
Build PEP training rows in a fixed order, so match-between-runs is reproducible

### LINKED ISSUE #1111: FlashLFQ peptide roll-up nondeterministically drops MBR intensities, making protein quant irreproducible run-to-run

acesnik:
🤖 Fixes #1111.

Two runs of `FlashLfqEngine` over identical inputs, match-between-runs on, returned different peptide intensities and different protein results. The peaks underneath were stable the whole time — 652 peaks, 145 of them MBR, 354 peptides, every run — and only the roll-up on top of them moved. A peptide whose sole peak was a genuine MBR transfer reported its intensity in one run and zero in the next.

Thanks @trishorts for a report that did most of the work: the observation that the peak lists were identical while the peptide table was not is what made this findable, and the six peptides listed in the issue are exactly the ones I ended up watching.

## The cause

Training rows for the PEP model were assembled by giving each parallel chunk its own list and appending it under a lock as it finished, so the row order depended on which chunk finished first:

```csharp
lock (ChromatographicPeakDataListLock)
{
    ChromatographicPeakDataList.AddRange(localChromatographicPeakDataList);
}
```

FastTree is configured to be as deterministic as it can be — `NumberOfThreads = 1`, `Seed = 42`, `FeatureSelectionSeed = 42`, `RandomStart = false`, and an `MLContext` seeded to 42 — but none of that helps when the *rows* arrive in a different order, because order changes its binning and split choices. So the model differed, and so did the PEP it predicted for every peak. `CreateChromatographicPeakDataIteration` had the same pattern, and it feeds the nine iterative training rounds, so an unstable order reached the trainer ten times per run.

Both now write into one slot per donor and flatten in donor order, which is what a sequential run would produce.

## Why a tiny PEP change moves a whole peptide

`AssignMbrQValues` sorts by `MbrPep` before walking targets and decoys, so a changed PEP changes the walk order, the running decoy counts, and every q-value downstream of the swap. `CalculatePeptideResults` then keeps an MBR peak only while `MbrQValue < MbrQValueThreshold`, so a peak that crosses the threshold drops out of the roll-up entirely and its peptide keeps the zero it was initialised with — reported as `NotDetected`.

The PEP values in play sit at the floating-point noise floor: `0`, `5.96e-08`, `1.19e-07`. Sorting those as if they were an order is what turns numerical noise into a categorical difference in `QuantifiedPeptides.tsv`.

## Measured

mzLib's own FlashLFQ test data — `AllPSMs.psmtsv` with the two K562 mzMLs, `MaxThreads = -1` — comparing two runs in one process, five times:

| | MBR peaks of 145 with a different PEP |
|---|---|
| before | 10, 11, 9, 7, 8 |
| after | 0, 0, 0, 0, 0 |

Peptide intensities differing across six runs went from 8 to 0. `MbrScore` was stable throughout, before and after, which is what pointed at the model rather than at peak finding.

On an older branch point, before mzLib moved to net10.0, the same measurement gave 95, 91, 5, 96, 7, 11, 12, 15, 142 and 94. The magnitude varies with the runtime; the defect does not.

## Test

`Test/FlashLFQ/TestMbrDeterminism.cs` quantifies twice and asserts that both the per-peak PEP and the peptide intensities agree.

It asserts on PEP deliberately. My first version compared only intensities, and it caught the defect **once in five attempts** — because intensities move only when a changed PEP happens to push a q-value across the reporting threshold. A regression test that passes four times in five when the bug is present is worse than none, since it reads as protection. Asserting on PEP catches it five times in five.

## Two things this does not do

**The cross-process irreproducibility in the report did not reproduce here.** Six fresh processes gave byte-identical output on this machine *before* the fix. What I could reproduce was two runs inside one process disagreeing, which shares the mechanism, so this may well be the same defect seen through a different door — but if the reporter still sees run-to-run differences between separate invocations after this lands, something else contributes and the issue should stay open.

**It does not touch the `NotDetected` labelling.** Reporting a peptide as not detected, intensity zero, when its only peak is a real MBR transfer that failed a q-value cutoff, is a labelling question rather than a determinism one: losing a transfer to a cutoff is defensible, calling it not detected is not. It deserves its own issue and its own before-and-after, and it is easier to argue about now that the behaviour is reproducible. I will file it separately.

## Verification

61 passed and 3 failed with this change; 60 passed and 4 failed without it, the extra failure being the new test. The three that fail either way are pre-existing at this branch point on macOS: `RealDataMbrTest` (expects 141 MBR peaks, finds 140), `TestFlashLfqDoesNotRemovePeptides`, and `TestSpectraFileInfoString`. `RealDataMbrTest` is worth singling out because an MBR count is exactly what this change could have moved — it fails identically with and without.

Incidentally, Thermo `.raw` reading does work on macOS arm64, contrary to what I first assumed when those tests failed; they pass once the fixtures are laid out correctly.


trishorts:
🤖 Reviewed locally against `smith/master @ 03081569`, PR head `0efac2ce` checked out and built on Windows. Everything below I ran; I have said where I did not.

**This is a good fix and an unusually good regression test, and I want to be specific about why rather than just saying so.**

I reverted `PepAnalysisEngine.cs` alone to master, kept the new test, and re-ran:

| | result |
|---|---|
| new test vs **master** engine | **fails** — `12 of 145 MBR peaks got a different posterior error probability on the second run` |
| new test vs **this branch**, 3 consecutive runs | **3/3 pass**, ~19 s each |

So the test genuinely regresses, which is rarer than it should be. The failure message names the peak and both values (`20100614_Velos1_TaGe_SA_K562_3|IVVNLTGR|2: 6.5565109252929688E-07 then 4.76837158203125E-07`), which is exactly what someone debugging a future regression needs. Your point in the description about asserting on PEP rather than intensity is borne out — 12 of 145 PEPs moved while the peptide table would have caught far fewer.

`RealDataMbrTest` **passes on Windows at this branch point** (36 s, 141 MBR peaks as asserted). That was the test most likely to move if this change had shifted results, so its being green is worth stating. The 140-vs-141 failure you saw is macOS-local, not something this PR causes.

## On cross-process reproducibility — corroborating you

Out of scope for this PR, but since #1111 hinges on it: I dumped per-peak PEP plus the raw stored peak sequence from **three separate processes** and diffed. All three byte-identical, both the upstream peak ordering and the PEP values.

That independently reproduces your six-process observation. If the reporter still sees cross-invocation differences after this lands, it is something environmental or something else in their pipeline — this dataset does not show it on Windows either.

## One thing I would add before merge

**The test has no guard against passing vacuously.** Three separate ways it can go green while checking nothing:

- if MBR produced zero peaks, both dictionaries are empty, `secondPeps.Count == firstPeps.Count` is `0 == 0`, and both `Where` filters yield nothing;
- if the PEP model never runs, every `MbrPep` is `null`, and `v != kv.Value` is `null != null` → false for every peak;
- `Assert.That(secondPeps.Count, Is.EqualTo(firstPeps.Count))` compares the two runs to each other, never to a floor.

None of this is happening today — I measured it: **145 MBR peaks, 0 null PEPs, 116 distinct PEP values**. So the test is doing real work right now. The gap is that nothing keeps it doing real work. A future change that quietly stops the PEP path from running (the `>100 peaks` guard in `FlashLfqEngine`, an exception swallowed upstream, a fixture that shrinks) turns this from a regression test into a green tick.

You already solved exactly this problem in #1156, with `Assert.That(byName, Has.Count.GreaterThan(10), "expected to scan the build output; the directory looks wrong or empty")`. The same one-liner here — assert the peak count is in the expected range and that the PEPs are non-null — would make this test as self-defending as that one.

## Something I checked and want to rule out explicitly

I looked hard at whether the PEP dictionary key `$"{file}|{modSeq}|{charge}"` can collapse two distinct MBR peaks — a donor's target and its `RandomRt` decoy share a donor `Identification`, so it looked plausible on paper. **It does not happen:** 145 MBR peaks produce 145 distinct keys, zero duplicates, which matches the 145 in your description exactly. Flagging it only so nobody else spends an afternoon on it.

## Not this PR, but found while tracing it

`FlashLfqEngine` appears to add the same peak to the results list twice. In `MatchBetweenRuns`, `_results.Peaks[acceptorFile].Add(best);` occurs inside `if (peakHypotheses.Count > 0)` and again unconditionally a few lines later, with no `continue` between them — so whenever there is more than one hypothesis, `best` is appended twice. `RunErrorChecking` then dedupes on apex and takes the MBR/MBR same-sequence branch, which calls `MergeFeatureWith` on the peak and itself.

I have **not** established that this inflates intensity — the final peak count comes out right (145), so the dedup is clearly catching it — and `FlashLfqEngine` is explicitly outside this PR's scope. But it may well be the answer to the `// Eliminate duplicate peaks (not sure where they come from)` comment further down the same method. Happy to open an issue with a proper before/after rather than leave it in a PR thread.

## Non-blocking

- `[ExcludeFromCodeCoverage]` — 5 of the 7 fixtures in `Test/FlashLFQ/` carry it; the new one doesn't. One line, purely for consistency.
- `ReadIdentifications` hand-rolls what `MzLibExtensions.MakeIdentifications` already does, which also sets `useForProteinQuant` and de-duplicates `ProteinGroup` by accession. In its favour, your version uses `SpectrumMatchTsvReader.ReadPsmTsv` and so avoids the hardcoded column indices that the two existing FlashLFQ tests already disagree about (`split[22]` vs `split[21]`) — so this may be the better pattern and the others the ones to change.
- `decoy: psm.DecoyContamTarget == "D"` misses the pipe-delimited ambiguous form (`"D|D|D"`), which would be read as a target. Same as the existing tests do it, so this is a pre-existing convention rather than something you introduced.
- The intensity comparison walks `firstIntensities` only, so a key present in the second run but absent from the first would not be reported. The PEP assertion has the same shape but is covered by the preceding count check.
- `[NonParallelizable]` is a no-op here — `mzLib/Test` has no assembly-level `[Parallelizable]` and no `.runsettings`, so NUnit is already `ParallelScope.None`. Harmless and there is precedent for it; just noting it isn't buying anything.

Runtime is a non-issue, in case anyone raises it: 18.9 s for the new test against `RealDataMbrTest`'s 36 s, and CI `build` on this PR was 9m20s against 9-10m on recent master jobs.

## What I did not check

I did not run on macOS or Linux, so the platform observations in your description stand on your testing. I also did not verify that the new fixed row order produces a *better* model than the old varying one — only that it produces a stable one, which is what the PR claims.

Alexander-Sol (approved):

mzLib/FlashLFQ/MBR/PEP/PepAnalysisEngine.cs
```diff
@@ -430,38 +426,40 @@ public IEnumerable<ChromatographicPeakData> CreateChromatographicPeakData(List<D
...
 
+            // One slot per donor, filled in place, then flattened in donor order. Appending each thread's
+            // chunk under a lock instead made the row order depend on which chunk finished first, and the
+            // trainer is sensitive to row order: FastTree runs single-threaded from a fixed seed, but the
+            // order rows arrive in changes its binning and split choices, so the model - and every PEP it
+            // predicts - differed between runs. Those PEPs order the target-decoy walk that assigns
+            // MbrQValue, and the peptide roll-up keeps an MBR peak only while its q-value is under the
+            // threshold, so a peptide whose only peak is an MBR transfer flipped between its real intensity
+            // and zero from one run to the next.
+            List<ChromatographicPeakData>[] rowsByDonor = new List<ChromatographicPeakData>[donorIndices.Count];
```
trishorts, line 437: Worth noting so it does not get re-litigated: this is the house idiom, not a new one. `FlashLfqEngine.QuantifyMs2IdentifiedPeptides` already pre-allocates `ChromatographicPeak[]` over the same `Partitioner.Create(0, n)` shape and flattens after the loop (`FlashLfqEngine.cs:488-551`), and `SpectrumMatchTsvReader.cs:45-79` does the same while deliberately leaving the order-insensitive `warnings` on a `ConcurrentBag`.

Minor: the `rows != null` filter on the way out is unreachable here — the partitioner covers every index in `[0, donorIndices.Count)`, each iteration assigns its own slot, and a throwing body would surface as an `AggregateException` rather than a partly-filled array. It does mirror `SpectrumMatchTsvReader.cs:78`, where it genuinely is reachable, so keep it if you prefer the symmetry.

mzLib/Test/FlashLFQ/TestMbrDeterminism.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <summary>
+        /// Per-MBR-peak posterior error probability, keyed so the same transfer is comparable across runs.
+        /// This is the sensitive measure: the peptide intensities only move when a changed PEP happens to
+        /// push a q-value across the reporting threshold, which made an intensity-only test miss the defect
+        /// four times in five.
+        /// </summary>
+        private static Dictionary<string, double?> QuantifyAndTakePeakPeps(out Dictionary<string, double> intensitiesOut)
+        {
+            var files = FileInfos();
+            var results = new FlashLfqEngine(ReadIdentifications(files), matchBetweenRuns: true, maxThreads: -1).Run();
```
trishorts, line 55: `maxThreads: -1` runs through `FlashLfqEngine.cs:98-105`, which maps it to `Environment.ProcessorCount - 1` and then clamps to a floor of 1. On a 1- or 2-core machine that resolves to 1, `MaxDegreeOfParallelism` is 1, and the loop this PR fixes runs sequentially — so the test goes green whether or not the fix is present.

It is fine on the current `windows-latest` runner, and I confirmed it fails 3/3 locally with `PepAnalysisEngine.cs` reverted. But the protection is silently core-count dependent, which is the same failure mode you argued against for the intensity-only version. Probably worth making explicit — e.g. `Assert.Ignore` when `Environment.ProcessorCount < 3`, so a small runner reports "not exercised" rather than a pass.

mzLib/Test/FlashLFQ/TestMbrDeterminism.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var peps = new Dictionary<string, double?>();
+            foreach (var fileAndPeaks in results.Peaks)
+            {
+                foreach (MbrChromatographicPeak peak in fileAndPeaks.Value
+                    .Where(p => p.DetectionType == DetectionType.MBR)
+                    .Cast<MbrChromatographicPeak>())
+                {
+                    Identification donor = peak.Identifications.First();
+                    peps[$"{fileAndPeaks.Key.FilenameWithoutExtension}|{donor.ModifiedSequence}|{donor.PrecursorChargeState}"] = peak.MbrPep;
```
trishorts, line 65: Indexer assignment, so a repeated key silently overwrites and shrinks what is compared — and the `Count` check on line 99 would not notice, because both runs would shrink the same way.

It is unique on this data today (I get 145 keys against the 145 MBR peaks), but nothing enforces that. One assertion would pin it: check the key count against the number of MBR peaks you enumerated.

mzLib/Test/FlashLFQ/TestMbrDeterminism.cs
```diff
@@ -0,0 +1,121 @@
...
+            new SpectraFileInfo(Path.Combine(DataDirectory, "20100614_Velos1_TaGe_SA_K562_4.mzML"), "a", 1, 0, 0),
+        };
+
+        /// <summary>
+        /// Two runs over the same inputs, with threading left at its default, have to agree on every MBR
+        /// peak's PEP and on every peptide intensity. Before the fix the PEPs differed on 5 to 142 of the
+        /// 145 MBR peaks on every attempt, and six peptides swung between a real intensity and zero.
+        /// </summary>
+        [Test]
+        [NonParallelizable] // two full quantifications; keep them off a loaded thread pool
```
trishorts, line 93: Nit: this is inert. Nothing configures assembly-level parallelism — no `[assembly: Parallelizable]` anywhere under `Test`, and no parallel settings in the runsettings — so NUnit already runs the suite sequentially. There is precedent for the attribute (`TestTimsTofFileReader.cs`), so no objection to keeping it, but the trailing rationale promises protection it is not providing: the concurrency that matters here is the engine's own `Parallel.ForEach`, which this attribute does not touch.

mzLib/Test/FlashLFQ/TestMbrDeterminism.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// differed. PEP orders the target-decoy walk that assigns MbrQValue, and the peptide roll-up keeps an
+    /// MBR peak only while its q-value is below the threshold.
+    /// </summary>
+    [TestFixture]
+    public static class TestMbrDeterminism
+    {
+        private static string DataDirectory => Path.Combine(
+            TestContext.CurrentContext.TestDirectory, "FlashLFQ", "TestData");
+
+        private static List<Identification> ReadIdentifications(List<SpectraFileInfo> files)
```
trishorts, line 29: Optional: `FileReader.ReadQuantifiableResultFile(path).MakeIdentifications(files)` (`FlashLFQ/ResultsReading/MzLibExtensions.cs:14`) covers this whole method — protein-group interning, the `decoy:` flag, file matching — and there is a two-line worked example in `Test/FileReadingTests/InternalFileReading/TestPsmFromTsv.cs:766-772`.

Not a drop-in, though: it also populates `qValue` and `psmScore`, which this loop leaves at their defaults, so the donor set and the 145-peak baseline would shift. Either is defensible for a determinism test — raising it only because the suite already carries four hand-rolled variants of this loop.

trishorts (approved):
Reviewed with a local build and targeted runs (Windows, net10), not just a read.

**Verified**
- `Test.csproj` builds clean at `64fe559c`.
- `TestMbrDeterminism` passes on this branch (20 s).
- Reverting **only** `PepAnalysisEngine.cs` to `smith/master` and keeping the new test: fails 3/3, with 13, 2 and 53 of 145 MBR peaks moving. So the test does catch the defect — and your "assert on PEP, not intensity" call was the right one: attempt 2 moved only 2 peaks, which an intensity-only assertion would very likely have slept through.
- Full `Test.FlashLFQ`: 114/114 green, including `RealDataMbrTest` — the one existing test whose exact MBR count this change could have moved. It is green here, so its macOS failure does look environmental rather than caused by this PR.

**On the approach** — this is the house idiom rather than a new one. `FlashLfqEngine.QuantifyMs2IdentifiedPeptides` already pre-allocates `ChromatographicPeak[]` over the same `Partitioner.Create(0, n)` shape and flattens after the loop (`FlashLfqEngine.cs:488-551`), and `SpectrumMatchTsvReader.cs:45-79` does the same thing while deliberately leaving the order-insensitive `warnings` on a `ConcurrentBag`. Worth stating explicitly, because it is the strongest argument for the change: `_randomSeed = 42` and `RandomStart = false` were already in place, so row order was the last remaining leak into FastTree.

**Out of scope and non-blocking, just a question:** two MBR accumulations still resolve ties by insertion order into collections filled from `Parallel.ForEach` — `FlashLfqEngine.cs:1034` (`MaxBy` returns the *first* maximum) and `:1060` (`OrderByDescending` is a stable sort, so `peakHypotheses.First()` inherits the source order). They evidently do not bite on this data, but they are the same shape as what you just fixed. Did you look at them, and is a follow-up issue worth filing?

Issue #1145, opened 2026-08-12
Add a CI check that every mzLib.nuspec <file src> path resolves


acesnik:
🤖 Filed at @trishorts's request from the review of #1141.

`mzLib.nuspec` hand-lists every packaged assembly with a literal build-output path:

```xml
<file src="BayesianEstimation\bin\Release\net10.0\BayesianEstimation.dll" target="lib\net10.0" />
```

There are 64 such entries. Nothing verifies they resolve. `nuget pack` does not fail on a `<file src>` that matches nothing — it just omits the file, so a stale path produces a **package that builds and publishes with assemblies missing**.

The failure is silent at the point of the mistake and loud somewhere unrelated. #1141 had to rewrite 126 of these lines for the `net8.0` → `net10.0` move; a missed one would surface as `CS0246: The type or namespace name 'Modification' could not be found` in a *consumer's* build, hundreds of errors deep, pointing at consumer source files and never mentioning NuGet. We saw exactly that shape of cascade in that PR's `integration` job (743 errors) for a different root cause, which is what makes the failure mode concrete rather than hypothetical.

Suggested check, cheap to run in the `build` job after a Release build:

- parse `mzLib.nuspec`, resolve every `<file src>` relative to the repo, fail listing any that do not exist
- assert each `target="lib\<tfm>"` matches the `<TargetFramework>` of the project the source path belongs to, so a retarget cannot leave `lib\net8.0` folders fed by `net10.0` output
- optionally assert the set of packaged assemblies equals the set of non-test projects, so a *new* project silently missing from the package is also caught

The third point is worth having: today, adding a project to the solution and forgetting the nuspec entry is undetectable until a consumer misses the type.

Related: #1141 also removed two `<dependency>` entries that had drifted (`Microsoft.Win32.Registry` was declared in both dependency groups while no project referenced it), so the dependency lists are worth validating on the same pass.

PR #1156, opened 2026-08-14
Drop the architecture-stamped Thermo MassPrecisionEstimator reference, and guard against it returning

acesnik:
🤖 `Readers.csproj` carries a reference to `ThermoFisher.CommonCore.MassPrecisionEstimator`. The
assembly is IL-only, but its PE machine field is stamped `AMD64`, and CoreCLR will not load an
architecture-stamped assembly on a different architecture. Because it is consumed as a managed
`<Reference>` rather than through P/Invoke, the failure would not be a missing-library error at a call
site — it would be a `BadImageFormatException` out of the assembly loader, and it would take Thermo
`.raw` reading down with it on arm64 and on any future non-x64 target.

Nothing has ever called into it, which is the only reason `.raw` reading works on Apple Silicon today.
The CLR resolves assemblies lazily, on first use of a type, so a reference that no code touches never
loads and its stamping never comes up.

This drops the reference, its copy-to-output item and its two `mzLib.nuspec` entries, and adds tests
so a future change cannot quietly re-arm it.

## Establishing that it really is unused

Three independent checks, because removing something from a published package deserves more than a
`grep`:

- **No source reference.** Not a single `.cs` reference to the assembly or its types anywhere in mzLib
  or MetaMorpheus.
- **No metadata reference.** Reading assembly references straight out of the PE metadata of every
  managed assembly in a full MetaMorpheus build output: nothing references it. This is the strong form
  of the check, since the C# compiler omits a reference that no code uses — its absence from metadata
  *is* the evidence that no code uses it.
- **No reflective load.** The `MassPrecision` strings inside `CommonCore.Data.dll` and
  `CommonCore.RawFileReader.dll` are all m/z formatting members — `FilterMassPrecision`,
  `defaultMassPrecisionFormat`, `GetRunHeaderFilterMassPrecision`. The assembly's own name appears
  nowhere, so nothing loads it by string.

While measuring, one more useful fact fell out: it is the **only** architecture-specific managed
assembly in the entire build output. Everything else mzLib and MetaMorpheus ship is neutral. The
remaining x64 binaries (`isodeclib`, `isogenmass`, `timsdata`, `baf2sql_c`, `msvcp110`, `svml_dispmd`)
are native libraries reached by P/Invoke, which is a different mechanism with a different failure mode.

## The guard

`TestNoArchitectureSpecificAssemblyReferences` resolves every assembly reference in the build output to
the file it points at and reads that file's PE headers. Architecture-neutral means `ILOnly` set and
machine `I386`, which the runtime treats as "any architecture":

| assembly | machine | ILOnly | verdict |
|---|---|---|---|
| `ThermoFisher.CommonCore.MassPrecisionEstimator` | Amd64 | True | flagged |
| `ThermoFisher.CommonCore.Data` | I386 | True | fine |
| `ThermoFisher.CommonCore.RawFileReader` | I386 | True | fine |
| `ThermoFisher.CommonCore.BackgroundSubtraction` | I386 | True | fine |

It deliberately tests the property rather than the assembly name. A name-based ban would fail against a
future Thermo package that shipped this assembly architecture-neutral — punishing a legitimate upgrade
and pressuring whoever hit it to delete the test. Checking the stamping means such a build passes on its
own merits, and it also catches anything *else* that arrives stamped, including a regression of mzLib's
own `PlatformTarget=x64` problem from MetaMorpheus #2503/#2504, which is the same stamping from a
different cause.

## Why the assembly stays in the tree

Unreferenced and no longer shipped, but still copied into the **test** output as the classifier's input.

A scan-based test can pass for the wrong reason: point it at an empty directory, or break its header
parsing, and it reports success. So `TheClassifierRecognisesAnArchitectureStampedAssembly` asserts that
a genuinely `AMD64`-stamped binary is classified as such, and this is the only one available to assert
it against. A second control asserts that a reference in real use (`Readers` → `CommonCore.Data`)
resolves and classifies as neutral, and the scan additionally requires that it examined more than ten
managed assemblies.

## Verification

Three tests pass. Then, rather than trusting that, I built the actual regression: restored the reference
and added `typeof(ThermoFisher.CommonCore.MassPrecisionEstimator.PrecisionEstimate)`, which is how
someone would really reach for this.

```
Failed NoAssemblyInTheBuildOutputReferencesAnArchitectureSpecificAssembly
  Test.dll -> ThermoFisher.CommonCore.MassPrecisionEstimator (machine=Amd64, ilOnly=True)
```

Fails, and names both the offender and the reason. `Test.csproj` targets `net10.0-windows`, so I ran
these through a throwaway `net10.0` harness compiling the same test file against `Readers`; Windows CI
will run them in place.

## Notes for review

- One deliberate hole: a reference that does not resolve to a file in the build output is skipped
  rather than flagged, because it cannot be classified. Such an assembly would fail to load for its own
  reasons anyway.
- The scan covers third-party assemblies too, not just mzLib's own. If a dependency bump ever
  introduced an architecture-stamped reference, this test would fail — which I think is the correct
  outcome rather than a false positive, since it would mean a real platform regression. Happy to narrow
  it to mzLib-authored assemblies if reviewers prefer.
- This changes the contents of the published package by removing a file, so it wants to go out in a
  release batch rather than on its own.

Context: MetaMorpheus's README claims Thermo `.raw` is Windows and Linux only, which a search run on
both Mac architectures disproves; that is a separate one-line PR over there.


acesnik:
🤖 Tracked by #1158, which records the full cross-platform picture this PR is one piece of — what works on macOS and Linux across x64 and arm64, the four measured blockers, and the order they have to be tackled in.

Two things from this PR that ended up in there: the assembly is IL-only stamped `AMD64` rather than mixed-mode C++/CLI, making it the same problem as #1127 and MetaMorpheus #2503/#2504 rather than a separate class; and it is the only architecture-specific managed assembly anywhere in the build output, which is what makes the guard here a narrow check rather than a broad one.

trishorts:
🤖 Reviewed this locally against `smith/master @ 03081569`, with the branch checked out and built. Everything below I ran rather than reasoned about, and I have said where I did not.

**The payload is right.** I confirmed the central claim four ways rather than taking the description's word for it — parsing the PE headers and the real `AssemblyRef` tables of all four vendored Thermo assemblies:

| vendored DLL | machine | ilOnly | Thermo assemblies it references |
|---|---|---|---|
| `ThermoFisher.CommonCore.Data.dll` | I386 | true | — |
| `ThermoFisher.CommonCore.BackgroundSubtraction.dll` | I386 | true | `...Data` |
| `ThermoFisher.CommonCore.RawFileReader.dll` | I386 | true | `...Data` |
| **`ThermoFisher.CommonCore.MassPrecisionEstimator.dll`** | **Amd64** | true | `...Data` |

It is a leaf — depends on `CommonCore.Data`, and nothing depends on it. No sibling Thermo assembly names it in its `AssemblyRef` table, so it cannot arrive transitively either. Worth noting `ilOnly` is **true** even for the AMD64 one, so the `&&` in `IsArchitectureNeutral` is load-bearing; an ILOnly-only check would sail straight past this file.

I also scanned the whole test output: **140 managed assemblies, exactly one not architecture-neutral** — the deliberate positive-control input. So the guard is not green by luck, and the `Has.Count.GreaterThan(10)` sentinel clears by a wide margin.

Three things I think are worth resolving before this merges.

---

## 1. This needs a companion MetaMorpheus PR, and that one has to merge first

MetaMorpheus's WiX installer lists this DLL **by name**, sourced straight from the GUI output directory — `MetaMorpheusSetup/Product.wxs:361-362`:

```xml
<Component Id="src_ThermoFisher.CommonCore.MassPrecisionEstimator.dll" Guid="03803256-...">
  <File ... Source="$(var.GUI_TargetDir)ThermoFisher.CommonCore.MassPrecisionEstimator.dll" />
</Component>
```

It is inside `ExecutablesAndLibrariesComponentGroup`, which is referenced into the shipping feature — live markup. `Product.wxs` uses `DoNotHarvest`, so every shipped file is hand-listed and nothing reconciles that list against the build output. The first MetaMorpheus build against a post-#1156 mzLib gets a `WIX0103` for a file that is no longer there — the same failure that #2713 and #2720 dealt with for the EF6 chain.

I have opened **smith-chem-wisc/MetaMorpheus#2726** for this. It also carries a piece that is not obvious from this side: removing the component alone would redden MetaMorpheus CI **immediately**, before any version bump, because `validate_installer_contents` fails on a DLL that is in the build output but not installed and not whitelisted (`InstallRunAndArtifact.yml:275`) — and mzLib 1.0.584 still delivers this one to the bin root (verified in a real Release build). So #2726 pairs the component removal with a whitelist entry, which makes it correct on today's mzLib and still correct after the bump.

Merge order: **#2726 → this PR → mzLib release → MetaMorpheus version bump.**

Not a defect in this PR — the change here is right. It just cannot land first.

## 2. Re-adding the reference does not fail CI, which is the invariant this PR states

The description's claim, and the thing the test is for, is that the reference cannot be quietly re-armed. I tested that directly:

| what I did | guard |
|---|---|
| PR as-is | 3/3 pass |
| **restored both the `<Reference>` and the `<None Update>` copy item** in `Readers.csproj` — the exact state this PR removes — rebuilt Readers and Test, confirmed the AMD64 DLL back in the test output | **3/3 still pass** |
| additionally added one line of real use (`new PrecisionEstimate()`) | **fails** |

The failure message, when it does fire, is exactly right:

```
Readers.dll -> ThermoFisher.CommonCore.MassPrecisionEstimator (machine=Amd64, ilOnly=True)
```

So the guard is sound for *use*, and the doc-comment is correct that an unused reference never loads. But the csproj entries can come back in full, the AMD64 binary lands in the build output again, and nothing turns red — because the compiler omits the `AssemblyRef` for an unused reference, and the scan reads references. Whoever re-adds it is one `using` away from the arm64 break with a green suite behind them.

Two ways out, either fine: narrow the claim (the guard catches *use*, not reintroduction), or add a presence assertion over the Readers output / the nuspec. Related: nothing guards the nuspec at all, and `nuget pack` — the only thing that would notice a re-added `<file>` entry — runs on tag push in `release.yml:54`, never on a PR. Six open PRs are currently editing that nuspec.

## 3. An x86-only assembly passes the guard — I built one to check

`IsArchitectureNeutral` is `ILOnly && Machine == I386`. A `/platform:x86` assembly is precisely that, *plus* `CorFlags.Requires32Bit`, and CoreCLR refuses to load it into any 64-bit process — the same `BadImageFormatException` this guard exists to prevent. I compiled a throwaway `PlatformTarget=x86` library and ran the PR's exact rule over it:

```
machine         : I386
CorFlags        : ILOnly, Requires32Bit
Requires32Bit   : True
PR's Stamp()    : machine=I386, ilOnly=True
PR classifies as: ARCHITECTURE-NEUTRAL (passes the guard)
```

`Stamp` reports it as `machine=I386, ilOnly=True`, which reads as innocuous, so even a human looking at the output would not catch it. x86-only is historically the most common form of architecture-stamped managed binary, so this is the case most likely to actually turn up in a dependency bump.

Treating `Requires32Bit` as architecture-specific fixes it — leaving `Prefers32Bit` alone, which is benign — and putting the 32-bit flags in `Stamp` makes the message self-explaining.

---

## Non-blocking

- **The offender path is executed by no test.** I drove it manually above and it works today, but nothing keeps it working: an inverted condition, a `TryGetValue` that stops matching, or a message format that drops `machine=` all leave the suite green.
- **The positive control asserts only `IsArchitectureNeutral(...) Is.False`**, which a two-term conjunction satisfies either way — so the machine-field half it is meant to anchor is not specifically pinned. Asserting `Machine == Machine.Amd64` on the known input would fix it.
- **Duplicate simple names resolve by sort order.** `TryAdd` over `OrderBy(path)` keeps one file per name, and duplicates genuinely exist in the output — `ThermoFisher.CommonCore.Data` and `...RawFileReader` each appear twice (root and `Thermo\`), plus 13-copy satellite resource sets. Harmless today since the copies are identical, but a RID-specific `runtimes\...\lib\` copy that is stamped while the root copy is neutral would be masked. Classifying every path, and reporting the resolved full path in the offender line, closes it.
- **Each referenced file is re-opened inside the innermost loop**, so the same assemblies are re-read once per referencing assembly. Classifying once at map-build time and sharing via `[OneTimeSetUp]` would remove nearly all of it.
- `Link="FileReadingTests\ArchitectureGuard\..."` names a folder matching neither the fixture's namespace (`ReadersInfrastructure`) nor any existing folder; and `<None Include ... Link=...>` is a new pattern here (Test.csproj has 266 `<None Update>` and no `<None Include>` or `Link=`).

## Things I checked that are fine — please don't let anyone "fix" these

- `[TestFixture]` / `[ExcludeFromCodeCoverage]` / `internal class` is a **verbatim** match for its three nearest neighbours in the same folder (`TestReaders.cs:15-17`, `TestResultFile.cs:12-14`, `TestSupportedFileExtensions.cs:12-14`). 23 internal fixtures already exist and NUnit discovers them — I ran it, all three found.
- Plain `Assert.That` is the current convention, not an omission: of the Test files added since mid-May, exactly one uses the `ClassicAssert` alias.
- `ReadersInfrastructure/` already exists and is the right home.
- Native binaries are handled correctly — `HasMetadata` plus the `BadImageFormatException` catch cleanly skipped all 32 in the output.
- No `[Category]` means it lands in the **required** CI job, which is where a build-output guard belongs.
- `check-thermo-rawfilereader.yml` is unaffected — it never checks out the repo.

## What I did not check

I did not build the MetaMorpheus MSI (no WiX here), so #2726 rests on reading `Product.wxs` and that workflow rather than on a green installer build; its own CI is the real check. I also did not evaluate whether removing a WiX component `Guid` has upgrade consequences for machines with the file already installed.

Context for anyone arriving later: #1158 tracks the wider cross-platform picture, though it does not currently mention the installer coupling.

trishorts:
This one I mostly want to endorse rather than pick at. Turning "we removed an architecture-stamped reference" into a standing guard test is the right shape — the reference would otherwise come back the next time someone copies a `<Reference>` block, and nothing would notice until an arm64 user did.

Two things I checked rather than assumed:

**Both nuspec entries went, not one.** `MassPrecisionEstimator` was listed twice — once under `lib\net10.0` and once under `lib\net10.0-windows7.0` — and the diff removes both alongside the `<Reference>` and the `<None Update>` in `Readers.csproj`. That matters because removing the copy step while leaving a nuspec `<file src>` behind would break `nuget pack` at release time, which is the worst moment to find out.

**Which is the argument for #1145.** This is now the second recent case of the nuspec and the build disagreeing, after #1179 declared four missing dependencies. A CI check that every `<file src>` path resolves would have caught both classes at PR time rather than at pack time. Not this PR's job — but this PR is good evidence for it, and worth a cross-reference.

One question, non-blocking: keeping the DLL as test data means the repo still carries the x64-stamped binary, so `git clone` still pays for it and the architecture guard has something real to assert against. That seems like the right trade to me — a guard test that asserts against a fixture it does not have would be theatre — but it is worth being explicit in the README you added that the file is deliberately retained, so nobody "cleans it up" later and quietly guts the test.

trishorts:
Verified locally rather than by reading, in a clean worktree off the PR head:

- The three new tests pass (3/3). They aren't vacuous — the pinned fixture lands in the test output, so the scan genuinely has an AMD64-stamped assembly to resolve and classify.
- The central claim holds. No source file, csproj, nuspec, or Thermo binary names `MassPrecisionEstimator` — I grepped all four vendored Thermo DLLs using the fixture itself as the positive control (2 hits in its own metadata, 0 in the other three), so it isn't being loaded reflectively either.
- Downstream fan-out is clear. MM #2726 is merged, and neither FlashLFQ nor ProteaseGuru hand-lists the assembly.
- The guard's coverage isn't leaky: every project the nuspec ships is reachable from `Test.csproj`, so a reference added anywhere that ships would land in the scanned output.

**The red `integration` check is stale** — that run is from 08-16, two days before MM #2726 removed the `Product.wxs` component. A re-run should be green; worth kicking before anyone reads the ✗ as a real failure.

trishorts (approved):
No comments — verified by running it rather than by reading, against 76a7291f.

The three new tests pass, and they aren't vacuous: the pinned fixture reaches the test output, so the scan has a genuinely AMD64-stamped assembly to resolve and classify. The claim that nothing needs the assembly holds — no source file, csproj, nuspec or Thermo binary names it, checked with the fixture itself as the positive control so the grep was known to be working. Every project the nuspec ships is reachable from Test.csproj, so the guard has no blind spot on shipped code. Downstream is clear: MM #2726 is merged, and neither FlashLFQ nor ProteaseGuru hand-lists the assembly.

`integration` was re-run and is green; the earlier failure was a pre-#2726 MetaMorpheus master, not this PR.

nbollis (approved):

Issue #1158, opened 2026-08-15
Cross-platform data reading on macOS and Linux (x64 and arm64): what works, what blocks, and in what order


acesnik:
🤖 Tracking issue for reading mass spectrometry data on macOS and Linux, on both x64 and arm64. It records what has been measured rather than assumed, so that the next person does not have to re-derive it, and so PRs in this area have something to attach to.

Nothing here is a regression. These are longstanding gaps that only became visible once the assemblies stopped being x64-stamped (#1127, MetaMorpheus #2503/#2504) and MetaMorpheus got far enough on Apple Silicon to reach them.

## Summary

| input | Windows | Linux x64 | Linux arm64 | macOS x64 | macOS arm64 |
|---|---|---|---|---|---|
| mzML, MGF | ✅ | ✅ | ✅ | ✅ | ✅ |
| Thermo `.raw` | ✅ | ✅ | ✅ | ✅ | ✅ |
| result files (psmtsv, MsFragger, TopPIC, MsPathFinderT, Crux, Casanovo, Dinosaur, FlashDeconv, SDRF …) | ✅ | ✅ | ✅ | ✅ | ✅ |
| Bruker timsTOF `.d` | ✅ | ❌ `timsdata.dll` | ❌ SQLite, then `timsdata.dll` | ❌ `timsdata.dll` | ❌ SQLite, then `timsdata.dll` |
| Bruker `.baf` | ✅ | ❌ `baf2sql_c` | ❌ | ❌ | ❌ |
| IsoDec deconvolution | ✅ | ❌ `isodeclib.dll` | ❌ | ❌ | ❌ |

Of roughly 29 supported file types, all but the two Bruker families are pure managed code and work everywhere. **Thermo `.raw` is in the working set** — verified by running a full search on both Mac architectures, 7/7 target PSMs at 1% FDR, identical results (MetaMorpheus #2723 corrects the README, which claimed Windows and Linux only).

## What is actually blocking, measured

Failures below are from running the readers on real fixtures (`Test/TestData/snippet.d`), on native arm64 and on x64 under Rosetta, plus `linux/arm64` and `linux/amd64` containers.

[...]

PR #1159, opened 2026-08-15
Build Bruker and timsTOF file paths with Path.Combine

acesnik:
🤖 Part of #1158.

Both Bruker readers joined the `.d` directory to the files inside it with a literal backslash:

```csharp
if (File.Exists(FilePath + @"\analysis.tdf") && File.Exists(FilePath + @"\analysis.tdf_bin"))
```

On macOS and Linux that builds `snippet.d\analysis.tdf`, which never exists. So `File.Exists` returned false and the reader reported this, about files sitting in the directory it was given:

```
Could not locate the the database (analysis.tdf, analysis.tsf) and/or binary
(analysis.tdf_bin, analysis.tsf_bin) files
```

`BrukerFileReader` was worse: a bare `throw new FileNotFoundException()` with no message and no path.

Nine occurrences across seven statements, in `TimsTofFileReader` (`.tdf`/`.tsf` existence checks and `GetSourceFile`) and `BrukerFileReader` (`OpenFileConnection` in both the static and dynamic paths, plus `GetSourceFile`). All now go through `Path.Combine`.

`TimsTofFileReader` already used `Path.Combine` when building the SQL connection string a few lines below the existence check, so the file disagreed with itself.

## What this does and does not do

**It does not make Bruker readable off Windows.** It removes the first of several blockers and stops the reader misreporting the reason. After this change a timsTOF `.d` gets past the existence check and fails on a native library instead — verified on four platforms:

| runtime | before | after |
|---|---|---|
| macOS arm64 | "Could not locate …" (files present) | `DllNotFoundException: SQLite.Interop.dll` |
| macOS x64 (Rosetta) | same false message | `DllNotFoundException: timsdata.dll` |
| Linux arm64 | same false message | `DllNotFoundException: SQLite.Interop.dll` |
| Linux x64 | same false message | `DllNotFoundException: timsdata.dll` |

The arm64/x64 split is not this change: `Stub.System.Data.SQLite.Core.NetStandard` ships no arm64 native, so arm64 trips on SQLite before reaching Bruker's library. #1158 records the full picture and the order the remaining blockers have to be tackled in.

Also included: both not-found paths now name the directory, and a missing `.d` directory throws `DirectoryNotFoundException` rather than `FileNotFoundException`.

## Verification, and why there is no test

Run against the real `Test/TestData/snippet.d` on a native arm64 runtime, an x64 runtime under Rosetta, and `linux/arm64` and `linux/amd64` containers, through `MsDataFileReader.GetDataFile(...)` so the dispatch is exercised too.

**No regression test, deliberately.** `Test.csproj` targets `net10.0-windows`, so tests run on Windows only — where a hardcoded `\` is correct and every test passes before and after. I looked for a variant that discriminates on Windows and could not construct one: Windows accepts forward slashes, mixed separators and doubled separators alike. So this class of bug is structurally uncatchable by the current test project, which is itself noted in #1158 as worth fixing separately.

That makes the four-platform table above the evidence, rather than a green check mark.


trishorts:
🤖 Reviewed locally against `smith/master @ 03081569`, branch checked out and built. Everything below I ran; I have said where I did not.

**The separator fix is correct and complete — I verified the count rather than trusting it.** Grepping both spacings of the concatenation pattern across production code, before and after:

| | production sites matching `+ @"\` |
|---|---|
| master | 7 statements / 9 occurrences, all in these two readers |
| this branch | **0** (the only remaining hit repo-wide is a regex literal in `IsoTrackerIdFilter.cs:47`) |

Worth noting for anyone re-checking: `BrukerFileReader.cs:70` was `FilePath+@"\analysis.baf"` with **no spaces** around the `+`. My first grep used `+ @"\` and missed it; a reviewer doing the same would wrongly conclude a site was left behind. It wasn't — the sweep is genuinely exhaustive.

Also confirmed: all timsTOF entry points funnel through the one fixed check (`LoadAllStaticData:350` just calls `InitiateDynamicConnection`), and Bruker's static path has its own guard which the PR also fixed. No missed entry point. 48 Bruker/timsTOF tests pass on this branch.

One correction to the description, since it will send a reviewer to the wrong place: the pre-existing `Path.Combine` is **not** "a few lines below the existence check". The check is at `TimsTofFileReader.cs:76-80` in `InitiateDynamicConnection`; the `Path.Combine` is at `:132-135` in a different method, `OpenSqlConnection`. The substantive point — the file already contradicted itself — is exactly right and is a good argument for this change.

---

## One thing worth changing before merge

### `DirectoryNotFoundException` diverges from how mzLib already types this exact condition — and the improved message doesn't reach users

Two separate problems that compound.

**It doesn't reach the path anyone takes.** I probed all three entry points on this branch:

```
MsDataFileReader.GetDataFile("fakePath.d")            -> FileNotFoundException      | 'Unable to find the specified file.'
new BrukerFileReader("fakePath.d").LoadAllStaticData()-> DirectoryNotFoundException | 'Could not locate the Bruker .d directory: fakePath.d'
new TimsTofFileReader("fakePath.d").LoadAllStaticData()-> FileNotFoundException     | 'Could not locate the database ... in fakePath.d'
```

`SupportedFileTypes.ParseFileType` (`Util/SupportedFileTypes.cs:98`) does `if(!Directory.Exists(filePath)) throw new FileNotFoundException();` — bare, no message, no path — in the `case ".d":` arm, *before* `GetDataFile` ever constructs a reader. So the user-facing complaint in the PR body ("a bare `throw new FileNotFoundException()` with no message and no path") is still exactly what a user gets. The improved Bruker message is reachable only by constructing the reader directly, or in the delete-between-dispatch-and-load race at `FlashLFQ/PeakIndexingEngine/PeakIndexingEngine.cs:36-37`, which calls `GetDataFile` then `LoadAllStaticData` with no catch.

**It splits the family.** `DirectoryNotFoundException` appears **nowhere** under `mzLib/Readers/` today — I checked master, zero hits — so this introduces it to the reader surface. And after this PR the same user error is typed three different ways depending on entry point: `FileNotFoundException` from the dispatcher at `SupportedFileTypes.cs:98`, `FileNotFoundException` from the sibling directory-based reader at `TimsTofFileReader.cs:86` (touched by this very PR), and `DirectoryNotFoundException` from Bruker. `DirectoryNotFoundException` is not a subclass of `FileNotFoundException` — both derive independently from `IOException` — so a `catch (FileNotFoundException)` silently stops matching.

Nothing goes red: no test constructs `BrukerFileReader` directly (`TestBruker.cs:33` goes through `GetDataFile` and is intercepted at `SupportedFileTypes.cs:98`), and MetaMorpheus and ProteaseGuru never name either type. So CI cannot see the divergence either way — which is part of why it's worth deciding deliberately rather than by default.

**Suggestion:** keep `FileNotFoundException` and just add the message and path. That delivers the entire user-visible benefit of this change with no contract divergence, and matches the shape the PR already uses two methods down at `BrukerFileReader.cs:130`. If `DirectoryNotFoundException` really is the right model, then `SupportedFileTypes.cs:98` is the line to change — it's the one users actually hit — and `TimsTofFileReader` should move with it so the library speaks with one voice. Changing Bruker alone is the option that's hard to defend.

Either of those is fine by me; this is a "was the alternative considered" finding, not a reject.

Separately and worth a moment: the message improvement in `InitiateDynamicConnection` — `new FileNotFoundException($"...", bafPath)` using the `(message, fileName)` overload with the *specific* missing file — is exactly the newest convention in this repo (`MslReader.cs:144`, `Mslfiletypehandler.cs:98`, `MslConverter.cs:80`). That half is unambiguously right.

---

## Non-blocking

- **Empty `FilePath` now resolves against the working directory.** Measured: old `"" + @"\analysis.baf"` → `\analysis.baf` (rooted); new `Path.Combine("", "analysis.baf")` → `analysis.baf` (relative). `FilePath` is unvalidated (`MsDataFile.cs:48-51`), so a reader constructed with `""` while the process is `cd`'d into a `.d` directory would now silently open that unrelated acquisition instead of failing. Narrow — needs direct construction — but the failure mode is reading the wrong file rather than an error. `Directory.Exists("")` is false, so Bruker's static path is unaffected; the dynamic paths are not.
- **Null `FilePath` now throws `ArgumentNullException`.** Measured: old `null + @"\analysis.baf"` → `\analysis.baf`; new → `ArgumentNullException` from `Path.Combine`. So a null path gets a worse diagnostic (naming `path1`) even as others improve. Very narrow, same reasoning.
- **The missing-directory case still reports the wrong thing in timsTOF.** If the `.d` directory doesn't exist at all, `File.Exists` on all four candidates is false and the user is told the database/binary files could not be located, rather than that the directory isn't there.
- **The no-test argument is right about separators, but doesn't cover the exception change.** I agree this class of bug is structurally uncatchable on Windows — I tried to construct a discriminating case and couldn't; Windows accepts `\`, `/`, mixed and doubled separators alike, and a `/` can't appear in a Windows directory name. But the type/message change *is* Windows-testable, and currently nothing covers the changed Bruker line in either direction.

If you want a guard for the separator class specifically, the machinery you built in #1156 is the shape that would work — a source-level invariant test rather than a behavioural one.

## Verified fine

Sweep for a shared path helper: mzLib has none (`MzLibUtil` has only a filename stripper), so raw `Path.Combine` is the correct and idiomatic choice — there is nothing you should have used instead. No hardcoded separators in the `DllImport` names either; those are bare library names, correctly left to #1158.

## What I did not check

I did not run on macOS or Linux, so the four-platform table in the description stands on your testing, not mine — my verification is Windows-side plus source and probe evidence. I also did not check the language bindings (pyMzLib / mzLibRust / mzLibR) for anything that maps mzLib exception types to error codes; given the changed line is unreachable through `GetDataFile`, exposure there is unlikely but it is the one place the type change could surface outside these repos.

No companion PR is needed in MetaMorpheus for this one — it names neither exception type anywhere, and `MyFileManager.cs:50` and `MetaDrawDataLoader.cs:138` both catch generically, so the message improvement reaches users unharmed.

nbollis (approved):

Alexander-Sol (approved):

PR #1162, opened 2026-08-15
Ship XML documentation so consumers get tooltips, and point the README at the wiki

### LINKED ISSUE #110: Field comments in each object
Each object should have some sort of metadata that explains the fields in each object. One example: "accession" in ModificationWithLocation is a Tuple<string, string>, but it's not clear unless looking at the mzLib code what each string represents.

### LINKED ISSUE #235: Updating readme
ProteinDatabaseLoader example is out of date, and perhaps some other items.

acesnik:
🤖 Closes #110. Closes #235.

Roughly 1,900 `<summary>` blocks exist across this repository and **none of them ever reached anyone outside it**. No project set `GenerateDocumentationFile` and the nuspec shipped no `.xml`, so someone writing MetaMorpheus code got no tooltip for any mzLib type despite the documentation being written. That is the delivery half of #110; the field comments it asked for in 2017 have largely been written since.

## Shipping it

One property in `Directory.Build.props`, and 33 nuspec entries putting each managed assembly's `.xml` beside its `.dll` in `lib\net10.0` and `lib\net10.0-windows`, which is where the IDE looks.

Only the 17 assemblies this repository builds get an entry. The natives shipped alongside them — `timsdata`, `baf2sql_c`, `isodeclib`, `isogenmass`, `libmmd`, `svml_dispmd` and the MSVC runtimes — emit no documentation, and a nuspec entry for a file that never exists **fails the pack**, so getting that wrong would have broken the next release.

Generation is scoped to projects that ship. An XML file for a test or benchmark assembly serves nobody, and generating it there accounted for **61 of the 95** CS1570/CS1587 warnings, all in code no consumer sees.

## Fixing what the diagnostics found

The remaining 34 were in shipped code, and most were the same mistake: **unescaped generics and comparisons inside comments** — `Dictionary<IBioPolymerWithSetMods, List<int>>`, `S <= s`, `(< 10 VCF columns)`, `"<-"`, a bare `&`. The XML parser reads those as tags, which silently truncates everything after them in a tooltip, so these were already degrading the documentation before it shipped.

The rest were structural: a `<returns>` never closed, a `<summary>` opened with `//` instead of `///`, a summary split by a bare line, doc comments sitting above commented-out code in FlashLFQ, and one `///` used as a trailing comment on an `if`.

Also degraded **14 unresolvable `cref` attributes** to `<c>`. Each was a link that could not resolve from where it was written, so it rendered dead; the text reads the same now without asserting a relationship that does not exist.

## Two things for a reviewer's judgement

**`SpectraFileInfo`'s equality documentation references a `ChannelLabel` member the type does not have.** I degraded the link rather than rewriting the sentence, because the wording may describe equality inaccurately and I would be guessing at intended semantics. Worth an owner's eye.

**243 public types still have no summary, and I left them deliberately.** Five of the most-used are now documented — `FlashLfqEngine`, `Identification`, `ChromatographicPeak`, `Peptide`, `MsDataFileReader` — chosen because I could establish their meaning from the code. For many of the rest I could not, and a confidently wrong tooltip is worse than an absent one: a reader trusts it. The audit is 930 public types, 248 undocumented before this change, concentrated in `Readers` (59), `Omics` (36), `Proteomics` (30), `MassSpectrometry` (29) and `FlashLFQ` (22). Filling them in is a good follow-up for whoever owns each area.

## README (#235)

The usage examples are **removed rather than updated**. They had drifted — one called a `ProteinDatabaseLoader` type that no longer exists, which is exactly what #235 reported — and the [wiki](https://github.com/smith-chem-wisc/mzLib/wiki) has 16 maintained pages covering the same ground while the README never linked to it. A second copy of the documentation is a second thing to keep current. Badges, image and existing links are unchanged; a short Contributing note records the `Category!=ExternalService` filter.

## Verification

Full **non-incremental** solution build: **0 errors, and 0 for every doc diagnostic except CS0419** — CS1570, CS1572, CS1574, CS1587, CS1711 and CS1734 are all clear. CS0419 (ambiguous cref) sits at 24; those resolve to an unspecified overload, so the link works and only the signature is arbitrary. Each assembly's `.xml` appears beside it and parses with real content — Omics 1012 members over 106 types, Readers 777 over 127, MassSpectrometry 249 over 36, FlashLFQ 141 over 14.

Worth noting the incremental build initially hid warnings from unchanged dependency projects, which is why the final count came from `--no-incremental`; a reviewer checking locally should do the same.

---

🤖 *Edited after review.* The original text said 60 of 103 doc-comment warnings, with 43 in shipped code. Re-measured against `upstream/master` with `GenerateDocumentationFile=true` forced on every project, non-incremental Release, `NoWarn` cleared, counting unique file+line+column: CS1570 41 (29 shipped / 12 test) and CS1587 54 (5 shipped / 49 test), so 95 total split 34 shipped / 61 test-and-Development. The test-side figure was right; the shipped-side one looks like transposed digits. Part of the gap may be that the original count predates the master merge. The verification line was also widened, because CS1572, CS1711, CS1734 and CS0419 are enabled by `GenerateDocumentationFile` too and were not being reported.


mzLib/UsefulProteomicsDatabases/Loaders.cs
```diff
@@ -213,7 +213,7 @@ public static IEnumerable<Modification> LoadUniprot(string uniprotLocation, Dict
 
         /// <summary>
         /// Shared client for the ontology downloads below. One instance rather than one per call, which is
-        /// what <see cref="PredictionClients"/>' Koina client and <see cref="PrideArchiveClient"/> both do,
+        /// what <c>PredictionClients</c>' Koina client and <c>PrideArchiveClient</c> both do,
```
trishorts: `PrideArchiveClient` is `public sealed class PrideArchiveClient` in `namespace UsefulProteomicsDatabases` (`PrideArchiveClient.cs:36`) -- the same namespace and the same assembly as `Loaders`, so this cref resolved.

The `<remarks>` block seven lines down still carries `<see cref="PrideArchiveClient"/>`, untouched by this PR, which settles it either way: if that one resolves then so did this one, and if it doesn't then the "0 CS1574" verification claim wouldn't hold.

Suggest restoring `<see cref="PrideArchiveClient"/>` here and leaving only `<c>PredictionClients</c>` degraded -- that one is genuinely unreachable, since `UsefulProteomicsDatabases.csproj` doesn't reference the project.

mzLib/Omics/SpectralMatch/BaseSpectralMatch.cs
```diff
@@ -113,7 +113,7 @@ protected BaseSpectralMatch(
 
         /// <summary>
         /// One-based positions of residues covered by matched fragment ions.
-        /// Null until <see cref="GetSequenceCoverage(IEnumerable{int}, IEnumerable{int})"/> is called.
+        /// Null until <c>GetSequenceCoverage(IEnumerable{int}, IEnumerable{int})</c> is called.
```
trishorts: The cref was correctly identified as unresolvable, but degrading it preserves a signature that doesn't exist. There is no two-parameter `GetSequenceCoverage` anywhere in `Omics` -- the real member is the parameterless `GetSequenceCoverage()` twelve lines below at line 123, and `IHasSequenceCoverageFromFragments.cs:16` already documents it with a working `<see cref="GetSequenceCoverage"/>`.

`FragmentCoveragePositionInPeptide` is a public property on a public abstract base type, so once the XML ships this tooltip tells every MetaMorpheus developer to call a two-argument overload that will not compile. A dead link is inert; a confidently wrong signature is actively misleading -- which is the same reasoning you gave for leaving 243 types undocumented.

Suggest `<see cref="GetSequenceCoverage"/>` rather than freezing the invented signature.

mzLib/Readers/BaseClasses/IMs1FeatureFile.cs
```diff
@@ -24,7 +24,7 @@ namespace Readers
     /// <item>
     ///     <description>
     ///     Mapping source file fields to the required interface properties:
-    ///     <see cref="ISingleChargeMs1Feature.MZ"/>,
+    ///     <c>ISingleChargeMs1Feature.MZ</c>,
```
trishorts: This cref failed on casing alone -- `ISingleChargeMs1Feature.cs:26` declares `double Mz { get; }`, not `MZ`, and cref lookup is case-sensitive. The interface's own remarks block at line 13 already uses the correct `<see cref="Mz"/>`, so the resolvable spelling was one file away.

Correcting the case silences CS1574 *and* keeps the link. As it stands, the one entry in this list that names a property which doesn't exist sits alongside four siblings (`Charge`, `RetentionTimeStart`, `RetentionTimeEnd`, `NumberOfIsotopes`) that remain live links -- and this list exists precisely to tell an implementer which interface properties to map.

mzLib/MzLibUtil/MrsNoiseEstimation/MRSNoiseEstimator.cs
```diff
@@ -77,8 +77,8 @@ public static bool MRSNoiseEstimation(double[] signal, double epsilon,
         /// <param name="noiseEstimate">The initial noise estimation of the signal.</param>
         /// <param name="noiseThreshold">A multiple of the noise estimate. noiseThreshold * noiseEstimate
         /// is the cutoff value determining whether or not a pixel is signal or noise.</param>
-        /// <returns>List<int[]>. Each element in the int[] is either a 0 or positive integer, representing a noise value (if 0)
-        /// or a signal value (if >9).</int></returns>
+        /// <returns>List of int[]. Each element in the int[] is either a 0 or positive integer, representing a noise value (if 0)
+        /// or a signal value (if &gt;9).</returns>
```
trishorts: Not about this line -- flagging it here because it is the nearest line in the diff for this file.

`ComputeStdevOfNoisePixels` at line 157 is declared `(double[] signal, int[] noiseIndices)`, but its doc block at line 147 still carries `/// <param name="wtOutput">The output of the mod wavelet transform.</param>`. No such parameter exists.

That was inert before, because nothing generated documentation. With `GenerateDocumentationFile` on it emits CS1572, which is not in the `NoWarn` list (see my note on `Directory.Build.props`). It also ships a `<param>` element that binds to nothing, so the IDE silently drops that line from the tooltip.

Deleting the `wtOutput` line is probably the fix -- the `<summary>` above it already mentions the ModWt output as provenance.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -1992,7 +1998,7 @@ internal XIC BuildXIC(List<Identification> ids, SpectraFileInfo spectraFile, boo
         /// The ChromPeaks will be stored in chromPeaksInThisSequence
         /// For example, we have four invaild peak in the XIC, then we look at the first peak, generate chormPeak from each file. The dictionary will be like: peak1: [chromPeak (run 1), chromPeak (run 2), chromPeak (run 3)...]
         /// </summary>
-        /// <param name="peaksInOneXIC"> The time window for the valid peak, format is <time < start, end>> </start></param>
+        /// <param name="peaksInOneXIC"> The time window for the valid peak, format is time &lt; start, end &gt; </param>
```
trishorts: The escaping fix on this line is right, but the tag still names `peaksInOneXIC`, and line 2002 names `chromPeaksInThisSequence`, while the method below takes `(sharedPeak, chromPeaksInSharedPeak, xICGroups)`.

A `<param>` whose name matches no parameter is dropped by the documentation generator, so the text this PR just made well-formed still renders against nothing -- and both tags now emit CS1572, which isn't suppressed.

Retargeting the two names to `sharedPeak` and `chromPeaksInSharedPeak` would make the escaping work actually land in the tooltip.

mzLib/Directory.Build.props
```diff
@@ -5,6 +5,27 @@
...
+         "malformed":
+           CS1591 missing XML comment for a public type or member - 248 of 930 public types have none,
+                  so this would bury every other warning. Worth filling in, not worth hiding the build.
+           CS1573 a parameter has no matching param tag - partial documentation is still useful
+                  documentation, and the tooltip renders.
+         Left visible, because these mean a comment may not render at all and so defeat the point of
+         shipping the file: CS1570 badly formed XML, and CS1587 comment not on a valid language
+         element. Roughly 200 of those exist today; the generated XML still parses, so they degrade
+         rather than break, but they are real defects worth fixing. -->
+    <NoWarn>$(NoWarn);CS1591;CS1573</NoWarn>
```
trishorts, line 33: Two things about the diagnostic policy this block sets out.

**CS1574 and CS1572 aren't in the accounting.** The comment presents a closed set of four -- CS1591/CS1573 suppressed, CS1570/CS1587 left visible -- but enabling `GenerateDocumentationFile` also turns on CS1574 and CS1572. CS1574 is load-bearing here: it is the diagnostic that surfaced all 14 crefs this PR degraded, yet it is never named. CS1572 is unhandled and does fire -- `MRSNoiseEstimator.cs:147` and `FlashLfqEngine.cs:2001-2002` all carry `<param>` tags for parameters that do not exist (commented separately). Since the stated verification is "0 CS1570 / CS1574 / CS1587", anyone reproducing the non-incremental build locally will see CS1572 warnings the PR says nothing about, and won't be able to tell whether they were known or missed.

**The number on line 26 doesn't reconcile.** It says roughly 200 CS1570/CS1587 "exist today", but the description reports the build at 0 for both and puts the total doc-comment warning count at 103 (60 in test/benchmark, 43 in shipped code). I can't make those three figures agree. This comment is what the next maintainer reads to decide whether a CS1570 they hit is a regression or expected background, so it's worth getting exact.

mzLib/FlashLFQ/ChromatographicPeak.cs
```diff
@@ -8,6 +8,12 @@
 
 namespace FlashLFQ
 {
+    /// <summary>
+    /// An integrated elution peak for one identified species in one spectra file: the isotopic envelopes
```
trishorts: `CalculateIntensityForThisFeature(bool integrate)` only sums envelope intensities when `integrate` is true; otherwise `Intensity = Apex.Intensity`. `FlashLfqParameters.cs:16` initialises `Integrate = false`, and the `FlashLfqEngine` constructor doc says "It is HIGHLY recommended this is set to FALSE".

So in the default and recommended configuration this type is not an integrated peak -- its `Intensity` is the single most intense isotopic envelope. The tooltip now tells a MetaMorpheus developer it's an integrated area, which is the kind of misreading that produces wrong cross-tool intensity comparisons.

Describing it as the grouped isotopic envelopes, with the intensity being the apex envelope unless `Integrate` was requested, would match the code.

mzLib/FlashLFQ/Peptide.cs
```diff
@@ -5,6 +5,12 @@
 
 namespace FlashLFQ
 {
+    /// <summary>
+    /// One peptide sequence's quantification across every spectra file: its intensity per file, how that
+    /// intensity was obtained per file, and the protein groups it belongs to. This is the roll-up of the
```
trishorts: `CalculatePeptideResults` doesn't aggregate those peaks. Per file it takes `sequenceWithPeaks.Value.Max(p => p.Intensity)` (`FlashLFQResults.cs:157`) and records that single best peak's intensity and apex retention time -- after filtering to peaks with `NumIdentificationsByFullSeq == 1`, non-decoy identifications, and MBR peaks passing `MbrQValueThreshold` with `RandomRt == false`.

"Roll-up" reads as summed or otherwise combined, so a consumer may assume peptide intensity equals the total of its chromatographic peaks and find the numbers don't reconcile against QuantifiedPeaks.

Stating that the per-file intensity comes from the single highest-intensity qualifying peak would match what the code does.

mzLib/Readers/MsDataFileReader.cs
```diff
@@ -9,6 +9,11 @@
 
 namespace Readers
 {
+    /// <summary>
+    /// Entry point for reading a mass spectrometry data file. Dispatches on the file's extension and
```
trishorts: `GetDataFile` dispatches on `filePath.ParseFileType()`, and for `.d` that method doesn't consider the extension at all. `SupportedFileTypes.cs:99-106` enumerates the directory and returns `BrukerD` when `analysis.baf` is present, `BrukerTimsTof` when `analysis.tdf`/`analysis.tsf` is, and otherwise throws `MzLibException("Bruker file type not recognized")`. One extension, two different readers -- so extension alone cannot select one, and a `.d` path can throw before any reader is constructed.

Since `.d` is named explicitly in this sentence, worth rewording to something like "identifies the file type from the path and, for container formats such as Bruker `.d`, from the files inside it".

The list also omits `_ms1.msalign`/`_ms2.msalign`, which `GetDataFile` does handle via `Ms1Align`/`Ms2Align` -- making it explicitly non-exhaustive would cover that.

mzLib/MassSpectrometry/SpectraFileInfo.cs
```diff
@@ -67,7 +67,7 @@ public SpectraFileInfo(string fullFilePathWithExtension, string condition, int b
         }
         /// <summary>
         /// Determines whether two <see cref="SpectraFileInfo"/> instances are equal.
-        /// Two samples are equal if they have the same <see cref="FullFilePathWithExtension"/> and <see cref="ChannelLabel"/>.
+        /// Two samples are equal if they have the same <see cref="FullFilePathWithExtension"/> and <c>ChannelLabel</c>.
```
trishorts: You flagged this one for an owner's eye, saying you would be guessing at the intended semantics. I don't think you have to guess -- the answer is in this file.

`Equals` at lines 98-107 compares `FullFilePathWithExtension`, `Condition`, `BiologicalReplicate`, `Fraction` and `TechnicalReplicate`; `operator==` delegates straight to it; `GetHashCode` combines the same five; and the class summary at the top of the file already says "Equality and hashing include all identity components to safely distinguish multiple entries that may share the same file path but differ by replicate/fraction".

So the sentence is wrong in two ways rather than one: `ChannelLabel` does not exist on this type, *and* it omits the four components that actually participate. Degrading the link keeps the build quiet but ships the wrong statement as a tooltip, which is the thing this PR is trying to stop.

I would just replace the sentence with the wording `Equals` already carries, and do the same on the `operator!=` block below.

mzLib/FlashLFQ/ChromatographicPeak.cs
```diff
@@ -8,6 +8,12 @@
...
+    /// that were grouped together, the resulting intensity, and how the peak was found. Peaks whose
+    /// <see cref="DetectionType"/> is MBR were transferred from another file by match-between-runs rather
```
trishorts: Minor, but `DetectionType.MBR` doesn't mean "transferred" on its own. `MbrChromatographicPeak.cs:36-40` constructs random-RT decoy peaks with `detectionType: DetectionType.MBR` as well, carrying `RandomRt == true`, and `CalculateFdrForMbrPeaks` deliberately retains them when PEP could not be run. That is why every downstream filter pairs the check with `!m.RandomRt` -- e.g. `FlashLFQResults.cs:149`.

Someone filtering `Peaks` on `DetectionType == MBR` on the strength of this tooltip would include decoy transfers and inflate their quantification. A clause noting that MBR peaks include random-RT decoys used for FDR estimation, or a pointer to `MbrChromatographicPeak`, would cover it.

mzLib/MzIdentML/mzIdentML1_2_0.cs
```diff
@@ -15,7 +15,7 @@
 // 
 
 
-/// <remarks/>
+// <remarks/>
```
trishorts: The fix itself is correct -- this `///` sits before the namespace and attaches to nothing, so CS1587 is real and downgrading it is the right call.

The catch is that this file's header says changes "will be lost if the code is regenerated", reinforced by `[GeneratedCodeAttribute("xsd", "4.7.2046.0")]` on every type, and `xsd.exe` emits this stray leading `/// <remarks/>` deterministically. So the next regeneration from the mzIdentML 1.2 schema reintroduces the warning, with nothing in the repo recording that a manual fix-up is expected. (I checked the siblings -- `mzIdentML1_1_0.cs` and `mzIdentML1_1_1.cs` don't carry the same file-scope `///`, so nothing else is affected.)

Since the project is entirely xsd-generated, scoping CS1587 off in `MzIdentML.csproj` would survive regeneration; failing that, a line wherever the regeneration steps are recorded would do. Not blocking -- just a small recurring tax otherwise.

trishorts (commented):
Reviewed the packaging change, the doc-comment repairs and the three new summaries against the code they describe.

The infrastructure side looks right. `$(MSBuildProjectName)` is the only scoping mechanism that actually works in `Directory.Build.props` -- `$(IsTestProject)` and `<IsPackable>` both evaluate empty there, because Test.Sdk imports after props are evaluated, so either would have silently no-opped. The 33 nuspec entries use the correct `lib\net10.0-windows7.0` spelling and correctly skip every native and vendored binary. The green `integration` job is real proof the paths resolve, since it runs a bare `nuget pack` where a missing `<file src>` is a fatal NU5019.

I also checked the MetaMorpheus seam, since a mzLib nuspec change has broken the installer before: **no paired MetaMorpheus PR is needed here.** `Product.wxs` uses `DoNotHarvest` with 149 hand-listed `<File>` elements, so WIX0103 fires only on *removal* (a dangling `Source=`); added files are simply never referenced. The installer-contents check in `InstallRunAndArtifact.yml` filters `-Filter *.dll` at all four enumerations, so `.xml` files are invisible to it, and both CLI zip steps already carry `-x!*.xml`. Two harmless consequences worth knowing: the XML files land in MetaMorpheus's build output but will *not* be in the MSI, and the Docker image grows slightly via the unfiltered `ADD CMD/bin/Release/net10.0/`.

Comments below; none are blocking. The ones I would actually act on are the three crefs that were repairable rather than dead (`Loaders`, `BaseSpectralMatch`, `IMs1FeatureFile`) and the `SpectraFileInfo` sentence you flagged for an owner -- the answer to that one is already in the file. The rest are notes.

One coordination note: #1131 splits Chronologer into its own package with its own nuspec. If it merges after this, that package ships with no `.xml` entry while the other 17 have them. Worth a heads-up on that PR either way.

mzLib/UsefulProteomicsDatabases/Loaders.cs
acesnik: 🤖 Agreed — done in b79f3e42. The inconsistency was inside this one file: line 216 degraded it to `<c>` while the remarks block at 223 still carried a live `<see cref="PrideArchiveClient"/>`. Confirmed `public sealed class PrideArchiveClient` at `PrideArchiveClient.cs:36` under `namespace UsefulProteomicsDatabases` (line 15), same assembly as `Loaders`. Cref restored on 216; only `<c>PredictionClients</c>` stays degraded.

mzLib/Omics/SpectralMatch/BaseSpectralMatch.cs
acesnik: 🤖 Agreed — done in b79f3e42. The only members are the parameterless `GetSequenceCoverage()` at line 123 and the extension in `HasSequenceCoverageFromFragmentsExtensions`; nothing takes two arguments. Now `<see cref="GetSequenceCoverage"/>`, and the build confirms it resolves rather than trading one dead link for another.

mzLib/Readers/BaseClasses/IMs1FeatureFile.cs
acesnik: 🤖 Agreed — done in b79f3e42. `double Mz { get; }` at line 26 of the interface, and line 13 of that same file already used `<see cref="Mz"/>`. One correction: the file is `MassSpectrometry/ISingleChargeMs1Feature.cs`, not under `Readers/BaseClasses/`. Casing fixed, so the entry joins its four live siblings instead of being the one dead link in the mapping list.

mzLib/MassSpectrometry/SpectraFileInfo.cs
acesnik: 🤖 Done in b79f3e42, and you were righter than you claimed — the correct sentence was not merely in the file, it was already on `Equals` itself at lines 92-95: "same file path, condition, biological replicate, fraction, and technical replicate". Reused verbatim on both `operator==` and `operator!=`.

mzLib/Directory.Build.props
acesnik: 🤖 Both points hold, and the first is bigger than stated. Fixed in b79f3e42; I also corrected the PR description, with the edit noted at the bottom of it.

**Seven diagnostics, not four.** `GenerateDocumentationFile` enables CS1570, CS1572, CS1573, CS1574, CS1587, CS1711, CS1734 and CS0419. The comment listed four. The ones you did not name fired too: CS1711 (typeparam tag with no such type parameter, 2), CS1734 (paramref with no such parameter, 1), and CS0419 (ambiguous cref, 24). Swept the tree — CS1572 39 → 0, CS1711 2 → 0, CS1734 1 → 0, CS1574 1 → 0. CS0419 left at 24, since those resolve to *an* overload: the link works and only the signature is arbitrary. The props comment now carries the full list.

**The CS1572 gap is not hypothetical — this branch creates two of them.** On master, `FlashLfqEngine.cs:1995` emits three CS1570s: the unescaped `< start, end >` aborts parsing of the entire block, so neither `peaksInOneXIC` nor `chromPeaksInThisSequence` is ever name-checked. The escaping fix makes the block well-formed, and both bad names then surface as CS1572. Net improvement, but it converts a listed diagnostic into an unlisted one, which is precisely the hole. Both retargeted to `sharedPeak` and `chromPeaksInSharedPeak`.

**Numbers.** Measured against `upstream/master` with `GenerateDocumentationFile=true` forced on every project, non-incremental Release, `NoWarn` cleared, counting unique file+line+column: CS1570 41 (29 shipped / 12 test), CS1587 54 (5 shipped / 49 test) — 95 total, split 34 shipped / 61 test-and-Development. So "roughly 200 of those exist today" was wrong in both tenses: 95 before this branch, 0 after. The description's 60 test-side figure was right; 43 shipped looks like transposed digits for 34, and part of the gap may be that the original count predates the master merge.

That also reconciles your 14: shipped CS1574 measures 15 at current master. Fourteen are the ones degraded here; the fifteenth is `SdrfValidation.cs:38`, which arrived with the merge after the verification ran. `SdrfDocument.Results` fails because `Results` is inherited from the generic base `ResultFile<TResult>`, and Roslyn will not follow a qualified cref through a generic base. Naming the base explicitly fixes it.

Figures are from macOS / .NET SDK 10.0.400 — worth a CI re-run before anyone leans on them.

nbollis (dismissed):

mzLib/Directory.Build.props
```diff
@@ -5,6 +5,32 @@
          net10.0, since that is the default from net9.0 on; stated anyway so retargeting a project
          downwards cannot quietly reduce auditing to direct dependencies only. -->
     <NuGetAuditMode>all</NuGetAuditMode>
+
+    <!-- Emit the XML documentation file next to each assembly, and ship it in the package, so that
+         consumers get tooltips for mzLib types in their IDE. Roughly 1,900 <summary> blocks already
+         exist across the tree and none of them reached anyone outside this repository before now. -->
+    <!-- Only for projects that ship. An XML documentation file for a test or benchmark assembly
+         serves nobody, and generating it there was surfacing 60 of the 103 doc-comment warnings in
```
trishorts: Two carry-overs from the earlier round on this block.

**Line 13 still carries the retracted figure.** "60 of the 103 doc-comment warnings" is precisely the number the PR description now retracts -- the edit note re-measures it as 95 total, split 34 shipped / 61 test-and-Development. So the comment that lives in the repo permanently states a count the description already corrects, and this comment is what the next maintainer reads first.

Line 29 I can reproduce exactly, for what it is worth: non-incremental `Release` build of the solution here gives 0 errors, 0 for all six visible diagnostics, and CS0419 at 24 unique sites. It is only line 13 that is stale.

**Nothing holds the six at zero.** `TreatWarningsAsErrors` and `WarningsAsErrors` appear nowhere in the repository -- no `.props`, no `.csproj`, no workflow -- so the first PR that reintroduces a CS1570 gets a green build and this block quietly becomes fiction. The six left visible are exactly the set defined here as "the comment renders wrongly or not at all", which makes them the natural candidates for a ratchet next to the `NoWarn` on line 33:

```xml
<WarningsAsErrors>$(WarningsAsErrors);CS1570;CS1572;CS1574;CS1587;CS1711;CS1734</WarningsAsErrors>
```

CS0419 stays out, since 24 of them are accepted. Getting to zero was most of the work here; keeping it there is one line.

mzLib/MzLibUtil/Tolerance.cs
```diff
@@ -38,7 +38,6 @@ public abstract class Tolerance
         /// <summary>
         /// Creates a new tolerance given a unit, value, and whether the tolerance is ±
```
trishorts: Deleting the phantom `<param name="unit">` is right, but it leaves the summary directly above asserting the same two things it described: "given a unit, value, and whether the tolerance is ±", against `protected Tolerance(double value)`. `PpmTolerance.cs:29` and `AbsoluteTolerance.cs:29` carry the identical sentence over the identical one-argument constructor, and all three lost the same tag in this PR.

Before this change the sentence was at least accompanied by a tag that bound to nothing. Now it is the only prose on the member, and it ships. Same category as the `SpectraFileInfo` sentence: it describes an API that is not there, in a tooltip a reader trusts.

"Creates a new tolerance with the given value" covers all three.

mzLib/Quantification/QuantificationEngine.cs
```diff
@@ -11,14 +11,14 @@ namespace Quantification;
 /// 1) Creates one SpectralMatchMatrix per file
 ///     1a) Normalize the SpectralMatch Matrix for each file
 /// 2) Roll up to peptides for each file
-///     2a) Map the PSMs to peptides, creating a Dictionary<IBioPolymerWithSetMods, List<int>> Map mapping peptides to the indices of their PSMs in the QuantMatrix
+///     2a) Map the PSMs to peptides, creating a Dictionary of IBioPolymerWithSetMods to List of int, mapping peptides to the indices of their PSMs in the QuantMatrix
 ///     2b) Roll-up. The roll-up strategy will take in a QuantMatrix of PSMs and the > map, and output a Peptide QuantMatrix
```
trishorts: Stray `>` on this line -- "takes in a QuantMatrix of PSMs and the > map". Lines 14 and 21 are where this PR de-mangled `Dictionary<...>` into prose, and 5b on line 21 says plainly "and the map", so this reads like the tail of an eaten generic rather than intended text.

It is legal XML and raises no diagnostic, which is why the sweep passed over it, but this is the class summary for `QuantificationEngine` and it is now shipping. One character.

trishorts (commented):
Second pass. Rebuilt from `d7241fbb` and reproduced the verification independently: non-incremental `Release` build of `mzLib.sln` gives 0 errors, 0 for CS1570 / CS1572 / CS1574 / CS1587 / CS1711 / CS1734, and CS0419 at 24 unique file+line+column sites. Every one of the 98 `<file src=...>` paths in `mzLib.nuspec` resolves after that build, including all 33 new `.xml` entries, so `nuget pack` will not break the next release. The 17 documented assemblies match the 17 shipping projects exactly, with `Test` and `Development` correctly the only two excluded. All 12 wiki links in the new README return 200.

MetaMorpheus is unaffected by the extra files: its installer comparison is DLL-only (`InstallRunAndArtifact.yml:275-278`) and the CommandLine zip already excludes `*.xml` (line 49).

Everything from the last round that touched code is confirmed fixed -- `PrideArchiveClient`, `GetSequenceCoverage`, the `Mz` casing, the `SpectraFileInfo` equality sentence, the `wtOutput` tag, and both `CollectChromPeakInRuns` param names.

Three new notes below, none of them blocking. Five from the last round are still open and unanswered; all are tooltip accuracy rather than build:

- `ChromatographicPeak` -- "an integrated elution peak", but `Integrate` defaults to false and is recommended off
- `ChromatographicPeak` -- `DetectionType.MBR` also covers the random-RT decoys
- `Peptide` -- "roll-up" reads as summed; the code takes the max qualifying peak per file
- `MsDataFileReader` -- `.d` is not dispatched on extension
- `mzIdentML1_2_0.cs` -- the fix is lost on the next `xsd.exe` regeneration

No objection if any of those are deliberate, but worth a line saying so rather than leaving them silent.

mzLib/MzLibUtil/MrsNoiseEstimation/MRSNoiseEstimator.cs
trishorts: Confirmed fixed -- the `wtOutput` tag is gone from the block at line 147, and CS1572 is at 0 in my own non-incremental Release build. Resolving.

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Confirmed fixed -- both tags now read `sharedPeak` and `chromPeaksInSharedPeak`, matching the signature, so the escaping work actually lands in the tooltip. CS1572 at 0 in my own build. Resolving.

trishorts:
@acesnik heads up — I pushed four commits to this branch rather than leaving them as comments, since they were all mechanical and the branch was conflicting:

- `9791b836` the three figures in the `Directory.Build.props` rationale (60/103 → 61/95 per your own re-measurement note, "seven" → nine diagnostics, 248 → 243 of 930)
- `3ee0ae82` the `Tolerance`, `ComputeStdevOfNoisePixels` and `QuantificationEngine` summaries, plus `BooleanizeLevels`' `>9` → `>0`
- `b23d90a6` merge of master — the one conflict was `ICollapseStrategy`'s summary, where I took master's side because it describes the `IAggregationStrategy` split the signature already reflects
- `b2512bef` the one CS1574 the merge exposed: `QuantificationResults` crefs `PeptideWithSetModifications`, which `Quantification` can't reference without inverting the layering, so it's inline code now

Non-incremental Release build is 0 errors with every doc diagnostic at 0 except CS0419 at 24 unique sites — matching what the comment claims. Revert or reword any of it freely, it's your prose.

Five threads from the earlier round are still open and need you:

- `FlashLFQ/ChromatographicPeak.cs:12` — "integrated" when `integrate` defaults off
- `FlashLFQ/ChromatographicPeak.cs:14` — `DetectionType.MBR` vs "transferred"
- `FlashLFQ/Peptide.cs:10` — `CalculatePeptideResults` doesn't aggregate
- `Readers/MsDataFileReader.cs:13` — `.d` dispatch isn't by extension alone
- `MzIdentML/mzIdentML1_2_0.cs:18`

One consequence worth knowing: the push dismissed nbollis's approval, so this is back to 0 of 2.

nbollis (dismissed):

mzLib/FlashLFQ/ChromatographicPeak.cs
acesnik: 🤖 Not deliberate — fixed. You're right that the default configuration makes this the apex envelope rather than an integrated area, and "integrated" is exactly the word that would produce a bad cross-tool intensity comparison. The summary now says the intensity is the apex envelope's unless the run asked for integration, and notes that `Integrate` defaults false with the engine recommending it stay there, so the integrated case is the exception.

mzLib/FlashLFQ/Peptide.cs
acesnik: 🤖 Fixed. Confirmed `CalculatePeptideResults` takes `Max(p => p.Intensity)` per file rather than combining anything, so "roll-up" was misleading in the specific direction you describe. It now says the per-file intensity is the single highest-intensity qualifying peak, states plainly that it will not reconcile against a total of QuantifiedPeaks, and lists what qualifying means — unambiguous by full sequence, non-decoy, and for MBR within the q-value threshold and not a random-RT decoy.

mzLib/Readers/MsDataFileReader.cs
acesnik: 🤖 Fixed, and taking both halves of your suggestion. The summary now says the format is identified from the path and, for container formats, from the files inside it, with a sentence spelling out why `.d` is the reason extension alone cannot work: one extension, two readers depending on `analysis.baf` versus `analysis.tdf`/`.tsf`, and a throw before any reader is constructed when neither is present. The format list also now reads as non-exhaustive, which covers the `_ms1.msalign`/`_ms2.msalign` omission.

mzLib/FlashLFQ/ChromatographicPeak.cs
acesnik: 🤖 Fixed. Confirmed `MbrChromatographicPeak` passes `detectionType: DetectionType.MBR` for random-RT decoys as well, so the tooltip was inviting exactly the filter that inflates a quantification. The summary now says an MBR detection type does not on its own mean a real transfer, that the random retention time decoys used for MBR FDR carry it too, and that callers pair the check with `RandomRt`.

mzLib/MzIdentML/mzIdentML1_2_0.cs
acesnik: 🤖 Taken, and done the durable way. `CS1587` is now scoped off in `MzIdentML.csproj` and the generated file is back to pristine, so there is nothing in it to reapply after the next `xsd.exe` run. The csproj comment records why it is suppressed rather than edited out, and notes the cost: a genuinely misplaced doc comment in the hand-written `MzidIdentifications.cs` would also go unreported. That seemed the better trade for a project that is otherwise entirely generated.

acesnik:
🤖 One thing that came out of rebuilding after the merges: CS1572 is no longer at zero, and it is not from anything reviewed here. `QuantificationEngine` has two `<summary>` blocks stacked, so `GetPsmToPeptideMap`'s documentation sits above `SingleIdentifiedBioPolymerOrNull` and attaches to it — hence param tags for `smMatrix` and `modifiedBioPolymers` that the method does not have, and `GetPsmToPeptideMap` documented nowhere.

It arrived with #1193 and is on master right now, silent there only because documentation generation is off. Turning it on is what surfaced it, which is a fair advertisement for this PR. The block is moved to its own method here, so the seven visible doc diagnostics are back to CS0419 alone at the documented 24 sites.

trishorts (approved):

nbollis (approved):

PR #1163, opened 2026-08-15
Binary search GetClosestOneBasedSpectrumNumber instead of walking every scan

### LINKED ISSUE #156: GetClosestOneBasedSpectrumNumber - implement binary search
it would be helpful (faster) if this method:

int GetClosestOneBasedSpectrumNumber(double retentionTime)

was implemented with binary search instead of iterating through scans

acesnik:
🤖 Closes #156, which asked for this in 2017 and which the method's own `TODO` also asked for.

`GetClosestOneBasedSpectrumNumber` walked scans from the first, stopping once a scan's distance to the target retention time grew. `GetMsScansInTimeRange` starts every range query here.

## Measured

2000 lookups at uniform random retention times, synthetic files:

| scans | linear | binary | |
|---|---|---|---|
| 1,000 | 11.98 ms | 0.47 ms | **26×** |
| 10,000 | 35.20 ms | 0.70 ms | **50×** |
| 100,000 | 143.24 ms | 0.92 ms | **156×** |

The shape matters more than the ratio: linear grows with scan count while binary barely moves.

## It does not add an assumption

Ascending retention times were already assumed. The linear version stopped at the first growth in distance, so it never looked past a local minimum — on unsorted input it returned a local answer, not the closest. This makes the existing assumption explicit rather than introducing one.

## Two behaviours preserved, and how the second was found

**A target falling exactly between two scans resolves to the later one.** That was the old behaviour and callers were written against it.

**Scans can share a retention time**, and the linear walk kept going while the distance stayed equal, so it came to rest on the **last** scan of such a run. A plain binary search lands on the **first**. The result now advances to the end of the run.

I did not reason my way to that second case — a differential test caught it. Comparing old against new across 400 random ascending files, exercising every scan's exact retention time, exact midpoints between neighbours, out-of-range queries and random values, the first version disagreed on **19 of 33,056 queries**, every one of them a run of equal retention times. After the fix: **33,056 queries, 0 disagreements**.

That is the argument for the change being safe, and it is stronger than the unit tests alone.

## Tests

`TestGetClosestOneBasedSpectrumNumber` pins exact hits, out-of-range clamping, nearer-neighbour on both sides, exact midpoints resolving later, equal-retention-time runs, a single-scan file, and a file that loads with no scans. All 18 asserted values were verified against the implementation before committing, and the two pre-existing assertions in `TestMzML` still hold.

The zero-scan case is narrower than it first looks, which CI is what showed me. `CheckIfScansLoaded()` is `Scans != null && Scans.Length > 0`, so a file with no scans always counts as *not loaded* and the method attempts a load before it can return anything. The `NumSpectra == 0` guard is therefore only reachable when a subclass's load succeeds and still yields zero scans — a genuinely empty mzML. My first version of the test used `FakeMsDataFile`, whose `LoadAllStaticData` throws, so it died there rather than reaching the guard; the fixture now uses a local double whose load is a no-op. That was my error in choosing the double, not a defect in the change, but the reachability limit is worth recording.

The tests also record a trap worth knowing: **`FakeMsDataFile` declares its own `GetClosestOneBasedSpectrumNumber` that hides rather than overrides this one**, and returns an insertion point rather than the closest scan. A test holding the derived type would quietly exercise the fake instead of mzLib. Its `LoadAllStaticData` also throws, which is a second reason it cannot serve here, and the two problems compound: the shadowing hides the method under test, and the throw blocks the one case that needs a successful load. Hence a local double rather than typing the variables as `MsDataFile`. That shadowing method looks worth deleting on its own, but it is not this change.

## What I could not run

`Test.csproj` targets `net10.0-windows`, so NUnit could not run here; it compiles, and every expected value was confirmed through a harness against the real class. Windows CI is what executes the fixture.


mzLib/MassSpectrometry/MsDataFile.cs
```diff
@@ -148,17 +148,43 @@ public virtual int GetClosestOneBasedSpectrumNumber(double retentionTime)
...
+            int low = 1;
+            int high = NumSpectra;
+            while (low < high)
             {
-                double diff = Math.Abs(GetOneBasedScan(i + 1).RetentionTime - retentionTime);
-                if (diff > bestDiff)
-                    return i;
-                bestDiff = diff;
+                int mid = low + (high - low) / 2;
+                if (GetOneBasedScan(mid).RetentionTime < retentionTime)
```
trishorts: **Major.** This probes an arbitrary scan slot through the virtual `GetOneBasedScan`, and two readers override it to index a *sparse* array.

`Mgf.GetOneBasedScan` (`Mgf.cs:80-83`) and `MsAlign` (`MsAlign.cs:87-89`) both return `IndexedScans[scanNumber - 1]`, where `IndexedScans` is sized to the **maximum** one-based scan number and deliberately null-padded -- the comment at `Mgf.cs:62-64` says so outright ("the final scans array will contain a null value"). `Scans` is the dense array, so `NumSpectra` is the dense count and the holes fall *inside* the searched range `[1, NumSpectra]`.

Concrete regression. Scan numbers `1,2,3,10,11,12,13,14`: `NumSpectra` is 8, `IndexedScans` has length 14, and slots 4-8 are null. For a target just above the first scan's retention time:

- old walk: probes slots 1 and 2, sees the distance grow, returns 1;
- new search: first probe is `mid = 1 + 7/2 = 4` -> `IndexedScans[3]` is null -> `NullReferenceException`.

Gaps arise whenever a converter skips a scan number, or when `filterParams` drops a scan in `Mgf.LoadAllStaticData` (`Mgf.cs:53`). The run-advance probe `GetOneBasedScan(low + 1)` on line 184 has the same exposure.

Suggest searching the dense `Scans` array directly rather than through the virtual indirection (making the array-position-to-one-based-number mapping explicit), or skipping null probes. Either way a gapped-scan-number fixture would pin it.

mzLib/MassSpectrometry/MsDataFile.cs
```diff
@@ -148,17 +148,43 @@ public virtual int GetClosestOneBasedSpectrumNumber(double retentionTime)
...
+                    low = mid + 1;
+                else
+                    high = mid;
```
trishorts, line 202: `NaN` retention times take this branch on every probe, and the answer flips from the last scan to the first.

`Mgf.cs:149` initialises `double rtInMinutes = double.NaN; //default when unknown`, so an MGF without `RTINSECONDS` yields scans whose retention times are all `NaN`. `NaN < retentionTime` is false, so every probe drives `high = mid` and `low` converges to 1. The run-advance on line 184 cannot compensate, because `NaN == NaN` is also false.

The old loop computed `diff = NaN`, and `NaN > bestDiff` is false too, so it never returned early and fell through to `return NumSpectra` -- the **last** scan. The new code returns **1**.

Neither answer is meaningful, but the flip is silent, and `GetMsScansInTimeRange` inherits it: it goes from yielding one scan to yielding every scan in the file. The differential test described in the PR body used random ascending non-`NaN` files, so it would not have covered this. Worth deciding the contract for non-finite retention times explicitly and pinning it, so the change is deliberate rather than incidental.

mzLib/MassSpectrometry/MsDataFile.cs
```diff
@@ -148,17 +148,43 @@ public virtual int GetClosestOneBasedSpectrumNumber(double retentionTime)
...
+                double distanceBefore = Math.Abs(GetOneBasedScan(low - 1).RetentionTime - retentionTime);
+                if (distanceBefore < distanceAfter)
+                    return low - 1;
+            }
+
+            // Scans can share a retention time. The linear version kept walking while the distance stayed
+            // equal, so it came to rest on the last scan of such a run; binary search lands on the first.
+            // Advancing to the end of the run keeps the answer identical to the previous implementation.
+            double closestRetentionTime = GetOneBasedScan(low).RetentionTime;
+            while (low < NumSpectra && GetOneBasedScan(low + 1).RetentionTime == closestRetentionTime)
```
trishorts: This keeps the worst case at O(n), so the method is O(log n + runLength) rather than the O(log n) claimed in the comment above.

The run length is unbounded: `MsAlign` assigns `retentionTime = DefaultErrorValue` (`-1`, `MsAlign.cs:16,436`) whenever the retention-time field fails to parse, so a file with unparseable retention times is a single run of length `NumSpectra`. Each call then degrades to the linear cost the PR set out to remove, and `GetMsScansInTimeRange` re-incurs it on every range query. Correctness is unaffected.

A second binary search for the upper bound of the run -- the first scan whose retention time exceeds `closestRetentionTime`, minus one -- keeps the whole method O(log n) regardless of run length.

Unrelated, on the same line: the exact `==` here is **correct** and I would not change it. The old walk continued only while `diff` was not strictly greater than `bestDiff`, which is itself an exact-equality condition, so a tolerance comparison would *diverge* from the implementation this PR is matching (retention times a few ULPs apart would be merged into one run where the old code split them). Worth saying so in the comment, because the usual house rule against float equality will make someone want to "fix" it later.

mzLib/MassSpectrometry/MsDataFile.cs
```diff
@@ -148,17 +148,43 @@ public virtual int GetClosestOneBasedSpectrumNumber(double retentionTime)
...
+            {
+                double distanceAfter = Math.Abs(GetOneBasedScan(low).RetentionTime - retentionTime);
+                double distanceBefore = Math.Abs(GetOneBasedScan(low - 1).RetentionTime - retentionTime);
+                if (distanceBefore < distanceAfter)
+                    return low - 1;
+            }
+
+            // Scans can share a retention time. The linear version kept walking while the distance stayed
+            // equal, so it came to rest on the last scan of such a run; binary search lands on the first.
+            // Advancing to the end of the run keeps the answer identical to the previous implementation.
```
trishorts: Worth recording that the behaviour being preserved here is a bug at the only production caller -- I am not asking for it to change in this PR.

`GetMsScansInTimeRange` (`MsDataFile.cs:128-144`) seeds from this method and then walks **forward only**: it increments past anything with `rt < firstRT` and yields until `rt > lastRT`. Because the closest-match answer is the *last* scan of an equal-retention-time run, the earlier members of that run are never visited even though they sit inside the window.

Retention times `[1.0, 5.0, 5.0, 5.0, 9.0]`, range `(5.0, 10.0)`: the seed is scan 4, so it yields scans 4 and 5. Scans 2 and 3 are also at RT 5.0 and inside the window, and are dropped. A whole-file window over `[5,5,5]` returns one scan of three. That reaches `MsDataFileExtensions.ExtractIonChromatogram` (`MsDataFileExtensions.cs:17`), which is every XIC mzLib extracts.

This is pre-existing -- the old walk came to rest on the last of the run too -- and preserving it is the right call for a performance PR. The reason it is worth a comment is that "identical to the old implementation" is exactly the criterion that hides it, and this is the one spot where a plain `LowerBound` returning the *first* member of the run would have fixed it for free. Filed as #1172 rather than expanding this PR's scope.

mzLib/MassSpectrometry/MsDataFile.cs
```diff
@@ -148,17 +148,43 @@ public virtual int GetClosestOneBasedSpectrumNumber(double retentionTime)
...
-            double bestDiff = double.MaxValue;
-            for (int i = 0; i < NumSpectra; i++)
+            if (NumSpectra == 0)
+                return 0;
+
+            // Binary search rather than a walk. Retention times ascending is not a new assumption: the
+            // linear version this replaces stopped as soon as a scan's distance to the target grew, so it
+            // never looked past the first local minimum either. Making it explicit turns O(n) into
+            // O(log n), which matters because callers such as GetMsScansInTimeRange start here.
+            int low = 1;
```
trishorts: mzLib already has a helper for this shape of search: `Chemistry.ClassExtensions.GetClosestIndex(this double[] sortedArray, double target, ArraySearchOption searchType = Closest)` at `Chemistry/ClassExtensions.cs:86`. It wraps `Array.BinarySearch`, bitwise-complements the miss, clamps both ends, and its `Closest` branch is `backwardsDiff < forwardsDiff` -- which resolves an exact midpoint to the **later** index, the same tie-break hand-rolled here. `MzSpectrum.GetClosestPeakIndex` (`MzSpectrum.cs:615`) is a one-line consumer, and `Averagine`/`ClassicDeconvolutionAlgorithm` use it too.

I am **not** asking you to switch to it. Two things make it a poor fit here, and both are worth having on the record:

1. it takes a `double[]`, so it needs a materialised array of retention times -- O(n) per call unless cached on `MsDataFile`, which would defeat the change and need invalidating on `LoadAllStaticData`;
2. `Array.BinarySearch`'s contract for duplicate keys is implementation-defined, so it cannot reproduce the last-scan-of-an-equal-run semantics you are deliberately preserving. `Mslindex.BinarySearchLowerBound` (`Mslindex.cs:1017`) hand-rolls its loop for exactly that reason and documents it.

So the hand-rolled loop is justified -- a line in the comment saying the existing helper was considered and why it does not fit would stop the next reviewer re-raising it.

Adjacent, for a follow-up rather than this PR: `IndexingEngine.GetXic` (`PeakIndexing/IndexingEngine.cs:94-104`) still converts a retention time to a scan index with a linear `foreach` over `ScanInfoArray`. Same problem, different class.

mzLib/Test/FileReadingTests/SpectraFileReading/TestGetClosestOneBasedSpectrumNumber.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <summary>
+        /// Scans can share a retention time, and the linear implementation kept walking while the distance
+        /// stayed equal, so it came to rest on the last scan of the run. A plain binary search lands on the
+        /// first instead. This is the case that caught the first version of the change: a differential test
+        /// over random data disagreed on 19 of 33,056 queries, all of them runs of equal retention times.
+        /// </summary>
+        [Test]
+        public static void RunsOfEqualRetentionTimesReturnTheLastScanInTheRun()
+        {
+            MsDataFile file = FileWithRetentionTimes(1.0, 5.0, 5.0, 5.0, 9.0);
```
trishorts, line 157: The equal-retention-time run is strictly interior here, so neither boundary of the new run-advance loop is exercised. With `(1.0, 5.0, 5.0, 5.0, 9.0)`, `low` is always `> 1` and the advance always stops before `low == NumSpectra` -- so the `low < NumSpectra` guard on `MsDataFile.cs:184` never does anything, and the path where the `low > 1` block is skipped entirely and the advance runs from index 1 is never taken. That is the exact failure mode the PR says the differential test caught, so it is the one most worth pinning.

Two cases would cover it:

- `(1.0, 5.0, 9.0, 9.0)` queried at `9.0` and above -> expect `NumSpectra`;
- `(5.0, 5.0, 9.0)` queried at `0` and at `5.0` -> expect the last scan of the leading run.

Two smaller things in this file:

- Nothing here drives `GetMsScansInTimeRange`, even though `ExactMidpointResolvesToTheLaterScan` justifies the tie-break by saying that caller was written against it. The claim that range queries are unaffected currently rests on `TestMzML.cs:738-739`, a two-scan file at retention times 1 and 2, which exercises neither midpoints nor equal-RT runs. One or two assertions on the returned scan numbers over these same fixtures would close that.
- `using System;` on line 1 is unused -- nothing in the file resolves through it.

trishorts (changes requested):
Nice change, and the differential test against the old implementation is the right way to argue for it -- that is a stronger safety case than the unit tests alone. The standing `TODO` on the old line 151 asked for exactly this.

One blocker: the search probes scan slots through the virtual `GetOneBasedScan`, which `Mgf` and `MsAlign` back with a null-padded array, so gapped files can throw where the linear walk answered. The rest are smaller, plus one note about an existing helper and one about a pre-existing bug this change preserves.

mzLib/MassSpectrometry/MsDataFile.cs
acesnik: 🤖 Confirmed and fixed in e6948c17 — the search now indexes `Scans` directly and converts back through `OneBasedScanNumber`.

Two things that make the case stronger than stated.

**The linear walk threw too.** On your fixture, only the 1.5 target survived on the old code; 3.0 and 13.0 both `NullReferenceException`, because the walk just happened to stop before the first hole. So this is pre-existing, and the binary search widened it from most targets to all of them rather than introducing it. Searching `Scans` repairs it rather than avoiding a regression.

**`tester.mgf` already does this.** No new fixture needed to demonstrate it — `Test/DataFiles/tester.mgf` has `SCANS=14`, `SCANS=19`, two entries with no `SCANS` line, then `SCANS=2`, so after ordering it is scan numbers 2, 14, 19, 20, 21 with `NumSpectra` 5 and 16 null slots. `GetClosestOneBasedSpectrumNumber` on it throws on master today and returns 14 with the fix.

Returning a position would only have moved the crash into the caller, since `Mgf.GetOneBasedScan(7)` is `IndexedScans[6]`. Converting through `OneBasedScanNumber` is safe everywhere else because `Mzml` rebuilds `Scans` indexed by scan number for out-of-order files (`Mzml.cs:205-213`), so the two only diverge for the readers that override the accessor. A gapped fixture pins it, and I checked it NREs against the unfixed code rather than passing vacuously.

**Beyond this method.** The same `1..NumSpectra` through `GetOneBasedScan` shape appears in `GetMS1Scans` (`:92`), `GetMsScansInTimeRange` (`:129`) and `MzmlMethods` (`:379`, `:469`). All four throw on `tester.mgf`; `GetAllScansList`, which reads `Scans` directly, does not. The `MzmlMethods` pair means converting a gapped MGF to mzML crashes. Out of scope here — I will open an issue, since the root cause is that `NumSpectra` and `GetOneBasedScan` disagree about what an index means, and fixing that once closes all four.

mzLib/MassSpectrometry/MsDataFile.cs
acesnik: 🤖 Right on both counts, fixed in e6948c17 with the second binary search you describe. Measured on a file where every scan shares one retention time, probes go 20,011 -> 31 at 20,000 scans (200 -> 17, 2,000 -> 23). Your point about the exact `==` is now in the comment, worded so the next reader can see it is deliberate rather than an oversight.

mzLib/MassSpectrometry/MsDataFile.cs
acesnik: 🤖 Checked both objections and they hold: it takes `double[]`, and its `Closest` branch is `backwardsDiff < forwardsDiff`, so it does resolve an exact midpoint to the later index the same way. The comment now records that it was considered and why it does not fit, citing `Mslindex.BinarySearchLowerBound` as the precedent for hand-rolling.

`IndexingEngine.GetXic` noted for the follow-up issue. It has a second problem worth folding in: with non-finite retention times `scan.RetentionTime < retentionTime` is false on the first iteration, so it breaks immediately and returns `scanIndex = -1`.

mzLib/Test/FileReadingTests/SpectraFileReading/TestGetClosestOneBasedSpectrumNumber.cs
acesnik: 🤖 All addressed in e6948c17: a run at the end of the file, a run at the start, a gapped-scan-number fixture, `GetMsScansInTimeRange` assertions over the same fixtures, and the unused `using` dropped. Confirmed the gapped assertion fails with `NullReferenceException` against the unfixed code rather than passing vacuously.

Also ran a differential against a verbatim copy of the pre-PR implementation over 400 random files with deliberate duplicate retention times: 24,000/24,000 queries agree.

mzLib/MassSpectrometry/MsDataFile.cs
acesnik: 🤖 Agreed, and left alone deliberately. It is now asserted in `RangeQueriesSeededByThisMethodYieldTheExpectedScans` with a pointer to #1172, so the current behaviour is pinned and changing it has to be a deliberate act rather than an accident.

mzLib/MassSpectrometry/MsDataFile.cs
acesnik: 🤖 Confirmed, including the `GetMsScansInTimeRange` consequence — an all-NaN file goes from yielding one scan to yielding all of them.

Not guarding on it, and 0bce7cf8 records why in the commit message rather than leaving it implicit. Detecting NaN needs either an O(n) scan per call or a cached flag, and `Scans` has a protected setter that any reader can reassign, so the flag could not be invalidated reliably. It would also give false confidence: `tester.mgf` interleaves NaN with real values, and readers order scans by scan number rather than by retention time, so a finite-but-unsorted file breaks the search just as thoroughly while passing a NaN check.

Instead the assumption is now stated on the method and the current answers are pinned in a test marked unspecified, so a future change surfaces as a deliberate edit rather than the silent flip you caught. One of them pins that a query past the end of an interleaved file returns a scan with no retention time at all, which makes the point better than prose does.

Warning from the reader would be better than either, but there is no logging facility to warn through — only `Debug.WriteLine` in the spectral-library readers, which compiles out in Release. That, and the remaining crash sites, are #1175.

nbollis (approved):

trishorts (approved):
All six threads addressed, and I checked the code rather than the replies.

- The blocker is gone: the search indexes `Scans` directly and converts back through `OneBasedScanNumber`, so a gapped-scan-number file can no longer land on a null slot in `IndexedScans`. `GappedScanNumbersAreSearchableAndTheAnswerRoundTrips` pins both the search and the round-trip through the reader's own accessor.
- The run advance is a second binary search, so the method is O(log n) even when the whole file is one run.
- `GetClosestIndex` considered-and-rejected, and the exact `==`, are both recorded in the comment, which is what stops them being re-raised.
- Non-finite retention times are documented as unspecified and pinned rather than guarded. I agree with that call: a NaN check would not catch a finite-but-unsorted file, which breaks the search just as thoroughly.
- #1172 is asserted rather than fixed, which is right for a performance PR, and #1175 captures the wider `NumSpectra`/`GetOneBasedScan` disagreement.

I re-ran the differential myself against a verbatim copy of the old walk over ~194,000 queries on random files seeded with duplicate retention times, exact hits, exact midpoints and out-of-range targets: no disagreements.

One cosmetic leftover for whenever you next touch the file: the doc comment for `RangeQueriesSeededByThisMethodYieldTheExpectedScans` ended up above `GappedScanNumbersAreSearchableAndTheAnswerRoundTrips`, so that test carries two consecutive `<summary>` blocks and the range test has none. Not worth a push on its own.

Issue #211, opened 2017-07-28
Add MGF Writer


trishorts:
mascot generic format

for reference: 
http://kirchnerlab.github.io/libmgf/
https://pythonhosted.org/pyteomics/api/mgf.html
https://proteomicsresource.washington.edu/mascot/help/data_file_help.html
http://ftp.mi.fu-berlin.de/pub/OpenMS/release-documentation/html/TOPP_MascotAdapter.html

PR #1164, opened 2026-08-15
Read negative charge states from MGF CHARGE lines

acesnik:
🤖 Fixes a sign-parsing bug in the MGF reader: `CHARGE=2-` parses as `+2`, so negative-mode files load with positive charge states and `Polarity.Positive`.

Found while scoping #211, reading the parser to work out what a writer would need to emit.

## The bug

`Readers/Mgf/Mgf.cs`, in the `CHARGE` branch:

```csharp
charge = Convert.ToInt32(entry.Substring(0, entry.Length - 1));
if (entry[entry.Length - 1].Equals("-"))
{
    charge *= -1;
}
```

`entry[entry.Length - 1]` is a `char` and `"-"` is a `string`, so this binds to `char.Equals(object)`, which returns true only for a `char` of equal value. Against a string it is always false and the negation never runs.

Polarity is derived from the sign of the charge:

```csharp
charge > 0 ? Polarity.Positive : Polarity.Negative
```

so the effect reaches polarity as well as charge state.

## Why it survived

Nothing exercised the path. All five MGF fixtures in `Test/DataFiles` use `CHARGE=2+` or `CHARGE=3+`, and all four polarity assertions in `TestMgf.cs` check `Polarity.Positive`.

## The fix

Compare `char` to `char`. One line, plus a fixture with one negative and one positive scan, and a test covering both.

## Verification

`Test.csproj` is `net*.0-windows`, so the suite does not run on macOS. I built a throwaway net10.0 console harness inside the repo tree against `Readers.csproj` and ran the new fixture through `MsDataFileReader` both ways.

Before:

```
scan1 (CHARGE=2-) charge=2  polarity=Positive
scan2 (CHARGE=3+) charge=3  polarity=Positive
```

After:

```
scan1 (CHARGE=2-) charge=-2 polarity=Negative
scan2 (CHARGE=3+) charge=3  polarity=Positive
```

The test asserts the positive scan alongside the negative one, so an inverted sign cannot pass it.

All 16 `TestMgf` tests were executed off Windows through a throwaway net10.0 harness that compiles the real test file, and pass on Windows CI.

## Also fixed here (added after review)

**The unsigned case was worse than this description first said, so it is fixed rather than deferred.** It was characterised as throwing, which holds only for a *single* digit. `Substring(0, Length - 1)` dropped the last character unconditionally, so a multi-digit unsigned line lost a digit and parsed *silently*:

```
CHARGE=12  ->  charge 1   (silent — 1 is a valid charge)
CHARGE=2   ->  FormatException
```

A wrong charge that loads is more dangerous than an exception that doesn't, so the suffix is now stripped only when it is actually a sign. The two failure modes get separate fixtures on purpose: in one file the single-digit throw aborts the load and masks the multi-digit case, so the assertion that fails is the exception rather than the value.

No existing expectation moves — every pre-existing `.mgf` fixture uses `2-`, `2+` or `3+`, so nothing encoded the truncated value.

**`TestDynamicMgf` was also testing nothing**, and is fixed here since it is the test file this PR touches. `GetOneBasedScanFromDynamicConnection` short-circuits to the static cache when scans are loaded, and the test loaded statically first, so every scan it "compared" was reference-identical to itself. It now uses a second reader with no static load and asserts the instances are distinct. `EliminateZeroIntensityPeaksFromMgfOnFileLoad` had the same shape and is fixed the same way. The equivalent defect in the mzML and Thermo tests is #1176.

## Not fixed here

- `CHARGE=2+ and 3+` still throws. It is genuinely spec-legal — `CHARGE=2+ and 3+` appears verbatim in the [Matrix Science data-file documentation](https://www.matrixscience.com/help/data_file_help.html), which defines the parameter as "N- to N+ where N is an integer and combinations" and permits it both in the header and locally inside a `BEGIN IONS` block. A header-level one is harmless here, because the static loader skips every line until `BEGIN IONS`; a local one reaches the parser and fails the whole file. It stays deferred because it needs a decision rather than a fix: the line means "this precursor's charge is ambiguous, try both", so taking the first candidate silently discards the other, and treating an ambiguous charge as *unknown* — letting deconvolution decide — may be the more faithful reading.

*Files: `Readers/Mgf/Mgf.cs`, `Test/FileReadingTests/SpectraFileReading/TestMgf.cs`, `Test/Test.csproj`, `Test/DataFiles/negativeModeCharge.mgf`.*


mzLib/Readers/Mgf/Mgf.cs
```diff
@@ -177,7 +177,8 @@ public static MsDataFile LoadAllStaticData(string filePath, FilteringParams filt
                 {
                     string entry = sArray[1];
                     charge = Convert.ToInt32(entry.Substring(0, entry.Length - 1));
```
trishorts: Not asking you to fix this here — you explicitly deferred it and I think that's the right call for reviewability. But **the deferred bug is worse than the description says**, and that changes how the follow-up should be prioritised.

The description characterises the unsigned case as throwing:

> `CHARGE=2` with no sign suffix — `Substring(0, Length - 1)` drops the digit and `Convert.ToInt32("")` throws.

That is true only for a **single-digit** charge. `Substring(0, entry.Length - 1)` drops the last character unconditionally, so a multi-digit unsigned line loses its last digit and parses silently:

```
CHARGE=12   -> Substring -> "1"   -> charge = 1     (silent, no exception)
CHARGE=2    -> Substring -> ""    -> FormatException
CHARGE=12+  -> Substring -> "12"  -> charge = 12    (correct)
```

I ran those three through the BCL to be sure rather than reasoning about it.

So the unsigned path is not "loud failure only" — `CHARGE=12` yields charge 1 and, via `:240`, `Polarity.Positive`, with nothing to indicate anything went wrong. A wrong charge that loads is a good deal more dangerous than an exception that doesn't, and unsigned `CHARGE=` lines are emitted by several MGF writers.

Worth capturing in the follow-up issue explicitly, because the loud-throw framing undersells it. The conditional strip handles all three cases:

```csharp
char last = entry[^1];
string digits = (last == '+' || last == '-') ? entry[..^1] : entry;
charge = Convert.ToInt32(digits);
if (last == '-') charge *= -1;
```

mzLib/Test/FileReadingTests/SpectraFileReading/TestMgf.cs
```diff
@@ -167,6 +167,28 @@ public void ReadsPrecursorIntensityWhenPresent()
...
+        /// the sign of that charge. The positive scan is asserted alongside it so an inverted sign test
+        /// cannot pass.
+        /// </summary>
+        [Test]
+        public void ReadsNegativeChargeAndPolarityFromChargeLine()
+        {
+            string path = Path.Combine(TestContext.CurrentContext.TestDirectory, "DataFiles", "negativeModeCharge.mgf");
+
+            var reader = MsDataFileReader.GetDataFile(path);
+            reader.LoadAllStaticData();
```
trishorts, line 185: Optional, and arguably past this PR's stated scope — but there's a pre-existing hole here you may want to know about regardless.

This asserts the fix through `LoadAllStaticData` only. `Mgf` has a second public entry, `InitiateDynamicConnection()` + `GetOneBasedScanFromDynamicConnection(...)`, which re-enters the same `GetNextMsDataOneBasedScanFromConnection` but from a `BuildIndex`-derived byte offset. The parser body is shared, so **your fix does apply to both paths** — this is a coverage gap, not a defect.

The trap, if you do decide to close it: adding `[TestCase("negativeModeCharge.mgf")]` to the existing `TestDynamicMgf` would not exercise anything, because `GetOneBasedScanFromDynamicConnection` short-circuits (`Mgf.cs:86-88`):

```csharp
if (CheckIfScansLoaded() && scanNumber <= IndexedScans.Length)
    return GetOneBasedScan(scanNumber);
```

and `TestDynamicMgf` calls `LoadAllStaticData()` *before* `InitiateDynamicConnection()`.

**Which means `TestDynamicMgf` does not currently test the dynamic path at all** — every scan it fetches comes back from the static cache, so it compares each scan against itself and all ~20 of its assertions are tautologies. That's pre-existing and entirely separate from your change, but it is a real hole in the file you're touching and probably deserves its own issue.

If you want dynamic coverage here, it has to open the fixture with `InitiateDynamicConnection()` and no preceding static load, then close the connection.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMgf.cs
```diff
@@ -167,6 +167,28 @@ public void ReadsPrecursorIntensityWhenPresent()
...
+
+            var negative = reader.GetOneBasedScan(1);
+            NUnit.Framework.Assert.That(negative.SelectedIonChargeStateGuess, Is.EqualTo(-2));
+            NUnit.Framework.Assert.That(negative.Polarity, Is.EqualTo(Polarity.Negative));
```
trishorts: Minor, and it does not weaken the test's central claim — but these two assertions cannot disagree, so the doc comment slightly oversells what is being pinned.

`MsDataScan`'s constructor normalizes the charge sign to the polarity (`MassSpectrometry/MsDataScan.cs:79-85`), and `Mgf.cs:240` derives the polarity it passes from the very same `charge` local. So whichever of the two the parser got right, the other is manufactured from it — the test pins one bit, the parsed sign, which *is* the thing this PR fixes.

The practical consequence is narrow: an alternative "fix" that set `Polarity.Negative` from some other signal while leaving `charge` at `+2` would still pass both assertions, because `MsDataScan` would flip the charge to match.

Either trim the doc comment to say what is actually pinned, or add one orthogonal assertion the normalization cannot manufacture — `GetAllScansList().Count == 2` plus the distinct retention times would also pin that each `BEGIN IONS` block parsed as its own scan.

While you're here: there's an established shape for exactly this concern two files over — `TestMsDataFile.cs:358-374` `NegativeModeSetsCorrectCharge_FromFile` — which puts a failure message on every assertion. Mirroring the naming and the message idiom would make this the MGF analogue of that test rather than a standalone.

trishorts (commented):
Good catch, and the diagnosis in the description is exactly right — I confirmed the pre-fix line at `smith/master @ 5ba13155`:

```csharp
179:   charge = Convert.ToInt32(entry.Substring(0, entry.Length - 1));
180:   if (entry[entry.Length - 1].Equals("-"))
```

`char.Equals(object)` against a `string` is always false, and `Mgf.cs:240` derives polarity straight from that charge, so negative-mode files really did load as `Positive`/`+2`. `int charge = 2; //default when unknown` at `:146` means unsigned scans land on `+2` rather than tripping the negative branch, so the description's account holds there too.

**The strongest argument for this fix is one the description doesn't make.** `MsDataScan` already codifies the invariant, `MassSpectrometry/MsDataScan.cs:79-85`:

```csharp
// Ensure the charge of the selected ion matches the polarity of the scan
SelectedIonChargeStateGuess = Polarity switch
{
    Polarity.Negative when selectedIonChargeStateGuess is > 0 => -selectedIonChargeStateGuess,
    Polarity.Positive when selectedIonChargeStateGuess is < 0 =>  Math.Abs(selectedIonChargeStateGuess.Value),
    _ => selectedIonChargeStateGuess
};
```

Mgf was the one reader silently violating a rule the core type already declares. Worth a line in the description — it makes this obviously correct rather than a matter of taste.

**I checked whether the same char-vs-string mistake exists elsewhere. It does not.** `git grep -P '\.Equals\(\x22.\x22\)'` over all production `.cs` at master returns exactly one hit — the line you fixed. The near-misses are all correct code (`DigestionMotif.cs:164` is char-vs-char; `ModificationLoader.cs:466` is `string[]` indexing). No sibling issue needs filing. Worth noting `Search-Repo.ps1` silently returns "no matches" for any term containing a double quote, so this family of bug is invisible to the usual tooling — `git grep -P` with `\x22` is the reliable form.

**Blast radius, since this changes loaded values.** Inside mzLib it is contained: `ToMass`/`ToMz` use `Math.Abs`/`Math.Sign` (`Chemistry/ClassExtensions.cs:40-64`), deconvolution reads `deconParams.Polarity` rather than `scan.Polarity`, and averagine scoring wraps charge in `Math.Abs`. The one output-visible change is mzML export — `MzmlMethods.cs:551` writes the signed value into `MS:1000041` and `:635` now emits `MS:1000129 "negative scan"`. The polarity term becomes correct; a signed `MS:1000041` is unusual versus most vendor converters, though the schema types it as a string so nothing fails validation. Round-trip is lossless (`Mzml.cs:653` parses a leading `-`). Downstream of mzLib, anyone reading negative-mode MGF previously got positive charges, so this is a real behaviour change for them — the `integration` job that clones MetaMorpheus should surface any fallout.

One style note in the other direction: the added comment narrates the defect that was removed rather than the contract the branch now implements. Reviewers here routinely ask for change-narration comments to be dropped, so restating it as a format fact — a trailing `-` denotes a negative charge state, from which polarity is derived — would probably save a round trip. Minor.

None of the three comments below block. Nothing here needs to change for this to merge; I've left it as a comment rather than an approval so you can decide on the follow-ups first.

Style points I deliberately checked and am **not** raising: `public void` (6 instance vs 4 static in this file, and uniformly instance in the newer siblings), fully-qualified `NUnit.Framework.Assert.That` (line 6 aliases `Assert` to `ClassicAssert`, and the file is 42/42 fully-qualified), and plain `[Test]`. All three match the file's own conventions.

acesnik:
🤖 Pushed `0a7b421d` and **edited the description** — flagging that here rather than letting it change silently.

Two sections changed:

- **"Not fixed here" → "Also fixed here".** The unsigned case is fixed, not deferred. Your review was right that it undersells the bug: `CHARGE=12` yields charge `1` *silently*, and only the single-digit case throws. I reproduced all seven forms before changing anything.
- **The "Not verified" line is gone.** The tests are now executed rather than replayed — see below.

What changed in the code, beyond the two you raised:

- The two unsigned failure modes get **separate fixtures**, which mattered more than expected. In one file the single-digit `FormatException` aborts the load and masks the multi-digit case, so the mutation test fails on the exception instead of on the value. Split, each fails on its own claim: `expected 12, was 1` versus the throw.
- `EliminateZeroIntensityPeaksFromMgfOnFileLoad` had the same static-cache shape as `TestDynamicMgf` — its "dynamic" assertions were re-reading the cache. Fixed the same way.
- The sign test now also asserts scan count and distinct retention times, since as you noted its two original assertions could not disagree.

On `CHARGE=2+ and 3+`: I checked, and it is spec-legal — it appears verbatim in the [Matrix Science docs](https://www.matrixscience.com/help/data_file_help.html), permitted both in the header and locally inside a block. A header-level one is harmless here (the static loader skips everything until `BEGIN IONS`), but a local one reaches the parser and fails the whole file. Still deferred, because it needs a decision rather than a fix: the line means the charge is *ambiguous*, so taking the first candidate silently discards the other, and treating it as unknown — letting deconvolution decide — may be more faithful.

**The equivalent dead-test defect is not MGF-specific.** `TestDynamicMzml`, `TestDynamicMzmlWithPeakFiltering` and `TestDynamicRaw` all do the same thing, so mzML and Thermo dynamic reading had no coverage either — #1176. No reader bug fell out of it; the paths were correct, just unverified.

Two things I checked and found clean, so nothing needs filing: the `char`-vs-`string` `.Equals` mistake appears exactly once across mzLib, MetaMorpheus *and* Spritz (the line here), and none of the other `Substring(0, Length - 1)` sites are the same bug — `DecoyProteinGenerator`'s are deliberate variant construction and `PrideArchiveClient`'s is already conditional.

Verification: 16 `TestMgf` tests run off Windows via a throwaway net10.0 harness compiling the real test file, each new test confirmed RED against the unfixed parser, and all of `build` / `integration` / `external-service-tests` green here.

nbollis (approved):

PR #1165, opened 2026-08-15
Add an MGF writer, and let MS level round-trip through the format

### LINKED ISSUE #211: Add MGF Writer

acesnik:
🤖 Closes #211, open since July 2017.

#1164 has merged, and this branch has been rebased onto it — the earlier stacking note no longer applies.

## What this adds

`MsDataFile.ExportAsMgf(path)`, delegating to a static `MgfMethods.WriteMgf`. That mirrors how mzML writing is already arranged — `ExportAsMzML` over `MzmlMethods.CreateAndWriteMyMzmlWithCalibratedSpectra` — rather than introducing a new shape. There is no spectra-writer interface to implement; `ResultFile.WriteResults` covers search-result files, not spectra.

Fields written per block, following the Matrix Science spec (no whitespace around `=`, parameters ahead of peaks, invariant culture):

| Line | Source |
|---|---|
| `TITLE` | `basename.scan.scan.charge File:"…", NativeID:"…"` |
| `MSLEVEL` | `MsnOrder` |
| `PEPMASS` | `SelectedIonMonoisotopicGuessMz ?? SelectedIonMZ`, plus `SelectedIonIntensity`; omitted for MS1 |
| `CHARGE` | magnitude from `SelectedIonChargeStateGuess`, sign from `Polarity`; omitted for MS1 |
| `SCANS` | `OneBasedScanNumber`, preserved rather than renumbered |
| `RTINSECONDS` | `RetentionTime * 60`; omitted when RT is `NaN` |

`PEPMASS` prefers the monoisotopic guess so an exported file carries a deconvoluted precursor where one is available.

## Reader change

The reader hardcoded `msnOrder: 2`, so any MS1 block the writer emitted would have been read back as MS2. It now reads `MSLEVEL` when present, and otherwise infers the level from whether the block has a precursor. `MSLEVEL` is not in the core Mascot spec; it is what SIRIUS, pyteomics and MsBackendMgf use.

MGF has no polarity field — polarity is only inferrable from the sign of `CHARGE`. Any block without a `CHARGE` line therefore reads back as positive, on both the static and the dynamic path. That covers MS1 scans and, equally, an MS2 whose charge state guess is null, since `CHARGE` is written only where a guess exists.

An earlier revision inherited polarity from the preceding block on the static path. Review caught that the dynamic reader cannot do the same — it seeks to a byte offset and cannot see what came before — so the two readers disagreed on the same scan. The inheritance was removed rather than replicated, which also restores exactly what master did before this branch. `MgfStaticAndDynamicReadsAgreeOnPolarity` pins the two paths together.

## What was verified, and how

`Test.csproj` is `net*.0-windows`, so the suite does not run on macOS. I built a throwaway net10.0 harness inside the repo tree against `Readers.csproj` and ran 43 checks; all passed. The harness was deleted before committing.

- Round trip of an MS1 + MS2 file: MS level, scan numbers, retention times, precursor m/z, charge, precursor intensity and both peak arrays all match. MS1 returns with `MsnOrder == 1` and a null precursor.
- Spec conformance with `CurrentCulture` set to `de-DE`: `PEPMASS=350.5`, not `350,5`. Also `CHARGE=2-` for a negative scan, and every parameter ahead of the first peak.
- Existing fixtures: `tester.mgf` still produces the nine values `TestLoadMgf` asserts; `withIntensity`, `withZeros`, `Tab_separated_peak_list` and `SmallCalibratibleYeast` (118 scans) all still read as MS2 with precursors; `withEmptySpectra` still skips empty blocks.

Not verified: the NUnit tests have not themselves been executed, only their assertions replayed in the harness. Windows CI is the first real run.

## Out of scope

- MetaMorpheus wiring — needs an mzLib release and a `PackageReference` bump.
- The two existing write paths that should reach this writer but do not. `MsDataFileToResultFileAdapter.WriteResults` appends `.mzML` and writes mzML regardless of type, so requesting an MGF through it yields `out.mgf.mzML` containing mzML; `SpectralAveraging/Algorithms/AveragedSpectraWriter.cs` throws on any output type it does not handle. Routing either through `ExportAsMgf` changes the behaviour of an existing path for every caller, so both are follow-up work rather than part of a writer PR.
- Fragment charge as a third peak column. The spec allows `<mz> <intensity> [charge]` and `ChargeArray` could supply it, but the reader would not restore it.
- Header parameters (`CLE`, `TOL`, `MODS`) — Mascot search settings, not spectral data.
- The two `CHARGE` parsing cases noted in #1164, `CHARGE=2` unsigned and `CHARGE=2+ and 3+`, both still throw.


nbollis (dismissed):

zhuoxinshi (dismissed):

RayMSMS (dismissed):

mzLib/Readers/Mgf/Mgf.cs
```diff
@@ -242,8 +259,24 @@ public static MsDataFile LoadAllStaticData(string filePath, FilteringParams filt
...
 
-            return new MsDataScan(spectrum, scanNumber, 2, true,
-                charge > 0 ? Polarity.Positive : Polarity.Negative,
+            // MSLEVEL when the writer supplied one; otherwise a block with a precursor is MS2 and a
+            // block without one is MS1. Files predating MSLEVEL all carry PEPMASS, so they read as before.
+            int msnOrder = msLevel ?? (sawPrecursorMz ? 2 : 1);
+
+            // MGF has no polarity field, so it can only be inferred from the sign of CHARGE. Blocks
+            // without one inherit the last polarity seen in the file.
+            Polarity polarity = sawCharge
```
trishorts: The polarity-inheritance mechanism is only wired into the static path, so this PR makes the static
and dynamic MGF reads disagree.

`LoadAllStaticData` threads `lastPolarity` through (:44, :55). But the dynamic reader's call site is
not in this diff:

```csharp
// Mgf.GetOneBasedScanFromDynamicConnection, unchanged
return Mgf.GetNextMsDataOneBasedScanFromConnection(_streamReader, new HashSet<int>(), filterParams, scanNumber);
```

Four arguments, so `fallbackPolarity` takes its default of `Polarity.Positive` on every dynamic read.
For a negative-mode file, `LoadAllStaticData` returns an MS1 with `Polarity.Negative` (inherited from
the preceding MS2), and `GetOneBasedScanFromDynamicConnection` returns the same scan as
`Polarity.Positive`. Same file, same scan number, two different answers depending on which reader you
used.

That's worth flagging particularly because it's the exact failure mode #1176 just built regression
coverage for on the mzML and Thermo readers — and the MGF dynamic path has no static-vs-dynamic test
either, so nothing here would catch it.

A dynamic read genuinely can't know the preceding block's polarity without scanning backwards, so
"inherit the last polarity seen" may just not be implementable on that path. If so, the honest fix is
probably to make the two agree by construction rather than to have the static path be cleverer — but
either way the current state is a silent divergence.

**Two related gaps while you're in here:**

The description says the round-trip caveat is about MS1 blocks, but an **MS2 with no charge guess**
loses polarity the same way. `CHARGE` is written only under
`if (scan.SelectedIonChargeStateGuess.HasValue)`, and the reader derives polarity solely from that
line — so a negative-mode MS2 with a null charge guess emits no `CHARGE` and comes back positive,
anywhere in the file rather than only before the first MS2.

And **nothing tests any of this.** The only `Polarity.Negative` in the new tests is the write-side
scan in `MgfWriterEmitsSpecConformantText`, which asserts `CHARGE=2-` in the emitted text. No test
reads a negative-mode file back, so the `lastPolarity` threading and the new `fallbackPolarity`
parameter have zero coverage for the case they exist to serve. A round trip of
MS1 -> MS2(negative) -> MS1 asserting the polarity of both MS1s would pin the intended behaviour and
would currently fail on the dynamic path.

mzLib/Readers/MsDataFileExtensions.cs
```diff
@@ -44,6 +44,16 @@ public static void ExportAsMzML(this MsDataFile file, string destinationPath, bo
             MzmlMethods.CreateAndWriteMyMzmlWithCalibratedSpectra(file, destinationPath, writeIndexed);
         }
 
+        /// <summary>
+        /// Writes the file as Mascot Generic Format. MGF carries a single precursor m/z per spectrum, so
+        /// the deconvoluted monoisotopic guess is preferred over the selected ion m/z where available.
+        /// Scans with no peaks are skipped. See <see cref="MgfMethods"/> for the field mapping.
+        /// </summary>
+        public static void ExportAsMgf(this MsDataFile file, string destinationPath)
```
trishorts, line 52: Now that this exists, the result-file write path is inconsistent with it rather than merely missing.

`SupportedFileType.Mgf` maps to `MsDataFileToResultFileAdapter`
(`Readers/Util/SupportedFileTypes.cs:247`), whose write method is:

```csharp
public void WriteResults(string outputPath)
{
    if (!outputPath.EndsWith(".mzml", StringComparison.InvariantCultureIgnoreCase))
        outputPath += SupportedFileType.MzML.GetFileExtension();
    _dataFile.ExportAsMzML(outputPath, true);
}
```

So asking that adapter to write an MGF produces `out.mgf.mzML`, containing mzML. Before this PR that
was "unimplemented"; after it, it's a branch that silently does the wrong thing while the capability
sits right here.

`SpectralAveraging/Algorithms/AveragedSpectraWriter.cs:32-39` is the second such site — it switches on
`parameters.OutputType` with `default: throw new MzLibException("Output averaged scans type not implemented")`,
which is the case that exists for exactly this.

I wouldn't hold the PR for either, but they're the two places a user would actually reach the new
writer from, so they're worth a follow-up issue or an explicit "out of scope" line in the description
alongside the MetaMorpheus wiring you already listed.

mzLib/Readers/Mgf/MgfMethods.cs
```diff
@@ -0,0 +1,114 @@
...
+            if (myMsDataFile.Scans == null)
+            {
+                myMsDataFile.LoadAllStaticData();
+            }
+
+            string sourceName = string.IsNullOrEmpty(myMsDataFile.FilePath)
+                ? Path.GetFileNameWithoutExtension(outputFile)
+                : Path.GetFileNameWithoutExtension(myMsDataFile.FilePath);
+
+            using StreamWriter output = new StreamWriter(outputFile);
```
trishorts: Skipping empty scans is right for the format, but it means the writer can emit a file mzLib's own
reader throws on.

If every scan is empty — or the file has none — no `BEGIN IONS` block is written, and reading that
back hits:

```csharp
var orderedScans = scans.OrderBy(x => x.OneBasedScanNumber).ToArray();
var indexedScans = new MsDataScan[orderedScans[^1].OneBasedScanNumber];
```

`orderedScans[^1]` on an empty array is an `IndexOutOfRangeException` with no message. The reader's
inability to handle an empty MGF is pre-existing, but until now mzLib had no way to *produce* one.

Related: skipping empties makes the surviving `SCANS=` values sparse, and the reader deliberately
allocates `new MsDataScan[maxScanNumber]` with nulls in the gaps (the comment above it says so). So
exporting a mostly-empty file with a high top scan number allocates by largest scan number rather
than by scans kept, and `GetOneBasedScan` returns null in the holes.

Both are edge cases; a guard here (or a documented "writes at least one block") would be cheaper than
diagnosing it later.

trishorts (commented):
Closing a 2017 issue with a writer that mirrors `ExportAsMzML` rather than inventing a new shape is the right call — I checked, and there is no spectra-writer interface to implement, so an extension method beside `ExportAsMzML` is the matching pattern. Culture handling, the streaming `StreamWriter`, and preserving `SCANS` rather than renumbering all match what the other text writers in `Readers` do.

One blocker inline, plus two smaller things.

The blocker is polarity: the inheritance mechanism is wired into `LoadAllStaticData` but not into the dynamic path, so the two readers now return different `Polarity` values for the same scan — and no test covers negative mode on the read side.

Worth noting the review engine flagged this one and could not self-verify it, so I checked it by hand: the dynamic call site is simply absent from the diff, which is why it still takes the parameter default.

acesnik:
🤖 Blocker fixed in `849161ff`, and thanks for hand-checking it — the call site really was just absent from the diff.

I removed the inheritance rather than wiring it into the dynamic path, which I think is the "agree by construction" you were pointing at. Details in that thread, including the alternative I rejected in case you'd rather have it.

Two smaller things also addressed: the writer now refuses to produce a blockless file, and the MS2-with-null-charge case is documented as behaving exactly like an MS1. The result-file adapter I've left alone per your note — follow-up issue rather than scope creep here.

I also **edited the description**: the polarity paragraph described the inheritance that is now gone, the "Stacked on #1164, merge that one first" line was stale since #1164 merged and this branch was rebased onto it, and Out of scope now names both write-path sites from your second comment.

One process note: the branch picked up two merges of master while I was working on this, so my commit sits on top of those rather than replacing them.

mzLib/Readers/Mgf/Mgf.cs
acesnik: 🤖 Confirmed, and removed rather than replicated.

You're right that a dynamic read cannot know the preceding block's polarity — it seeks straight to a byte offset. So the inheritance was only ever implementable on one of two paths, which is the divergence. Both paths now derive polarity from `CHARGE` alone.

The reason that is safe rather than a behaviour change: master already did this. `int charge = 2; //default when unknown` plus `charge > 0 ? Positive : Negative` means a `CHARGE`-less block read as **positive** on master, on both paths. Removing the inheritance restores exactly that, so nothing regresses for blocks that carry a `CHARGE`.

You are also right that the description understated the round-trip caveat: an MS2 with a null charge guess writes no `CHARGE` and loses polarity exactly like an MS1, anywhere in the file. That is now documented on `MgfMethods` as a limit of the format, and it is asserted.

`MgfStaticAndDynamicReadsAgreeOnPolarity` writes MS1 → negative MS2 with a charge guess → negative MS2 *without* one → trailing MS1, then compares the static and dynamic reads scan by scan and pins which scans can keep their sign. Against the previous commit it fails on scan 3: `Expected: Negative, But was: Positive` — the divergence you described.

**One alternative I considered and rejected**, in case you would prefer it: resolve a single file-level polarity from the first `CHARGE` in the file and use it as the fallback on both paths. `BuildIndex` already walks the whole file so the dynamic half is nearly free, and it would make a negative-mode file's MS1s read as negative — more *correct*, where what I have done is merely *consistent*. It needs a matching pre-pass on the static side, and it still guesses on a mixed-polarity file. Happy to switch if you think the accuracy is worth it.

mzLib/Readers/MsDataFileExtensions.cs
acesnik: 🤖 Confirmed both sites. `MsDataFileToResultFileAdapter.WriteResults` does append `.mzML` and write mzML, so asking it for an MGF produces `out.mgf.mzML` — and you are right that this PR changes it from "unimplemented" to "silently wrong while the capability sits next door".

Taking your advice and leaving both alone here: routing the adapter through `ExportAsMgf` changes the behaviour of an existing path for every caller, and `AveragedSpectraWriter`'s `default: throw` is a third decision about what averaged output should default to. Neither belongs in a writer PR.

Both are now named explicitly in the description's Out of scope section, and I am filing a follow-up issue so they do not get lost.

mzLib/Readers/Mgf/MgfMethods.cs
acesnik: 🤖 Good catch, fixed. `WriteMgf` now collects the scans that will produce blocks first and throws `MzLibException` if there are none:

```
Cannot write an mgf with no spectra: every scan was empty, and mgf requires at least
one fragment peak per block.
```

Checked *before* the `StreamWriter` is opened, so a rejected export leaves nothing on disk — `MgfWriterRefusesToWriteAFileWithNoBlocks` asserts that too, since a half-written empty file would be its own trap.

On the sparse-`SCANS` half: that one I have left. Preserving scan numbers rather than renumbering is deliberate, and the reader's allocate-by-max-scan-number with nulls in the gaps is pre-existing and documented in a comment there. Fixing it means either renumbering on write, which loses the mapping back to the source file and seems worse, or changing the reader's indexing strategy. Worth its own issue if it bites someone; I did not want to bundle a reader change into this.

nbollis (approved):

zhuoxinshi (approved):

Issue #245, opened 2017-10-02
Single spectrum deconvolution does not pick correct peak if within tolerance


stefanks:
In case there are two peaks within the deconvolution tolerance, the single spectrum deconvolution picks the nearest one to the theoretical, instead of the one that matches the expected intensity the closest.

PR #1166, opened 2026-08-16
Consider all in-tolerance peaks when matching isotopes, not only the nearest

acesnik:
🤖 Draft. This addresses the isotope-matching half of #245. See **Scope** for what is deliberately left unchanged.

`FindIsotopicEnvelope` looked up the single nearest peak to each sought isotope and ended the isotope walk if that peak failed the tolerance, intensity-ratio or already-claimed checks. A nearer peak that failed those checks masked a usable isotope sitting behind it, and because the walk `break`s on the first miss it truncated the entire envelope rather than swapping one peak.

The window is now scanned for the nearest peak that passes all the existing checks — but only when the nearest peak itself is unusable. No predicate changed.

### Accuracy

Verified on a constructed spectrum (z=2, 20 ppm, exact averagine envelope; real isotope +12 ppm with the predicted intensity, interferer −4 ppm at 20× the predicted intensity):

| | monoisotopic mass | peaks |
|---|---|---|
| truth | 1600.0000 | 6 |
| before | 1601.0017 | 4 |
| after | 1600.0000 | 6 |

On real data there is no ground truth, so the table below reports deltas only. **These are not validated as improvements** — envelope count moves in opposite directions on the two files.

| file | tolerance | envelopes | matched isotope peaks |
|---|---|---|---|
| TDYeastFractionMS1 (455 MS1) | 4 ppm | 228,583 → 228,583 | 617,823 → 617,823 |
| TDYeastFractionMS1 | 20 ppm | 325,487 → 325,535 | 960,326 → 963,270 (+0.31%) |
| FXN11…scans716_718 | 4 ppm | 355 → 355 | 1,294 → 1,297 |
| FXN11…scans716_718 | 20 ppm | 487 → 474 (−2.7%) | 1,791 → 1,812 (+1.2%) |

**At 4 ppm this is a measured no-op on real data** — byte-identical output. The behaviour only diverges at wide tolerance, because the condition itself is tolerance-dependent (fraction of isotope lookups with >1 peak in tolerance):

| tolerance | TDYeastFractionMS1 | FXN11…716_718 |
|---|---|---|
| 4 ppm | 0.002% (148 / 8.08M) | 0.034% (11 / 32,195) |
| 10 ppm | 0.140% (20,469 / 14.6M) | 1.57% (631 / 40,214) |
| 20 ppm | 5.53% (1.31M / 23.7M) | 6.68% (3,365 / 50,405) |
| 50 ppm | — | 19.8% (17,999 / 90,827) |

### Why not select by closest intensity match

The issue proposes taking the peak whose intensity best matches the prediction. That rule and a combined mass+intensity rule were both implemented and measured against this one. TDYeast, 20 ppm; "surrendered" is the extra mass error accepted by overruling the nearest peak, "poor trades" are overrules giving up >5 ppm for <2× intensity gain:

| rule | envelopes | peaks | overrules | median ppm surrendered | poor trades |
|---|---|---|---|---|---|
| previous behaviour | 325,487 | 960,326 | — | — | — |
| closest intensity match | 325,453 | 962,570 | 648,430 | 8.18 | 409,534 (63%) |
| **nearest passing peak (this PR)** | **325,535** | **963,270** | 0 | — | — |
| combined mass+intensity | 325,561 | 963,290 | 159,060 | 2.77 | 24,814 (16%) |

The intensity-ratio check already screens out implausible abundances, so among peaks that survive it, mass error carries more information. The combined rule scores marginally better than this one but introduces a weighting with no validation behind it.

Note the 0 in this PR's row is definitional, not a result: this rule never overrules the nearest passing peak, so the metric cannot fire. Its own error mode is the mirror image — declining a ≥2× better intensity match that sits farther out:

| tolerance | ≥2× better match declined | median extra ppm to reach it |
|---|---|---|
| 4 ppm | 0 | — |
| 10 ppm | 288 of 20,469 ambiguous (1.4%) | 1.35 |
| 20 ppm | 51,462 of 1.31M ambiguous (3.9%) | 7.97 |

### Speed

455 top-down yeast MS1 scans, minimum of 4–6 rounds with the variants interleaved inside one process (cross-process comparison on this machine was dominated by noise; spreads ran up to 2× the minimum):

| | 4 ppm | 20 ppm |
|---|---|---|
| **this PR** | **+1.0%** | **+2.4%** |
| helper called on every lookup | +5.4% | +5.6% |
| helper with an early-return fast path | +4.7% | +5.8% |

The last two explain the structure: the cost was calling an 8-argument out-of-line helper 8–24M times, not the window arithmetic. Keeping the nearest-peak test inline and calling the helper only on the rare miss recovers most of it. Part of the remaining 2.4% at 20 ppm is real extra work, since 5.5% of lookups take the slow path there.

### Scope — what this does not do

- **Does not implement the criterion in the issue body** (closest intensity match); measured above.
- **Does not fix the same pattern in `FindChargeStateOfMass`**, so the issue *title* is only partly addressed. Measured at 0–35 ambiguous lookups out of up to 5M, and no averagine intensity prediction exists for an adjacent charge state's abundance.
- **Does not change the `break`.** One unmatched isotope still ends the walk; this only reduces how often a spurious miss triggers it.
- The `[Obsolete]` copy in `MzSpectrum.Deconvolute` is unchanged.

### Testing and what could not be run

4 new cases in `TestDeconvolutionPeakChoiceWithinTolerance`; 3 fail against unfixed code, the 4th pins that the single-candidate path is unchanged. No existing test was modified.

`Test` is `net10.0-windows` and cannot run on macOS, so only the deconvolution test files were executed here, via a throwaway `net10.0` harness. The full suite has not been run — CI covers it. Benchmarks are one machine, two files.

mzLib ships to MetaMorpheus by NuGet; downstream search output has not been measured.

### DIFF
--- a/mzLib/MassSpectrometry/Deconvolution/Algorithms/ClassicDeconvolutionAlgorithm.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Algorithms/ClassicDeconvolutionAlgorithm.cs
@@ -188,26 +188,40 @@ namespace MassSpectrometry
             for (int indexToLookAt = 1; indexToLookAt < theoreticalIntensities.Length; indexToLookAt++) //cycle through all theoretical peaks in this envelope from most intense to least intense
             {
                 double theorMassThatTryingToFind = theoreticalMasses[indexToLookAt] + differenceBetweenTheorAndActualMass; //get the expected mass of the next most intense peak
-                int closestPeakToTheorMass = spectrum.GetClosestPeakIndex(theorMassThatTryingToFind.ToMz(chargeState)); //find the experimental peak for that mass
-                double closestPeakmz = spectrum.XArray[closestPeakToTheorMass];
-                double closestPeakIntensity = spectrum.YArray[closestPeakToTheorMass];
-                double closestPeakMass = closestPeakmz.ToMass(chargeState);
-                //if the peak is within the deconvolution tolerance, has the correct intensity, and hasn't already been observed
-                if (Math.Abs(closestPeakMass - theorMassThatTryingToFind) / theorMassThatTryingToFind * 1e6 <= deconvolutionTolerancePpm
-                    && Peak2satisfiesRatio(theoreticalIntensities[0], theoreticalIntensities[indexToLookAt], candidateForMostIntensePeakIntensity, closestPeakIntensity, intensityRatioLimit)
-                    && !listOfObservedPeaks.Contains((closestPeakmz, closestPeakIntensity)))
-                {
-                    //Found a match to an isotope peak for this charge state!
-                    listOfObservedPeaks.Add((closestPeakmz, closestPeakIntensity)); //add to observed list
-                    totalIntensity += closestPeakIntensity; //add intensity
-                    listOfRatios.Add(theoreticalIntensities[indexToLookAt] / closestPeakIntensity); //add ratio
-                    double monoisotopicMassFromThisPeak = monoisotopicMass + closestPeakMass - theorMassThatTryingToFind;
-                    monoisotopicMassPredictions.Add(monoisotopicMassFromThisPeak);
-                }
-                else
+                int matchedPeakIndex = spectrum.GetClosestPeakIndex(theorMassThatTryingToFind.ToMz(chargeState)); //find the experimental peak for that mass
+                double matchedPeakMz = spectrum.XArray[matchedPeakIndex];
+                double matchedPeakIntensity = spectrum.YArray[matchedPeakIndex];
+                double matchedPeakMass = matchedPeakMz.ToMass(chargeState);
+
+                //the nearest peak is out of tolerance, and it bounds every other peak's distance from the sought mass
+                if (Math.Abs(matchedPeakMass - theorMassThatTryingToFind) / theorMassThatTryingToFind * 1e6 > deconvolutionTolerancePpm)
                 {
                     break;
                 }
+
+                //the nearest peak is unusable, but issue #245: a usable isotope may sit just behind it, still in tolerance
+                if (!Peak2satisfiesRatio(theoreticalIntensities[0], theoreticalIntensities[indexToLookAt], candidateForMostIntensePeakIntensity, matchedPeakIntensity, intensityRatioLimit)
+                    || listOfObservedPeaks.Contains((matchedPeakMz, matchedPeakIntensity)))
+                {
+                    matchedPeakIndex = FindUsablePeakNearTheorMass(theorMassThatTryingToFind, chargeState, deconvolutionTolerancePpm,
+                        theoreticalIntensities[0], theoreticalIntensities[indexToLookAt], candidateForMostIntensePeakIntensity,
+                        intensityRatioLimit, listOfObservedPeaks);
+                    if (matchedPeakIndex < 0)
+                    {
+                        break;
+                    }
+
+                    matchedPeakMz = spectrum.XArray[matchedPeakIndex];
+                    matchedPeakIntensity = spectrum.YArray[matchedPeakIndex];
+                    matchedPeakMass = matchedPeakMz.ToMass(chargeState);
+                }
+
+                //Found a match to an isotope peak for this charge state!
+                listOfObservedPeaks.Add((matchedPeakMz, matchedPeakIntensity)); //add to observed list
+                totalIntensity += matchedPeakIntensity; //add intensity
+                listOfRatios.Add(theoreticalIntensities[indexToLookAt] / matchedPeakIntensity); //add ratio
+                double monoisotopicMassFromThisPeak = monoisotopicMass + matchedPeakMass - theorMassThatTryingToFind;
+                monoisotopicMassPredictions.Add(monoisotopicMassFromThisPeak);
             }
 
             return new IsotopicEnvelope(listOfObservedPeaks, monoisotopicMass, chargeState, totalIntensity, listOfRatios.StandardDeviation());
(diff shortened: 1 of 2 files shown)


mzLib/MassSpectrometry/Deconvolution/Algorithms/ClassicDeconvolutionAlgorithm.cs
```diff
@@ -294,6 +308,62 @@ private bool FindChargeStateOfMass(IsotopicEnvelope originalEnvelope, int zToInv
                 spectrum.XArray.GetClosestIndex(maxX, ArraySearchOption.Previous));
         }
 
+        /// <summary>
+        /// Nearest peak to <paramref name="theorMass"/> that is in tolerance, matches the predicted intensity and is
+        /// unclaimed, or -1 if there is none. Called only when the closest peak itself is unusable (issue #245).
+        /// </summary>
+        private int FindUsablePeakNearTheorMass(double theorMass, int chargeState, double deconvolutionTolerancePpm,
```
trishorts, line 315: `MzSpectrum.GetPeakIndicesWithinTolerance(double x, Tolerance tolerance)` already exists
(`MassSpectrometry/MzSpectra/MzSpectrum.cs:620`) and returns every peak index inside a tolerance
window, empty list on a miss. `IsoDecAlgorithm` is currently its only caller — Classic has never used
it, which is probably why it didn't surface.

Your two window-walking loops are that method, hand-rolled.

The catch is that it takes an m/z and a `Tolerance`, whereas the comment above explains you want the
window bounded in the mass domain so the ppm check isn't clipped by the proton offset — and Classic
carries a raw `double deconvolutionTolerancePpm` rather than a `PpmTolerance`, so reuse means
constructing one at the call site. That's a real reason not to reuse it, not a dodge.

So: not asking you to change it, just flagging that the helper exists so the choice is deliberate. If
the local version stays, one line saying why it isn't `GetPeakIndicesWithinTolerance` would save the
next person the same investigation.

mzLib/MassSpectrometry/Deconvolution/Algorithms/ClassicDeconvolutionAlgorithm.cs
```diff
@@ -294,6 +308,62 @@ private bool FindChargeStateOfMass(IsotopicEnvelope originalEnvelope, int zToInv
...
+            double peak1TheorIntensity, double peak2TheorIntensity, double peak1Intensity,
+            double intensityRatioLimit, List<(double, double)> alreadyObservedPeaks)
+        {
+            int nearestIndex = spectrum.GetClosestPeakIndex(theorMass.ToMz(chargeState));
```
trishorts, line 319: This recomputes something the caller already has. `FindIsotopicEnvelope` evaluates
`spectrum.GetClosestPeakIndex(theorMassThatTryingToFind.ToMz(chargeState))` into `matchedPeakIndex`,
then on the unusable-peak path calls this method, which runs the identical expression again.

That's an extra binary search per rejected isotope peak, and Classic is the default algorithm — it is
on MetaMorpheus's precursor and top-down deconvolution paths, `MassIndexingEngine` (which feeds
FlashLFQ), and `Deconvoluter.DeconvoluteWithDecoys`, which re-runs the whole thing on decoy
parameters. Passing `matchedPeakIndex` in as a parameter removes it entirely.

trishorts (commented):
Early input while it's a draft. Two small comments inline, plus one thing worth deciding before this leaves draft.

**Blast radius.** `ClassicDeconvolutionAlgorithm` is the default algorithm and the only one doing per-isotope nearest-peak matching in managed code — `IsoDecAlgorithm` matches inside native `process_spectrum`, `FromFile` reads pre-computed envelopes, and the template is a stub. But `MultipleDeconvolutionAlgorithm` fans out through `Deconvoluter.Deconvolute`, so it inherits this whenever Classic is one of its sub-parameters, and `DeconvoluteWithDecoys` re-runs Classic on decoy parameters — so decoy score distributions and therefore deconvolution FDR move too, not just target envelope counts. Downstream that reaches `MzSpectrum`, `MsDataScan.GetIsolatedMassesAndCharges` (used by `BrukerFileReader`), `MassIndexingEngine` into FlashLFQ, and `DeconvolutionScorer`/`DeconvolutionQValueCalculator`.

That's not an argument against the change — issue #245 is real, and the `break` in the old loop meant one bad nearest peak truncated the whole envelope rather than just skipping an isotope, which is a genuinely bad failure mode. It's an argument for showing before/after envelope counts and score distributions on a real file, since "more peaks matched" is not automatically "better" once it feeds an FDR calculation.

Also worth knowing: #1045 and #1042 both touch `Deconvoluter.cs`, so there's some ordering to think about.

Note the review tooling hit a session limit on this PR, so this is from my own reading of the diff rather than a full pass — I may re-run it later and follow up.

mzLib/Test/Deconvolution/TestDeconvolutionPeakChoiceWithinTolerance.cs
```diff
@@ -0,0 +1,135 @@
...
+            Assert.That(targetIndex, Is.GreaterThanOrEqualTo(0));
+            intensities[targetIndex] = built.ExpectedIntensity * 20;
+
+            var envelopes = Deconvoluter
+                .Deconvolute(new MzSpectrum(mz, intensities, false), parameters, WholeSpectrum)
+                .ToList();
+
+            Assert.That(envelopes.Any(e => Math.Abs(e.MonoisotopicMass - 1600.0) < 0.01
+                                           && e.Peaks.Count == 6), Is.False,
+                "the full ladder must not be recovered when the only in-tolerance peak has the wrong intensity");
```
trishorts, line 132: This is the only test pinning "the single-candidate path is unchanged", and its entire assertion is a
negative: `Assert.That(envelopes.Any(...), Is.False)`. That is satisfied by *any* result lacking a
mono≈1600 / 6-peak envelope — including an empty one. If a future change to
`FindUsablePeakNearTheorMass` or the restructured loop made Classic return nothing at all for this
spectrum, this test stays green and the invariant it exists to protect is gone.

Your own sibling in this folder already guards exactly this, twice
(`TestDeconvolutionDuplicateMz.cs:65` and `:87`):

```csharp
Assert.That(clean, Is.Not.Empty, "the ladder itself must deconvolute, or the test proves nothing");
```

Same idea here: assert `envelopes` is non-empty, and ideally pin what the truncated walk *does*
produce (charge, mono, peak count), keeping the `Is.False` as the secondary check rather than the
only one.

mzLib/Test/Deconvolution/TestDeconvolutionPeakChoiceWithinTolerance.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                peaks.Add((targetMz * (1 + interfererPpmError / 1e6), expectedIntensity * interfererFoldIntensity));
+            }
+
+            var ordered = peaks.OrderBy(p => p.Mz).ToList();
+            return (new MzSpectrum(ordered.Select(p => p.Mz).ToArray(), ordered.Select(p => p.Intensity).ToArray(), false),
+                targetMz, expectedIntensity);
+        }
+
+        private static string[] Describe(IEnumerable<IsotopicEnvelope> envelopes) => envelopes
```
trishorts, line 60: `Describe` here emits `z=`, `mono=` (F4) and `peaks=`. The sibling helper at
`TestDeconvolutionDuplicateMz.cs:46` also emits `intensity={e.TotalIntensity:F2}` — and that omission
costs you the assertion that matters most for half the new cases.

With intensity dropped, the `Is.EqualTo(expected)` comparison can't see *which* of the two in-window
peaks was chosen: the envelope has 6 peaks and z=2 either way, and `SetMedianMonoisotopicMass` takes
the median of six predictions of which five are identical, so one displaced isotope cannot move the
median at 4 dp. Peak identity is asserted only in
`TheEnvelopeContainsTheRealIsotopeNotTheNearerInterferer`, which runs only
`interfererFoldIntensity: 20`. So the `TestCase(1.0 / 20.0)` branch — the too-weak interferer, which
is precisely the half where "nearest but unusable" and "usable but further" diverge — has no
assertion anywhere that the real isotope rather than the interferer ended up in the envelope.

Adding `TotalIntensity` to `Describe` fixes it on its own, since the real peak and a 20x/0.05x
interferer differ enormously in intensity.

trishorts (commented):
Follow-up: the review tooling hit a session limit on my first pass, so I re-ran it. Two test-quality points it surfaced that I'd missed, both about assertions that can't fail. The code side of my earlier comments is unchanged.

PR #1169, opened 2026-08-16
Read mzIdentML 1.3.0

acesnik:
🤖 [mzIdentML 1.3.0](https://www.psidev.info/mzidentml) was released in June 2024. `MzidIdentifications` recognises 1.1.0, 1.1.1 and 1.2.0, so a 1.3.0 file fails to deserialize against any of them and every accessor then throws `NullReferenceException`.

## The generated file is derived, not regenerated

`mzIdentML1_3_0.cs` is a copy of `mzIdentML1_2_0.cs` with three textual substitutions, plus one hand-added property. That is deliberate on two grounds: `xsd.exe` is Windows-only, and the schema diff is five lines.

This also mirrors how the schema itself was produced. The first commit on `mzIdentML1.3.0.xsd` (2023-04-28) is `copy 1.2.0 xsd, reanming to 1.3.0`, followed by `add cvParams after cvList` and the two `use="required"` changes. The schema has been unchanged since 2024-05-01.

| Schema change | Effect on generated C# |
|---|---|
| target namespace → `http://psidev.info/psi/pi/mzIdentML/1.3` | 67 `Namespace=` strings |
| `MzIdentMLType` gains `cvParam` (`CVParamType`, 0..unbounded) between `cvList` and `AnalysisSoftwareList` | one new property |
| `InputSpectraType.spectraData_ref` → `use="required"` | none — a required string attribute generates the same property |
| `SearchDatabaseRefType.searchDatabase_ref` → `use="required"` | none, same reason |
| version pattern `(1\.2\.\d+)` → `(1\.3\.\d+)` | none — the facet is not surfaced |

The substitutions were `mzIdentML120.Generated` → `130` (1 occurrence), `MzIdentMLType120` → `130` (2), and the namespace URI (67). Zero `120` references remain, so the derivation is reproducible and reviewable rather than a 4,647-line blob from a differently-versioned generator.

The new property sits between `cvList` and `AnalysisSoftwareList` because these classes carry no `Order=` attributes, so `XmlSerializer` uses declaration order.

## Wiring

Every accessor dispatches by trying each version and catching, ending at 1.2.0 with no further catch. Each of the 24 cascades gains a 1.3.0 level at the innermost point, and the constructor gains a fourth deserialization attempt. That is +635 lines in `MzidIdentifications.cs`, and is why the diff is large for what it does.

## Test fixture

`noncovalently_assoc_1_3_0.mzid` is `examples/1_3examples/crosslinking/noncovalently_assoc_1_3_0_draft.mzid` from the HUPO-PSI repository, renamed on the way in. The upstream `_draft` suffix dates from before the June 2024 release and was never removed; the file validates cleanly against the released schema:

```
$ xmllint --noout --schema mzIdentML1.3.0.xsd noncovalently_assoc_1_3_0.mzid
noncovalently_assoc_1_3_0.mzid validates
```

## Verified

Against a harness referencing `MzIdentML.csproj` — `Test.csproj` is `net*.0-windows` and does not run on macOS.

- The 1.3.0 fixture now reads: `Count` 1, sequence `AYALMTDIHWDDCFCR`, charge 3, experimental m/z 1392.897440641436, calculated 1392.567094980103, accession `P15640`, not decoy.
- `SmallCalibratible_Yeast.mzID` is unchanged: `Count` 65 and the same values the existing tests assert.
- `TestParentTolerance.mzid` throws `NullReferenceException` on `Count` and `FragmentTolerance`. This is pre-existing — I checked out master's `MzidIdentifications.cs`, rebuilt and got identical exceptions. That fixture contains no `SpectrumIdentificationResult` elements.

Not verified: the NUnit tests have not been run, only their assertions replayed in the harness. Windows CI is the first real run.

## Scope

1.3.0's headline feature is crosslinking, and `MzidIdentifications` has no crosslinking accessors. This makes 1.3.0 files readable through the existing surface; it does not expose anything new.

## Worth a separate look

The version cascade is now four levels of `try`/`catch` using `NullReferenceException` as the dispatch signal, duplicated across 24 accessors. It works, and matching it kept this diff mechanical, but a 1.4.0 will be worse again. A small dispatch abstraction would collapse it — better as its own change than smuggled in here.


mzLib/MzIdentML/MzidIdentifications.cs
```diff
@@ -58,11 +59,23 @@ public MzidIdentifications(string mzidFile)
...
+                            dd120 = _indexedSerializer.Deserialize(stream) as mzIdentML120.Generated.MzIdentMLType120;
+                        }
+                    }
+                    catch
+                    {
+                        using (Stream stream = new FileStream(mzidFile, FileMode.Open, FileAccess.Read, FileShare.Read))
+                        {
+                            XmlSerializer _indexedSerializer = new XmlSerializer(typeof(mzIdentML130.Generated.MzIdentMLType130));
+                            // Read the XML file into the variable
+                            dd130 = _indexedSerializer.Deserialize(stream) as mzIdentML130.Generated.MzIdentMLType130;
```
trishorts, line 77: This adds a whole generated namespace, and the existing reflection sweep won't see it.

`Test/MzIdentML/MzIdentMlReflectionCoverageTests.cs` enumerates the generated namespaces by hand:

```csharp
[TestCase("mzIdentML110.Generated")]
[TestCase("mzIdentML111.Generated")]
[TestCase("mzIdentML120.Generated")]
public void EveryGeneratedType_PropertyGetSet_RoundTrips(string generatedNamespace)
```

`mzIdentML130.Generated` isn't there, so every type in these 4,647 new lines is excluded from the
one test that covers generated code at all. That file isn't in your diff, which is exactly why it's
easy to miss — the sweep is opt-in per namespace, not `assembly.GetTypes()` over everything.

Adding one line fixes it:

```csharp
[TestCase("mzIdentML130.Generated")]
```

Conveniently that also closes the other gap I'd have raised separately: `MzIdentMLType130.cvParam`
is the single hand-written property in an otherwise mechanically-derived file — the one line most
likely to be wrong and the only one no test asserts. The reflection sweep round-trips every property
getter/setter, so the same `[TestCase]` covers it.

(The sweep asserts `"No instantiable generated types found in {generatedNamespace}"`, so if the
namespace string is ever wrong the test says so rather than passing empty.)

mzLib/MzIdentML/MzidIdentifications.cs
```diff
@@ -301,9 +387,18 @@ public double QValue(int sirIndex, int siiIndex)
...
-                    return cvParam == null ? -1 : Convert.ToDouble(cvParam.value, CultureInfo.InvariantCulture);
+                    try
+                    {
+                        var cvParam = dd120.DataCollection.AnalysisData.SpectrumIdentificationList[0].SpectrumIdentificationResult[sirIndex].SpectrumIdentificationItem[siiIndex].cvParam.
+                            Where(cv => cv.accession == "MS:1002354").FirstOrDefault();
+                        return cvParam == null ? -1 : Convert.ToDouble(cvParam.value, CultureInfo.InvariantCulture);
+                    }
+                    catch
+                    {
+                        var cvParam = dd130.DataCollection.AnalysisData.SpectrumIdentificationList[0].SpectrumIdentificationResult[sirIndex].SpectrumIdentificationItem[siiIndex].cvParam.
```
trishorts, line 398: Before adding a fourth level, it's worth knowing what the cascade already costs — because this makes
it 50% worse and the shape is now load-bearing for a third release in a row.

20 of the 23 accessors are indexed per match — `ProteinAccession(sirIndex, siiIndex)`,
`CalculatedMassToCharge`, `QValue`, `ModificationMass(sirIndex, siiIndex, i)` — so they're called in
loops over every identification. Each one dereferences `dd110` first (`ProteinAccession` at :898 on
master). For any file newer than 1.1.0 that's a `NullReferenceException` thrown and swallowed by a
bare `catch`, then the same for `dd111`, then `dd120`. A 1.3.0 file pays **three thrown exceptions per
accessor call**, up from two today.

On the 18,023-match file in your description, a loop reading ~10 fields per match is on the order of
half a million thrown-and-caught exceptions, purely to work out which of four fields is non-null.
That's seconds of wall clock doing nothing. (The engine also flagged `ids.Count` being re-evaluated
in a loop condition in the new tests — same root cause, since `Count` is itself a cascade.)

The constructor already knows which version it parsed. A `private readonly` discriminator set right
here — or even just `if (dd110 != null) ... else if (dd111 != null) ...` in the accessors — turns each
cascade into a flat branch with no exceptions, and is behaviour-preserving for well-formed files.

It also fixes a correctness problem the fourth level inherits and relocates. Because the catches are
bare and untyped, a *genuine* NRE inside a correctly-matched version is indistinguishable from "wrong
version": a 1.1.1 file with a missing `ParentTolerance` throws at :90, is swallowed at :95, and dies
in the 1.2.0 arm — the reported exception names the wrong version and the original isn't chained.
After this PR that fallthrough sink becomes `dd130`, so every existing stack trace or bug report
mentioning `dd120` changes meaning. Switching to typed catches wouldn't help, since the NRE you must
swallow and the NRE you mustn't are the same type; only a null check on the field separates them.

You called this out yourself under "Worth a separate look" and I agree it's a separate change — I'd
just rather see it land before or alongside the fourth level than after, since each version doubles
down. Not blocking; an issue would be enough so it doesn't get lost when 1.4.0 arrives.

trishorts (commented):
The derivation argument is convincing, and I checked the mechanical part rather than trusting it: `dd130` appears 50 times against `dd120`'s 50, and there are 24 new catch blocks for 23 accessors plus the constructor — so no accessor arm was missed, which is the main risk in a hand-repeated 24-way edit. Documenting the substitutions and the upstream schema's own copy-then-edit history makes this reviewable in a way a regenerated blob wouldn't be.

Also worth saying: adding `OpenxQuest_example_1_2_0.mzid` and `mzidLib_xtandem_fdr_1_2_0.mzid` closes a real pre-existing hole — the 1.2.0 arm had **zero** fixtures on master (both existing mzid files are `.../1.1.1`-namespaced), so that whole third arm was untested before this PR.

Two comments inline: one gap I'd fix before merge (one line), and one I'd at least open an issue for.

Note this stacks with #1171 — on the merged tree without that fix, a compliant 1.1 file falls through to the *1.3.0* deserializer and throws there, which will look like a 1.3.0 bug. You already flagged the `MzIdentMLType130` -> `.../1.3` row for `MzIdentMlNamespaceTests`; whichever merges second should carry it.

mzLib/MzIdentML/MzidIdentifications.cs
acesnik: 🤖 Fixed — added `[TestCase("mzIdentML130.Generated")]`.

You were right that it was excluded, and about the size of it: `mzIdentML1_3_0.cs` was at 38.83% line coverage (577/1486); with the row it is at 100%, the same as the other three trees. Measured both ways with coverlet against a harness referencing `MzIdentML.csproj`.

That does cover `MzIdentMLType130.cvParam`, so the one hand-written property now has a test asserting its getter and setter round-trip.

I also updated the class summary, which still said "three ... trees" and "all three versions", and noted there that the namespace list is opt-in — so the next version tree does not land in the same hole.

mzLib/MzIdentML/MzidIdentifications.cs
acesnik: 🤖 Agreed, and agreed it should not ride along here — done as #1183, stacked on this, rather than rewriting all 24 cascades under a 1.3.0 change.

Two parts of your analysis I checked rather than just took.

The exception count is right. A 1.3.0 file throws and swallows three NREs per accessor call — `dd110`, `dd111`, `dd120` — before reaching the arm that works, where a 1.2.0 file today throws two.

The fallthrough point is the one I would have missed, and it turned out to be measurable. Holding the old implementation side by side with the flattened one and comparing every accessor, every index including one past the end on each axis, across all six checked-in fixtures: 3,642 comparisons, **0 value differences**, and 1,458 differences in thrown exception type — every one of them `NullReferenceException` becoming `IndexOutOfRangeException` on an out-of-range index. That is your point, counted: the old cascade was swallowing the real exception and letting the last arm's NRE out in its place. The two 1.3.0 fixtures show 0 such changes, because for them the last arm was already the matching one, which is a decent check that the mechanism is what we think it is.

Your point that typed catches cannot separate the two NREs is exactly why the null-check-on-the-field version is the one to write.

acesnik:
🤖 Pushed `0c2e9bea`. The reflection-sweep row is in, with the before/after numbers in that thread; the cascade cost is #1183, stacked on this branch, rather than folded in here.

On #1171 — agreed, and there is nothing for me to add on this branch, since `MzIdentMlNamespaceTests.cs` does not exist here yet. Whichever merges second adds the `MzIdentMLType130` → `.../1.3` row to both `[TestCase]` tables. I have confirmed `mzIdentML1_3_0.cs` declares `.../1.3` on both `XmlRootAttribute` and `XmlTypeAttribute`, so that row passes as written. Happy to carry it here if #1171 lands first.

mzLib/MzIdentML/MzidIdentifications.cs
trishorts: Verified at head: `[TestCase("mzIdentML130.Generated")]` is in `MzIdentMlReflectionCoverageTests`, and the class summary now says the list is opt-in. Resolving.

mzLib/MzIdentML/MzidIdentifications.cs
trishorts: Agreed on splitting it out, and #1183 is the right shape — null check on the field, not typed catches. The 1,458 NRE-to-IndexOutOfRange differences across 3,642 comparisons is a better demonstration of the fallthrough sink than my argument for it was. Resolving here; the discussion continues on #1183.

nbollis (approved):

trishorts (approved):
Both threads I opened are addressed and verified:

- `[TestCase("mzIdentML130.Generated")]` is in `MzIdentMlReflectionCoverageTests`, taking `mzIdentML1_3_0.cs` from 38.83% to parity with the other three trees, and the class summary now records that the list is opt-in;
- the cascade cost is correctly split out to #1183 rather than rewritten under a 1.3.0 change — the null-check-on-the-field shape is the right one, and that discussion continues there.

The only commit since my last pass is `7875365d`, a clean base-only merge of master with no conflict resolution. CI is green. Approving.

Issue #353, opened 2018-07-11
add unit test to read metamorpheus gptmd db modifications


trishorts:
too unstable right not to write a useful unit test.

acesnik:
We have tests that read, write, and read a database and check that the second read gives the same result as the first.

acesnik:
Which XML are you referring to?

acesnik:
🤖 The unit test this asks for exists on current master (`5ba13155`):

- `TestProteomicsReadWrite.AddModsDirectlyToProteinDbWriter` (`Test/DatabaseTests/TestProteomicsReadWrite.cs:341`) covers the GPTMD path directly — modifications supplied through the `Dictionary<accession, HashSet<(position, Modification)>>` argument, written with `ProteinDbWriter.WriteXmlDatabase`, read back with `ProteinDbLoader.LoadProteinXML`, asserting `OneBasedPossibleLocalizedModifications` goes from 0 to 1.
- `Test_write_with_custom_mods` (`:411`) reads an XML, writes it, reads it again, and asserts the modification count on both sides.
- `TestWriteProSightCompatibleMods_WithAdditionalMods` (`:780`) covers the same additional-mods path for the ProSight-compatible writer.
- `TestDatabaseLoaders.LoadingIsReproducible` (`Test/DatabaseTests/TestDatabaseLoaders.cs:96`) loads a real `cRAP_databaseGPTMD.xml` fixture.

Closing as done rather than fixed, since the work is already in place.

The separate crash noted in this thread — a correct two-letter field code with an empty value — did reproduce on current master, so it is tracked in #1170 with a repro rather than here. Of the thirteen field codes, six threw: `TG`, `CF`, `DR`, `TR`, `KW` with `NullReferenceException` and `MM` with `ArgumentOutOfRangeException`, at `Omics/Modifications/IO/ModificationLoader.cs:172-182`. `ID`, `MT` and `PP` already produced an error entry instead, and `NL`, `DI`, `FT` and `AC` already tolerated it.

Reopen if a GPTMD database exercises a modification path these tests miss — a concrete file whose mods are read back wrong would be the signal.

PR #1170, opened 2026-08-16
Guard ptmlist field codes that carry no value

acesnik:
🤖 Draft. Found while checking whether #353 is still current; that issue's own request is already satisfied, so this covers only the crash noted in its thread.

A field code with an empty value left `modValue` null. `ModificationLoader.cs:172-182` only assigns it when the line is longer than five characters, and it calls `.Trim()` *before* `.Substring(5)`, so any line whose trimmed content is five characters or fewer falls through the `try` and stays null. Six of the thirteen cases then dereferenced it.

Measured on current master by emptying each field code in turn, in both the bare (`TG`) and column-padded (`TG␣␣␣`) forms:

| behaviour | field codes |
|---|---|
| **unhandled throw** | `TG`, `CF`, `DR`, `TR`, `KW` → `NullReferenceException`; `MM` → `ArgumentOutOfRangeException` |
| error entry, no throw | `ID`, `MT`, `PP` |
| tolerated | `NL`, `DI`, `FT`, `AC` |

### Reachable through our own writer

Not a hand-edited-file-only problem. `Modification.ToString()` null-guards every field, but empty-but-non-null values pass the guard, so mzLib writes two forms it cannot read back:

| modification | written as | read back as |
|---|---|---|
| empty keyword | `KW␣␣␣` | `NullReferenceException` |
| empty database reference | `DR␣␣␣;␣` | `IndexOutOfRangeException` |

Nothing shipped triggers it — no text file in the repo contains a bare field code, and `ChemicalFormula.ParseFormula("")` throws, so the `CF` path cannot be produced by the writer.

### The guard

An empty value is treated as an absent line, so a mandatory field that is present but blank fails the same validation it would if the line were missing. That reuses the existing path rather than adding a new error mode, and it is what `ID`, `MT` and `PP` already did.

`NL` and `DI` are excluded by an explicit set, because both are documented to be meaningful with no value, and the `//` terminator must still reach the switch — a blanket skip would swallow it and never close the record. `DR` and `TR` additionally check the value splits into a `key; value` pair before indexing, which is what the `DR␣␣␣;␣` case needs; emptiness alone does not cover it.

### Two copies

The parsing loop is duplicated in `UsefulProteomicsDatabases/PtmListLoader.cs:180`, which is `[Obsolete]` but still public and still exercised by tests, and which crashes identically. Both are guarded. My first pass patched only the obsolete copy — worth knowing if you touch this area, since the class comment says functionality "moved to" `ModificationLoader` when it was in fact duplicated.

### Testing

16 new cases in `TestPtmListLoader`. Against the unguarded loaders 9 fail — the six field codes, the two malformed-pair cases, and the writer round-trip; the other 7 pass before and after, pinning that `ID`/`MT`/`PP`/`NL`/`DI`/`FT`/`AC` behaviour is unchanged.

Full solution builds clean. The new tests run on macOS via a throwaway net10.0 harness; the rest of the suite is `net10.0-windows` and has not been run locally, so CI covers it.

### DIFF
--- a/mzLib/Omics/Modifications/IO/ModificationLoader.cs
+++ b/mzLib/Omics/Modifications/IO/ModificationLoader.cs
@@ -9,6 +9,11 @@ using System.Xml.Serialization;
 namespace Omics.Modifications.IO;
 public static class ModificationLoader
 {
+    // Field codes whose handlers below require a value. NL and DI are absent on purpose: both
+    // accept an empty value.
+    private static readonly HashSet<string> FieldCodesRequiringAValue =
+        new() { "TG", "CF", "MM", "DR", "TR", "KW" };
+
 
     #region PTM List Loading (from ptmlist.txt format)
 
@@ -182,6 +187,14 @@ public static class ModificationLoader
                     }
                 }
 
+                    // A field code with no value is treated as an absent line (#353). Without this the cases
+                    // below dereference a null modValue. NL and DI are excluded deliberately: both are
+                    // documented to be meaningful with no value, and "//" must still reach the switch.
+                    if (string.IsNullOrWhiteSpace(modValue) && FieldCodesRequiringAValue.Contains(modKey))
+                    {
+                        continue;
+                    }
+
                 switch (modKey)
                 {
                     case "ID": // Mandatory
@@ -237,6 +250,10 @@ public static class ModificationLoader
 
                     case "DR": // External database links!
                         var splitStringDR = modValue.TrimEnd('.').Split(new string[] { "; " }, StringSplitOptions.None);
+                        if (splitStringDR.Length < 2)
+                        {
+                            break; //not a "key; value" pair
+                        }
                         if (_databaseReference == null)
                         {
                             _databaseReference = new Dictionary<string, IList<string>>();
@@ -254,6 +271,10 @@ public static class ModificationLoader
 
                     case "TR": // External database links!
                         var splitStringTR = modValue.TrimEnd('.').Split(new string[] { "; " }, StringSplitOptions.None);
+                        if (splitStringTR.Length < 2)
+                        {
+                            break; //not a "key; value" pair
+                        }
 
                         if (_taxonomicRange == null)
                         {
--- a/mzLib/Test/Omics/Modifications/TestPtmListLoader.cs
+++ b/mzLib/Test/Omics/Modifications/TestPtmListLoader.cs
@@ -2,6 +2,7 @@
 using NUnit.Framework;
 using Assert = NUnit.Framework.Legacy.ClassicAssert;
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Omics.Modifications;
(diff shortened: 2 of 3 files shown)


mzLib/Omics/Modifications/IO/ModificationLoader.cs
```diff
@@ -9,6 +9,11 @@
 namespace Omics.Modifications.IO;
 public static class ModificationLoader
 {
+    // Field codes whose handlers below require a value. NL and DI are absent on purpose: both
+    // accept an empty value.
+    private static readonly HashSet<string> FieldCodesRequiringAValue =
+        new() { "TG", "CF", "MM", "DR", "TR", "KW" };
```
trishorts, line 15: The carve-out rationale doesn't hold: NL and DI don't "accept an empty value", they *lose data* on one.

Both cases pass `modValue` straight into `DiagnosticIonsAndNeutralLosses` (:284 and :292), whose first
statement is:

```csharp
string[] nlOrDiEntries = oneEntry.Split(new string[] { " or " }, StringSplitOptions.None);
```

On a null that throws NRE, which the method's blanket `catch` swallows — and the handler is
`dAndNDictionary = null`. The caller then assigns that back:

```csharp
_neutralLosses = DiagnosticIonsAndNeutralLosses(modValue, _neutralLosses);
```

So a bare `NL` line silently discards every neutral loss already accumulated for that modification:

```
NL   H2O
NL
```

leaves `_neutralLosses` null rather than `{AnyActivationType: [18.0106]}`. Same for `DI`.

The strongest argument that these belong *inside* the guard is master's own comment on the DI case:

```csharp
case "DI": // Masses of diagnostic ions. Might just be "DI"!!! If field doesn't exist, create an empty list!
```

The code creates the empty dictionary on the line above and then nulls it. Bare `DI` is broken in
exactly the way that comment says it should work — which is presumably the behaviour issue #353's
thread was circling.

`TestFieldCodeWithNoValueDoesNotThrow` covers `NL` and `DI` and passes, because it only asserts
`DoesNotThrow`. That's the part I'd change regardless of what you decide about the guard: a test that
proves only the absence of a throw can't see a silent wipe. Asserting that a preceding well-formed
`NL` survives a following empty one would catch it.

Everything else here looks right to me — the `continue`-and-let-`ValidModification`-report shape is
the correct one (it routes through the existing `filteredModificationsWithWarnings` sink instead of
inventing a new error path), and `TestBlankMandatoryFieldIsTreatedAsAnAbsentLine` comparing against
the absent-line result rather than a hardcoded message is a nice touch. Good catch fixing
`PtmListLoader` in lockstep too — that obsolete copy duplicates the loop rather than delegating, so a
guard on one really isn't a guard on the other.

mzLib/UsefulProteomicsDatabases/PtmListLoader.cs
```diff
@@ -19,6 +19,11 @@ public static class PtmListLoader
     {
         private static readonly Dictionary<string, char> AminoAcidCodes;
 
+        // Field codes whose handlers below require a value. NL and DI are absent on purpose: both
+        // accept an empty value.
+        private static readonly HashSet<string> FieldCodesRequiringAValue =
+            new() { "TG", "CF", "MM", "DR", "TR", "KW" };
```
trishorts, line 25: Heads-up on a merge collision rather than a code comment: **#1162** ("Ship XML documentation…") also
edits both `mzLib/Omics/Modifications/IO/ModificationLoader.cs` and
`mzLib/UsefulProteomicsDatabases/PtmListLoader.cs`. It's a 55-file PR, so whichever of the two lands
second will conflict right around the parse loop you're guarding here. Worth deciding the order
deliberately — this one is much smaller and probably wants to go first.

trishorts (commented):
Early input since it's still a draft — take or leave as you see fit.

I confirmed the underlying defect on master: the length guard tests the raw `line` while `Substring(5)` runs on the shorter `Split('#')[0].Trim()`, so there are two ways to reach a null `modValue` (short line falls past the `if`; commented line throws inside the `try` and is swallowed by the bare `catch`). Six cases then dereference it, and the `MM` case is worse than the others because the `ArgumentOutOfRangeException` comes from `line.Substring(5)` *inside the throw that reports the error*.

One substantive comment inline, plus a merge-ordering heads-up. The design of the fix looks right to me; it's the NL/DI exclusion I'd revisit.

mzLib/Test/Omics/Modifications/TestPtmListLoader.cs
```diff
@@ -165,5 +167,138 @@ public static void TestInvalidModTypeError()
...
+            foreach ((string name, ReadModsFromString read) in Loaders)
+            {
+                var fromMalformed = read(malformed, out var malformedErrors).ToList();
+
+                NUnit.Framework.Assert.That(fromMalformed.Count, Is.EqualTo(1),
+                    $"{name} discarded the whole record over a malformed {fieldCode}");
+                NUnit.Framework.Assert.That(malformedErrors.Count, Is.EqualTo(0));
+                NUnit.Framework.Assert.That(fromMalformed[0].IdWithMotif, Is.EqualTo("testmod on X"));
+                NUnit.Framework.Assert.That(fromMalformed[0].MonoisotopicMass, Is.Not.Null,
+                    $"{name} lost a field that came after the malformed {fieldCode}");
```
trishorts, line 232: This assertion doesn't check what its message says, so it isn't testing that parsing survived the
malformed line.

`ValidMod` is `ID / MT / PP / TG / CF / //`, and the malformed line is spliced in where `//` was — so
it lands *after* `CF   H1`. `MonoisotopicMass` comes from that `CF`, which is **before** the malformed
`DR`/`TR` line, not after it. The check passes whether or not the loop kept going past the bad line.

That matters because "the guard skips the field without eating the rest of the record" is the exact
property this test exists to prove, and it's the one thing here that isn't currently pinned. Asserting
on a field placed after the malformed line — or splicing the malformed line before `CF` rather than
after it — would make it real.

(Same wording appears in the `Is.Not.Null` message; the other assertions in this test are fine.)

trishorts (commented):
Follow-up after re-running the review tooling (it hit a session limit on my first pass). One concrete test-logic point below.

For the record, it also flagged that `FieldCodesRequiringAValue` omits `BM` and `BL` — I checked and that's correct behaviour, not a gap: both already guard with `string.IsNullOrWhiteSpace(modValue)` in their own cases, so they're self-protecting. Might be worth a word in the comment next to the set, since the omission reads as an oversight.

trishorts:
@acesnik — flagging a merge-order requirement rather than a review finding; I have not reviewed the change itself yet.

**#1270 must merge before this PR, and this PR must be green against it.**

This PR guards ptmlist field codes that carry no value, and three of the codes in that set — `CF`, `MM` and `DR` — are exactly the inputs the formal-charge adjustment branches on:

```csharp
bool useChemFormulaForMM    = _monoisotopicMass == null && _chemicalFormula != null;
bool adjustWithFormalCharge = _monoisotopicMass != null && _databaseReference != null;
```

So a guard that changes whether a valueless `MM` or `DR` leaves its field null changes which branch fires, and therefore what a modification weighs. That is not a criticism of the guard — skipping a valueless field is right — it is that the blast radius reaches mass arithmetic where a wrong answer is silent: the modification still loads, still searches, still reports identifications, against a mass one proton out, with nothing downstream that would notice.

#1270 pins the invariant that has to survive:

```
trimethyllysine, MM 43.054227 + FormalCharge 1+  ->  42.046951 / C3H6   (Unimod Trimethyl)
```

UniProt writes that modification as `CF   C3 H7`, the quaternary ammonium as it exists on the protein — permanently charged, needing no proton to become so — so the formula cannot be added to a peptide as written, and dropping the adjustment leaves every trimethyl site a proton heavy. The test asserts it through both `ModificationLoader` and `PtmListLoader`.

#1270 is test-only with no production changes, so it can land on its own. Same requirement is on #1182, which deletes a copy of that arithmetic.

trishorts:
@acesnik — the gate I put on this PR is cleared. #1270 merged (`828c06fb`, now master's tip), so the trimethyllysine formal-charge invariant is pinned on master and this PR is free to move.

I also checked rather than just announcing it: merged current master into `decc7d12` locally — clean, no conflicts — and ran both gate tests against the result.

```
Passed TrimethylLysine_FormalChargeAdjustmentYieldsTheNeutralMassIncrement
Passed TrimethylLysine_FormalChargeAdjustmentIsTheSameThroughPtmListLoader
```

So this PR does not disturb the invariant. That was the open question, given it guards `CF`, `MM` and `DR` — the three fields the adjustment branches on — and a guard that changed whether a valueless `MM` stays null would change which branch fires.

Nothing else from me here; I have not reviewed the change itself yet. Update the branch when you are ready and I will take a proper look.

Still outstanding from your earlier replies, and not blocking: the `IFilteringParams` extension issue collapsing the six `ShouldTrim` copies, and the exact-token switch for the mgf header chain. Both were on #1227 rather than here, but neither has been opened yet and both are the kind of thing that evaporates once the thread scrolls away.

Issue #526, opened 2019-12-05
"format of the URI could not be determined" crash


rmillikin:
Something to do with the SourceFile constructor missing a URI

acesnik:
🤖 Fixed, and the same underlying defect as #300 — the two titles are just the two messages `new Uri(string)` produces for a malformed `sourceFile/@location`:

```
file://.         -> Invalid URI: The hostname could not be parsed.              (#300)
Company          -> Invalid URI: The format of the URI could not be determined. (this issue)
../relative/dir  -> Invalid URI: The format of the URI could not be determined.
```

Both came from the same unguarded `new Uri(simpler.location)` in `Mzml.GetSourceFile`, replaced by a guarded `Uri.TryCreate` in #1060 (`mzLib/Readers/MzML/Mzml.cs:296`). A follow-up PR adds a regression fixture covering both message classes.

On the note about the `SourceFile` constructor: that constructor has used `Uri.TryCreate` since September 2017, two years before this report, so it cannot throw and was not the crash site. It did point at something real, though — the overload taking neither a path nor a `Uri` leaves `Uri` null, and the mzML writer dereferenced it unguarded, which throws a `NullReferenceException` when writing mzML from such a file. The follow-up PR fixes that too.

Closing as already fixed. Reopen if a `location` value still aborts a load.

Issue #300, opened 2018-02-26
Read .mzML file report: invalid uri the hostname could not be parsed


lonelu:
I found this is caused by file with something like **location="file://."** , which is not able to be parsed.
I change it to **location="file://E:/"** Then it is OK. 
In case anyone meet the same problem in the future.

rmillikin:
can you give an example? I'm not sure I understand

I usually just use something like
myFilePath = @"C:\Users\rmillikin\myfile.txt";

lonelu:
I cannot search some public .mzML files in MetaMorpheus, because it report: invalid uri the hostname could not be parsed. 
Then I check it and found it is the mzLib  Mzml.LoadAllStaticData  is related with this problem. 
My guess is that these .mzML files contains something that could not be read by the following code: Mzml.LoadAllStaticData > sourceFile = new SourceFile > new Uri(simpler.location). It turns out it is location="file://." in the .mzML files need to be changed.

rmillikin:
strange. that path is in the mzml file itself? sounds like @dippman might be able to fix this in mzLib

lonelu:
The path is in the mzML file. It is maybe related to filepath created by linux.  Because "." means current filepath in linux, which could not read by Windows. What a small weird bug.

acesnik:
🤖 Fixed. `Mzml.GetSourceFile` no longer calls `new Uri(location)` directly — it uses a guarded `Uri.TryCreate`, falling back to the opened file's own path and then to a placeholder (`mzLib/Readers/MzML/Mzml.cs:296`). That landed in #1060.

Verified on .NET 10 that the guard is still doing real work rather than being belt-and-braces: `new Uri("file://.")` continues to throw `UriFormatException: Invalid URI: The hostname could not be parsed.`, but reading an mzML carrying `location="file://."` now loads normally.

Nothing pinned this behaviour, so a follow-up PR adds a regression fixture covering `file://.` along with `file://`, `Company`, `Z:\some\windows\path`, `../relative/dir` and an empty value.

Closing as already fixed. Reopen if a `sourceFile/@location` value still aborts a load.

Issue #259, opened 2017-10-16
Proteome Exchange Submission fail


stefanks:
FILE: 12-10-16_A17A_yeast_BU_fract5_rep1-Calibrated.mzML
VALIDATION MESSAGE: (1) Non-fatal XML Parsing error detected on line 13, (2)
Non-fatal XML Parsing error detected on line 48, (3) Error message:
cvc-datatype-valid.1.2.1: '12-10-16_A17A_yeast_BU_fract5_rep1-Calibrated' is
not a valid value for 'NCName'., (4) Error message: cvc-attribute.3: The value
'12-10-16_A17A_yeast_BU_fract5_rep1.raw' of attribute 'id' on element
'sourceFile' is not valid with respect to its type, 'ID'., (5) Error message:
cvc-datatype-valid.1.2.1: '12-10-16_A17A_yeast_BU_fract5_rep1.raw' is not a
valid value for 'NCName'., (6) Error message: cvc-attribute.3: The value
'12-10-16_A17A_yeast_BU_fract5_rep1-Calibrated' of attribute 'id' on element
'run' is not valid with respect to its type, 'ID'.

FILE: 12-10-16_A17A_yeast_BU_fract5_rep1-Calibrated_1mm.mzid
VALIDATION MESSAGE: (1) Fatal Error message: Content is not allowed in prolog.,
(2) FATAL XML Parsing error detected on line 1

PR #1171, opened 2026-08-16
Write schema-valid mzML ids and fix the mzIdentML 1.1.0 namespace

acesnik:
🤖 Four defects that each make mzLib's XML output unusable to a schema-validating consumer, plus regression tests for two issues that were already fixed but never pinned. Validated against the mzML and mzIdentML schemas from HUPO-PSI.

This is the mzLib half of #259. The `.mzid` half also needs a MetaMorpheus change and a release, so #259 stays open — see the bottom of this description.

## 1. mzML `run/@id` and `sourceFile/@id` are not NCNames

Both are `xs:ID`, so a value may not start with a digit, `.` or `-`, and may not contain `:` or a space. `MzmlMethods` derives them from file names, and dated instrument names like `12-10-16_yeast_rep1.raw` break the first rule — the original #259 report.

A guard already existed, added in #263 as the response to #259:

```csharp
string idTitle = char.IsNumber(title[0]) ? "id:" + title : title;
```

`:` is exactly what NCName forbids, so a digit-leading name traded one violation for another. The `sourceFile` guard had a second problem — `"id:" + FileName[0]` takes a single character, so the id becomes `id:0` and the filename is discarded.

`ToValidXmlId` replaces both: illegal characters become `_`, and a leading `_` is prepended when the first character cannot start an NCName. Names already valid are returned unchanged. The real filename stays on `sourceFile/@name` and `@location`.

**The vignette's own output is invalid today.** A full Calibrate → GPTMD → Search run on the two vignette fractions produces calibrated mzML like this:

```
mzML/@id       = 'id:04-30-13_CAST_Frac4_6uL-calib'   INVALID NCName
sourceFile/@id = 'id:0'                               INVALID NCName
run/@id        = 'id:04-30-13_CAST_Frac4_6uL-calib'   INVALID NCName
schema validation: FAIL (3 errors)
```

3 errors per file on both fractions, including a failed `KEY_SOURCEFILE_ID` identity constraint. `id:0` is the `FileName[0]` bug erasing the source filename.

Sweeping all 24 checked-in mzML fixtures through a round-trip: **7 had ids rewritten, and all 7 were previously invalid.** No already-valid name changed, so the blast radius is exactly the output that was broken.

Three fixtures remain schema-invalid afterwards for a reason that is not a writer bug: `BinGenerationTest.mzML` and the two Velos `noRt` files carry duplicate spectrum nativeIDs *in the input* and the writer copies them faithfully.

## 2. An empty `sourceFileList`

`fileDescription.sourceFileList` was always constructed but only populated when `NativeIdFormat`, `MassSpectrometerFileFormat` and `FileChecksumType` are all set. Otherwise an empty `<sourceFileList/>` was written, which is invalid: the element requires a `count` attribute and at least one `sourceFile` child once present. It is `minOccurs="0"`, so it is now omitted. `tester.mzML` was this case and now validates. This is also the MGF→mzML path, since `Mgf` leaves `FileChecksumType` null.

## 3. The mzIdentML 1.1.0 root namespace

The schema version and the target namespace are different strings, and the root type had them confused:

| schema | targetNamespace |
|---|---|
| 1.1.0 | `http://psidev.info/psi/pi/mzIdentML/1.1` |
| 1.1.1 | `http://psidev.info/psi/pi/mzIdentML/1.1.1` |
| 1.2.0 | `http://psidev.info/psi/pi/mzIdentML/1.2` |
| 1.3.0 | `http://psidev.info/psi/pi/mzIdentML/1.3` |

`MzIdentMLType110` declared `.../1.1.0`; the other 65 namespace references in the same file already said `.../1.1`. Introduced in #910 as a 4-line change, shipped in 1.0.584.

- **Read.** `MzidIdentifications` threw `InvalidOperationException` on a real mzIdentML 1.1 file. With the fix it reads it — 18,023 matches. No generated root type matched `.../1.1`, so a compliant file failed every version attempt and the last one's exception escaped the constructor.
- **Write.** Every `.mzID` MetaMorpheus currently produces is in `.../1.1.0`; all six files from a full 5-task vignette run confirm it. Such a file cannot be validated against the mzIdentML schema at all, so ProteomeXchange rejects it outright.

Not caught because both fixtures added in #910 are `.../1.1.1` files, which never exercise the 1.1.0 type. The new tests assert the declared *and* the serialized namespace for every generated root type — 1.1.0, 1.1.1, 1.2.0 and 1.3.0 — so the whole mapping is pinned rather than the one line that broke.

### Files mzLib already wrote in `.../1.1.0` stay readable

Correcting the namespace would otherwise have made every `.mzID` mzLib has already produced unreadable: `XmlSerializer` matches namespaces exactly, and after the fix no type declares `.../1.1.0`. So the constructor gains a **last** attempt that deserializes the 1.1.0 types through `LegacyMzidNamespaceReader`, an `XmlTextReader` that reports `.../1.1.0` as `.../1.1`.

- **Remapped in the reader, not with an `XmlRootAttribute` override.** An override renames only the root, so the root deserializes while every child stays bound to `.../1.1` and is silently dropped — measured: `id` came through, `cvList` and `AnalysisSoftwareList` came back null. That trades a crash for a file of identifications reading back empty. `NestedElementsSurviveTheNamespaceRemap` reads a `ParentTolerance` three levels down, which is what separates the two fixes.
- **Last, after every real format version**, so a namespace rewrite can never pre-empt a document that parses as a genuine version. The order is 1.1.0 → 1.1.1 → 1.2.0 → 1.3.0 → legacy rewrite.
- **DTDs prohibited.** `XmlTextReader` defaults to `DtdProcessing.Parse`, while the reader `XmlSerializer` builds for the `Stream` overload used by the other arms prohibits DTDs. Without an explicit `Prohibit`, a document carrying an internal DTD subset was refused by the first four attempts and then accepted by this one — in the compliant namespace too, since the remap is a no-op for `.../1.1`. `ADocumentTypeDeclarationIsRejectedInEitherNamespace` failed in both namespaces before the fix.

One consequence to know when reading a bug report: a file no attempt can parse now surfaces the legacy attempt's exception. For a malformed 1.3.0 file that means the 1.3.0 parse error is swallowed and the reported cause is `<MzIdentML xmlns='.../1.3'> was not expected` — the symptom of a version mismatch, not of what is actually wrong with the file.

### Interaction with #1169 and #1183

Both have merged, and this branch is current with them. #1169 added a 1.3.0 attempt at the same innermost position this PR added the legacy rewrite, so that merge conflicted in `MzidIdentifications.cs`; it was resolved by nesting both, 1.3.0 first. #1183 (version dispatch by null check) merged in cleanly — the legacy rewrite populates `dd110`, which #1183's accessors check first.

## 4. A null `SourceFile.Uri`

The `SourceFile` overload taking neither a path nor a `Uri` leaves `Uri` null — `Mgf.cs:77` uses it — and `Uri.TryCreate` leaves it null on failure. The required `location` attribute dereferenced it unguarded, giving a `NullReferenceException` when writing mzML from such a file. It now falls back to `file:///unknown-source`, matching what the reader already does.

The same constructor leaves `FileName` null, and `name` is also `use="required"` — a null string attribute is omitted by `XmlSerializer`, so that output was schema-invalid too. It now falls back to `unknown-source`. Both values are constants on `SourceFile` (`UnknownLocation`, `UnknownName`), shared by the reader and the writer so the two cannot drift apart.

## Regression tests for #300 and #526

These are the same defect reported twice, distinguishable only by which message `new Uri` produced:

```
file://.        -> "The hostname could not be parsed."              #300
Company         -> "The format of the URI could not be determined." #526
../relative/dir -> "The format of the URI could not be determined." #526
```

Both hit the old unguarded `new Uri(simpler.location)`, fixed by #1060 (`Mzml.cs:296`) with nothing pinning it. The new fixture reads mzML carrying all three plus `file://`, a bare Windows path and an empty value, and asserts the raw parses still throw so the test speaks up rather than passing vacuously if a future runtime changes.

A caveat on #526: its body says *"SourceFile constructor missing a URI"*, but that constructor has used `Uri.TryCreate` since September 2017, two years before the report, so it cannot throw. The title's message maps only to the relative-location case. The original crash could not be reproduced — no stack trace or file in the issue — so that mapping is inference rather than measurement. Item 4 above is the one live defect matching the body's wording.

## Scope

Fixes the mzML file in #259. Its `.mzid` file needs this **plus** a MetaMorpheus writer change (empty `ProteinDetectionHypothesis` elements) **plus** an mzLib release and a `PackageReference` bump before MetaMorpheus output is submittable, so **#259 should stay open**.

## Verification

**29 test cases** across four fixtures:

| fixture | cases |
|---|---|
| `TestMzmlXmlIdValidity.cs` | 6 |
| `TestMzmlSourceFileLocation.cs` | 9 |
| `MzIdentMlNamespaceTests.cs` | 8 |
| `MzIdentMlLegacyNamespaceTests.cs` | 6 |

The tests for each fix were confirmed to fail against the code without it before being trusted: the namespace tests fail on exactly the two 1.1.0 cases (1.1.1, 1.2.0 and 1.3.0 were never wrong and pass either way), the `name`/`location` test fails against the unguarded writer, and the DTD test fails in both namespaces without `Prohibit`.

Run on Windows (`net10.0-windows`) against the current head, which includes #1169 and #1183: all 217 mzIdentML and mzML tests pass.


nbollis (dismissed):

mzLib/MzIdentML/mzIdentML1_1_0.cs
```diff
@@ -22,8 +22,10 @@ namespace mzIdentML110.Generated
     [System.SerializableAttribute()]
     [System.Diagnostics.DebuggerStepThroughAttribute()]
     [System.ComponentModel.DesignerCategoryAttribute("code")]
-    [System.Xml.Serialization.XmlTypeAttribute(Namespace = "http://psidev.info/psi/pi/mzIdentML/1.1.0")]
-    [System.Xml.Serialization.XmlRootAttribute("MzIdentML", Namespace = "http://psidev.info/psi/pi/mzIdentML/1.1.0", IsNullable = false)]
+    // mzIdentML 1.1.0's targetNamespace is .../1.1, not .../1.1.0 -- the schema version and the
+    // namespace differ. Every other type in this file already says .../1.1.
+    [System.Xml.Serialization.XmlTypeAttribute(Namespace = "http://psidev.info/psi/pi/mzIdentML/1.1")]
+    [System.Xml.Serialization.XmlRootAttribute("MzIdentML", Namespace = "http://psidev.info/psi/pi/mzIdentML/1.1", IsNullable = false)]
```
trishorts, line 28: The namespace correction itself is right — I checked the file at head and all 67 references now read
`.../1.1` consistently, which matches the schema's `targetNamespace`.

But the same change silently drops the ability to read every `.mzID` mzLib has already written.

After this diff there are **zero** declarations of `.../1.1.0` anywhere in the assembly (I grepped the
head commit: `mzIdentML/1.1.0` → 0 hits). `XmlSerializer` matches the root element namespace exactly,
so a document whose root is `<MzIdentML xmlns="http://psidev.info/psi/pi/mzIdentML/1.1.0">` now fails
the 1.1.0 attempt, fails 1.1.1, fails 1.2.0 — and the innermost attempt in the constructor has no
`catch`, so `new MzidIdentifications(path)` throws `InvalidOperationException` rather than returning.

Your own description is what makes this concrete: *"Every `.mzID` MetaMorpheus currently produces is
in `.../1.1.0`; all six files from a full 5-task vignette run confirm it."* That is the exact corpus
that stops loading. Anything that re-reads previous search results through `MzidIdentifications`
goes from working to a hard constructor throw on upgrade.

This looks additive to fix, without touching a generated type or adding public API — one more
attempt in the cascade with an explicit root override:

```csharp
new XmlSerializer(typeof(mzIdentML110.Generated.MzIdentMLType110),
                  new XmlRootAttribute("MzIdentML") { Namespace = "http://psidev.info/psi/pi/mzIdentML/1.1.0" })
```

so compliant `.../1.1` files read (the fix) *and* legacy `.../1.1.0` files still read (the
compatibility). Worth a line in the description either way, since right now the read-side
consequence for existing files isn't mentioned — only the write side is.

Note this interacts with #1169: after that merges the innermost arm is 1.3.0, so the same legacy file
will throw from the *1.3.0* deserializer, which will read like a 1.3.0 bug.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzmlXmlIdValidity.cs
```diff
@@ -0,0 +1,215 @@
...
+            Assert.That(sourceFile.Uri, Is.Null, "this test is pointless if the constructor sets a Uri");
+
+            string outPath = Path.Combine(TestContext.CurrentContext.TestDirectory, "no_source_uri.mzML");
+
+            try
+            {
+                Assert.That(() => MzmlMethods.CreateAndWriteMyMzmlWithCalibratedSpectra(
+                    new GenericMsDataFile(new[] { OneScan() }, sourceFile), outPath, false), Throws.Nothing);
+
+                Assert.That(ReadIdAttributes(outPath).ContainsKey("sourceFile"), Is.True);
```
trishorts: This test can't fail for the thing it exists to pin.

It's the only coverage for the new `location = ... ?? "file:///unknown-source"` fallback, but it
asserts `Throws.Nothing`, that *a* `sourceFile` element has *an* `id`, and that the file re-reads with
one spectrum. It never reads `location` back.

Revert the writer to a bare `Uri?.ToString()` and this still passes: `location` is
`[XmlAttribute(DataType="anyURI")]`, so a null is omitted rather than throwing, and
`Mzml.GetSourceFile` treats a missing attribute as `IsNullOrWhiteSpace` and falls back to the opened
file path — so `NumSpectra` is still 1. The half-fix produces exactly the schema-invalid output this
PR exists to eliminate, and ships green.

Two additions would close it:
- read the written `sourceFile/@location` back and assert it is exactly `file:///unknown-source`;
- run the written `sourceFile/@id` through `XmlConvert.VerifyNCName`, as the other fixtures here do —
  this is the only test that exercises the null-`FileName` branch of `ToValidXmlId`, and it currently
  only checks the key exists.

While you're in here: this `SourceFile` is built with the id-only constructor, so `FileName` is null,
and the writer does `name = myMsDataFile.SourceFile.FileName` unguarded. A null there means the
`name` attribute is omitted from the written `sourceFile`. Worth running this specific output through
the schema you validated the rest against — if `sourceFile/@name` is `use="required"`, then the one
new test for the null path is itself producing an invalid file.

mzLib/Test/MzIdentML/MzIdentMlNamespaceTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        private const string Namespace120 = "http://psidev.info/psi/pi/mzIdentML/1.2";
+
+        [Test]
+        [TestCase(typeof(mzIdentML110.Generated.MzIdentMLType110), Namespace110)]
+        [TestCase(typeof(mzIdentML111.Generated.MzIdentMLType111), Namespace111)]
+        [TestCase(typeof(mzIdentML120.Generated.MzIdentMLType120), Namespace120)]
+        public void RootTypeDeclaresTheSchemaTargetNamespace(Type rootType, string expectedNamespace)
+        {
+            var root = (XmlRootAttribute)Attribute.GetCustomAttribute(rootType, typeof(XmlRootAttribute));
+            var type = (XmlTypeAttribute)Attribute.GetCustomAttribute(rootType, typeof(XmlTypeAttribute));
```
trishorts: `Attribute.GetCustomAttribute(Type, Type)` is the `inherit: true` overload, and `MzIdentMLType110`
derives from `IdentifiableType`, which carries its own `[XmlType(Namespace = ".../1.1")]`
(mzIdentML1_1_0.cs:886).

So if someone deleted the `XmlTypeAttribute` from `MzIdentMLType110`, this lookup would walk up and
find the base's — which now declares the same `.../1.1` — and the assertion would pass vacuously.
The same holds for the 1.1.1 and 1.2.0 cases, since each base agrees with its derived type.

The `XmlRootAttribute` half is fine — that one is only on the root types, not on `IdentifiableType`.
Passing `inherit: false` for the `XmlTypeAttribute` lookup would make the test fail on deletion, which
is presumably the intent given the whole point is pinning a mapping that a 4-line change broke once
already.

mzLib/Readers/MzML/MzmlMethods.cs
```diff
@@ -112,14 +151,14 @@ public static void CreateAndWriteMyMzmlWithCalibratedSpectra(MsDataFile myMsData
...
-                    myMsDataFile.SourceFile.FileName;
+                string idName = ToValidXmlId(myMsDataFile.SourceFile.FileName);
                 mzML.fileDescription.sourceFileList.sourceFile[0] = new Generated.SourceFileType
                 {
                     id = idName,
                     name = myMsDataFile.SourceFile.FileName,
-                    location = myMsDataFile.SourceFile.Uri.ToString(),
+                    // SourceFile.Uri is null whenever the constructor without a path was used, or
+                    // when that path failed to parse; location is required, so emit a placeholder.
+                    location = myMsDataFile.SourceFile.Uri?.ToString() ?? "file:///unknown-source",
```
trishorts: `file:///unknown-source` now exists in two places: here on the write side, and `Mzml.cs:301`
(`catch { sourceUri = new Uri("file:///unknown-source"); }`) on the read side.

Matching the reader is the right call — but the two string literals *are* the contract, and nothing
declares that. A shared `const` somewhere both can see (`SourceFile` would be the natural home) would
keep them from drifting apart.

Unrelated, and only so you know it was checked rather than assumed: I looked at whether dropping the
unconditional `sourceFileList` leaves anything dangling. It doesn't — all seven dereferences
(:118-140 on master) are inside the `if` gate, and no `defaultSourceFileRef` is ever written, so
there's no IDREF pointing at the omitted element. That change is clean.

trishorts (commented):
Read this one closely — the diagnosis work behind it is excellent, and the three defects are all real. The `ToValidXmlId` replacement is a clear improvement over both ad-hoc guards, and catching that `"id:" + FileName[0]` was taking a single character is a nice find.

One blocking concern and a few test-quality notes.

**The blocker is back-compatibility, not correctness.** The `.../1.1` namespace is the right value, but nothing in the diff keeps the files the old value already produced readable. Details inline.

I verified the parts I'm not raising, so you know they were checked rather than skipped: the `sourceFileList` removal leaves no null deref and no dangling IDREF; `mzML/@id` doesn't need sanitising (`mzMLType.id` is a plain attribute, not `xs:ID`, so only `run/@id` and `sourceFile/@id` are load-bearing); and `run/@id` vs `sourceFile/@id` can't collide in the normal flow, since one comes from `GetFileNameWithoutExtension` and the other keeps its extension.

mzLib/MzIdentML/mzIdentML1_1_0.cs
acesnik: 🤖 Confirmed and fixed, but the suggested fix doesn't work as written — and it fails in the dangerous direction, so it's worth spelling out.

Your diagnosis reproduces exactly. At head `4187f9de`: zero `mzIdentML/1.1.0` occurrences in any `.cs`, 67 `.../1.1` in `mzIdentML1_1_0.cs`. And the cascade is worse than degrading — the 1.2.0 attempt at `MzidIdentifications.cs:60-67` is the body of the last `catch`, so nothing is below it. A legacy document threw `InvalidOperationException` at (2, 2) out of the constructor.

An `XmlRootAttribute` override renames only the *root*. A legacy file carries `xmlns` on the root, so every descendant is in `.../1.1.0` too; those stay bound to `.../1.1` and are dropped. I measured all three:

```
plain (bound to .../1.1)              THREW InvalidOperationException
XmlRootAttribute override -> 1.1.0    obj=ok  id=demo  cvList=null  software=<null>
namespace-rewriting XmlReader         obj=ok  id=demo  cvList=1     software=MetaMorpheus
```

The middle row is a `.mzID` full of identifications reading back empty — silent loss where we currently get a crash. So the fourth arm uses a 5-line `XmlTextReader` subclass overriding `NamespaceURI` instead, which remaps at all depths; a compliant document passes through untouched.

Tests pin both directions, including a nested `ParentTolerance` three levels down — a root-only remap returns null there, which is what distinguishes the two fixes. Description updated for the read-side consequence. And agreed on #1169: once that lands this surfaces as a 1.3.0 bug, which is the wrong file to go looking in.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzmlXmlIdValidity.cs
acesnik: 🤖 Right that it can't fail, and the schema question you flagged turns it into a writer bug rather than a test-strength one.

mzML 1.1.0 from HUPO-PSI, `SourceFileType`:

```
id        xs:ID       use="required"
name      xs:string   use="required"
location  xs:anyURI   use="required"
```

`name` is required, `MzmlMethods.cs:158` set it from `FileName` unguarded, and this test's `SourceFile` has no `fileName` — so the attribute was omitted and the file this test produced was schema-invalid. The test meant to prove the fix was shipping the defect.

Fixed with a `SourceFile.UnknownName` fallback beside the location one, and the test now reads `location`, `name` and `id` back and runs `id` through `XmlConvert.VerifyNCName`. Verified red against the unfixed writer:

```
unfixed:  id=id, location=file:///unknown-source          name -> <omitted>
fixed:    id=id, location=file:///unknown-source, name=unknown-source
```

Worth noting `NumSpectra == 1` passed in both, which is your point exactly.

mzLib/Test/MzIdentML/MzIdentMlNamespaceTests.cs
acesnik: 🤖 Confirmed — `IdentifiableType` carries its own `[XmlTypeAttribute(Namespace = ".../1.1")]` at `mzIdentML1_1_0.cs:888` and declares `[XmlInclude(typeof(MzIdentMLType110))]`, so the two-arg lookup walks up and the assertion survives a deleted attribute. Switched to `inherit: false`, after checking all three root types declare their own `[XmlType]` so this doesn't fail against the real code. Left the `XmlRootAttribute` lookups alone — you're right that it isn't on the base.

mzLib/Readers/MzML/MzmlMethods.cs
acesnik: 🤖 Agreed — both literals are now `SourceFile.UnknownLocation`, with `SourceFile.UnknownName` next to it for the `name` case above.

Thanks for separating what you checked from what you raised; the `sourceFileList` and `run/@id` vs `sourceFile/@id` notes saved me re-deriving them. I confirmed the `mzML/@id` reasoning against the schema too — `mzMLType/@id` is `xs:string` and optional, `RunType/@id` is `xs:ID` and required.

mzLib/MzIdentML/MzidIdentifications.cs
```diff
@@ -21,12 +21,35 @@
...
+        /// bound to ".../1.1" and is silently dropped. Measured on a legacy document -- id came through,
+        /// cvList and AnalysisSoftwareList came back null. Remapping in the reader fixes all depths.
+        /// A compliant ".../1.1" document passes through untouched, so this is safe either way.
+        /// </remarks>
+        private sealed class LegacyMzidNamespaceReader : XmlTextReader
+        {
+            private const string LegacyNamespace = "http://psidev.info/psi/pi/mzIdentML/1.1.0";
+            private const string SchemaNamespace = "http://psidev.info/psi/pi/mzIdentML/1.1";
+
+            public LegacyMzidNamespaceReader(Stream stream) : base(stream) { }
```
trishorts, line 47: Minor, and the only thing I found in the new arm: this reader is the one place in the cascade where DTD processing is on.

The other three arms call `XmlSerializer.Deserialize(Stream)`, which builds its own `XmlTextReader` and sets `XmlResolver = null`, `Normalization = true`, `WhitespaceHandling.Significant`. Subclassing `XmlTextReader` and passing the instance to `Deserialize(XmlReader)` skips all of that, and `XmlTextReader.DtdProcessing` defaults to `Parse`.

Measured on a probe with the same subclass shape:

```
INTERNAL entity, Deserialize(Stream)   THREW XmlException: For security reasons DTD is prohibited in this XML document.
INTERNAL entity, subclass reader       child=[hello]        DtdProcessing=Parse
billion-laughs, Deserialize(Stream)    THREW in 0ms   (DTD prohibited)
billion-laughs, subclass reader        THREW in 101ms (MaxCharactersFromEntities)
```

Two things I checked so you don't have to: it is **not** an XXE — `XmlTextReader`'s resolver is null by default on .NET Core, so an external entity resolved to empty rather than reading the file — and the entity bomb is capped by `MaxCharactersFromEntities`. I also checked `Normalization`/`WhitespaceHandling`, expecting a difference in attribute values and CRLF handling; there was none (`61-09-62` and `78-0D-0A-79` on both paths), so that half of my guess was wrong.

So the practical effect is narrow: a legacy `.mzID` carrying a DTD parses through arm 4, while the identical file in the compliant namespace is rejected by arms 1-3. `DtdProcessing = DtdProcessing.Prohibit;` in the constructor makes the last arm agree with the other three, and costs nothing — a real 1.1.0 `.mzID` has no DTD.

mzLib/MzIdentML/mzIdentML1_1_0.cs
trishorts: Verified at `8fffa2b4`. You are right that the `XmlRootAttribute` override is the worse fix and I should have caught that it only renames the root — the remapping reader handles all depths, and `MzIdentMlLegacyNamespaceTests` pins the nested `ParentTolerance` case that distinguishes the two. `dd110` is tried first in every property cascade, so the legacy read plugs in where it should. One narrow follow-up on the new reader posted separately (DTD handling); not blocking. Resolving this.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzmlXmlIdValidity.cs
trishorts: Verified at `8fffa2b4`: `sourceFile/@name` now falls back to `SourceFile.UnknownName`, and the test reads `location`, `name` and `id` back with `XmlConvert.VerifyNCName` on the id. Resolving.

mzLib/Test/MzIdentML/MzIdentMlNamespaceTests.cs
trishorts: Verified at `8fffa2b4`: `inherit: false` on the `XmlTypeAttribute` lookups, `XmlRootAttribute` left alone. Resolving.

mzLib/Readers/MzML/MzmlMethods.cs
trishorts: Verified at `8fffa2b4`: both literals are `SourceFile.UnknownLocation` now, with `UnknownName` beside it. Resolving.

nbollis (dismissed):

mzLib/MzIdentML/MzidIdentifications.cs
trishorts: I went and measured this on net10.0 rather than leaving it as an assertion, and the result narrows it — it is a consistency point, not a security one. Withdrawing the security framing.

What reproduces:

```
XmlTextReader.DtdProcessing = Parse
plain-stream arm  (the other three): REJECTED -> XmlException:
    "For security reasons DTD is prohibited in this XML document."
legacy-reader arm (this one):        ACCEPTED the DTD, dd110 set
```

So the asymmetry is real: a document carrying an internal DTD subset is refused by the three `Deserialize(Stream)` arms and accepted by this one.

What does **not** reproduce — and this is the part I had implied:

```
external-dtd (SYSTEM id pointing at a nonexistent path):
    constructed without error, no fetch attempted
```

The external subset is never resolved, so there is no XXE/SSRF path here, and general entities stay unexpanded (`id="&xxe;"` came back literal) so there is no expansion-bomb path either. My original comment was right that DTD processing is on and wrong to imply that mattered.

All that is left is that one of four arms accepts input the other three reject. If you want them uniform, `DtdProcessing = DtdProcessing.Prohibit` in the `LegacyMzidNamespaceReader` constructor is a one-liner. Non-blocking either way — resolving this and approving.

trishorts (dismissed):
All five threads are now closed. The four substantive ones were fixed and verified at `8fffa2b4`:

- the legacy `.../1.1.0` namespace is remapped in the reader, which handles all depths — you were right that the `XmlRootAttribute` override I suggested was the worse fix, since it renames only the root and silently drops every child;
- `sourceFile/@name` falls back to `SourceFile.UnknownName`, and the test now reads `location`, `name` and `id` back with `XmlConvert.VerifyNCName`;
- the reflection lookups use `inherit: false`, so a deleted `[XmlType]` no longer passes by inheriting the base's;
- both `unknown-source` literals are `SourceFile.UnknownLocation`, with `UnknownName` beside it.

The fifth was my DTD note, which I have withdrawn — I measured it and the exploit paths do not reproduce on net10.0 (no external fetch, no entity expansion). What remains is only that this arm accepts an internal DTD subset the other three reject; `DtdProcessing.Prohibit` would make them uniform, but it is not worth holding the PR for.

The only commit since is `311ae1f7`, a clean base-only merge with no conflict resolution. CI is green. Approving.

trishorts:
@acesnik heads up — this now conflicts with master. #1169 merged (`165abaab`), and `update-branch` refuses with `422 merge conflict between base and head`, so the branch needs a manual merge before the button will light up. Both approvals are still standing, and a base-only merge does not dismiss them on this repo, so updating costs nothing.

One conflict, one file: `mzLib/MzIdentML/MzidIdentifications.cs`. Nothing is actually incompatible — you and #1169 each appended a new innermost arm to the same nested `try/catch` deserialization cascade, so git can't see that they're additive rather than competing.

I resolved it locally to confirm it's mechanical. Nesting them both, 1.3.0 first and the legacy-namespace 1.1.0 last:

```csharp
                    catch
                    {
                        try
                        {
                            using (Stream stream = new FileStream(mzidFile, FileMode.Open, FileAccess.Read, FileShare.Read))
                            {
                                XmlSerializer _indexedSerializer = new XmlSerializer(typeof(mzIdentML130.Generated.MzIdentMLType130));
                                // Read the XML file into the variable
                                dd130 = _indexedSerializer.Deserialize(stream) as mzIdentML130.Generated.MzIdentMLType130;
                            }
                        }
                        catch
                        {
                            // Last arm: an mzIdentML 1.1.0 document declaring the namespace mzLib itself used
                            // to write, ".../mzIdentML/1.1.0", instead of the schema's ".../mzIdentML/1.1".
                            // Every .mzID this library produced before that was corrected is in the old
                            // namespace, and XmlSerializer matches namespaces exactly, so without this arm
                            // all four attempts above fail and the constructor throws on files we wrote.
                            // Tried last: it is a bug-compatibility fallback, so a real 1.3.0 document must
                            // never reach it.
                            using (Stream stream = new FileStream(mzidFile, FileMode.Open, FileAccess.Read, FileShare.Read))
                            using (LegacyMzidNamespaceReader reader = new LegacyMzidNamespaceReader(stream))
                            {
                                XmlSerializer _indexedSerializer = new XmlSerializer(typeof(mzIdentML110.Generated.MzIdentMLType110));
                                dd110 = _indexedSerializer.Deserialize(reader) as mzIdentML110.Generated.MzIdentMLType110;
                            }
                        }
                    }
```

The ordering is the only judgement in it, and it is yours to overrule: 1.3.0 is a real schema version while the 1.1.0 arm is a bug-compatibility fallback for files this library mis-wrote, so a valid 1.3.0 document must never reach the legacy reader. Reversed, the legacy reader would get first refusal on them. Your comment's "all three attempts above" becomes "all four".

Verified on my end: builds clean, and 165/165 tests pass across the `MzIdentML`/`Mzid`/`Mzml` filters — both PRs' suites together, so `MzidIdentificationsTests` and `MzIdentMlReflectionCoverageTests` from #1169 alongside your `MzIdentMlLegacyNamespaceTests`, `MzIdentMlNamespaceTests` and `TestMzmlXmlIdValidity`.

I did not push it, since it is your branch and the ordering is a call you may want to make yourself. Happy to push the resolution if you would rather not hand-merge — just say so.

Also worth expecting: #1176 is the third of this stack and is likewise `BEHIND`, so it may want the same treatment once this one lands.

acesnik:
🤖 Merged master to clear the conflict. One file conflicted, `MzidIdentifications.cs`, and it turned out both sides were adding an arm to the same version cascade rather than disagreeing: #1169 added an mzIdentML 1.3.0 attempt, this branch added a fallback that rewrites the legacy `.../mzIdentML/1.1.0` root namespace mzLib used to emit.

Kept both, nested one level deeper, with 1.3.0 tried first and the legacy rewrite last. The ordering is the only real decision here — a namespace rewrite must not pre-empt a document that parses as a genuine format version, or a malformed 1.3.0 file would get silently reinterpreted as legacy 1.1.0 instead of failing.

Checked that neither arm shadows the other rather than assuming it. 26 tests across `MzIdentMlLegacyNamespaceTests`, `MzIdentMlNamespaceTests` and #1169's own `MzidIdentificationsTests` pass together. The two that carry the argument are `BothNamespacesAreReadable(".../1.1.0", "legacy.mzid")`, which shows the legacy arm is still reachable now that something precedes it, and `MzIdentML130_IsReadThroughEveryVersionCascade`, which shows the 1.3.0 arm still wins when it should.

Run through a throwaway net10.0 harness, since `Test.csproj` is `net*.0-windows` via `mzPlot` and cannot execute on macOS — so CI is the first real run of the full suite.

trishorts (dismissed):
Re-approving. My approval on `311ae1f7` was dismissed as stale by the master merge `84c0a0a1`, so I re-checked what that merge brought in rather than just re-stamping it.

It is not a no-op: master gained mzIdentML 1.3.0 (`mzIdentML1_3_0.cs`, and `MzidIdentifications.cs` +512/-135), and that is one of this PR's own files. The integration is correct — the legacy-namespace fallback is nested inside the 1.3.0 catch, so the arm order is 1.1 → 1.2 → 1.3 → legacy rewrite, and the rewrite still cannot pre-empt a document that parses as a real format version. That was the property I cared about in the original review and it survived the merge.

All five of my threads stay resolved and CI is green.

trishorts:
@acesnik — notes from bringing #1183 current, relevant here:

**The bug this fixes is live on #1183's head.** Taking `TestParentTolerance.mzid` and rewriting only its root `xmlns` to `http://psidev.info/psi/pi/mzIdentML/1.1` makes `new MzidIdentifications(path)` throw on every version arm:

```
InvalidOperationException: There is an error in XML document (2, 2).
  inner: <MzIdentML xmlns='http://psidev.info/psi/pi/mzIdentML/1.1'> was not expected.
```

No checked-in `.mzid` fixture declares that namespace (two are `/1.2`, two `/1.3`, one `/1.1.1`), and the 1.1.0 tests write their files through the generated types, so the existing suite round-trips the wrong namespace and cannot catch it.

**#1183 and this PR don't conflict.** A trial `git merge --no-commit` of this branch onto #1183 at `4de34406` completed automatically, even though both edit `MzidIdentifications.cs`. I'm reviewing this one next.

trishorts:
@acesnik — following up on my note above. Reviewed, and I pushed two small commits to the branch rather than leaving them as threads:

- **`12978417` — `DtdProcessing.Prohibit` on `LegacyMzidNamespaceReader`, plus the `MzIdentMLType130` → `.../1.3` rows.** This is the one-liner from my 08-26 DTD thread, and applying it showed the gap was wider than that thread said: a *compliant* `.../1.1` document carrying a DTD was also accepted, because the first four arms refuse it and the legacy arm's remap is a no-op for `.../1.1`. The new `ADocumentTypeDeclarationIsRejectedInEitherNamespace` failed in **both** namespaces before the fix. The 1.3 rows are the follow-up your description promised once #1169 merged.
- **`c8d7ee20` — master merged in, now including #1183.** Clean; the legacy arm populates `dd110`, which #1183's accessors check first.

```
mzIdentML + mzML tests   217 passed, 0 failed   (net10.0-windows, on c8d7ee20)
```

I also brought the description current: the 1.3 row in the namespace table, a section on the legacy-namespace arm and its ordering/DTD handling, the `name` fallback beside `location`, the actual #1169/#1183 history in place of "no conflict — disjoint files", and 29 cases across four fixtures in place of 21 across three.

Two things I left alone: `8fffa2b4` added a UTF-8 BOM to `Mzml.cs` and `MzmlMethods.cs` (a line-1 diff, harmless), and a file that no arm can parse now reports the legacy arm's exception rather than the one from its real version — noted in the description so a future bug report isn't misread.

PR #1173, opened 2026-08-16
Share the PRIDE pager, and close the understated total_records truncation

trishorts:
## The nut

`GetProjectFilesAsync` trusted `total_records` in both directions. An **understated** one truncated the tail: the loop stopped at `files.Count >= total` and every file past the reported total was dropped, with no error. That is the mirror of the capped-`pageSize` bug #1102 fixed, and @Alexander-Sol named it on that PR as the one truncation path that was neither documented nor pinned.

This PR closes it — and first moves the paging loop somewhere it only has to be closed once.

### 1. `refactor` — lift the pager into `GetAllPagesAsync<T>`

The loop was inline in `GetProjectFilesAsync`. PRIDE uses the same envelope (bare JSON array + `total_records` header) on more than one endpoint, and dataset search is about to need it, so it moves into a private generic helper rather than being copied.

Copying was the alternative, and this loop's history argues against it: #1102 needed two review rounds, and round one's own hardening *reintroduced* the silent manifest loss it was meant to prevent. A second copy is a second place for that to happen.

**Behavior-preserving.** Every termination rule keeps its order and its comment: empty-page stop, raw-body page identity, the identical-page `MzLibException` checked *before* the total comparison, `total_records` authoritative, short-page only as the no-header fallback, `MaxPages` backstop. **No test file is touched in this commit** — the existing 145 PRIDE offline tests and 8 live canaries pass unchanged, which is the proof. @acesnik verified this independently by running the same paging scenarios against base and the refactor separately; they agree on every one.

One deliberate text change: the identical-page message ended *"the project's file list may have changed mid-fetch"*, which would be wrong on a non-manifest endpoint; it now reads *"the result set"*. Everything else is verbatim — the accession still reaches the message, through a `subject` parameter, so each endpoint's failures stay as specific as they were.

### 2. `fix` — the tail probe

An understated total is only distinguishable by asking for one more page, so the loop asks — **but only when the answer is in doubt**. Truncation needs the last page to have been *full*: a server with more to give fills the page it is giving. So a page that is not full ends the fetch on the header's word, and a page that could be full buys one speculative request first.

### 3. `fix` — what the probe cost, and what it duplicated

Review (@acesnik) found three ways the commit above overstated itself. All three reproduce against a stub handler; all three are corrected here rather than reworded around.

**It fired on every single-page fetch.** `largestPageSeen` is updated before it is compared against, so "shorter than the largest page seen" is false on any page that *is* the largest — and page 0 always is. 50 files at the default `pageSize` of 100 cost two requests. Most PRIDE projects are single-page, so the stated cost was not merely wrong, it was backwards.

Fullness is now measured against different evidence on page 0 than afterwards, because the two pages carry different information:

```csharp
bool pageCouldBeFull = page == 0
    ? servedCount >= pageSize
    : servedCount >= largestPageSeen;
```

On a **later** page `largestPageSeen` is the only sound measure — PRIDE caps `pageSize` server-side, so a page of 100 against a requested 500 is full, and #1102 settled that the requested size cannot tell the two apart. On **page 0** there is no such evidence, so the requested `pageSize` decides: a first page shorter than what was asked for is the server running out. Exact whenever `pageSize` is at or below the cap, which is every default call. One gap remains, and only one — a single-page fetch requested *above* the cap comes back capped, so it looks short when it may be full. That is the same above-the-cap caveat the docstring already carries.

**It could duplicate a record.** A result set that grows between two requests repages itself: page 0 of a 2-record set is `[a, b]`, page 1 of the 3-record set it became is `[b, c]`. The server is correct throughout — the empty-page and identical-page comments both decline to fail on exactly this — but the probe appended anything not byte-identical, returning `[a, b, b, c]` where before the probe it returned `[a, b]`. A duplicate the caller cannot see is a worse failure than the truncation the probe exists to prevent, and unlike that truncation it lands on a *correct* server.

A probe's records are now matched against the page before them and the repeats dropped, on the **whole parsed JSON record** rather than any field of it. That distinction is the loop's own argument, not a loophole around it: a field is refused as identity because a value may repeat or be absent, so two *different* records can share one; a complete record cannot be confused with a different record that way. Two byte-equal records in one manifest are indistinguishable from each other anyway, so dropping one of a straddling pair costs nothing a caller could act on. Parsed rather than raw also means whitespace and property order no longer defeat it.

**"Never throw where it did not throw already" was too strong.** A server that ignores `page` and does not return byte-stable bodies never tripped the identical-page discard, so the fetch ran to the `MaxPages` backstop — 10,000 requests by default — where the total check used to stop it at page 0. The parsed record comparison recovers the whitespace and property-order cases. What is left is a server that ignores `page` **and** varies the *content* of the records it re-serves; nothing short of a record key the DTO does not carry would tell those pages apart. The claim is narrowed in the comment to say exactly that.

## What it costs now

| scenario | before this PR | after |
|---|---|---|
| 131 files, `pageSize` 100 | 2 requests, 131 files | unchanged |
| 50 files, `pageSize` 100 | 1 request | unchanged |
| record count an exact multiple of the page size | 1 request too few — silent truncation risk | +1 request |
| `total_records` understated (4 records reported as 2) | 2 of 4, no error | all 4 |
| manifest changes between requests | `[a, b]` | `[a, b, c]` |

The probe is spent only on a fetch that ends on a page that could be full. PXD012345 (131 files at the default page size) does not pay it; neither does any project holding fewer files than the page size.

## The test I flipped, and why it is not quiet

`GetProjectFilesAsync_TotalRecordsExactlyMatched_DoesNotFetchExtraPage` asserted exactly this behavior should not exist, and its own comment nominates it as the tripwire for weakening the `total_records` check. So it is flipped deliberately:

- renamed to `..._ProbesOncePastTheReportedTotal`, request count `1` → `2`;
- it **keeps** the ignores-`page` stub, because that is the probe's worst case — the probe walks straight into the identical-page guard, and returning the two files rather than throwing is precisely what makes the probe non-regressive;
- it now also asserts the repeated page is **discarded, not appended**.

Five tests added, each verified RED against the loop it pins:

- `..._TotalRecordsUnderstated_ReturnsEveryRecordAnyway` — 2 of 4 files in 1 request before. Asserts the request count too, so it discriminates: the file list alone would also pass if the loop ignored `total_records` entirely, which is not the fix.
- `..._TotalRecordsMatchedOnAShortPage_DoesNotProbe` — the ordinary multi-page fetch pays nothing.
- `..._SinglePageShorterThanRequested_DoesNotProbe` — 50 files at the default `pageSize` of 100, 1 request. This is the shape most callers hit, and nothing exercised it before, which is why the cost claim went unchallenged.
- `..._ResultSetGrowsBeforeTheProbe_DoesNotDuplicateARecord` — `[a, b, c]`, not `[a, b, b, c]`.
- `..._ProbeRepeatsARecordFormattedDifferently_StillDropsIt` — reordered properties and inserted whitespace are still recognised as a repeat.

## Rejected alternatives

- **A docstring caveat instead of a guard** (the lighter option @Alexander-Sol offered). Rejected because the loop now has exactly one home, which is the cheapest this fix will ever be, and because a documented silent truncation is still a silent truncation.
- **Always probing / paging until an empty page.** Correct, and it makes `total_records` unnecessary for termination — but it costs one extra request on *every* call, including the single-page common case, and changes the request count in most existing tests.
- **Learning the effective page size from page 0** for termination. Already rejected in #1102: it cannot distinguish a capped page from a genuinely short last page. Note `largestPageSeen` here decides only whether to *probe*, never when to terminate — a misjudgment costs a request, not a record.
- **Keying the dedup on a record field** (`fileName`, `accession`). Refused for the same reason the identical-page guard refuses it, and refused again here: a value may legitimately repeat or be absent, so it cannot carry a record's identity. The whole record can.
- **Guarding the dedup against an unparseable body or a mismatched length.** Both are unreachable through the public API — the caller reaches that line only by having deserialized the body element-for-element, on a page that follows another. Written as an invariant in the remarks rather than shipped as uncoverable branches, the same call made on the null-`Key` guard in #1102.

## Verification

- Full offline suite **5741 passed / 0 failed** (29 skipped)
- PRIDE offline **150**, live PRIDE **8/8** against the real service
- No public API change; no existing caller behaves differently

Stacked work: dataset search (`GET /search/projects`) consumes `GetAllPagesAsync<T>` and will name this PR.

### DIFF
--- a/mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
+++ b/mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
@@ -67,6 +67,31 @@ public class PrideArchiveClientTests
 
     private static string Array(params string[] fileJson) => "[" + string.Join(",", fileJson) + "]";
 
+    /// <summary>
+    /// The same record <see cref="FileJson"/> produces, written with its properties in a different
+    /// order and different whitespace — the same record to anything that parses it, a different one to
+    /// anything that compares bytes.
+    /// </summary>
+    private static string FileJsonReordered(string fileName) =>
+        $$"""
+        {   "totalDownloads": 19,
+            "fileName": "{{fileName}}",
+            "additionalAttributes": [], "compress": false,
+            "updatedDate": "2019-01-15T09:47:55.000+00:00",
+            "publicationDate": "2025-07-13T23:01:04.308+00:00",
+            "submissionDate": "2019-01-15T09:42:57.000+00:00",
+            "fileSizeBytes": 96358400,
+            "publicFileLocations": [
+              { "@type": "CvParam", "cvLabel": "PRIDE", "accession": "PRIDE:0000469", "name": "FTP Protocol", "value": "ftp://ftp.pride.ebi.ac.uk/{{fileName}}" },
+              { "@type": "CvParam", "cvLabel": "PRIDE", "accession": "PRIDE:0000468", "name": "Aspera Protocol", "value": "prd_ascp@fasp.ebi.ac.uk:{{fileName}}" }
+            ],
+            "checksum": "",
+            "fileCategory": { "value": "SEARCH", "name": "category", "accession": "PRIDE:0000408", "cvLabel": "PRIDE", "@type": "CvParam" },
+            "accession": "hashid",
+            "projectAccessions": ["PXD012345"]
+        }
+        """;
+
     // ---- deserialization ----------------------------------------------------
 
     [Test]
@@ -319,6 +344,84 @@ public class PrideArchiveClientTests
             Assert.That(files.Select(f => f.FileName), Is.EqualTo(new[] { "a", "b", "c" }));
             Assert.That(handler.RequestedUris.Count, Is.EqualTo(2), "a short final page needs no probe");
         });
+    }
+
+    /// <summary>
+    /// The single-page fetch is the common case -- most PRIDE projects hold fewer files than the
+    /// default page size -- so it is the one the probe must not tax. Page 0 is trivially the largest
+    /// page seen, so "as full as any served" is vacuously true there and would probe every such fetch;
+    /// the requested pageSize is what decides it instead. 50 of a requested 100 is the server running
+    /// out, not a full page, so the header is believed and nothing further is requested.
+    /// </summary>
+    [Test]
+    public async Task GetProjectFilesAsync_SinglePageShorterThanRequested_DoesNotProbe()
+    {
+        string[] names = Enumerable.Range(0, 50).Select(i => "f" + i).ToArray();
+        var handler = new StubHandler(request => request.RequestUri.ToString().Contains("page=0")
+            ? JsonResponse(Array(names.Select(n => FileJson(n)).ToArray()), totalRecords: 50)
+            : JsonResponse("[]", totalRecords: 50));
+        using var client = new PrideArchiveClient(new HttpClient(handler));
+
+        var files = await client.GetProjectFilesAsync("PXD012345"); // the default pageSize of 100
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(files, Has.Count.EqualTo(50));
+            Assert.That(handler.RequestedUris.Count, Is.EqualTo(1),
+                "50 files at a requested page size of 100 is a short page; the probe is not for it");
+        });
+    }
+
+    /// <summary>
+    /// A result set that GROWS between two requests repages itself, so the probe can be handed records
+    /// the fetch already holds: page 0 of a 2-record set is [a, b], and page 1 of the 3-record set it
+    /// became is [b, c]. The server is behaving correctly throughout -- the empty-page and
+    /// identical-page comments both decline to fail on exactly this -- so the answer cannot be to
+    /// throw, and it cannot be to append either: that hands the caller a duplicate "b", which it has no
+    /// way to see, and which a caller acts on by downloading the same file twice or reporting the wrong
+    /// count. The repeat is dropped by matching the whole JSON record, and only "c" is appended.
+    /// </summary>
+    [Test]
+    public async Task GetProjectFilesAsync_ResultSetGrowsBeforeTheProbe_DoesNotDuplicateARecord()
+    {
+        var handler = new StubHandler(request =>
+        {
+            string uri = request.RequestUri.ToString();
+            if (uri.Contains("page=0")) return JsonResponse(Array(FileJson("a"), FileJson("b")), totalRecords: 2);
+            if (uri.Contains("page=1")) return JsonResponse(Array(FileJson("b"), FileJson("c")), totalRecords: 3);
+            return JsonResponse("[]", totalRecords: 3);
+        });
+        using var client = new PrideArchiveClient(new HttpClient(handler));
+
+        var files = await client.GetProjectFilesAsync("PXD012345", pageSize: 2);
+
+        // Not [a, b, b, c]. The record the probe repeats is dropped; the one it genuinely adds is kept,
+        // so the probe still does the job it exists for on a result set that moved underneath it.
+        Assert.That(files.Select(f => f.FileName), Is.EqualTo(new[] { "a", "b", "c" }));
+    }
+
+    /// <summary>
+    /// The repeat is matched on the whole record as PARSED JSON, not as raw bytes, so a server that
+    /// re-serves the same record with different whitespace or a different property order is still
+    /// recognised. Byte identity alone would miss it and append the duplicate.
+    /// </summary>
+    [Test]
+    public async Task GetProjectFilesAsync_ProbeRepeatsARecordFormattedDifferently_StillDropsIt()
+    {
+        var handler = new StubHandler(request =>
+        {
+            string uri = request.RequestUri.ToString();
+            if (uri.Contains("page=0")) return JsonResponse(Array(FileJson("a"), FileJson("b")), totalRecords: 2);
+            // Same two records, reformatted: whitespace inserted and the properties reordered.
+            if (uri.Contains("page=1"))
+                return JsonResponse("[\n  " + FileJson("a") + " ,\n  " + FileJsonReordered("b") + "\n]", totalRecords: 2);
+            return JsonResponse("[]", totalRecords: 2);
+        });
+        using var client = new PrideArchiveClient(new HttpClient(handler));
+
+        var files = await client.GetProjectFilesAsync("PXD012345", pageSize: 2);
+
+        Assert.That(files.Select(f => f.FileName), Is.EqualTo(new[] { "a", "b" }));
     }
 
     [Test]
--- a/mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
+++ b/mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
@@ -229,15 +229,22 @@ public class PrideArchiveClientTests
         });
     }
 
+    /// <summary>
+    /// The counterpart to the overstated case, and the pin on what the tail probe costs. The fetch ends
+    /// on a page as full as any served, which is indistinguishable from a total that is one page short,
+    /// so one further page is requested before the header is believed.
+    /// </summary>
+    /// <remarks>
+    /// The stub ignores <c>page</c> and serves the same two files forever, never an empty page. That is
+    /// deliberate and does double duty: it makes total_records the only thing that can stop this loop,
+    /// so weakening that check shows up here; and it is the worst case for the probe, because the probe
+    /// walks straight into the identical-page guard. Returning the two files rather than throwing is
+    /// what makes the probe strictly non-regressive -- a server that ignores paging behaves exactly as
+    /// it did before the probe existed.
+    /// </remarks>
     [Test]
-    public async Task GetProjectFilesAsync_TotalRecordsExactlyMatched_DoesNotFetchExtraPage()
+    public async Task GetProjectFilesAsync_TotalRecordsExactlyMatched_ProbesOncePastTheReportedTotal()
     {
-        // The counterpart to the overstated case. Note the stub ignores `page` and serves the same
-        // two files forever, never an empty page: that makes the total_records check the ONLY thing
-        // that can stop this loop on the first request, so if it is ever weakened this test does not
-        // fail by one wasted request -- the second request re-serves an identical page and trips the
-        // identical-page guard instead. PagesUntilTotalRecords cannot pin that, because its fallback
-        // returns [] and the empty-page check would rescue it.
         const long total = 2;
         var handler = new StubHandler(_ => JsonResponse(Array(FileJson("a"), FileJson("b")), totalRecords: total));
         using var client = new PrideArchiveClient(new HttpClient(handler));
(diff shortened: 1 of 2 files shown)


mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -706,6 +623,168 @@ private static void RemoveNullElements(PrideProject project)
...
+                // downward. So it must never be able to make things worse than trusting the header
+                // would have been. A probe that brought nothing new ends the fetch and lets the header
+                // stand -- deliberately NOT the identical-page throw, which is reserved for a server
+                // contradicting a total that says more is still to come. A server that ignores `page`
+                // entirely therefore still returns its records rather than throwing, exactly as it did
+                // before this guard existed.
+                if (verifyingTail && !pageAdvanced)
+                    break;
+
+                items.AddRange(pageItems);
```
acesnik, line 764: Appends the probe page unconditionally once it's byte-different, which duplicates records when the manifest changes between requests — the case the comment above declines to throw on.

`page=0` → `[a, b]`, `total_records: 2`; a file is inserted; `page=1` → `[b, c]`. The server pages correctly throughout, so nothing here is a fault. Returns `[a, b, b, c]`; before the probe this request was never made and it returned `[a, b]`.

The flipped test's "discarded, not appended" holds only for a byte-identical repeat. A duplicate manifest entry means a caller can download the same file twice or report the wrong count.

No clean fix to offer — byte identity can't see an overlap, and keying on a record field is what this loop rightly refuses to do.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -706,6 +623,168 @@ private static void RemoveNullElements(PrideProject project)
...
+
+                // A tail probe (see the total_records branch below) is speculative: total_records has
+                // already said the fetch is done, and this page is only being read in case it lied
+                // downward. So it must never be able to make things worse than trusting the header
+                // would have been. A probe that brought nothing new ends the fetch and lets the header
+                // stand -- deliberately NOT the identical-page throw, which is reserved for a server
+                // contradicting a total that says more is still to come. A server that ignores `page`
+                // entirely therefore still returns its records rather than throwing, exactly as it did
+                // before this guard existed.
+                if (verifyingTail && !pageAdvanced)
```
acesnik: This discard only fires on a byte-identical page, so a page-ignoring server whose body isn't byte-stable (whitespace, field order, the embedded timestamp raised a few lines up) never trips it. `verifyingTail` stays set, `pageAdvanced` stays true, and the loop runs to the `MaxPages` backstop and throws — where before, the total check stopped it at page 0 and returned the records. `MaxPages` defaults to 10000, so that's 10,000 requests to PRIDE before the exception.

Worth softening "never throw where it did not throw already" either way.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -706,6 +623,168 @@ private static void RemoveNullElements(PrideProject project)
...
+                    // believing the header.
+                    //
+                    // The probe cannot cost correctness, only a single request: it may add records the
+                    // header disclaimed, and the guard above makes an empty or repeated answer mean
+                    // "the header was right after all". The one request is spent only when a fetch ends
+                    // exactly on a full page boundary, which for an honest server means the record
+                    // count is an exact multiple of the page size.
+                    if (items.Count >= total)
+                    {
+                        if (pageItems.Count < largestPageSeen)
```
acesnik: `largestPageSeen` is updated on line 719, before this compares against it, so this is false on any page that is the largest seen — which page 0 always is. The probe fires on **every** single-page fetch, not just an exact page-boundary one: 50 files at the default `pageSize` of 100 costs 2 requests, and 50 isn't a multiple of 100.

The code looks right — `pageSize` can't be used here, exactly as #1102 established, since a capped page is short but not final. It's the stated cost that's off: with the default at 100, most PRIDE projects are single-page, so this is a doubling on the common path.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
```diff
@@ -100,6 +100,11 @@ private PrideArchiveClient(HttpClient httpClient, bool ownsHttpClient)
         /// page against, and termination falls back to the first short page — under which a requested
         /// size above the server cap can still return a partial manifest. Leave this at the default
         /// unless you have a reason not to; the full manifest is returned either way at or below the cap.
+        /// <para>
+        /// A <c>total_records</c> that misreports the count in either direction is tolerated: an
+        /// overstated one stops on the empty page past the end, and an understated one is caught by
+        /// fetching one further page whenever the fetch would otherwise end on a full page.
```
acesnik: "whenever the fetch would otherwise end on a full page" — it's whenever the last page is no smaller than the largest seen, which includes every single-page fetch regardless of size. See the note on the `largestPageSeen` comparison.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
```diff
@@ -246,8 +253,71 @@ public async Task GetProjectFilesAsync_TotalRecordsExactlyMatched_DoesNotFetchEx
...
+            return JsonResponse("[]", totalRecords: total);
+        });
+        using var client = new PrideArchiveClient(new HttpClient(handler));
+
+        var files = await client.GetProjectFilesAsync("PXD012345", pageSize: 2);
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(files.Select(f => f.FileName), Is.EqualTo(new[] { "a", "b", "c" }));
+            Assert.That(handler.RequestedUris.Count, Is.EqualTo(2), "a short final page needs no probe");
```
acesnik, line 345: This pins the multi-page no-probe path, and `..._ProbesOncePastTheReportedTotal` uses `pageSize: 2` with 2 files, so between them they only cover the exact-multiple case. Nothing pins the single-short-page cost — 50 files at `pageSize: 100`, say — which is what most callers will actually pay.

acesnik (commented):
🤖 Verified the refactor independently rather than from the diff: same paging scenarios run through a stub handler against base (`5ba1315`), the refactor (`4d51eb4`) and the fix (`0b536e7`). Base and refactor agree on every one, so the behaviour-preserving claim holds. The understated-total fix works — 4 records reported as 2 returns all four.

Three things from the fix commit, inline. The first is the one I'd want resolved before merge: the probe appends any page that isn't **byte-identical** to its predecessor, and two cases the loop's own comments promise to tolerate aren't byte-identical.

| scenario | before | after |
|---|---|---|
| 131 files, `pageSize` 100 | 2 requests, 131 files | unchanged ✅ |
| 4 files, `total_records: 2` | 2 of 4 | **all 4** ✅ |
| 50 files, `pageSize` 100 | 1 request | **2 requests** |
| manifest changes between requests | `[a, b]` | **`[a, b, b, c]`** |
| page-ignoring server, unstable body | `[a, b]`, 1 request | **throws after `MaxPages`** |

Not verified: whether PRIDE's responses are byte-stable across repeated requests, or how often its manifests change mid-fetch — both need live probing I didn't do. Didn't run the suite locally (`Test` is `net10.0-windows`); CI is green and I took that at face value.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Confirmed, and thank you — this is the one I'd have shipped without noticing. Reproduced it before touching anything: 50 files at the default `pageSize` of 100 cost **2 requests**. So the stated cost wasn't merely imprecise, it was backwards — the exact-multiple case is rare and the case that actually pays is the common one.

You're also right that `pageSize` can't replace `largestPageSeen` here, for the reason you give. But the two pages carry *different* information, and that's what the fix leans on:

- On a **later** page, `largestPageSeen` is the only sound measure — a page of 100 against a requested 500 is full, and #1102 settled that the requested size can't tell that from a genuinely short last page. Unchanged.
- On **page 0** there is no such evidence: the first page is trivially the largest seen, so the test is vacuously true. The requested `pageSize` is the only evidence there is, so it's what decides — a first page shorter than what was asked for is the server running out.

```csharp
bool pageCouldBeFull = page == 0
    ? servedCount >= pageSize
    : servedCount >= largestPageSeen;
```

That's exact whenever `pageSize` is at or below the server cap, i.e. every default call. It leaves one gap and only one: a single-page fetch requested *above* the cap comes back capped, so it looks short when it may be full, and an understated total would still truncate it. That's the same above-the-cap caveat the docstring already carries, and it buys back a request on every ordinary call — which given your 50-of-100 example is the right trade.

Pinned by `..._SinglePageShorterThanRequested_DoesNotProbe`, verified RED against the previous loop (2 requests).

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Confirmed exactly as you wrote it — `[a, b]` + an insert + `[b, c]` returns `[a, b, b, c]`, where before the probe the request was never made and it returned `[a, b]`. Agreed this is the one to resolve before merge: a duplicate the caller can't see is a worse failure than the truncation the probe exists to prevent, and it lands on a *correct* server, which the truncation doesn't.

There is a fix, and I think it's inside the line the loop already draws. The probe's records are now matched against the page before them and the repeats dropped — matching on the **whole JSON record, parsed**, not on any field of it:

```csharp
if (verifyingTail)
{
    if (!pageAdvanced)
        break;
    NullOutRecordsRepeatedFromPreviousPage(pageItems, content, previousPageBody);
    if (pageItems.All(x => x is null))
        break; // the probe held nothing this fetch did not already have
}
```

The distinction from "keying on a record field" is the loop's own argument, not a loophole around it: a field is refused as identity because a value may repeat or be absent, so two *different* records can share one. A complete record can't be confused with a different record that way. And two byte-equal records in one manifest are indistinguishable from each other anyway, so dropping one of a straddling pair costs nothing a caller could act on.

Parsed rather than raw also buys the whitespace/property-order case from your other comment. Your scenario now returns `[a, b, c]` — the repeat dropped, the record the probe genuinely found still kept, so the probe still does its job on a result set that moved underneath it.

Two tests, both verified RED against the previous loop: `..._ResultSetGrowsBeforeTheProbe_DoesNotDuplicateARecord` and `..._ProbeRepeatsARecordFormattedDifferently_StillDropsIt`.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Confirmed — I reproduced it with a stub that re-serves the same two records with per-response whitespace variance, and it runs to the backstop and throws where the total check used to stop it at page 0. 10,000 requests at the default is not a cost worth hiding behind a sentence.

Partly recovered, partly narrowed:

- The record comparison added for the duplicate finding compares **parsed** records, so it sees past whitespace and property order. Both of those cases now dedup to nothing, the fetch ends, and the header stands — no throw, records returned, same as before the probe.
- What's left is a server that ignores `page` **and** varies the *content* of the records it re-serves (your embedded-timestamp example, if the timestamp is per-record rather than per-envelope). Those pages are new by every measure available here, and nothing short of a record key the DTO doesn't carry would tell them apart.

So you're right that the claim needed softening rather than defending, and I've narrowed it in the comment instead of removing it, since what survives is worth stating precisely:

> One case is NOT recovered, and the claim is limited to match: a server that both ignores `page` AND varies the CONTENT of the records it re-serves (a per-record timestamp, say) produces pages that are new by every measure available here, so the fetch runs to the MaxPages backstop and throws where the total check used to stop it. That server already defeated the identical-page guard above before this probe existed; what is new is that the total check no longer rescues it.

Same correction made in the PR body, which claimed this outright.

mzLib/UsefulProteomicsDatabases/PrideArchiveClient.cs
trishorts: Right — the docstring inherited the same wrong claim as the PR body. Rewritten to describe what the code does, including the case it can't catch:

> A `total_records` that misreports the count in either direction is tolerated: an overstated one stops on the empty page past the end, and an understated one is caught by fetching one further page whenever the fetch would otherwise end on a page that could be full — a first page holding everything that was asked for, or a later page as large as the largest the server has served. That costs one extra request per fetch of exactly that shape; a fetch ending on a page that is visibly short pays nothing. The one case it cannot catch is an understated total on a single page requested ABOVE the server cap, where the capped page that comes back looks short but may not be — the same above-the-cap caveat as the paragraph above.

mzLib/Test/DatabaseTests/PrideArchiveClientTests.cs
trishorts: Agreed, and the gap in the tests is exactly why the cost claim went unchallenged — nothing exercised the shape most callers actually hit, so nothing contradicted the PR body.

Added `..._SinglePageShorterThanRequested_DoesNotProbe`: 50 files at the default `pageSize` of 100, asserting **1 request**. Verified RED against the previous loop, where it was 2. It's your example, more or less literally.

trishorts:
@acesnik — all three reproduce. I rebuilt each scenario against a stub handler before changing anything rather than reasoning from the diff, and every one behaved as you described. Pushed `f6d6e6a` on top.

Between them they falsify three of this PR's central claims: *"one request only on an exact multiple"*, *"cannot cost correctness, only a request"*, and *"never throw where it did not throw already"*. The PR body and the docstring have been corrected rather than reworded around, and the loop is fixed where fixing it was possible.

| scenario | before | after `f6d6e6a` |
|---|---|---|
| 131 files, `pageSize` 100 | 2 requests, 131 files | unchanged |
| 4 files, `total_records: 2` | 2 of 4 | all 4 |
| **50 files, `pageSize` 100** | **2 requests** | **1 request** |
| **manifest changes between requests** | **`[a, b, b, c]`** | **`[a, b, c]`** |
| page-ignoring server, unstable **whitespace / field order** | throws after `MaxPages` | returns `[a, b]`, no throw |
| page-ignoring server, unstable **record content** | throws after `MaxPages` | unchanged — see below |

**The cost.** `page == 0` now measures fullness against the requested `pageSize` instead of `largestPageSeen`, which on the first page is vacuously satisfied. `largestPageSeen` still governs every later page, for the reason you gave and #1102 established. The gap this leaves is a single-page fetch requested *above* the server cap — the same above-the-cap caveat the docstring already carried.

**The duplicate.** A probe's records are matched against the page before them and repeats dropped, on the whole parsed JSON record rather than any field of it. That's the loop's own distinction: a field is refused as identity because a value may repeat or be absent, so two different records can share one; a complete record can't be confused with a different record that way.

**The throw.** Parsed matching recovers the whitespace and field-order cases. A server that ignores `page` *and* varies record content still runs to `MaxPages`; nothing short of a record key the DTO doesn't carry would separate those pages. The comment now says that instead of claiming otherwise.

Three tests added, each verified RED against the previous loop. The two defensive branches the dedup wanted — unparseable body, mismatched length — are unreachable through the public API, so they're stated as an invariant in the remarks rather than shipped as uncoverable lines.

On the two you flagged as unverified: I haven't probed PRIDE for byte-stability across repeated requests or for how often a manifest changes mid-fetch either, and the fix is built so neither answer matters — both cases are handled without knowing the frequency. Suite re-run locally: 5741 offline / 0 failed, PRIDE offline 150, live PRIDE 8/8 against the real service.

acesnik (approved):
🤖 Reproduced all three fixes against a stub handler: the single-page tax is gone (1 request for 50 files at `pageSize` 100), the mid-fetch overlap now returns `[a, b, c]` instead of `[a, b, b, c]`, the whitespace-variance case returns its records rather than throwing, and moving `RemoveAll` after the dedup keeps `currentRecords[i]` aligned — a `null` element on the probe page no longer costs a record. One minor thing, worth a line in the comment rather than a change: a byte-equal duplicate that straddles the probe boundary is now dropped, so whether it survives depends on where the page boundary happens to fall.

nbollis (approved):

PR #1174, opened 2026-08-16
Build an SDRF-Proteomics file from a search

trishorts:
# Build an SDRF-Proteomics file from a search

Fifth and last of the SDRF stack. #1138 gave mzLib the ability to read, write, validate and pool
SDRF-Proteomics files. This adds the ability to **author** one from mzLib-native inputs, plus a
coverage metric that says how much a document actually says.

**Counterpart PR: `smith-chem-wisc/MetaMorpheus` — the SDRF call site (not yet open; will be linked
here).** That is the only consumer, and it is why this API exists in this shape. Without that link
the new public surface here reads as dead code.

## What it adds

**`SdrfBuilder.Build(rows, options)` → `SdrfDocument`.** Takes `Tolerance`, `DigestionAgent`,
`Modification`, `DissociationType`, `CvParam` — nothing application-specific — so a caller passes
what it already holds rather than translating. That is what keeps the MetaMorpheus side a ~10-line
call rather than a second implementation.

Two rules govern every value it writes:

- **Terms are resolved, never copied.** Accessions come from the pinned vocabularies added in #1139,
  not from whatever an input document happened to say. The curated corpus is full of drift a copying
  builder would launder — bare `4` for `UNIMOD:4` in 39 documents, `Unimod:35` in 22, and two files
  that give one instrument a different accession on every row.
- **It will not invent a sample fact.** If a caller cannot supply a required sample value, the build
  fails rather than writing `not available` — unless the caller explicitly opts out via
  `SdrfBuilderOptions.RequireSampleMetadata`, which is what a caller does when a run is already
  under way and failing would throw away a finished search.

**`SdrfCoverage`** reports per-column fill rate with the SDRF reserved words counted as **unfilled**,
plus a distinct-value count so a column filled with one repeated value cannot look healthy. Without
it an SDRF that is 90% `not available` produces zero validation errors and zero drift findings, and
reads as flawless: the validator treats reserved words as the spec-correct way to state an absence,
and the drift lint skips them outright.

The curated corpus gives the benchmark to hold output to — every mandatory column is 80–100% filled
there, with `characteristics[organism part]` weakest at 80% and 642 distinct values.

## Public API — one line per type, per the by-consumer rule

Every type below is called from `PostSearchAnalysisTaskSdrf.cs` in the counterpart MetaMorpheus PR.
Nothing is public without a named cross-assembly caller.

| Type | Called by |
|---|---|
| `SdrfBuilder` | `WriteSdrf()` — `SdrfBuilder.Build(rows, options)` |
| `SdrfBuilderOptions` | same call site — sets `Software`, `SoftwareVersion`, `RequireSampleMetadata` |
| `SdrfRowInput` | `BuildSdrfRows()` — yields one per spectra file |
| `SdrfSample` | `BuildSdrfRows()` — the sample half of each row |
| `SdrfAssay` | `BuildSdrfRows()` — the assay half of each row |
| `SdrfCoverage` | `ReportSdrfCoverage()` — `SdrfCoverage.Uninformative(collection)` |
| `SdrfColumnCoverage` | same — the returned rows, read for `.Column` and `.FillRate` |

`SdrfReserved` is **internal**: it is referenced three times, all inside `SdrfBuilder.cs`, and by
nothing outside this assembly. The reserved words are fixed by the specification rather than by this
library, so a public constant would imply mzLib is the authority on their spelling.

## A note on one behaviour, because it looks like a bug and is not

`SdrfBuilder` always emits the characteristics columns SDRF requires, even when the caller supplies
no value for them — writing the reserved word rather than omitting the column.

**Absent is strictly worse than empty, because absent is invisible.** `SdrfCoverage` measures the
fill rate of columns that *exist*, so it has nothing to report for a column that is not there; the
document reads as complete exactly where it says least. Empty is neither invisible nor wrong: the
specification marks `characteristics[organism part]` required but explicitly permits the reserved
words, and about a fifth of the cells in the curated corpus are one.

This was found by the counterpart MetaMorpheus work, not by this library's own tests, and the reason
is worth stating: `TestSdrfBuilder`'s fixture has always populated organism part, so every test here
proved the builder writes a column it was **handed** — never that it writes one it was **not**. A
fixture that always supplies the input cannot test the caller that does not. Both cases are now
pinned.

## Testing

`+1,130 / -0` across 5 files, of which **492 lines are tests**.

- Full suite **5,758 passed / 0 failed / 30 skipped** on `net10.0`.
- 97 SDRF tests, 22 of them new here.
- Behaviours pinned: terms resolved rather than copied; a required characteristic emitted whether or
  not supplied; the strict path still refusing an unsupplied value; zero-based replicates rejected;
  an unresolvable dissociation type refused under the strict path and reserved-worded under the
  lenient one; ragged documents impossible (multi-cardinality columns sized to the widest row).

## Not in scope

- `comment[label]` for isobaric labelling. SDRF wants one row per sample per channel and MetaMorpheus
  has no channel-to-sample mapping; guessing one would invent an experimental design. The column is
  left unresolved and both a warning and the coverage report say so.
- Sample vocabularies (UBERON, MONDO, CL). `SdrfSample.Characteristics` takes already-resolved
  `CvParam`s, so whoever supplies the sample metadata supplies the terms with it. Embedding a
  corpus-scoped subset is scoped separately.


nbollis (dismissed):

acesnik (commented):
🤖 Checked this against the SDRF-Proteomics specification (`bigbio/proteomics-sample-metadata`, `sdrf-proteomics/README.adoc`) as well as the surrounding SDRF stack. One finding changes every modification cell this writes; the rest are small.

## `comment[modification parameters]` writes the motif into `NT=`

`SdrfBuilder.cs:269` takes the name from `mod.IdWithMotif`. `Modification`'s constructor sets `IdWithMotif = OriginalId + " on " + Target` whenever a target is present (`Omics/Modifications/Modification.cs:78`), so every cell this authors reads:

```
NT=Oxidation on M;AC=UNIMOD:35;TA=M;MT=Variable
```

The spec's two examples for that column are:

```
NT=Oxidation;MT=Variable;TA=M;AC=Unimod:35
NT=Carbamidomethyl;TA=C;MT=fixed;AC=UNIMOD:4
```

`NT=` is the bare modification name and the residue is carried by `TA=`, so the motif is both duplicated and in the wrong field. `SdrfCell.cs:8` already documents the target form — `NT=Oxidation;AC=UNIMOD:35;TA=M;MT=Variable` — so the builder is the one thing here that doesn't produce it. `mod.OriginalId` is the right source; the `?? mod.OriginalId` fallback currently only fires for a target-less modification, which isn't what MetaMorpheus passes.

Worth noting why the suite doesn't catch it: the other columns are pinned exactly (`Is.EqualTo("NT=Q Exactive;AC=MS:1001911")`), but both modification assertions are `Does.Contain("AC=UNIMOD:4")`, which passes either way. An exact-form assertion alongside the fix would close that.

## `comment[sdrf version]` can emit `vV1.1.0`

`SdrfBuilder.cs:197` — `"v" + options.SdrfVersion.TrimStart('v')` only strips lowercase. The spec asks for `vMAJOR.MINOR.PATCH` (`v1.1.0`), and nothing downstream validates the column, so `SdrfVersion = "V1.1.0"` goes out malformed and silent.

## Tolerances are culture-formatted

`SdrfBuilder.cs:281` — `$"{tolerance.Value} ppm"`. The spec's grammar for these columns is a numeric value plus a unit (`10 ppm`, `0.02 Da`, `20 mmu`); on a `de-DE` machine this writes `0,02 Da`. `Positive` at `:319` already uses `InvariantCulture`, so this reads as an omission rather than a decision. `TolerancesAreWrittenAsValueAndUnit` would fail there too.

## Coverage counts repeated columns per cell, not per row

`SdrfCoverage.cs:100` — `rows[column]++` sits inside the loop over header *positions*, so a column appearing N times gets a denominator of rows × N. That collides with a behaviour this PR pins on purpose: multi-cardinality columns are sized to the widest row, the padding is `SdrfReserved.NotApplicable`, and `IsAnswer` counts reserved words as unfilled. Four rows where one file has 8 modifications and three have 2 → 8 `comment[modification parameters]` columns, 14 of 32 cells real → FillRate 0.44 → reported uninformative, although every modification searched is recorded.

## The lenient path can still throw

`SdrfCell.ToCell` throws `ArgumentException` when name and accession are both empty (`SdrfCell.cs:119`), so a non-null `Instrument` whose name and accession are both empty gets past the `is null → Missing` guard at `SdrfBuilder.cs:224` and throws with `RequireSampleMetadata = false` — the option whose purpose is not throwing away a finished search. Same shape at `:240` for a `DigestionAgent` with an empty `Name`. I confirmed the throw is reachable in principle; I did not confirm any reader actually yields such an instrument, so this may be theoretical.

## Smaller

- `comment[label]` is a required column in the spec and the isobaric case is out of scope here by design — but `Missing` gates it through the same `RequireSampleMetadata` switch as the sample columns, so a caller that legitimately can't resolve the label has to disable the organism guard too, and the throw text says "sample metadata" when the label or dissociation method is what's missing.
- `Build` null-checks `rows`, then dereferences `r.Assay.FixedModifications` and `r.Sample.Characteristics` unguarded (`:93`), so a null `Assay` or `Characteristics` surfaces as `NullReferenceException` from a LINQ lambda rather than the `ArgumentException` the entry point otherwise establishes.
- `values[column]` is `StringComparer.Ordinal` while `IsAnswer` compares case-insensitively (`SdrfCoverage.cs:106`), so `"liver"` and `"Liver"` count as two distinct values — the case the distinct count exists to catch.

## Checked and correct

Recording these because they're the parts a reviewer would otherwise assume: `SdrfValidator.ReservedWords` matches the spec exactly (four words, lowercase); `ToCell` emits `NT` then `AC` then extras, which follows the spec's "key order MUST be `NT` (name) first, followed by `AC` (accession), then any additional keys" and is stricter than the spec's own examples, which put `MT`/`TA` ahead of `AC`; and the tolerance columns are value-plus-unit rather than the MS:1001412/1001413 pair.

Not verified: I couldn't run the suite (`Test` targets `net*.0-windows`), so I'm taking 5,758/0 on trust — every finding above is one a passing suite wouldn't catch. I also didn't check output against the curated corpus, and the counterpart MetaMorpheus PR isn't open yet, so I haven't seen the call site.

trishorts:
Thanks — this was a good catch list, and the framing was the useful part: every one of these is something a green suite would not have caught, which is itself the finding. All eight are fixed in 5b3039dc.

**Each fix has a test that fails against the code you reviewed.** I stashed the two source files, kept the tests, and re-ran: 10 failed / 96 passed. So these are not assertions written to agree with whatever the code now does.

Full suite on the fix: **5,771 passed / 0 failed / 36 skipped**. You noted you couldn't run it (`Test` targets `net*.0-windows`) — it's green, for whatever that's worth given your point that a passing suite wouldn't catch any of this.

---

### RESOLVED — `comment[modification parameters]` wrote the motif into `NT=`

Correct, and it was every cell the builder authored. `NT=` now comes from `mod.OriginalId`.

Two things worth recording beyond the fix:

- **The fallback was also backwards.** `mod.IdWithMotif ?? mod.OriginalId` could never reach `OriginalId`, because `Modification`'s constructor only ever populates `IdWithMotif` when `OriginalId` is also set (`Modification.cs:57-84`). So the `??` was dead either way. It reads `mod.OriginalId ?? mod.IdWithMotif ?? ""` now.
- **`OriginalId` is the house convention, not just the correct value.** `ProFormaConverter.cs:159` already does exactly this — `new ProFormaDescriptor(ProFormaKey.Name, mod.OriginalId)` — when it emits a modification name into an external format. This builder was the outlier.

On why the suite was blind to it: exactly as you said. The two modification assertions were the only `Does.Contain` in a fixture where every other column is pinned exactly. They're exact now:

```csharp
Assert.That(mods[0], Is.EqualTo("NT=Carbamidomethyl;AC=UNIMOD:4;TA=C;MT=Fixed"));
Assert.That(mods[1], Is.EqualTo("NT=Oxidation;AC=UNIMOD:35;TA=M;MT=Variable"));
```

plus `TheModificationNameIsBareAndTheResidueAppearsOnlyInTa`, which asserts the input really does carry the motif (`IdWithMotif == "Oxidation on M"`) before asserting the cell contains no `" on "` — otherwise the test would pass for the wrong reason if `Modification` ever stopped setting it.

### RESOLVED — `comment[sdrf version]` could emit `vV1.1.0`

`TrimStart('v', 'V')`. `TheSdrfVersionPrefixIsWrittenOnceWhateverCasingTheCallerSupplied` pins all three inputs (`1.1.0`, `v1.1.0`, `V1.1.0`).

### RESOLVED — tolerances were culture-formatted

`InvariantCulture`. And you read it right that this was an omission rather than a decision: `PpmTolerance.cs:38` and `AbsoluteTolerance.cs:38` both already format their own values with `InvariantCulture`, as does `Positive()` twenty lines below in this same file.

Their `ToString()` is still not reused, and I noted why in the summary so the next reader doesn't try: it renders `±10.0000 PPM` / `±0.0200 Absolute`, which is not the grammar SDRF asks for.

The test carries `[SetCulture("de-DE")]` **and a guard on the fixture itself**:

```csharp
Assert.That(0.02d.ToString(), Is.EqualTo("0,02"),
    "guard on the fixture itself: this test proves nothing unless the culture took");
```

Without that, a test that silently failed to switch culture would pass forever on an en-US box — the same class of blindness as the `Does.Contain` above.

### RESOLVED — coverage counted repeated columns per cell, not per row

Yours is the sharpest of these, because it collides with behaviour this PR pins deliberately, exactly as you laid out. `rows[column]++` sat inside the loop over header *positions*, so the format's own padding landed in the denominator.

A row now asks one question per column *name*: did it say anything here at all. `Rows` counts once per row per name, `Filled` counts a row once however many of its repeats answered.

Your four-file example is now the test (`ARepeatedColumnIsCountedOncePerRowNotOncePerCell`): one file with four modifications and three with one each gives `Rows=4`, `Filled=4`, `FillRate=1.0`, `IsUninformative=false`. The other half is pinned too (`ARepeatedColumnEmptyAcrossTheWholeRowStillCountsAsAbsentOnce`) — a row silent across every repeat is absent exactly once, not N times.

The `<param>` docs on `SdrfColumnCoverage.Rows` and `Filled` now say this, since "Rows in those documents" was true of the contract and not of the code.

### RESOLVED — the lenient path could still throw

Confirmed reachable, and fixed at all three sites — I added a third you didn't name.

`SdrfCell.ToCell` refuses a term with neither name nor accession, which is right. What was wrong is that the refusal could be how a **lenient** build ends, and `RequireSampleMetadata = false` exists precisely so a finished search is never thrown away. All three now route through `Missing()`:

- `InstrumentCell` — non-null instrument, name and accession both empty (your case)
- `CleavageAgentCell` — `DigestionAgent` with an empty `Name` (your case)
- `ModificationCell` — a modification with no name and no UNIMOD reference (same throw, same shape; folded in since I was already there)

You were right to say you hadn't confirmed a reader actually yields such an instrument. I haven't either. The guard is cheap and the option's entire purpose is not throwing, so I'd rather not depend on that being true. `ALenientBuildDoesNotThrowOnATermThatNamesNothing` asserts `Throws.Nothing` and then that all three cells read `not available`; `TheStrictPathRefusesATermThatNamesNothingWithTheUsualException` pins that the strict path still refuses it as an `MzLibException` rather than as an `ArgumentException` escaping a formatting helper.

### RESOLVED — `Build` null-checks `rows`, then dereferences unguarded

`RequireComplete(input, index)` now checks `Sample`, `Assay`, `Characteristics` and both modification lists, and names the row index and the specific member. Since those collection properties default to empty, reaching it with a null means the caller set one to null on purpose, so the message says to pass an empty list or dictionary instead.

### RESOLVED — `values[column]` was `Ordinal` while `IsAnswer` is case-insensitive

`OrdinalIgnoreCase`, and trimmed for the same reason `IsAnswer` trims. `DistinctValuesIgnoresCasingAndSurroundingWhitespace` runs `"liver" / "Liver" / "LIVER" / " liver "` through and expects `DistinctValues = 1` and `IsUninformative = true` — a column that says "liver" four ways cannot group anything, which is the case the count exists to catch.

The casing difference is a real finding, but it belongs to `SdrfDriftLint`, and that's now recorded on the `<param>` so the two comparers don't drift apart again.

---

### `comment[label]` — half resolved, and I want to be explicit about which half

**Fixed:** the message. `Missing()` no longer asserts that sample metadata is what's absent, which for `comment[label]` and `comment[dissociation method]` was simply untrue. It names the column and stops there.

**Not done:** splitting `RequireSampleMetadata` so the label gates separately from the sample columns.

That's a design change to a `public` option on a PR that already carries a compatibility commitment, not a wording fix, and there's no caller today that needs the halves apart — the MetaMorpheus side always resolves a label. I've documented it as a deliberate decision on `Missing()` rather than leaving it silent, so the next reader hits the reasoning instead of the gap:

> NOTE that the switch is named for the columns it exists for, but it guards EVERY column routed through here — comment[label] and comment[dissociation method] among them.

If you'd rather it split now, say so and I'll do it here rather than deferring — I'd just rather that be your call than mine, since it changes a public option's meaning.

---

### On your "checked and correct" section

Recording those was worth more than it might look, so: thank you. The `ToCell` key-order note in particular — that we're stricter than the spec's own examples, which put `MT`/`TA` ahead of `AC` — is the kind of thing that gets "helpfully" relaxed a year from now by someone matching the examples. It's a MUST in README.adoc:261 and the comment on `ToCell` says so.

Two open items you flagged that are still open, so they're not lost:

- **The curated corpus check.** Still not run against builder output. The corpus is a regression suite and a vocabulary source here, never a scoring target, but a round-trip pass over it is worth having and isn't in this PR.
- **The MetaMorpheus call site.** Still not open — it's built and tested locally against this branch, but it waits on this merging *and* shipping in a release. So you're right that you haven't seen the consumer, and that remains true after this push.

acesnik (approved):
🤖 Re-checked against the previous review. All eight findings are fixed, and I verified they're actually pinned rather than just present — I reverted each fix and confirmed a named test fails:

- `NT=` now takes `mod.OriginalId` — caught by `TheModificationNameIsBareAndTheResidueAppearsOnlyInTa` and `ModificationsCarryAccessionTargetAndFixedOrVariable`
- `TrimStart('v', 'V')` — caught by `TheSdrfVersionPrefixIsWrittenOnceWhateverCasingTheCallerSupplied`
- Tolerances in `InvariantCulture` — caught by `TolerancesAreWrittenInInvariantCultureWhateverTheMachineIsSetTo`, which sets `de-DE` explicitly
- Coverage counted once per row per column name — caught by `ARepeatedColumnIsCountedOncePerRowNotOncePerCell`
- Distinct values `OrdinalIgnoreCase` — caught by `DistinctValuesIgnoresCasingAndSurroundingWhitespace`

The lenient-throw path and the null-part guard have tests too, and the exact-form modification assertion replaced the `Does.Contain` that let the original bug through in the first place.

On `comment[label]`: taking the wording fix and declining the switch split, with the reasoning left in the code, is the right call. Nothing further from me there.

I could also run the tests this time, which I couldn't before — 56/56 in the two files this touches. For anyone else trying, the harness assembly has to be named `Test`, since `SdrfValidator` and `SdrfDriftLint` are internal behind `InternalsVisibleTo`.

Approving without waiting for the counterpart MetaMorpheus PR, because this stands alone: the diff is additive apart from one unused `using` in a test file, no production code in mzLib touches the new types, and the full build and SDRF suite pass. Nothing existing changes behaviour.

The one thing that outlives this PR is that the public surface becomes a compatibility commitment once it ships in a package. If the call site turns out to want a different shape, changing it afterwards is a breaking change rather than an edit — so it is worth building the MetaMorpheus side against this before we cut the release that carries it, rather than after.

nbollis (approved):

PR #1176, opened 2026-08-17
Make the mzML and Thermo dynamic-connection tests actually read dynamically

acesnik:
🤖 The defect trishorts found in `TestDynamicMgf` on #1164 exists in three more places. Fixing those is all this PR does.

## The nut

`GetOneBasedScanFromDynamicConnection` short-circuits to the static cache when scans are already loaded:

```csharp
if (CheckIfScansLoaded() && scanNumber <= IndexedScans.Length)
    return GetOneBasedScan(scanNumber);
```

So a test that calls `LoadAllStaticData()` and then `InitiateDynamicConnection()` **on one reader** gets the very same `MsDataScan` instance back, and every static-vs-dynamic comparison is a scan against itself. Confirmed with `ReferenceEquals` before changing anything, rather than inferred from the short-circuit.

| test | dead assertions |
|---|---|
| `TestDynamicMzml` | ~25 × 8 cases |
| `TestDynamicMzmlWithPeakFiltering` | ~25 |
| `TestDynamicRaw` | ~25 × 3 cases |

**mzML and Thermo dynamic reading had no regression coverage at all.**

## The fix

Each test now uses a second reader with no static load, and asserts the two instances are distinct so it cannot silently revert:

```csharp
NUnit.Framework.Assert.That(ReferenceEquals(dynamicScan, staticScan), Is.False,
    "The dynamic read returned the statically cached instance, so nothing below is being compared.");
```

**`TestDynamicMzmlWithMixedBits` needed care rather than the same treatment.** Its `staticScan2`-vs-`staticScan` assertions are a genuine cross-file comparison of 32-bit against mixed-bit encoding, and are kept exactly as they are — only its two dynamic reads were dead, so each file gets its own unloaded reader. A blanket rewrite would have thrown away the one part of that test that was working.

## What this did and did not find

**No bug in the readers.** The tests pass as written, so the dynamic paths were correct all along — just unverified. That is the good outcome, but it is worth saying plainly rather than implying this fixed a live defect.

Verified the other direction: restoring the single-reader shape fails **all 13 cases** on the new guard.

## Deliberately not touched

Three tests do this on purpose and say so in their names — they exercise the short-circuit itself:

- `TestBruker.TestDynamicConnection_AfterStaticLoading`
- `TestReaders.TestLoadingDynamicAfterStaticLoading`
- `TestMsAlign.GetOneBasedScanFromDynamicConnection_ExistingScanNumberOnLoadedData_ReturnsMsDataScan`

`TestReaders.cs` even carries a comment about the short-circuit, so the trap was known — it just was not applied uniformly.

## Verification

Run off Windows through a throwaway net10.0 harness that compiles the real test files (named `Test` so the `InternalsVisibleTo` reaches `ThermoRawFileReader.ThrowIfNotReadable`), so these are executed results rather than replayed assertions:

- **68 tests pass** across `TestMzML` and `TestRawFileReader`
- 13/13 fail on the guard when reverted to a single reader

No production code changes; no public API change.

Stacked on the same review as #1164, which fixes the MGF case.


nbollis (dismissed):

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
```diff
@@ -1478,13 +1478,24 @@ public static void TestDynamicMzml(string fileName, bool writeIndexed = false)
...
-            foreach (MsDataScan staticScan in reader.GetAllScansList())
+            var dynamicReader = MsDataFileReader.GetDataFile(filePath);
+            dynamicReader.InitiateDynamicConnection();
+
+            foreach (MsDataScan staticScan in staticReader.GetAllScansList())
             {
-                MsDataScan dynamicScan = reader.GetOneBasedScanFromDynamicConnection(staticScan.OneBasedScanNumber);
+                MsDataScan dynamicScan = dynamicReader.GetOneBasedScanFromDynamicConnection(staticScan.OneBasedScanNumber);
+
+                NUnit.Framework.Assert.That(dynamicScan, Is.Not.Null,
```
trishorts, line 1494: Confirming the `ReferenceEquals` guard does what it says — I checked the four converted tests and the two-reader split is right.

One consequence worth chasing though: several assertions below were written when `dynamicScan` and `staticScan` were literally the same object, and they are no longer safe now that the two are independently constructed.

**The `||` guards dereference both sides.** At :1530 and :1542 (and :1627/:1639, and TestRawFileReader.cs:393):

```csharp
if (dynamicScan.IsolationWidth != null || staticScan.IsolationWidth != null)
{
    if (!double.IsNaN(dynamicScan.IsolationWidth.Value) || !double.IsNaN(staticScan.IsolationWidth.Value))
```

Under the old shape `a != null || b != null` was equivalent to `a != null`, so the dereference could not fail. Now a one-sided null enters the block and throws. That is reachable rather than theoretical: the dynamic path sets isolation width only when `msOrder > 1` in the hand-rolled `XmlReader` loop (`Mzml.cs:795`), while the static path builds it elsewhere, and `MsDataScan.IsolationRange` (`MsDataScan.cs:135-149`) is null unless *both* `IsolationWidth` and `IsolationMz` have values — so the two parsers can legitimately disagree on null-ness. The net effect is that the exact divergence these newly-live assertions exist to catch surfaces as a bare `NullReferenceException` instead of a labelled failure. Asserting the two sides agree on null-ness first, then `&&` before dereferencing, keeps the diagnostic.

**`NoiseData` is still dead.** `Assert.That(dynamicScan.NoiseData == staticScan.NoiseData)` (:1522, :1619, TestRawFileReader.cs:381) compares `double[,]` with `==`, i.e. reference equality. It was a tautology on one instance before; across two instances it can only hold when both are null — which they always are, since neither reader populates the field (`ThermoRawFileReader.cs:519` passes `noiseData: null` with a `//TODO`). So this one is not among the assertions the PR revives, and it will go red on every scan the moment either reader starts populating noise data, for reasons unrelated to a static-vs-dynamic mismatch. Worth either an element-wise comparison or a comment saying it is inert today.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
```diff
@@ -1668,8 +1695,13 @@ public static void TestDynamicMzmlWithMixedBits()
                 }
 
                 MsDataScan staticScan2 = reader2.GetOneBasedScan(staticScan.OneBasedScanNumber);
-                MsDataScan dynamicScan1 = reader1.GetOneBasedScanFromDynamicConnection(staticScan.OneBasedScanNumber);
-                MsDataScan dynamicScan2 = reader2.GetOneBasedScanFromDynamicConnection(staticScan.OneBasedScanNumber);
+                MsDataScan dynamicScan1 = dynamicReader1.GetOneBasedScanFromDynamicConnection(staticScan.OneBasedScanNumber);
+                MsDataScan dynamicScan2 = dynamicReader2.GetOneBasedScanFromDynamicConnection(staticScan.OneBasedScanNumber);
+
+                NUnit.Framework.Assert.That(ReferenceEquals(dynamicScan1, staticScan), Is.False,
```
trishorts, line 1740: This one is missing the null guard its three siblings have. `TestDynamicMzml` (:1493), `TestDynamicMzmlWithPeakFiltering` and `TestDynamicRaw` all pair the new `ReferenceEquals` check with a preceding `Assert.That(dynamicScan, Is.Not.Null, ...)`; `TestDynamicMzmlWithMixedBits` has only the `ReferenceEquals` checks.

`ReferenceEquals(null, staticScan)` is `false`, so a null dynamic read passes straight through both new guards and then throws `NullReferenceException` on `dynamicScan1.OneBasedScanNumber` at :1706. Null is a reachable return — `Mzml.GetOneBasedScanFromDynamicConnection` returns it whenever the scan is absent from `ScanNumberToByteOffset` (`Mzml.cs:468-471`).

Adding the same `Is.Not.Null` assertion for `dynamicScan1` and `dynamicScan2` would make all four converted tests fail identically for a missing dynamic scan, which is presumably the intent.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
```diff
@@ -1558,19 +1569,30 @@ public static void TestDynamicMzmlWithPeakFiltering()
...
-            reader.InitiateDynamicConnection();
+            // Two readers, deliberately. GetOneBasedScanFromDynamicConnection short-circuits to the
+            // static cache when scans are already loaded, so a single reader that has called
+            // LoadAllStaticData hands back the very same MsDataScan instance and every comparison
+            // below is a scan against itself.
+            var staticReader = MsDataFileReader.GetDataFile(filePath);
+            staticReader.LoadAllStaticData(filteringParams);
+
+            var dynamicReader = MsDataFileReader.GetDataFile(filePath);
+            dynamicReader.InitiateDynamicConnection();
```
trishorts, line 1594: Drive-by, and **pre-existing rather than introduced here** — I checked the diff and these `InitiateDynamicConnection()` calls were already unbalanced before the rename, so this is not a regression from your change.

That said, you are rewriting these exact lines: of the four dynamic readers, only `TestDynamicMzml` closes its connection (:1563), and that close is not in a `finally`. `Mzml.InitiateDynamicConnection` opens a `StreamReader` (`Mzml.cs:878`) and the Thermo one opens an `IRawDataPlus` handle (`ThermoRawFileReader.cs:232`), and these fixtures are shared with `TestRawFileReader`. This test additionally `break`s at scan 10, so it exits with the handle open even on a clean pass.

Entirely your call whether it belongs in this PR — mentioning it because the lines are already open in front of you.

trishorts (commented):
Reviewed the four converted tests against the head commit. The core change is correct - I confirmed each test now uses a second reader with no static load, so the static-vs-dynamic comparisons are real rather than a scan against itself, and the `ReferenceEquals` guard prevents a silent revert.

Three follow-ups below, all Minor; none of them block. The first is the substantive one: some assertions in these loops were safe only because the two scans used to be the same object.

I also checked the three tests you deliberately left alone (`TestBruker.TestDynamicConnection_AfterStaticLoading`, `TestReaders.TestLoadingDynamicAfterStaticLoading`, `TestMsAlign.GetOneBasedScanFromDynamicConnection_ExistingScanNumberOnLoadedData_ReturnsMsDataScan`) - they do exercise the short-circuit on purpose, so excluding them is right.

Alexander-Sol (dismissed):

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
acesnik: 🤖 Both confirmed and both fixed. The `||` count is five rather than two — `IsolationWidth` at :1530/:1627 and `IsolationRange` at :1542/:1639 plus `TestRawFileReader.cs:393`, all dereferencing both sides.

Your reachability argument holds: `IsolationRange` is null unless both `IsolationWidth` and `IsolationMz` have values (`MsDataScan.cs:143`), and the dynamic path only sets width when `msOrder > 1` (`Mzml.cs:793`). Each site now asserts the two agree on null-ness first, then uses `&&` before dereferencing, so a divergence fails with a message instead of an NRE.

`NoiseData` is now compared element-wise and null-safe, for the reason you give — it can only pass today because both sides are always null, and it would have gone red on every scan the moment either reader started populating it.

13/13 dynamic cases still pass, so this is diagnostics rather than a behaviour change.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
acesnik: 🤖 Right, and added. `ReferenceEquals(null, staticScan)` being false meant a null dynamic read passed both guards and threw on `dynamicScan1.OneBasedScanNumber` at the next line. `dynamicScan1` and `dynamicScan2` now get the same `Is.Not.Null` assertion its three siblings carry, so all four converted tests fail identically for a missing dynamic scan.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
acesnik: 🤖 Confirmed, and agreed it's pre-existing — I checked the diff too. Leaving it out of this PR to keep the change to the one defect, but it's worth doing: `Mzml.InitiateDynamicConnection` holds a `StreamReader` and the Thermo one an `IRawDataPlus`, the fixtures are shared with `TestRawFileReader`, and `TestDynamicMzml` `break`s at scan 10 so it exits with the handle open even on a pass. Opening an issue for wrapping all four in `try`/`finally` rather than burying it here. Thanks for checking it wasn't a regression from the rename.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
trishorts: Verified at `0b71b1be`. All five dereference sites now assert null-ness agreement first and use `&&` — `IsolationWidth` and `IsolationRange` in both TestMzML loops, plus `IsolationRange` in TestRawFileReader (the `IsolationWidth` comparison there is a plain `==` on the nullables, so it was never at risk). `NoiseData` goes through the new element-wise, null-safe `AssertNoiseDataEqual`. Resolving.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
trishorts: Verified at `0b71b1be`: `TestDynamicMzmlWithMixedBits` now asserts `Is.Not.Null` on both `dynamicScan1` and `dynamicScan2` before the `ReferenceEquals` guards, matching its three siblings. Resolving.

mzLib/Test/FileReadingTests/SpectraFileReading/TestMzML.cs
trishorts: Agreed on leaving it out of this PR, and #1218 captures it. Resolving here.

nbollis (approved):

trishorts (approved):
All three threads I opened are addressed and verified at `0b71b1be`:

- the five `||` dereference sites now assert null-ness agreement first and use `&&`, and `NoiseData` goes through the null-safe element-wise helper;
- `TestDynamicMzmlWithMixedBits` carries the `Is.Not.Null` guards its three siblings had, so a null dynamic read fails identically across all four converted tests;
- the unbalanced `InitiateDynamicConnection` calls are correctly left out of scope and captured as #1218.

The only commit since is `c5ca5712`, a clean base-only merge of master with no conflict resolution, so the code I verified is the code at head. CI is green. Approving.

PR #1180, opened 2026-08-19
Full width at half maximum per chromatographic peak

acesnik:
🤖 Peak width is the most direct readout of how the chromatography is behaving, and right now nothing in FlashLFQ's output reports it in a usable form. This adds full width at half maximum per chromatographic peak, plus a per-file median to aggregate it. Addresses smith-chem-wisc/MetaMorpheus#1624, opened in 2019 off a comment at Keystone that FWHM per peptide would be helpful for QC.

## Why FWHM is worth having for QC

A few things it gives you that peptide counts do not:

- **Column and emitter health as a trend.** Columns broaden as they age, and a fitting that develops dead volume broadens everything downstream of it. A median FWHM per run, tracked across runs, moves before anything else does.
- **Problems that ID count hides.** A partially blocked emitter or a badly mixed gradient broadens peaks while identification counts hold up for a while. Quantitative precision degrades first, and width is the thing that shows it.
- **Triage between sample and instrument.** IDs down but FWHM unchanged points at digestion or sample prep. IDs down and FWHM up points at the LC. That split is most of what a QC panel is for.
- **Whether the duty cycle matches the chromatography.** The usual criterion for reliable areas is a healthy number of MS1 points across the peak, and "across the peak" conventionally means across FWHM, not across the full base. This exposes FWHM alongside the apex-charge point count, which makes points-across-FWHM computable per peak for the first time.
- **Objective method comparison.** Comparing two gradients or two columns without eyeballing chromatograms.
- **Interference on a single peptide.** A peak much wider than that peptide's own history suggests co-elution, which is worth knowing when one peptide's quant looks wrong.

There is also a pleasant symmetry: mzLib already *reads* an `fwhm` field from Dinosaur output (`DinosaurFeature.cs:49`) but has never been able to produce one.

## What this adds

- `PeakWidth` — FWHM of a peak, with a status enum saying whether it is measurable and why not when it isn't.
- `ChromatographicPeak.PeakWidth` — computed property.
- Two columns in `AllQuantifiedPeaks.tsv` and the per-file `_QuantifiedPeaks.tsv`: **`Peak FWHM`** and **`Peak FWHM Status`**, inserted after `Peak RT End`, written from both `ChromatographicPeak` and `MbrChromatographicPeak`. No existing column changes value, though a positional parser will see the shift.
- `PeakWidthSummary.Summarize` — median and quartiles per file, optionally binned by retention time.

Purely additive: 749 lines, no deletions. Nothing re-reads spectra — `ChromatographicPeak.IsotopicEnvelopes` already keeps the whole trace after quantification, so this is one pass over envelopes per peak. The 16,119-peak run below quantified in 26 s including it.

## It measures something the output does not already have

The existing `Peak RT Start`/`End` columns look like a width, but they are `Min`/`Max` over **every** envelope regardless of charge state, and they extend to wherever the trace happened to stop.

Vignette mouse CAST fractions, 2 files, 176 min gradient, 30,620 identifications at 1% q-value, 16,119 peaks:

| | p10 | median | p90 | max | p90/p10 |
|---|---|---|---|---|---|
| **FWHM** | 0.185 | **0.305** min | 0.724 | 9.78 | **3.9×** |
| base width (`RT End − RT Start`) | 0.344 | 0.794 min | 3.493 | **101.10** | 10.2× |

The ratio between them spans 0.003 to 0.988 with a median of 0.424, so one does not predict the other. A base width of 101 minutes is over half the gradient. FWHM is 2.6× smaller and 2.6× tighter in spread, which is what makes it trendable.

Coverage on the same run:

| | measured | tooFewPoints | censored | noApex |
|---|---|---|---|---|
| all peaks | **61.5%** | 22.5% | 13.7% | 2.3% |
| MS/MS peaks | 71.1% | | | |
| MBR peaks | 20.8% | | | |

## Measured on the apex charge state

`IsotopicEnvelopes` interleaves charge states, so walking it in retention-time order alternates between charges and describes a sawtooth rather than a chromatogram. This uses the apex charge alone — the same filter `CutPeak` already applies for valley detection (`FlashLfqEngine.cs:1755`). One of the tests is an interleaved second charge state that must not change the answer.

## Why the summary can bin by retention time

The two files' overall medians agree to 0.004 min — 0.302 against 0.306 — which looks reassuringly stable. But median FWHM by retention time decile is U-shaped: **0.372 min early, 0.260 at 75–85 min, 0.393 late**. A single median per file averages over a real trend, and late-gradient broadening is normal while early broadening is not, so the two should not be pooled. Hence the bin count.

Bins hold equal numbers of peaks rather than spanning equal time, so a sparse early or late stretch does not produce a median resting on a handful of peaks.

## Why the point floor is 5

It matches `CutPeak`'s own floor, and it sits on the flat part of the curve rather than being picked:

| floor | measured | censored |
|---|---|---|
| 3 | 62.8% | 3,594 |
| **5** | **61.5%** | **2,215** |
| 10 | 41.6% | 549 |

Relaxing to 3 buys 1.3 points of coverage and adds 1,379 censored peaks — it mostly relabels `TooFewPoints` as `Censored`. `Measure(peak, minimumTimePoints)` is public so this is re-checkable.

## Other decisions that change the answer

- **Crossings are linearly interpolated.** Taking the bracketing sample instead would quantise every width to a multiple of the MS1 cycle time.
- **Crossings are the nearest either side of the apex**, not the first and last of the trace, so a shoulder that dips below half maximum and comes back is not swallowed into one very wide peak.
- **A truncated trace reports `Censored`**, not a number computed off the edge. FlashLFQ stops extending an XIC after `MissedScansAllowed` misses and `CutPeak` truncates at valleys, so this is expected rather than a failure.
- **Status is an enum, not a sentinel width**, so a consumer can separate "not measurable here" from a measurement and can tell the two reasons apart. A bin with nothing measurable reports `NaN`, not `0`, since `0` would read as infinitely sharp peaks.

## A defect this went through, worth knowing about if you touch these columns

The first push failed CI, and the failure was more interesting than a shifted golden string.

`MbrChromatographicPeak` overrides `ToString()` and **duplicates the entire row** rather than extending the base. Adding the width columns to `ChromatographicPeak` alone therefore left every MBR row two fields out of step with the header from `Peak MZ` onward — 3,054 of the 16,119 rows in the run above. The measurement was never wrong, since it comes from the property, but the file would have been, and misaligned in a way no spot check of a single row would reveal unless that row happened to be MBR.

Fixed by writing the fields from the subclass too, and guarded by a test asserting each row carries as many fields as the header — for the base peak and the subclass, with and without an apex, since those branches write different literals. It fails on all four combinations against the unfixed subclass.

The underlying hazard outlives this PR: any future column has the same trap, and the duplicated row is the reason. Probably worth a separate issue to have the subclass extend the base row instead. I have not done that here.

While in there, `ChromatographicPeakTests.TestDetectionType` indexed the detection type as a hardcoded column 16; it now resolves the column from `TabSeparatedHeader` by name, so the next column added ahead of it does not break the test the same way.

## One prediction of mine was wrong

I expected censoring to hit wide, co-eluting peaks. It does the opposite: censored peaks have a median base width of **0.416** against 0.794 for measured, and a median intensity of **9.6e5** against 3.5e6. It tracks weak, short traces instead. Same direction on the checked-in K562 data, where only 35.5% were measurable — that file is a 4-minute retention time slice with a median of 4 apex-charge points per peak, so it understates coverage badly.

## Tests

17, and every width is known arithmetically rather than by fit: a triangle, a gaussian against `2√(2ln2)σ`, an interpolation case whose answer is not reachable by snapping to sample points, one- and two-sided censoring, the point floor, a shoulder, an interleaved second charge state, and the drift case the binning exists for — ten narrow peaks early and ten wide late give a flat median of 0.5 that describes neither population, and 0.2 and 0.8 once binned.

## Not verified

- The 17 tests were run on macOS arm64 through a throwaway `net10.0` harness, because `Test.csproj` is `net10.0-windows`. **They have not been run through `Test.csproj`** — CI covers that.
- Coverage was measured on one 2013 instrument's data plus the K562 slice. Not measured on fast-MS1 modern data, DIA, or timsTOF, so 61.5% is one data point rather than a general figure.
- MBR's 20.8% is an observation, not a diagnosis. I did not chase why MBR traces fare worse.
- No MetaMorpheus-side change here. Nothing consumes `PeakWidthSummary` yet, and reaching MetaMorpheus needs an mzLib release plus a `PackageReference` bump.
- I have not audited whether anything else duplicates the peak row the way `MbrChromatographicPeak` does. The row-width test covers those two types; a third would need adding to it.


Alexander-Sol (approved):

Alexander-Sol (dismissed):

mzLib/FlashLFQ/PeakWidthSummary.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (retentionTimeBins < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(retentionTimeBins), "at least one bin is required");
+            }
+
+            var summaries = new List<PeakWidthSummary>();
+
+            foreach (var byFile in peaks
+                .Where(p => p?.SpectraFileInfo != null)
+                .GroupBy(p => p.SpectraFileInfo.FilenameWithoutExtension)
```
nbollis: [synthesis] QC summary groups by basename instead of full file identity

`FilenameWithoutExtension` collapses distinct files that share a basename into one summary bucket, so per-file peak-width QC can silently merge unrelated runs and report corrupted counts and retention ranges.

Suggestion: Group by the full source-file identity rather than basename alone.

mzLib/FlashLFQ/PeakWidthSummary.cs
```diff
@@ -0,0 +1,133 @@
...
+                .OrderBy(g => g.Key, StringComparer.Ordinal))
+            {
+                List<ChromatographicPeak> ordered = byFile.OrderBy(p => p.ApexRetentionTime).ToList();
```
nbollis: [synthesis] Null-apex peaks sort into the first retention-time bin

Peaks without an apex sort as retention time `-1`, which pollutes the first summary bin and inflates counts with peaks that cannot contribute a meaningful width measurement.

Suggestion: Exclude peaks without an apex from retention-time binning or handle them in a separate bucket.

nbollis (commented):
## Review Summary

- 1 major issue and 1 minor issue in `mzLib/FlashLFQ/PeakWidthSummary.cs`.
- No summary-only findings.

The main risk is silent corruption of per-file peak-width QC summaries when multiple source files share a basename.

mzLib/FlashLFQ/PeakWidthSummary.cs
trishorts: Fixed in `81fc5c0f`. Grouped by the `SpectraFileInfo` itself now, ordered by full path; `FileName` still reports the base name, so the output shape is unchanged.

Test covers two files with the same name in different directories — previously one summary of four peaks with a median between the two runs, now two summaries of two.

mzLib/FlashLFQ/PeakWidthSummary.cs
trishorts: Fixed in `81fc5c0f`. Confirmed the mechanism: `ApexRetentionTime` is `Apex?.IndexedPeak.RetentionTime ?? -1`, so an apexless peak sorted ahead of every real one.

It took the first bin's `RetentionTimeStart` down to -1 and counted toward `TotalPeakCount` for a bin it could not contribute a width to — so the denominator of "how much of the file the median rests on" was wrong by however many such peaks there were.

Excluded from the summary entirely rather than binned separately: a peak with no apex has nothing to say about width.

nbollis (dismissed):

mzLib/FlashLFQ/PeakWidth.cs
```diff
@@ -0,0 +1,166 @@
...
+        public double FullWidthAtHalfMaximum { get; init; }
+
+        /// <summary>Interpolated retention time of the rising half-maximum crossing, or NaN.</summary>
+        public double HalfMaximumStart { get; init; }
+
+        /// <summary>Interpolated retention time of the falling half-maximum crossing, or NaN.</summary>
+        public double HalfMaximumEnd { get; init; }
+
+        /// <summary>Time points on the apex charge state that the measurement was made over.</summary>
+        public int TimePointsOnApexCharge { get; init; }
```
trishorts, line 70: **This is the one I would like changed before merge.** It is public API in a NuGet package, so a rename after release is a break, and the fix now is a line.

`TimePointsOnApexCharge` is assigned `trace.Count` (`:128`) — every point on the apex-charge trace. The summary says "Time points on the apex charge state that the measurement was made over", which reads as the points spanning the measurement, i.e. between the two half-maximum crossings.

That matters because of the case the PR description makes for this feature:

> the usual criterion for reliable areas is a healthy number of MS1 points across the peak, and "across the peak" conventionally means across FWHM, not across the full base

Someone applying that criterion will reach for this property, and get a count across the full base — the very thing the description distinguishes it from. It will read as comfortably above threshold when the points across FWHM are not.

Either is fine by me:

- keep the value and make the name say it — `TimePointsOnTrace`, or leave the name and change the doc to "points on the apex-charge trace, of which the measurement spans a subset"; or
- keep the name and add the number the description actually argues for, as a second member — points strictly between `HalfMaximumStart` and `HalfMaximumEnd`.

The second is more useful for the QC panel this is aimed at, but the first closes the trap and costs nothing.

mzLib/FlashLFQ/PeakWidth.cs
```diff
@@ -0,0 +1,166 @@
...
+            int apexIndex = 0;
+            for (int i = 1; i < trace.Count; i++)
+            {
+                if (trace[i].Intensity > trace[apexIndex].Intensity)
+                {
+                    apexIndex = i;
+                }
+            }
+
+            double halfMaximum = trace[apexIndex].Intensity / 2.0;
```
trishorts, line 120: Non-blocking. `apex / 2.0` takes the half-maximum against a zero baseline, which is the right assumption here — FlashLFQ's XICs carry no persistent baseline, since envelopes below detection are simply absent from the trace rather than sitting at some offset.

But it is an assumption, in a method whose whole job is a metric with a textbook definition, and a reader coming from instrument software will expect `(apex + baseline) / 2`. One sentence in the remarks saying the baseline is taken as zero and why would stop someone "fixing" it later.

mzLib/FlashLFQ/PeakWidth.cs
```diff
@@ -0,0 +1,166 @@
...
+                }
+
+                IsotopicEnvelope below = trace[i];
+                IsotopicEnvelope above = trace[i - direction];
+
+                double intensitySpan = above.Intensity - below.Intensity;
+                double timeSpan = below.IndexedPeak.RetentionTime - above.IndexedPeak.RetentionTime;
+
+                // Equal intensities would mean above is also at half maximum; take that point.
+                if (intensitySpan <= 0)
```
trishorts, line 165: Non-blocking, and only worth mentioning because the comment will mislead someone.

This branch is unreachable except for an all-zero trace. The loop `continue`s while `trace[i].Intensity > halfMaximum`, so on exit `below <= halfMaximum` and `above > halfMaximum` strictly — meaning `intensitySpan` is always positive. The comment explains the equal-intensity case ("above is also at half maximum"), but equal intensities cannot get here: if `above` were exactly at half maximum the loop would have stopped on it instead.

The only way in is `apex.Intensity == 0`, which makes `halfMaximum` zero so nothing is ever `> halfMaximum`.

Keep the guard — it is cheap and the degenerate case is real — but the comment would be more useful as "only reachable when the apex intensity is zero".

trishorts (dismissed):
This is a careful implementation and I want it in — requesting changes on one point only, because it is public API and the cost of fixing it moves sharply once this ships in a package.

Things I checked that are right, since they were the plausible places for this to go wrong:

- **TSV columns line up.** Both `ChromatographicPeak` and `MbrChromatographicPeak` emit seven columns in *both* branches of `ToString`, with `NoApex` landing in the `Status` slot. A mismatch there would shift every column after it, silently.
- **The apex is re-found within the filtered trace** rather than trusting `peak.Apex` by reference — correct, since the trace is re-sorted, and the comment says why.
- **Crossings walk outward to the nearest one**, not the first and last of the whole trace, so a shoulder that dips below half maximum and recovers is not measured as one wide peak.
- **Apex at a trace edge** returns null and lands in `Censored` rather than throwing.
- Measuring on the apex charge state alone, because `IsotopicEnvelopes` interleaves charge states and an RT-ordered walk over all of them describes a sawtooth. Right call, well explained.

Two smaller notes are inline and **not** blocking — fold them in if you are touching the file anyway, or leave them.

For the record: commit `81fc5c0f` on this branch is mine, pushed while working through nbollis's two findings (basename grouping, and null-apex peaks binning at RT −1). So this review covers a change I wrote as well as yours.

mzLib/FlashLFQ/PeakWidth.cs
trishorts: Pushed the doc half myself in `9f99d0bc`, since it was a line and it is public API — the name is untouched, so nothing breaks for anyone already using it.

It now says the count is the whole apex-charge trace, not the points between the two crossings, and why that distinction matters for the duty-cycle criterion the description argues from.

I deliberately did **not** take the other option — adding a points-within-FWHM member. That is a design choice on your feature rather than a correction, and it should be yours. If you want it, it is a small follow-up; the doc now points at where the number comes from.

The two non-blocking notes further down (zero-baseline assumption, unreachable guard comment) are still untouched and still optional.

trishorts (approved):

PR #1182, opened 2026-08-19
Make PtmListLoader forward to ModificationLoader, and fix the two dissociation keys they disagreed on

acesnik:
## Problem

🤖 `UsefulProteomicsDatabases/PtmListLoader.cs` is marked `[Obsolete]` and its own summary says *"Legacy wrapper for backwards compatibility. All functionality moved to Omics.Modifications.ModificationLoader."* It is not a wrapper. It holds a separate 484-line copy of the parser — amino-acid table, `ReadMod`, formal-charge adjustment and all — and the two copies have drifted.

They disagree on two dissociation-type keys in diagnostic ion and neutral loss entries:

| entry | `PtmListLoader` | `ModificationLoader` |
|---|---|---|
| `MPD:` | `IRMPD` | `MPD` |
| `Any:` | `AnyActivationType` | the modification loses **all** its diagnostic ions and neutral losses |

**`MPD`** — `ModificationLoader` resolves the type with a plain `Enum.TryParse`, and `MPD` is itself an enum member (MS:1000435 photodissociation), so the parse succeeds and returns that. A guard immediately below intends to map it to `IRMPD`:

```csharp
DissociationType? dt = Enum.TryParse(entryKeyValue[0], true, out DissociationType parsedDt) ? parsedDt : null;

if (dt is null && entryKeyValue[0].Equals("MPD", StringComparison.InvariantCultureIgnoreCase))
    dt = DissociationType.IRMPD;
```

`dt` is never null for `"MPD"`, so the guard never runs.

**`Any`** — no such enum member exists; the enum spells it `AnyActivationType`. `TryParse` fails, and an unresolvable type throws rather than skipping: the surrounding `catch` swallows the exception and sets the whole dictionary to `null`, so one unrecognised key discards every diagnostic ion and neutral loss on that modification.

## Severity: latent

Nothing mzLib ships hits either case. Across all mods files in the repository the only dissociation keys present are `HCD` (7,410) and `ETD` (322), and both resolve identically under either implementation. Reaching this needs a custom mods file using `MPD:` or `Any:`.

The reason to deal with it now is that it blocks the deletion below: two copies of a file parser is how they drift, and these already had.

## Solution

`ModificationLoader.ParseDissociationType` tests the two special cases ahead of the generic parse. Everything else still goes through `Enum.TryParse` and remains case-insensitive.

With the two implementations agreeing, `PtmListLoader` becomes the wrapper it already claimed to be — every member forwards. **517 lines out, 54 in.** The only internal caller was one test.

`ModDissociationType` is now case-insensitive where the copy matched exactly: more permissive rather than different, since every spelling the old switch accepted resolves to the same value.

## One call worth a second opinion

`MPD → IRMPD` narrows a general term to a specific one. From the PSI-MS ontology, MS:1000435 photodissociation is *"absorption of one or more photons"* and MS:1000262 IRMPD is *"absorption of multiple infrared photons"*, with IRMPD's ancestors being `dissociation method → photodissociation`. So IRMPD is a **child** of photodissociation, and mapping a file's `MPD:` to IRMPD asserts infrared where the file only said photons — a UVPD-derived entry would be mislabelled.

I kept it: it is the behaviour the format has always had, and what the unreachable guard was reaching for. If `MPD:` should mean the parent term, that is a deliberate decision rather than something to inherit from a bug fix.

## Verification

Every mods file in the test data loaded through both entry points and compared on the full modification sets — id, type, monoisotopic mass, location restriction, diagnostic ions, neutral losses — plus warning counts. They match.

Restoring the pre-fix parsing fails exactly 4 of the 14 new cases: `Any`, `any`, `MPD`, `mpd`, the last two reporting `Expected: IRMPD, But was: MPD`. The other 10 pass either way, as regression guards should.

`Test.csproj` compiles clean. It targets `net10.0-windows`, so the tests were run on macOS through a throwaway `net10.0` harness compiling the committed methods verbatim; the full Windows suite has not been run locally.


nbollis (dismissed):

trishorts (commented):
The `MPD` case is a genuinely good catch — an enum member shadowing the intended mapping is the kind of thing that parses cleanly and is wrong, and it would never have shown up as an error. Deleting 474 lines of duplicated loader in the same breath is worth having.

One mechanical problem inline, and one null path.

mzLib/Omics/Modifications/IO/ModificationLoader.cs
```diff
@@ -429,6 +429,27 @@ private static double AdjustMonoIsotopicMassForFormalCharge(double? _monoisotopi
     /// </summary>
     /// <param name="oneEntry"></param>
     /// <returns></returns>
+    /// <summary>
```
trishorts, line 426: **The new doc block landed inside the existing one, so it belongs to the wrong method and the old method lost its documentation.**

As it stands at head:

```csharp
/// <summary>
/// Parse diagnostic ion and neutral loss strings     <- meant for DiagnosticIonsAndNeutralLosses
/// </summary>
/// <param name="oneEntry"></param>
/// <returns></returns>
/// <summary>                                          <- new, meant for ParseDissociationType
/// Resolves the dissociation type named at the start ...
/// </summary>
/// <remarks> ... </remarks>
public static DissociationType? ParseDissociationType(string dissociationTypeName) => ...

public static Dictionary<DissociationType, List<double>> DiagnosticIonsAndNeutralLosses(...)   <- now undocumented
```

So `ParseDissociationType` carries two `<summary>` tags and a `<param name="oneEntry">` for a parameter it does not have, and `DiagnosticIonsAndNeutralLosses` has no comment at all. Worth fixing beyond tidiness: mzLib ships XML documentation to consumers (#1162), so this reaches tooltips, and duplicate/mismatched tags are a doc-comment warning in a repo that has been actively clearing warnings (#1192).

The fix is just moving the new block down past the old `<returns>` — the new text itself is the most useful comment in the file.

**Separately:** `ParseDissociationType` dereferences its argument before testing it —

```csharp
_ when dissociationTypeName.Equals("Any", ...)
```

— so a null name throws rather than returning `null`. Since the method's contract is "or null when it names nothing recognised", null arguably *is* nothing recognised. Whether that can happen depends on how the callers split the entry, but the public signature invites it now that this is a `public static` helper rather than inline code.

acesnik: 🤖 Both fixed in e76fe1f.

The doc block moves down to `DiagnosticIonsAndNeutralLosses`, and both methods now document every parameter — the old block only had `<param name="oneEntry">`, so restoring it verbatim would have traded CS1572 for CS1573 the moment #1162 turns generation on. Your read of why it matters beyond tidiness was right, and it isn't the only instance: the identical stacked-summary bug is sitting on master in `QuantificationEngine`, which surfaced while rebuilding #1162.

On the null path — agreed, and it isn't reachable from the loader: the caller splits on `":"` and `Split` never yields null. It's the public signature that invites it, as you say, so `ParseDissociationType` now returns null for a null name. Behaviour through `DiagnosticIonsAndNeutralLosses` is unchanged either way, since the `NullReferenceException` and the `MzLibException` it now takes instead are swallowed by the same `catch`.

acesnik:
🤖 Took the patch coverage while I was in here. It was 70.59%, and all five uncovered lines were in the wrapper — four of its six members had no caller at all. Each now runs against its `ModificationLoader` counterpart.

The formal-charge overload needed more than an agreement check: a forward that silently dropped the dictionary would still return modifications and would still agree with a replacement handed the same argument. So that test loads a probe entry with a `DR` line and asserts the mass comes back one proton lighter.

The probe carries an explicit `MM` line for a reason worth recording. `adjustWithFormalCharge` is computed from `_monoisotopicMass != null` **before** the mass is taken from the chemical formula, so a CF-only modification never reaches the adjustment — `ModsWithComments.txt`'s Sodium entry was my first attempt and the delta came back exactly 0. Pre-existing, and byte-identical in the copy this PR deletes, so nothing here changes it. Noting it because it is invisible from the call site.

Six mutants to confirm the tests aren't inert: drop the null arm, drop the formal-charge dictionary, stub out the StreamReader and string overloads, forward a different entry from `DiagnosticIonsAndNeutralLosses`, un-forward `ModDissociationType`. All six fail the matching test.

Run on macOS through a throwaway net10.0 harness compiling the committed methods verbatim, 21 of 21. `Test.csproj` compiles clean under `-p:EnableWindowsTargeting=true`; the Windows suite still has not run locally.

The branch is behind master — say the word if you want it merged forward for a fresh codecov run before this goes in.

trishorts (dismissed):
Both points confirmed fixed at `e76fe1fa`, read from the source rather than taken on the summary.

The doc blocks are unstacked: `ParseDissociationType` now carries one `<summary>`, a `<param name="dissociationTypeName">` that matches its actual parameter, a `<returns>` and the `<remarks>`; `DiagnosticIonsAndNeutralLosses` has its own block again. Documenting `dAndNDictionary` rather than restoring the old block verbatim was the right call -- the original only had `<param name="oneEntry">`, so it would have traded CS1572 for CS1573 once #1162 turns generation on, and #1162 is now on master.

`ParseDissociationType` takes `null => null` as its first switch arm, so the name is no longer dereferenced before it is tested, and the method matches the contract its own `<returns>` states. Agreed it is unreachable from the loader -- `Split` does not yield null -- but it is a `public static` helper now.

Approving. Two things that are not objections:

- The branch conflicts with master in `PtmListLoader.cs`, from #1162 having landed there since. Resolving it is a content-changing push and `dismiss_stale` is on for master, so this approval will come off and need redoing -- worth resolving the conflict first and collecting both approvals after.
- The coverage work in your follow-up comment is more than was asked for, and the note about `adjustWithFormalCharge` being computed from `_monoisotopicMass != null` before the chemical formula is read is a genuinely useful thing to have written down. Pre-existing and byte-identical in the copy this PR deletes, so nothing to do here.

trishorts:
@acesnik — I have withdrawn my approval, on merge-order grounds rather than anything in the review coming undone. Both fixes from the last round are still verified good at `e76fe1fa`. Stating the requirement here so it is visible rather than buried in a dismissal notice.

**#1270 must merge before this PR, and this PR must be green against it.**

## Why this PR counts as formal-charge code

It deletes an entire second copy of the formal-charge arithmetic from `PtmListLoader` — `useChemFormulaForMM`, `adjustWithFormalCharge`, `AdjustMonoIsotopicMassForFormalCharge` and the `ProtonMass` subtraction — and replaces it with a forward. I had been describing this PR as "a dedup refactor that does not touch formal-charge behaviour", and that was too casual: it does not *intend* to change behaviour, which is not the same thing.

That code path decides what a trimethylated lysine weighs. UniProt writes it as `CF   C3 H7`, the quaternary ammonium as it exists on the protein — permanently charged, needing no proton to become so — so the formula cannot be added to a peptide as written. `PSI-MOD MOD:00083` carries `FormalCharge 1+`, and subtracting it is what produces the neutral increment:

```
MM 43.054227 + FormalCharge 1+  ->  42.046951 / C3H6      (Unimod Trimethyl)
adjustment bypassed             ->  43.054227 / C3H7      one proton heavy
```

A wrong answer here is silent and systematic. The modification still loads, still searches, and still reports identifications — against the wrong mass, with no exception and nothing downstream that would notice. That is why this gets a test in front of it rather than a review comment.

## What this means in practice

#1270 is test-only, no production changes, so it can land on its own. I am adding an arm to it that runs the same assertions through `PtmListLoader` as well as `ModificationLoader`, so it covers this PR's own entry point directly and the forward is *shown* to preserve the invariant rather than asserted to.

This PR has to resolve its `PtmListLoader.cs` conflict with master anyway — #1162 landed there — so picking up #1270 in the same update costs nothing extra.

Re-approving as soon as it sits on top of that test and is green. Nothing else outstanding from my side.

trishorts:
@acesnik — the gate is cleared. #1270 merged (`828c06fb`, now master's tip), so the trimethyllysine formal-charge invariant is pinned on master.

I checked that this PR passes it rather than just lifting the block, and also worked out the conflict so you do not have to re-derive it.

**The conflict is one hunk, and the resolution is unambiguous.** Master's side is a three-line fragment of the *old* `ReadModsFromFile(StreamReader, ...)` body — the one this PR already replaced with a forward at the top of the file. Your side is the new `ModDissociationType` forward. Keep yours, drop the fragment; nothing of master's is lost, because master's other changes to `PtmListLoader.cs` auto-merge cleanly around it.

Resolved that way locally, current master merged in:

```
Passed TrimethylLysine_FormalChargeAdjustmentYieldsTheNeutralMassIncrement
Passed TrimethylLysine_FormalChargeAdjustmentIsTheSameThroughPtmListLoader

Full offline suite: 5908 passed, 0 failed, 29 skipped
```

The second test is the one that matters here, and it is worth saying why. After this PR, `PtmListLoader.ReadModsFromFile` **is** the forward — so that test is no longer checking a parallel implementation, it is exercising your forward and proving the copy you deleted is not missed. It pins the value at both entry points rather than just comparing them, so it cannot pass by both going wrong together, which is exactly the failure a shared-code refactor makes possible.

That is the whole reason for the gate, and it holds. Nothing in your change disturbs the invariant.

Not pushed: resolving a conflict means choosing which side of a deletion wins, and that is your call on your PR, so I have left the branch alone. Happy to push the resolution if you would rather — it is the one hunk above and I have already verified the result.

Once it is resolved on the branch and green, I will re-approve. Both review points from the last round were verified good at `e76fe1fa` and nothing has come undone; withdrawing was purely about merge order.

trishorts:
@acesnik — pushed the conflict resolution onto this branch (`17690539`) rather than leaving it sitting. Nothing of yours changed; it is a merge commit only.

**How the hunk was resolved, so you can check my reasoning rather than take it.** Master's side was a nine-line fragment of the old `ReadModsFromFile(StreamReader, ...)` body — the method this branch already replaced with a forward at the top of the same file. Keeping it would have reintroduced a partial copy of the very implementation the branch exists to delete, sitting next to the forward that supersedes it. So your side won and the fragment went. Master's other changes to the file auto-merged around the hunk untouched.

I asserted that rather than assuming it: the resolution refused to drop master's side unless it actually contained the old body, so a genuinely new change landing in that region would have stopped the merge instead of being quietly swallowed. Resulting file is the three forward overloads and nothing else, 54 lines.

**Against the gate, now that #1270 is on master:**

```
Passed TrimethylLysine_FormalChargeAdjustmentYieldsTheNeutralMassIncrement
Passed TrimethylLysine_FormalChargeAdjustmentIsTheSameThroughPtmListLoader

Full offline suite: 5908 passed, 0 failed, 29 skipped
```

The second is the one this branch is measured by. `PtmListLoader.ReadModsFromFile` **is** the forward now, so that test stopped comparing two implementations and started exercising yours — and it pins the value at both entry points rather than just asserting they agree, so it cannot pass by both going wrong together. The copy you deleted is provably not missed.

Re-approving once CI comes back. Both review points from the last round were verified good at `e76fe1fa` and nothing came undone; withdrawing was only ever about merge order, and that order is now satisfied.

trishorts (approved):
Re-approving at `17690539`, all five checks green and the branch mergeable.

Both conditions I set when I withdrew are now met: #1270 has merged, so the trimethyllysine invariant is on master, and this branch is green against it rather than merely compatible with it.

The two review points from the last round stand as verified at `e76fe1fa` and nothing came undone in the merge: `ParseDissociationType` carries one `<summary>` and a `<param>` matching its actual parameter, `DiagnosticIonsAndNeutralLosses` has its own block documenting both parameters, and `null => null` is the first switch arm.

What the gate actually bought, now that it can be stated as a result rather than a worry: `TrimethylLysine_FormalChargeAdjustmentIsTheSameThroughPtmListLoader` was written while `PtmListLoader` still had its own copy of the formal-charge arithmetic. After this PR that method **is** the forward, so the test stopped comparing two implementations and started exercising this one — and because it pins the value at both entry points rather than asserting the two agree, it cannot pass by both going wrong together. That is the specific failure a shared-code refactor makes possible, and the reason a deleted copy of mass arithmetic deserved a test in front of it rather than a review comment.

Withdrawing the approval was never a judgement on the change. It was that an approval standing over a branch which had not yet been measured against the invariant would have been an approval of something untested.

trishorts (commented):
Read the diff directly rather than running the engine (2 source files plus tests). All five checks are green, and I confirmed the shape of the change is sound: `PtmListLoader`'s public surface is exactly six members on master (`ReadModsFromFile` x3, `ReadModsFromString`, `ModDissociationType`, `DiagnosticIonsAndNeutralLosses`) and all six forward, so nothing is removed from the shipped API. `ModificationLoader.ReadModsFromFile(string, out)` carries the same `.OrderBy(b => b.IdWithMotif)` the deleted copy had, so the ordering contract survives. `BackboneModification` and `BaseModification` both derive from `Modification`, so the element type of the returned sequence is still assignable. And `grep -rn PtmListLoader` over MetaMorpheus returns nothing, so there is no in-org consumer of the deprecated path at all.

This does not withdraw my earlier approval. The one thing I want changed before merge is the description, not the code.

**The drift inventory is incomplete.** The PR says the two copies "disagree on two dissociation-type keys" and grades the whole delta latent. Diffing master's `PtmListLoader` against `ModificationLoader` with whitespace ignored, there is a third disagreement, and it is bigger: `ModificationLoader.ReadMod` has `case "BM"` (backbone modification) and `case "BL"` (base loss) and yields `BackboneModification` / `BaseModification` for them. Master's `PtmListLoader.ReadMod` has neither case, and its `default: break;` at line 327 drops them silently.

That one is not latent. `mzLib/Omics/Resources/RnaMods.txt` carries 4 `BM` and 5 `BL` lines today, and it is copied to `Resources/RnaMods.txt` in the Test output -- I count 9 matches in `Test/bin/Debug/net10.0-windows/Resources/RnaMods.txt`. So after this PR the deprecated entry point parses nine lines it used to ignore, on a file mzLib ships, and returns two runtime types it never returned before.

It is a widening, both new types are subtypes, and nothing consumes the path, so I am not asking you to change the behaviour -- forwarding to the real parser is the point of the PR. I am asking the body and the `ModDissociationType` remark to say so, because right now they tell a reader the delta is two dissociation keys and unreachable. Inline notes below on where the new tests could have caught it and on one contract wording in the new public helper.

mzLib/Omics/Modifications/IO/ModificationLoader.cs
```diff
@@ -423,11 +423,36 @@ private static double AdjustMonoIsotopicMassForFormalCharge(double? _monoisotopi
...
+    /// </remarks>
+    public static DissociationType? ParseDissociationType(string dissociationTypeName) =>
+        dissociationTypeName switch
+        {
+            null => null,
+            _ when dissociationTypeName.Equals("Any", StringComparison.InvariantCultureIgnoreCase)
+                => DissociationType.AnyActivationType,
+            _ when dissociationTypeName.Equals("MPD", StringComparison.InvariantCultureIgnoreCase)
+                => DissociationType.IRMPD,
+            _ => Enum.TryParse(dissociationTypeName, true, out DissociationType parsed) ? parsed : null
```
trishorts, line 447: `Enum.TryParse` accepts the underlying numeric representation, so this arm resolves any integer string: `ParseDissociationType("99")` returns `(DissociationType)99`, an undefined value. Verified against the framework -- `Enum.Parse(typeof(DayOfWeek), "9", true)` succeeds and `Enum.IsDefined` on the result is `False`.

That behaviour is pre-existing inside `DiagnosticIonsAndNeutralLosses`, so I am not calling it a regression. But this is a new `public` method with a new documented contract -- "or null when it names nothing recognised" -- and an integer string names nothing recognised while returning non-null. Either tighten it:

```csharp
_ => Enum.TryParse(dissociationTypeName, true, out DissociationType parsed)
     && Enum.IsDefined(typeof(DissociationType), parsed)
     ? parsed : null
```

or say in the `<returns>` that numeric strings are accepted as raw enum values. Tightening changes what a mods file with `7:H2O` does, so the doc route is the conservative one if you would rather not touch behaviour in a refactor PR.

mzLib/UsefulProteomicsDatabases/PtmListLoader.cs
```diff
@@ -1,484 +1,54 @@
...
+                out filteredModificationsWithWarnings, fileLocation);
+
+        /// <inheritdoc cref="ModificationLoader.ReadModsFromString(string, out List{ValueTuple{Modification, string}})"/>
+        public static IEnumerable<Modification> ReadModsFromString(string storedModifications,
+            out List<(Modification, string)> filteredModificationsWithWarnings) =>
+            ModificationLoader.ReadModsFromString(storedModifications, out filteredModificationsWithWarnings);
+
+        /// <inheritdoc cref="ModificationLoader.ParseDissociationType(string)"/>
+        /// <remarks>
+        /// Now case-insensitive, where the copy this replaced matched exactly. More permissive rather
```
trishorts, line 43: "More permissive rather than different" undersells this by a lot. Case-insensitivity is the smaller half.

The switch this replaced had nine cases and `default: return null`. `DissociationType` has 22 members, so `ModDissociationType` used to return null for `PD`, `PSD`, `SID`, `BIRD`, `IRMPD`, `SORI`, `ISCID`, `UVPD`, `NETD`, `LowCID`, `aEPD`, `Unknown`, `AnyActivationType` and `Autodetect` -- fourteen of them -- and now resolves every one. Plus numeric strings, per the note on `ParseDissociationType`.

Every spelling the old switch accepted does still resolve to the same value, which is the sentence you wrote and it is true. The part missing is that the accepted *set* grew from 9 keys to all 22 plus integers. Worth stating outright here, since this remark is the only place a reader of the deprecated type would learn what changed.

mzLib/Test/DatabaseTests/TestDatabaseLoaders.cs
```diff
@@ -1752,5 +1752,237 @@ public static void DecoyWritingLoading_Xml()
...
+        /// <summary>
+        /// PtmListLoader forwards to ModificationLoader rather than carrying its own copy of the parser.
+        /// These pin the two together on every mods file mzLib ships, so a future change to one cannot
+        /// silently diverge from the other while the deprecated entry point is still public.
+        /// </summary>
+        [Test]
+        [TestCase("DataFiles", "Ontologies", "ptmlist.txt")]
+        [TestCase("ModificationTests", "", "ModsWithComments.txt")]
+        [TestCase("ModificationTests", "", "ProteaseMods.txt")]
+        [TestCase("Transcriptomics", "TestData", "RnaMods.txt")]
```
trishorts, line 1765: This is the `RnaMods.txt` under `Test/Transcriptomics/TestData`, which has zero `BM` and zero `BL` lines. The one that has them is `Omics/Resources/RnaMods.txt` -- 4 and 5 respectively -- and it lands in the test output as `Resources/RnaMods.txt` (9 matches in `Test/bin/Debug/net10.0-windows/Resources/RnaMods.txt`).

So the file that would have exposed the third drift is the one file not in this list. Adding

```csharp
[TestCase("Resources", "", "RnaMods.txt")]
```

covers the `BM`/`BL` path. It cannot fail today -- both sides are the same parser now -- but it is the case that makes the guard mean something for the shipped RNA mods file rather than only for the trimmed test copy.

mzLib/Test/DatabaseTests/TestDatabaseLoaders.cs
```diff
@@ -1752,5 +1752,237 @@ public static void DecoyWritingLoading_Xml()
...
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(viaDeprecated.Select(Describe).OrderBy(s => s),
+                    Is.EqualTo(viaReplacement.Select(Describe).OrderBy(s => s)));
+                Assert.That(deprecatedWarnings, Has.Count.EqualTo(replacementWarnings.Count));
+            });
+        }
+
+        private static string Describe(Modification m) =>
```
trishorts, line 1786: `Describe` covers id, type, mass, location restriction, diagnostic ions and neutral losses, but not the runtime type. With `BM`/`BL` in play the forward changes `Modification` into `BackboneModification` / `BaseModification`, and this comparison would report the two as equal through that change.

```csharp
private static string Describe(Modification m) =>
    $"{m.GetType().Name}|{m.IdWithMotif}|{m.ModificationType}|{m.MonoisotopicMass}|{m.LocationRestriction}|"
    + $"DI:{Flatten(m.DiagnosticIons)}|NL:{Flatten(m.NeutralLosses)}";
```

One field, and it makes the guard sensitive to the subtype split that the `BM`/`BL` cases produce.

mzLib/Test/DatabaseTests/TestDatabaseLoaders.cs
```diff
@@ -1752,5 +1752,237 @@ public static void DecoyWritingLoading_Xml()
...
+        }
+
+        /// <summary>
+        /// The deprecated ModDissociationType is a forward now, so it must give the same answer as the
+        /// method it forwards to for every key the old hand-written switch handled.
+        /// </summary>
+        [Test]
+        public static void DeprecatedModDissociationTypeAgreesWithItsReplacement()
+        {
+            foreach (string key in new[]
```
trishorts, line 1846: After the refactor this asserts `x == x`. `ModDissociationType` is an expression body that returns `ParseDissociationType(key)` verbatim, so the two sides of the equality are the same call and no change to `ParseDissociationType` can make this fail.

It would have had teeth against master's hand-written switch. As a forward guard it needs the expected values written down instead:

```csharp
[TestCase("Any", DissociationType.AnyActivationType)]
[TestCase("CID", DissociationType.CID)]
[TestCase("MPD", DissociationType.IRMPD)]
// ... ECD, PQD, ETD, HCD, EThcD, Custom
public static void DeprecatedModDissociationTypeStillResolvesEveryKeyTheOldSwitchHandled(
    string key, DissociationType expected)
```

with `NotADissociationType` kept as a separate null case. That pins the nine keys the old switch actually handled, which is the compatibility claim the test's summary is making.

PR #1183, opened 2026-08-19
Dispatch the mzIdentML version by null check, not by thrown exception

acesnik:
🤖 Follow-on to #1169, raised in review there. Stacked on it rather than folded in, so the 1.3.0 derivation stays reviewable on its own. Merge after #1169.

The branch is cut from #1169's head, so until that merges the diff here shows its commits too. The change this PR is asking you to review is the single commit `Dispatch the mzIdentML version by null check, not by thrown exception`, which touches only `MzidIdentifications.cs`.

`MzidIdentifications` holds one deserialized DTO per schema version — `dd110`, `dd111`, `dd120`, `dd130` — of which exactly one is non-null. Every accessor discovered which by dereferencing them in order and catching the `NullReferenceException`. 20 of the 23 accessors are indexed per match, so that ran once per identification per field read, and a 1.3.0 file threw and swallowed three NREs per call.

Each cascade is now a flat `if`/`else` on field nullity. The constructor keeps its `try`/`catch` — that is where the version is actually detected.

**It also stops the cascade misreporting genuine failures.** The catches were bare and the innermost arm had none, so a real exception from the correctly-matched version was swallowed and whatever the last arm threw escaped in its place, naming the wrong version.

## Verification

The four generated trees are structurally identical, so the transform was scripted and then checked by comparing every arm against the 1.1.0 arm with the version substituted. 20 of 23 matched exactly; the three that did not were inspected by hand and were cosmetic — line wrapping, indentation, a commented-out predecessor. `IsDecoy` was converted by hand: its 1.1.0 arm expressed the predicate as a nested `if` where the other three used a single `&&`.

The pre-refactor implementation was then held side by side against the new one and compared over every accessor, every index including one past the end on each axis, and all six checked-in fixtures:

| | |
|---|---|
| comparisons | 3,642 |
| value differences | **0** |
| thrown-type differences | 1,458 |

Every one of the 1,458 is `NullReferenceException` → `IndexOutOfRangeException` on an out-of-range index — the exception the access actually raises, rather than the one the last arm produced after the real one had been swallowed. The two 1.3.0 fixtures and `TestParentTolerance.mzid` show 0 of them, since for those the last arm was already the matching one.

## Behaviour change to be aware of

A caller catching `NullReferenceException` to detect a bad index will no longer see it. Nothing in mzLib does. I have not checked MetaMorpheus, which consumes this through the NuGet package.

The differential harness is not in the diff — it needed a verbatim copy of the old file, which is not something to keep in the tree.


trishorts:
@acesnik — merged master into this branch (`d8c70c78`) so the diff shows what the PR actually does. **Leaving it as a draft — that call is yours**, and I have not reviewed the change itself.

This was stacked on `feature/mzidentml-1.3.0` and had not been updated since #1169 merged, so its merge-base with master was `5ba13155` — 15 August. The diff was carrying all of #1169 alongside your own change:

```
before   7951 insertions, 475 deletions, 9 files
after     592 insertions, 761 deletions, 1 file   (net -169)
```

Nobody should have to read 7951 lines of already-merged code to find the 592 that matter, which is the only reason I did this before you asked.

**The conflict was 24 hunks but one decision.** Every one was the same shape — your `if (dd111 != null) ... else if (dd120 != null) ... else` against master's nested try/catch. That is the point of the branch repeated once per property, so your side won throughout.

Checked rather than assumed that this loses nothing from master: `MzidIdentifications.cs` has changed on master *only* through #1169, and this branch was stacked on #1169, so your side already contained that work. After resolving, the file still carries 50 `dd130` references, so 1.3.0 handling is intact.

```
mzIdentML tests    121 passed, 0 failed
Full offline suite 5887 passed, 0 failed, 29 skipped
```

Two things for you rather than me:

- The three remaining `try/catch` sites are the constructor's version detection, which runs once per file. I read that as deliberate — what this branch removes is the dispatch in the ~24 property accessors, which threw and caught on *every* property access for anything that was not 1.1.0, two nested throws per read for a 1.3.0 file. Worth saying so in the PR body, because "not by thrown exception" in the title reads like all of them went.
- The constructor populates `dd110`, but the property dispatch handles `dd111`/`dd120`/`dd130` with no `dd110` branch. That predates this PR as far as I can tell, and I have not touched it — but if it is the reason this is still a draft, it is the thing I would want to know.

Ping me when you mark it ready and I will review it properly.

trishorts (approved):
Approving. I merged master in again (`4de34406`, 13 behind, clean) and reviewed on that head.

**Both points from my 09-01 comment are addressed:** all 23 accessors now carry a `dd110` arm, and the body says the constructor's `try`/`catch` is deliberate version detection.

```
mzIdentML + mzML reader tests   158 passed, 0 failed   (net10.0, build output verified fresh)
```

**Checked whether a mismatched parse can leave every `ddXXX` null with no exception** — the case where the final `else` arm would NRE on `dd130`. It can't: `XmlSerializer` throws on a root-namespace mismatch, so the constructor either populates exactly one field or throws. Probed directly on this head:

| input | result |
|---|---|
| not XML | `InvalidOperationException` → `XmlException: Data at the root level is invalid` |
| empty file | `InvalidOperationException` → `XmlException: Root element is missing` |
| missing path | `DirectoryNotFoundException` |

So the flat `if`/`else` is safe without an all-null guard.

**One thing that is not this PR's:** a file declaring the schema's real namespace `http://psidev.info/psi/pi/mzIdentML/1.1` is rejected by every arm (`TestParentTolerance.mzid` with only its `xmlns` rewritten → `<MzIdentML xmlns='…/1.1'> was not expected`). No checked-in fixture uses that namespace, so the suite can't see it. #1171 fixes it, and a trial merge of #1171 onto this head is **conflict-free** despite both touching `MzidIdentifications.cs` — so the order they land in doesn't matter.

PR #1185, opened 2026-08-20
feat(nuget): ship the modification data files alongside the assemblies

acesnik:
🤖 Eight modification data files are maintained both here and in MetaMorpheus, or live only there despite not being application-specific. Five of the shared six have drifted, and one of the differences is a defect this repo inherited and never received the fix for. This packs the files so a consuming application can take them from the package instead of keeping its own copies.

## The drift

| file | mzLib | MetaMorpheus |
|---|---|---|
| `unimod.xml` | 2,233,804 B | identical blob |
| `Mods.txt` | 667 lines | 644 — 27 differing |
| `RnaMods.txt` | 204 lines | 195 — 11 differing |
| `ptmlist.txt` | UniProt 2026_01, 748 entries | 2018_01, 626 entries |
| `PSI-MOD.obo.xml` | 2002 terms | 2001 |
| `TMT.txt` | 16 `DR Unimod;` cross-references | absent |

Of the 27 differing lines in `Mods.txt`, **25 are lines mzLib already has** and 2 are MetaMorpheus's side of a two-way difference — so exactly one line needed changing here. The rest resolve on the MetaMorpheus side by deleting its copy.

## This changes output

`Omics/Resources/Mods.txt` was snapshotted from a MetaMorpheus tree predating December 2023, so it still carries the hydroxyproline diagnostic ion that two MetaMorpheus PRs were opened to correct:

```
170.060123533 -> 171.07642    smith-chem-wisc/MetaMorpheus#2324 ("wrong mass of course")
171.07642     -> 170.069143   smith-chem-wisc/MetaMorpheus#2325 (the value shipping today)
```

`26131d14` applies the correction. It is a separate commit so it can be split out or reviewed on its own.

The ion is reachable wherever the `Mods` registry feeds scoring or annotation — `PeptideWithSetModifications`, `OligoWithSetMods`, the ProForma converter, the external-result readers, the Koina clients. I have **not** measured the downstream effect on any of those; I am reporting the data change and its reach, not a quantified result.

The one remaining two-way difference is `Water Loss on E`, `DR Unimod; 27.` here (Glu→pyro-Glu) against `23.` there (the generic Dehydrated). Checked against the `unimod.xml` this PR ships, and this repo is right, so nothing changes here.

The entry is `TG E` with `PP Peptide N-terminal.`. Record 27 has exactly one specificity, `E` at Any N-term. Record 23 lists C, D, N, Q, S, T and Y — it has no `E` specificity at all. So the two are not interchangeable and 23 is not a looser-but-valid choice; it cites a record that does not cover the residue. The convention is also more consistent than it first looked: `Dehydroalanine on S` uses 23 because S *is* an Anywhere site for 23, while `Dehydroalanine on C` uses 368, so the rule is to cite the most specific record whose specificity actually matches.

That leaves MetaMorpheus's `23.` as the wrong value rather than the alternative one, resolved by smith-chem-wisc/MetaMorpheus#2749 deleting that copy.

## Two files adopted from MetaMorpheus

`c6f05629` brings over `substitutions.txt` (378 amino-acid substitution mods) and `surfactants.txt`, byte-for-byte. Both are plain modification lists in this repo's own ptmlist format with no MetaMorpheus-side parser — they are read by `ModificationLoader` through a generic directory loop. Amino-acid substitutions and surfactant adducts are not application-specific concepts, and they were in MetaMorpheus only because that is where they were written.

Payload only, deliberately **not** `EmbeddedResource`: mzLib does not load either file today, and embedding them would change the default set `Mods.AllModsKnown` exposes. Where the files live and what the library loads by default are separate decisions.

## A protease cleavage modification gets a real name

`protease_mods.txt` defined a modification called **`Test on M`**, which reads like a leftover fixture. It is not — `proteases.tsv:119` uses it as the cleavage modification for the shipped protease `CNBr_N`. It is the same chemistry as homoserine lactone (`C-1 H-4 S-1`) applied at the peptide N-terminus, because `CNBr_N` cleaves before M rather than after it.

`db95a520` renames it to **`Homoserine lactone N-terminal on M`** across the definition, the `proteases.tsv` reference, two test fixtures and the test file. Six files, all in this repo; MetaMorpheus has no reference to the old name.

The name is a proposal. It describes what the entry encodes, but anyone with a stronger claim on the nomenclature should overrule it in review.

This matters now because MetaMorpheus is about to start surfacing these modifications in its searchable list (smith-chem-wisc/MetaMorpheus#2749), where `Test on M` would be user-visible.

Verified against the built assembly rather than by reading:

```
embedded protease mods: 2
  IdWithMotif='Homoserine lactone on M'              OriginalId='Homoserine lactone'             loc=Peptide C-terminal.
  IdWithMotif='Homoserine lactone N-terminal on M'   OriginalId='Homoserine lactone N-terminal'  loc=Peptide N-terminal.

CNBr:     CleavageMod='Homoserine lactone on M'
CNBr_N:   CleavageMod='Homoserine lactone N-terminal on M'
CNBr_old: CleavageMod=<null>
```

## Mechanism

The original six stay `EmbeddedResource` for mzLib's own use — correct for a library, which cannot assume a file layout around it. All eight are additionally packed as a payload:

```
build/mzLib.targets
modificationData/Mods/{Mods,RnaMods,TMT,substitutions,surfactants}.txt
modificationData/Data/{ptmlist.txt,PSI-MOD.obo.xml,unimod.xml}
```

`build/mzLib.targets` links them into `Mods\` and `Data\` in the consumer's output. One physical file, two build actions; nothing about mzLib's own loading changes.

**Off by default.** The payload is ~6.8 MB and most consumers only want the embedded copies. Opt in with `<MzLibIncludeModificationData>true</MzLibIncludeModificationData>`.

The targets file lives at `mzLib/nuget/mzLib.targets` rather than `mzLib/build/`, because `.gitignore`'s `[Bb]uild/` rule swallows the latter — the nuspec would pack a path that isn't tracked.

## Verification

- `dotnet build mzLib.sln -c Release -p:EnableWindowsTargeting=true` — 0 errors.
- Packed and inspected the nupkg: all nine entries present at the paths above.
- Consumed it from a local feed in a two-level project chain (library → app) and confirmed the files reach the **app's** output, not just the directly-referencing library's. That transitive hop is what the MetaMorpheus side depends on.

Not run: this repo's test suite, which is Windows-targeted.

## Follow-up

MetaMorpheus deletes its copies and opts in, which needs a release carrying this payload first. That PR is up as a draft and pinned to a placeholder version.


nbollis (dismissed):

nbollis (commented):
Approved as it does not break anything, but what is the utility of shipping these mod files if the code will always read from the embedded resource?

acesnik:
Hi Nic, here's the accompanying PR in MM to remove the embedded resource and use the shipped mods lists https://github.com/smith-chem-wisc/MetaMorpheus/pull/2749. This will keep them from diverging by having them in one place.

trishorts:
Design looks right to me, and I checked it against its first consumer rather than in the abstract.

**Opt-in defaulting to false is the correct call**, and MetaMorpheus#2749 — the PR that deletes MetaMorpheus's bundled copies and takes them from here — does set it:

```xml
<MzLibIncludeModificationData>true</MzLibIncludeModificationData>
```

in `EngineLayer.csproj`. Worth stating because the failure mode if it had not is silent: the files simply would not appear, and `GlobalVariables` would come up with no modifications rather than erroring. Anyone reviewing #2749 should check that line is still there.

**The path math checks out.** `mzLib.targets` lands at `build\mzLib.targets`, so `$(MSBuildThisFileDirectory)..\modificationData\` resolves to the package root, which is where the nuspec `target` attributes put the two folders.

Two things I would want settled:

**1. `#2749` is blocked on this *plus a release*, not just on this merging.** It bumps to `1.0.586`, and `1.0.586`'s nuspec has no data-file entries — this PR is what adds them. So the order is: merge here → cut a release → then #2749's pin means something. Same shape as the queue #2718 and #2755 are already in. Worth saying in the description so nobody merges #2749 on the strength of the bump alone.

**2. `TMT.txt` is in the shipped set, which makes this the answer to half of #1213.** That issue is about mzLib having no production isobaric channel table; the blocker recorded there is that MetaMorpheus's `IsobaricMassTag` resolves reporter-ion m/z at runtime from `GlobalVariables.AllModsKnown` and mzLib has no equivalent source. Shipping `Omics\Resources\TMT.txt` gives mzLib one. Not a change request — but if you are already here, a line in the description connecting the two would save the next person the archaeology.

Related, since this is the second recent nuspec/file mismatch after #1179: this PR is a decent argument for #1145, the CI check that every `<file src>` in the nuspec resolves.

mzLib/mzLib.nuspec
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0"?>
+﻿<?xml version="1.0"?>
```
trishorts: This picked up a UTF-8 BOM — master starts `3c 3f 78`, this starts `ef bb bf`. `nuget pack` doesn't care, but it's an editor artifact rather than part of the change, and it's noise on a file two other open PRs (#1156, #1162) also touch. Worth reverting line 1.

mzLib/nuget/mzLib.targets
```diff
@@ -0,0 +1,29 @@
+<Project>
+
+  <!--
+    Ships the modification data files that mzLib embeds so a consuming application can also
+    expose them as editable files on disk. Off by default: the payload is ~6.8 MB, and most
```
trishorts: "Off by default: the payload is ~6.8 MB" reads as if the 6.8 MB were conditional. The `<file>` entries put all eight files in the .nupkg unconditionally; the property only gates copy-to-output. Every consumer downloads the 6.8 MB whether they opt in or not — worth saying that plainly here, since it's the part a consumer can't opt out of.

trishorts (commented):
Verified rather than read: the `Test on M` -> `Homoserine lactone N-terminal on M` rename is safe across the seam — MM master has no reference to the old name and ships no `proteases.tsv` or `protease_mods.txt` of its own. And `substitutions.txt`/`surfactants.txt` really aren't embedded: `Omics.csproj` uses an explicit `Include` list, and the built `Omics.dll` reports 6 manifest resources, not 8, so `Mods.AllModsKnown` is unchanged as the commit message claims.

One thing not worth its own thread: those two are now the only files in `Omics/Resources/` that aren't `EmbeddedResource`, and nothing in `Omics.csproj` records that as deliberate — the reasoning lives only in the commit message.

mzLib/nuget/mzLib.targets
```diff
@@ -0,0 +1,29 @@
...
+    Files land in Mods\ and Data\ beside the application, and flow to the output of any
+    project that references the one opting in.
+  -->
+
+  <PropertyGroup>
+    <MzLibIncludeModificationData Condition="'$(MzLibIncludeModificationData)' == ''">false</MzLibIncludeModificationData>
+    <MzLibModificationDataRoot>$(MSBuildThisFileDirectory)..\modificationData\</MzLibModificationDataRoot>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(MzLibIncludeModificationData)' == 'true'">
```
trishorts, line 26: Nothing in the repo exercises this file, so if the `modificationData\` layout in the nuspec and `MzLibModificationDataRoot` ever drift apart, pack still succeeds and consumers just silently get nothing.

I ran it against a synthetic package layout and all three claims in the comment above hold: a default build copies nothing, `-p:MzLibIncludeModificationData=true` lands all 8 files in `Mods\` and `Data\`, and they flow transitively to a project that merely references the one opting in. So this is a request for a regression test, not a bug report — something that asserts those two paths still agree.

mzLib/Test/ProteomicsTests/ProteolyticDigestion/TestProteaseDictionaryEmbeddedMods.cs
```diff
@@ -158,15 +158,15 @@ public static void LoadEmbeddedProteaseMods_ContainsHomoserineLactone()
         }
 
         /// <summary>
-        /// Verifies that the "Test on M" modification is loaded correctly.
+        /// Verifies that the "Homoserine lactone N-terminal on M" modification is loaded correctly.
         /// </summary>
         [Test]
         public static void LoadEmbeddedProteaseMods_ContainsTestOnM()
```
trishorts: The rename stopped at the strings — `LoadEmbeddedProteaseMods_ContainsTestOnM`, and the `testMod` locals at 167/225/297, still say "Test on M": the name this PR is removing precisely because it reads like a leftover fixture. Renaming the method finishes the job.

mzLib/mzLib.nuspec
acesnik: 🤖 Confirmed and fixed: this branch started `ef bb bf`, master starts `3c 3f 78`. Editor artifact, now stripped.

mzLib/nuget/mzLib.targets
acesnik: 🤖 You're right, and the wording was actively misleading. The `<file>` entries are unconditional, so every consumer downloads the 6.8 MB regardless; the property only decides whether the files get copied out beside the application. Comment rewritten to say which of the two is optional, since the non-optional half is the part a consumer would want to know about.

mzLib/nuget/mzLib.targets
acesnik: 🤖 Added, as `Test/NuGetPackagingTests.cs`. It reads the nuspec and the targets file as text rather than packing, so it runs anywhere, and asserts that the data root and its subfolders agree and that every file the nuspec ships exists.

Checked it actually fails on drift rather than merely passing today, with a distinct message per mode: renaming the root gives "the targets file reads from a different folder than the nuspec writes to", and renaming one subfolder gives "the nuspec ships modificationData subfolders that mzLib.targets does not copy".

The comment in `mzLib.targets` now also points at the test, so the next person to rename a folder finds out where the other half lives.

mzLib/Test/ProteomicsTests/ProteolyticDigestion/TestProteaseDictionaryEmbeddedMods.cs
acesnik: 🤖 Fixed — the method and all eight `testMod` locals. They are now `LoadEmbeddedProteaseMods_ContainsHomoserineLactoneNTerminal` and `nTerminalMod`, which also stops them reading as interchangeable with the `homoserineMod` local sitting next to them, since the two refer to different mods.

acesnik:
🤖 On the last point in your review — `Omics.csproj` now records why `substitutions.txt` and `surfactants.txt` are the only files in `Resources\` that are not `EmbeddedResource`, rather than that reasoning living only in a commit message. The note also says what the consequence of adding a file to `Resources\` is, since the explicit list means a new file does nothing until it is named there.

Also worth recording here, since your review asked about it indirectly: the `Water Loss on E` cross-reference is settled, and the answer is that this repo is already right. The entry is `TG E` with `PP Peptide N-terminal.`, and in the `unimod.xml` this PR ships, record 27 (Glu->pyro-Glu) has exactly one specificity — `E`, Any N-term — while record 23 (Dehydrated) lists C, D, N, Q, S, T and Y and does not include E at all. So 23 is not a matter of preference there, it cites a record that does not cover the residue. Nothing changes here; MetaMorpheus's copy is the one that is wrong, and #2749 removes it.

trishorts (approved):
All four points answered, and two of them better than I proposed.

- BOM gone; the nuspec starts `3c 3f 78` again.
- The targets comment now says the ~6.8 MB is not optional and that the property gates only the copy-out. That is the honest framing — I had suggested a footnote, this says it in the place a consumer will actually read.
- `NuGetPackagingTests` is exactly the missing gate: `ModificationDataLayoutAgreesBetweenNuspecAndTargets` fails if the nuspec `target=` paths and `MzLibModificationDataRoot` ever drift, which was the silent-breakage path I was worried about.
- `LoadEmbeddedProteaseMods_ContainsHomoserineLactoneNTerminal` finishes the rename.

Nothing further from me.

nbollis (approved):

acesnik:
🤖 Adding the specifics behind the answer above, since "keep them from diverging" undersells it: they have already diverged. Five of the six files maintained in both repos differ today — `Mods.txt` by 27 lines, `RnaMods.txt` by 11, `ptmlist.txt` is UniProt 2026_01 here against 2018_01 there, `PSI-MOD.obo.xml` by one term, and `TMT.txt` is missing its 16 Unimod cross-references there.

One of those differences is a defect rather than a lag. `Mods.txt` here was snapshotted from a MetaMorpheus tree predating December 2023, so it still carries a hydroxyproline diagnostic ion that two MetaMorpheus PRs were opened to correct. The fix never crossed back, and this PR applies it.

On the loading question specifically: you're right that mzLib reads its embedded copies and that does not change. Shipping the files is for consumers, not for us — an application that wants to expose modifications as user-editable files on disk currently has to vendor its own copy, which is exactly how the drift started. #2749 is that pattern applied to MetaMorpheus: delete its copies, take them from the package.

trishorts:
> 🤖 Adding the specifics behind the answer above, since "keep them from diverging" undersells it: they have already diverged. Five of the six files maintained in both repos differ today — `Mods.txt` by 27 lines, `RnaMods.txt` by 11, `ptmlist.txt` is UniProt 2026_01 here against 2018_01 there, `PSI-MOD.obo.xml` by one term, and `TMT.txt` is missing its 16 Unimod cross-references there.
> 
> One of those differences is a defect rather than a lag. `Mods.txt` here was snapshotted from a MetaMorpheus tree predating December 2023, so it still carries a hydroxyproline diagnostic ion that two MetaMorpheus PRs were opened to correct. The fix never crossed back, and this PR applies it.
> 
> On the loading question specifically: you're right that mzLib reads its embedded copies and that does not change. Shipping the files is for consumers, not for us — an application that wants to expose modifications as user-editable files on disk currently has to vendor its own copy, which is exactly how the drift started. #2749 is that pattern applied to MetaMorpheus: delete its copies, take them from the package.

i don't want the user to have any editable files. I prefer that they have companion custom files. but there is no reason they can't see the files.

PR #1186, opened 2026-08-20
Make the proline suffix follow the community convention (|P was inverted)

### LINKED ISSUE #1106: proteases.tsv: 'trypsin' and 'trypsin|P' mean the reverse of the MaxQuant/Mascot convention
`proteases.tsv` defines two tryptic entries:

```
trypsin      K|,R|
trypsin|P    K[P]|,R[P]|
```

Reading `DigestionMotif.ParseDigestionMotifs`, the bracket is `preventingCleavage` — the comment above it says *"find preventing cleavage"*. So:

| mzLib name | behaviour |
|---|---|
| `trypsin` | cleaves after K/R **always**, including before proline |
| `trypsin\|P` | cleaves after K/R **except** before proline — the classic Keil rule |

**This is the opposite of the convention every major search engine uses.** In MaxQuant, Mascot and Comet, `Trypsin/P` is the name for the rule that *ignores* proline — it "also cleaves at the carboxyl side of lysine or arginine, also if a proline follows" ([MaxQuant documentation](https://pharmazie.uni-greifswald.de/storages/uni-greifswald/fakultaet/mnf/pharma/biotechno/dokumente/MaxQuant_Infos_and_Tutorial_07.pdf); [Galaxy MaxQuant tutorial](https://training.galaxyproject.org/training-material/topics/proteomics/tutorials/maxquant-label-free/tutorial.html)), while plain `Trypsin` applies the Keil rule.

So the two names mean precisely the reverse of what a user coming from any of those tools expects, in **both** directions:

- Someone who selects `trypsin` expecting standard no-P behaviour gets cleavage before proline.
- Someone who selects `trypsin|P` expecting MaxQuant's `Trypsin/P` gets the proline rule enforced.

Neither produces an error. Both produce a plausible peptide list.

### Measured

Digesting P02768 (human serum albumin) with mzLib, 2 missed cleavages, minimum length 7, no modifications:

```
trypsin      200 peptides
trypsin|P    193 peptides
             22 peptides unique to 'trypsin'
             15 peptides unique to 'trypsin|P'
```

37 peptides differ out of ~200 — around 18% of the digest — determined entirely by which of two similarly-named entries was picked.

### Why this is worth changing

This is not a correctness bug: both rules are implemented correctly, and having both available is right. It is a naming and communication problem, and it is the kind that does not announce itself — the user gets a complete, sensible-looking peptide list either way and no indication that a choice was made on their behalf.

Worth noting that it caught a maintainer of this library, not just an outsider, which is roughly the definition of a trap worth removing.

### Suggested change

**Not a rename** — that would break every existing `.toml` and saved task in the wild. Two additive changes instead:

1. **Aliases matching the community convention**, so the familiar spelling does the familiar thing:
   - `Trypsin/P` → the same motif as the current `trypsin` (`K|,R|`, cleaves before proline)
   - `Trypsin` → the same motif as the current `trypsin|P` (`K[P]|,R[P]|`, Keil rule)

   Existing names keep working; new users get what they meant.

2. **A header comment in `proteases.tsv`** stating the motif syntax, in particular that `[X]` denotes a *preventing* residue rather than a required one. The file currently opens with a `# ===` divider and no explanation, so the notation has to be reverse-engineered from `DigestionMotif.cs`.

Happy to open a PR for either or both if the approach seems right — though the choice of which alias maps to which motif is a scientific call for the maintainers rather than mine.

### Context

Found while building a Python wrapper over mzLib ([pyMzLib](https://github.com/smith-chem-wisc/pyMzLib)), where a user asked for "trypsin not followed by proline" and the obvious-looking name turned out to be the wrong one.

trishorts:
> **Project of origin:** none — standalone fix (issue #1106)

Closes #1106.

## The problem

`proteases.tsv` used `|P` to mark the enzyme that **does not** cleave before proline. Every other search engine uses `/P` to mark the enzyme that **does**. So the names meant the reverse of what a user coming from MaxQuant, Mascot or Comet expects, in both directions:

| name | motif | mzLib behaviour | what the name says elsewhere |
|---|---|---|---|
| `trypsin` | `K\|,R\|` | cleaves before proline | Keil rule (does not) |
| `trypsin\|P` | `K[P]\|,R[P]\|` | Keil rule | — |

Neither name errors, and both produce a plausible peptide list. Digesting P02768 with 2 missed cleavages and minimum length 7, the two differ by **37 of ~200 peptides** — about 18% of the digest, decided entirely by which of two similarly-named entries got picked. #1106 has the measurement.

## Where it came from

Correction to an earlier version of this description, which claimed this PR restored a convention that two January 2026 commits had walked backwards. That was wrong — those commits (`bcbc8391`, `e2065d89`, `f8d0463d`) are PR-branch commits that were squashed and never entered master in that form.

What actually happened on master:

- Bare `trypsin` has been `K|,R|` — cleaving before proline — since at least `b36356fd` (2022-03-09). Four years.
- `#1005` (`eace48f3`, 2026-02-06) replaced the old parenthetical names with the `|P` suffix and **added** `trypsin|P` as a new restricted entry. It did not touch bare `trypsin`.
- The same commit **dropped** the unrestricted `chymotrypsin (cleave before proline)` (`F|,W|,Y|,L|`) and `Lys-C (cleave before proline)` (`K|`) variants, which have no replacement today.

So `trypsin` naming the enzyme that cleaves before proline is long-standing behaviour, not a recent regression. **This PR changes it rather than restoring it**, and that distinction matters for how it should be landed.

## The change

Six rows trade names. **No motif changes and no accession changes** — every enzyme in this file digests exactly as it did before.

```diff
-chymotrypsin|P   F[P]|,W[P]|,Y[P]|,L[P]|
+chymotrypsin     F[P]|,W[P]|,Y[P]|,L[P]|
-elastase|P       I[P]|,S[P]|,…
+elastase         I[P]|,S[P]|,…
-Lys-C|P          K[P]|
+Lys-C            K[P]|
-subtilisin|P     N[P]|,S[P]|,…
+subtilisin       N[P]|,S[P]|,…
-trypsin          K|,R|
-trypsin|P        K[P]|,R[P]|
+trypsin          K[P]|,R[P]|
+trypsin/P        K|,R|
```

The bare name is the proline-restricted enzyme; `/P` marks the variant that ignores the restriction. `/P` also matches the PSI-MS cleavage agent terms verbatim (`Trypsin/P` MS:1001313, `Lys-C/P` MS:1001310), which should make the accession column easier to get right.

`proteases.tsv` gains a header block stating the convention and the history, so the notation no longer has to be reverse-engineered from `DigestionMotif.cs`.

## Two variants restored

The `X (cleave before proline)` mapping above only means something if the target exists, and for the two names that actually shipped it did not. `#1005` dropped the unrestricted variants outright:

```diff
+chymotrypsin/P   F|,W|,Y|,L|
+Lys-C/P          K|
```

Both are the pre-`#1005` motifs restored verbatim. Before this, `"chymotrypsin (cleave before proline)"` and `"Lys-C (cleave before proline)"` — real names from real saved settings — had nothing to resolve to at all, and mzLib had no way to digest with unrestricted chymotrypsin or Lys-C.

Their accession columns are left **blank on purpose**. The old rows carried `MS:1001306`/`Chymotrypsin` and `MS:1001310`/`Lys-C/P`; the second is right and the first is not, but accessions are being corrected in a separate PR and this one should not pre-empt it.

## Backward compatibility

`|P` shipped in **1.0.578 through 1.0.585**, so those names are in saved settings. ProteaseGuru's default digestion agent is `"trypsin|P"` and it hardcodes the `|P` spellings in about eight places; MetaMorpheus master has `chymotrypsin|P` in `Task1-SearchTaskconfig.toml` and `Lys-C|P` in `MultiProteaseParsimonyTest`. All of them keep working:

- `NormalizeProteaseName` now maps `X|P` → `X` and `X (don't cleave before proline)` → `X`, and adds `X (cleave before proline)` → `X/P`.
- `DigestionParams` used the raw `ProteaseDictionary.Dictionary[...]` indexer, which **bypassed the normalization shim entirely** — a legacy name threw `KeyNotFoundException` rather than being remapped. It now calls `GetProtease`. This is what actually makes old tomls load.

Chose `/P` rather than swapping `|P`'s meaning precisely so that this works. Had `|P` been kept for the unrestricted variant, `"trypsin|P"` would have stayed a live key and silently inverted — ProteaseGuru's default would have flipped from Keil to unrestricted with nothing to detect it.

## The one breaking change, and its measured cost

**Bare `trypsin` now applies the proline restriction.** It has named the unrestricted enzyme for four years. Both spellings are live dictionary keys, so this cannot be remapped or detected at lookup time — it is documented in `proteases.tsv` and belongs in the release notes. Settings that meant the unrestricted enzyme must now say `trypsin/P`.

This is the intended correction — a user who writes `trypsin` almost certainly means the Keil rule, since that is what the name means in every other search engine — but it is not free, and the cost is larger than "a naming fix" suggests.

**Measured against MetaMorpheus master, on one machine, same worktree:**

| build | result |
|---|---|
| MetaMorpheus master + released mzLib 1.0.585 | **0 failed**, 1771 passed |
| MetaMorpheus master + this PR (+ the toml fix below) | **205 failed**, 1566 passed |

The failures are result changes, not lookup errors. Representative:

```
Assert.That(results[7], Is.EqualTo("All target PSMs with q-value <= 1: 84"))
Expected: "All target PSMs with q-value <= 1: 84"
But was:  "All target PSMs with q-value <= 1: 83"
```

MetaMorpheus has 203 test call sites that construct `DigestionParams` without naming a protease, so they take the default — which this PR redefines. **This PR therefore cannot merge on its own**; it needs a coordinated MetaMorpheus change, and it changes default search output for every MetaMorpheus user who does not pick a protease explicitly. That is a call for the maintainers, not something this PR should slip through.

## A separate MetaMorpheus bug this exposed

`TaskLayer/MetaMorpheusTask.cs:62` deserializes the protease from every task toml with the raw indexer:

```csharp
.FromToml(tmlString => ProteaseDictionary.Dictionary[tmlString.Value])
```

That bypasses `NormalizeProteaseName`, so the compatibility shim mzLib has carried since the rename has never applied on the one path that needs it. Because 1.0.585 ships `chymotrypsin|P` and not the parenthetical spelling, **any pre-2026-02 saved toml already throws on MetaMorpheus master today**, independent of this PR. `GetProtease` fixes it, exists in 1.0.585, and can merge into MetaMorpheus now against its current pin. Filed as smith-chem-wisc/MetaMorpheus#2750.

## Tests

Full suite green under CI's filter: **5564 passed, 0 failed, 29 skipped**.

Existing tests were renamed mechanically (`trypsin|P` → `trypsin`, and so on), which preserves the enzyme each one was exercising. Three GAPDH tests in `PeptideWithSetModsFullSequenceTests` had pinned exact peptide lists against the *unrestricted* enzyme via bare `"trypsin"`; they now say `"trypsin/P"`, so their expected values are untouched. No golden data was regenerated in this PR.

New tests in `ProteaseLoadingTests` pin the convention itself so it cannot invert a third time:

- `ProlineSuffixFollowsTheCommunityConvention` — direction of the suffix, not just the motifs
- `BareNameDoesNotCleaveBeforeProlineButSlashPDoes` — behavioural, digesting `KPEPTIDER`
- `NoShippedProteaseUsesThePipePSuffix` — guards against a `|P` name returning
- `HistoricalNamesResolveToTheSameEnzyme` — 9 cases, each asserted through `NormalizeProteaseName`, `TryGetProtease` **and** `DigestionParams`
- `BareTrypsinIsNotRemapped` — pins the one deliberate exception

## Out of scope

`proteases.tsv` also carries wrong PSI-MS accessions — `trypsin|P` claimed MS:1001313 (*Trypsin/P*) while being the restricted enzyme, `subtilisin|P` claims MS:1001312 which is *TrypChymo*, and `singleN`/`singleC` cite MS:1001957/MS:1001958, which are regex terms rather than cleavage agent names. Those are being fixed separately; this PR deliberately leaves the accession columns untouched.

## Downstream

MetaMorpheus and ProteaseGuru both load through `DigestionParams`, so they continue to work unchanged via the shim. Both would still benefit from a follow-up moving their hardcoded `|P` strings and GUI switch cases onto the current names.


trishorts:
Downstream migration issues filed for the two consumers that match protease names as strings:

- MetaMorpheus: smith-chem-wisc/MetaMorpheus#2750
- ProteaseGuru: smith-chem-wisc/ProteaseGuru#94

Neither is blocking — MM pins mzLib 1.0.585 and ProteaseGuru pins 1.0.580, so both are insulated until their pins are bumped past the release carrying this PR.

ProteaseGuru is the sharper case and is worth a look from anyone reviewing this: `trypsin|P` is its default digestion agent, and it compares names by string equality against `DigestionAgentName`, by dictionary key, and via the raw `ProteaseDictionary.Dictionary[...]` indexer. The indexer sites throw; the string-equality sites fail silently (half the default enzyme selection stops being pre-selected, and the individual-protein analyzer opens with no enzyme chosen). That asymmetry is the concrete reason this PR introduced `/P` instead of swapping what `|P` means — keeping `|P` for the unrestricted variant would have left `"trypsin|P"` a live key in ProteaseGuru that silently inverted meaning, changing every coverage number with nothing to detect it.

Filing also turned up a pre-existing MetaMorpheus bug unrelated to this PR: its GUI "recommended settings" switch still cases on `"chymotrypsin (don't cleave before proline)"` and `"elastase"`, names retired by the January 2026 rename, so those prompts have silently stopped firing. Details in MetaMorpheus#2750.

mzLib/Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
nbollis, line 1: Is this a general solution to the rename problem or does it only work for trypsin? What about the other enzymes that were renamed previously?

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
```diff
@@ -98,15 +121,15 @@
...
+elastase	I[P]|,S[P]|,L[P]|,V[P]|,N[P]|,T[P]|,K[P]|,A[P]|,G[P]|,Y[P]|,Q[P]|,R[P]|,F[P]|,D[P]|	full		Elastase	
 Glu-C	E|	full			
 Glu-C (with asp)	E|,D|	full			
-Lys-C|P	K[P]|	full	MS:1001309	Lys-C	
+Lys-C	K[P]|	full	MS:1001309	Lys-C	
 Lys-N	|K	full			
-trypsin	K|,R|	full	MS:1001313	Trypsin	
-trypsin|P	K[P]|,R[P]|	full	MS:1001313	Trypsin	
+trypsin	K[P]|,R[P]|	full	MS:1001313	Trypsin	
+trypsin/P	K|,R|	full	MS:1001313	Trypsin
```
nbollis: We are changing from |P to /P ? What is the intention with this change?

acesnik:
🤖 The direction of the inversion checks out, verified through the mechanism rather than the table: `[X]` parses into `DigestionMotif.PreventingCleavage` (`DigestionMotif.cs:78`) and a fitting cut is rejected when that residue is present (`:137-149`). So `K[P]|` is the Keil rule and `K|` is not, and master's line 108 (`trypsin  K|,R|`) really does name the unrestricted enzyme while line 109 (`trypsin|P  K[P]|,R[P]|`) names the restricted one. Backwards from `/P` everywhere else, in both directions, as you say.

Also confirmed the two compatibility claims: `DigestionParams.cs:20` uses the raw `ProteaseDictionary.Dictionary[protease]` indexer and so has never gone through `NormalizeProteaseName`, and `GetProtease`/`NormalizeProteaseName` are both on master — so the MetaMorpheus-side fix really can land against the current pin, independent of this.

**One thing the "no accession changes" framing hides.** It is true that no accession *string* moves, but the accession-to-behaviour pairing does, and for the row people actually use it moves the wrong way:

```
before:  trypsin      K|,R|          MS:1001313  Trypsin     <- MS:1001313 IS Trypsin/P, so correct
after:   trypsin      K[P]|,R[P]|    MS:1001313  Trypsin     <- now the restricted enzyme claiming Trypsin/P
after:   trypsin/P    K|,R|          MS:1001313  Trypsin     <- correct
```

So bare `trypsin` goes from correctly accessioned to incorrectly accessioned, and both rows still carry `MS:1001313`, meaning anything resolving a protease by accession cannot tell the restricted from the unrestricted one either before or after. Deferring the accession corrections is reasonable, but this particular row regresses in the interim and it is the default for most users — worth either fixing that one cell here (`MS:1001251` for the restricted trypsin) or saying explicitly in the description that bare `trypsin` is knowingly mis-accessioned until the accession PR lands.

For contrast, `Lys-C` and `chymotrypsin` come out *right* after the swap — `MS:1001309`/`MS:1001306` are the restricted terms — so trypsin is the only row where the naming fix and the accession disagree. That asymmetry is the argument for not treating them as fully orthogonal.

**I have not approved, and it is not the code.** You have already said this cannot merge alone; I agree, and the reason is worth restating as the review outcome rather than a caveat. Redefining bare `trypsin` changes default search output for every MetaMorpheus user who does not pick a protease, and 203 test sites take that default. That is a maintainers' call about changing published behaviour, not something a second approval should quietly clear. Happy to approve as soon as there is a paired MetaMorpheus PR to point at, since the mzLib half looks right to me.

The five new tests in `ProteaseLoadingTests` are the right set — `NoShippedProteaseUsesThePipePSuffix` and `BareTrypsinIsNotRemapped` in particular, since those are the two things a future edit would get wrong silently.

mzLib/Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
trishorts: The mapping is general; the **reach** is not, and your question found that.

`NormalizeProteaseName` is enzyme-agnostic — it strips either parenthetical form off whatever name precedes it, so `chymotrypsin (don't cleave before proline)`, `Lys-C (don't cleave before proline)` and the `|P` suffix all normalize for every enzyme, not just trypsin. Exactly one name is deliberately not remapped, bare `trypsin`, because both spellings are live dictionary keys and the change cannot be detected from the string; that carve-out is documented in the method.

But `integration` is red on this PR, and it is red for a reason that answers you better than I can:

```
Failed to convert TOML object with key 'Protease', type 'string' and value 'chymotrypsin|P'
  to object property with name 'Protease' and type 'Proteomics.ProteolyticDigestion.Protease'
```

MetaMorpheus's TOML converter looks the name up directly rather than going through `GetProtease`, so the compatibility shim never runs on the one path where old names actually persist — saved task files. A general mapping that the important caller does not call is not general in the way that matters. That is a real gap, not a test artifact.

I checked whether those failures might be pre-existing: `integration` passes on #1233 and #1244, so every failure on this PR is this PR's.

There are two of them, and both are consequences of the rename rather than bugs in it:

- **old TOMLs stop loading** — the shim gap above;
- **PSM and peptide counts move** (`428`, `374`, `143` in `AllResultsAndResultsTxt*`), because bare `trypsin` now names the restricted motif, so the digest genuinely differs.

Neither is fixable from mzLib alone. This needs a paired MetaMorpheus PR — route the TOML converter through `ProteaseDictionary.GetProtease`, and update the count expectations — landing with the mzLib release that carries this. I would rather agree that shape with you before writing it than push a cross-repo rename and have it argued afterwards.

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
trishorts: Not a rename of `|P` to `/P` — they mark opposite things, and that is the whole point of the change.

`|P` marked the enzyme that **does not** cleave before proline. Everywhere else — MaxQuant, Mascot, Comet, and the PSI-MS cleavage agent terms — the suffixed name marks the enzyme that **does**, and the bare name is the Keil rule. So mzLib's suffix meant the reverse of what the same-looking name means in every other engine, in both directions at once:

| name | motif | mzLib today | what that name means elsewhere |
|---|---|---|---|
| `trypsin` | `K\|,R\|` | cleaves before proline | Keil rule (does *not*) |
| `trypsin\|P` | `K[P]\|,R[P]\|` | Keil rule | cleaves before proline |

After this PR: bare `trypsin` is the Keil rule, and `trypsin/P` is the unrestricted one. `|P` is retired rather than renamed — a file that says `trypsin|P` normalizes to bare `trypsin`, which is the same motif it always named.

Why it is worth the disruption: neither name errors, and both produce a plausible peptide list, so choosing the wrong one is silent. Digesting P02768 at 2 missed cleavages and minimum length 7, the two differ by **37 of ~200 peptides** — about 18% of the digest, decided entirely by which of two similarly-named entries got picked. #1106 has the measurement.

The `/P` suffix also restores something #1005 dropped: `chymotrypsin (cleave before proline)` and `Lys-C (cleave before proline)` lost their unrestricted variants in February and have had no replacement since. `chymotrypsin/P` and `Lys-C/P` are those entries back under the conventional spelling.

See the other thread for where this stands — `integration` is red, and it needs a paired MetaMorpheus PR before it can go anywhere.

trishorts:
@nbollis both of your questions are answered above and I've resolved those threads — but I owe you a correction on the CI picture before you spend time here, because what I told you on 08-29 understated it.

I said there were "two of them, and both are consequences of the rename". There are **221 failing tests**, and I re-ran against fresh MM master (the branch was `behind`; I updated it and got the identical set) so this is not stale-master noise.

The 221 overstates the number of independent problems, though. Breaking it down:

- **136** are cascades: one fixture's `OneTimeSetUp` fails (`SpectrumMatchPlotTests.Setup`, on `metadrawLogic.FilteredListOfPsms.Any()`) and every test in it reports failed.
- **1** is the shim gap I described — `TestChymotrypsinCleaveAfterLeucine`, on `Protease = "chymotrypsin|P"` in a saved TOML, because MetaMorpheus's converter looks the name up directly instead of going through `GetProtease`.
- The remaining **~85** cluster on searches returning no PSMs at all — `FilteredListOfPsms` empty, `Sequence contains no elements`, and missing output files (`AllQuantifiedPeptides.tsv`, `AllTranscriptGroups.tsv`, protein group files).

That last cluster is what changes the shape of the fix. "Counts move because bare `trypsin` now names the restricted motif" explains numbers going *down*; it does not explain a search producing nothing, and it does not explain RNA and oligo tests failing at all, since they never touch trypsin. So I do not yet know that these are all consequences of the rename rather than a defect in it, and I am not going to propose updating expectations to match numbers I cannot account for.

Where that leaves the paired MetaMorpheus PR:

1. **Route the TOML converter through `ProteaseDictionary.GetProtease`.** This is worth doing on its own merits and does not depend on #1186 at all — `GetProtease` is already on master — so it can be a standalone MM PR that merges now.
2. **Diagnose the empty-PSM cluster** before agreeing on anything else. If it is the rename, the fix may belong in this PR rather than in expectations.
3. **Count expectations** last, and only for the ones that genuinely moved, landing with the mzLib release that carries this.

I'd still rather agree that shape with you than push a cross-repo rename and argue it afterwards — and step 2 might change my mind about step 3 entirely.

trishorts:
@nbollis diagnosis done. Short version: the 221 is real and it is this PR's, but it is roughly three problems, not 221, and only one of them is a defect.

**It is this PR.** I suspected MM master, since every passing control run predated the failing one by ~20h. So I re-ran #1247's integration unchanged against today's MM master: it passes. Not master. I also checked the `Unrecognized mod Oxidation on M` line that looked incriminating — it appears **117 times in the passing control run too**. Red herring.

I then built this branch as `mzLib 9.9.1186`, pinned a MetaMorpheus worktree at `86bfc8d8` to it, and ran the failures locally.

**1. Genuine, reproduces in isolation — the digest changed.** `TestProteinPrunedWithModSelection` and `...AndVariants` fail with `Sequence contains no elements`; `TestModificationAnalysisEngine` fails on `Carbamidomethyl on C\t84`. These run live searches with the *default* protease, which is exactly the name whose meaning this PR changes. Measured on a proline-rich sequence at 2 missed cleavages, min length 7: `trypsin` gives 17 peptides, `trypsin/P` gives 22, **4 shared**. MetaMorpheus's small curated fixtures were built around identifications that the restricted motif no longer produces. These are expectation updates, and they are legitimately this PR's bill.

**2. One real defect, one line.** `MetaMorpheusTask.cs:62`:

```csharp
.FromToml(tmlString => ProteaseDictionary.Dictionary[tmlString.Value])
```

The raw indexer, so `NormalizeProteaseName` never runs on the one path where old names actually persist. I verified the shim itself is fine — on this branch `trypsin|P`→`trypsin`, `chymotrypsin|P`→`chymotrypsin`, `Lys-C|P`→`Lys-C`, `(don't cleave before proline)`→bare, `(cleave before proline)`→`/P`, all resolving. mzLib is doing its job; MetaMorpheus is not calling it. **This is worth fixing on master today, independent of this PR** — old TOMLs saying `(don't cleave before proline)` already fail to load, and `GetProtease` has been there all along.

**3. The rest is cascade, not signal.** 136 of the 221 are a single fixture's `OneTimeSetUp` failing (`SpectrumMatchPlotTests.Setup`) and taking every test in it down. And the RNA/oligo tests — the ones I could not explain from a protein-only diff, and was right not to accept — **pass locally in isolation against this exact build**. `FindsSimpleSixmer` and `GptmdTask_Rna_AddsExpectedMethylationMods` both green. They only fail inside the full 1862-test run, which makes them order-dependent on the failures above rather than consequences of the rename. I have not pinned the exact pollution path, and I am not going to touch RNA expectations on the strength of a failure I cannot reproduce standing alone.

So the plan I proposed earlier survives, with step 2 answered:

1. **MM PR now, standalone:** route the TOML converter through `ProteaseDictionary.GetProtease`. Independent of this PR, mergeable immediately.
2. **MM PR paired with the mzLib release:** update the protein search expectations in (1) above — the ones that genuinely moved.
3. **Re-run integration after 1 and 2** and see what is actually left before anyone writes a line about RNA.

Happy to open (1) now if you agree it stands on its own.

acesnik (commented):
🤖 The diagnosis in your last two comments is the part I'd want on the record. Going from "two failures, both consequences of the rename" to 221, then refusing to accept your own explanation because it did not account for RNA and oligo tests failing on a protein-only diff, is what made the number usable — and re-running #1247's integration unchanged against today's MM master to rule master out, then finding `Unrecognized mod Oxidation on M` 117 times in the *passing* control run, is how the red herring got caught rather than written up. Splitting 221 into 136 cascade, ~85 real and 1 shim gap turns an unmergeable-looking PR into three tractable pieces.

Step 1 stands on its own, and I checked the specific claim rather than taking it: `MetaMorpheusTask.cs:62` is `.FromToml(tmlString => ProteaseDictionary.Dictionary[tmlString.Value])` on current MM master — the raw indexer, so `NormalizeProteaseName` never runs on the one path where retired names actually persist in the wild. `GetProtease` has been on master all along. That PR fixes TOMLs saying `(don't cleave before proline)` that already fail to load today, independent of anything here, so it should not wait on this.

What I have to add is in the accession column, which I don't think has come up. Both consumers of it write to files rather than to a screen: `MzIdentMLWriter.cs:588` puts `PsiMsAccessionNumber` into every `.mzID` MetaMorpheus emits, and `SdrfBuilder.cs:282` into SDRF — and `Readers.Sdrf` ships in the current pin, so that path is live now.

Two things there. The first is not yours: `trypsin` and `trypsin|P` both carry `MS:1001313` on master, so the two variants have never been distinguishable by accession. But this PR flips which motif the *default* name resolves to, and `MS:1001313` is the `Trypsin/P` term, so bare `trypsin` goes from an accession that describes its motif to one that does not — in an interchange file, on the path taken by every user who doesn't pick a protease. (I'm taking the CV meanings from the PSI-MS ontology rather than verifying them here; there's no OBO in the checkout. The structural half doesn't depend on that reading.)

The second is this PR's, and it's the cheaper one to settle: the three new `/P` rows don't agree with each other. `trypsin/P` carries `MS:1001313`, while `chymotrypsin/P` and `Lys-C/P` carry nothing at all. Since both writers emit the field unconditionally, a search with `chymotrypsin/P` produces an `accession` attribute with an empty value — which I'd expect to be schema-invalid where a `cvParam` requires one, though I haven't run it against the XSD.

The restricted rows come out right, incidentally, which is the argument against treating the naming and the accessions as fully orthogonal: `chymotrypsin` keeps `MS:1001306` and `Lys-C` keeps `MS:1001309`, both correct for the restricted motif after the swap. Trypsin is the only row where the swap and the accession disagree, and it's the default.

Holding the approval rather than blocking, for the same reason as before and now a second: the accession cells want a push, and I'd rather the enzyme term in an exported mzID be right on the same commit that changes what the default enzyme does. The five new `ProteaseLoadingTests` are the right set — `NoShippedProteaseUsesThePipePSuffix` and `BareTrypsinIsNotRemapped` especially, since those are the two a later edit would break silently.

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
```diff
@@ -98,15 +121,17 @@
...
+elastase	I[P]|,S[P]|,L[P]|,V[P]|,N[P]|,T[P]|,K[P]|,A[P]|,G[P]|,Y[P]|,Q[P]|,R[P]|,F[P]|,D[P]|	full		Elastase	
 Glu-C	E|	full			
 Glu-C (with asp)	E|,D|	full			
-Lys-C|P	K[P]|	full	MS:1001309	Lys-C	
+Lys-C	K[P]|	full	MS:1001309	Lys-C	
+Lys-C/P	K|	full			
 Lys-N	|K	full			
-trypsin	K|,R|	full	MS:1001313	Trypsin	
-trypsin|P	K[P]|,R[P]|	full	MS:1001313	Trypsin	
+trypsin	K[P]|,R[P]|	full	MS:1001313	Trypsin
```
acesnik: `MS:1001313` is the `Trypsin/P` term, and after the swap this row is the restricted motif — so the default protease now exports an accession that contradicts its own cleavage rule. It reaches files, not just the table: `MzIdentMLWriter.cs:588` and `SdrfBuilder.cs:282` both write `PsiMsAccessionNumber`. `MS:1001251` is the restricted term. (The two rows sharing one accession is pre-existing; what's new is which motif the default name lands on.)

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
```diff
@@ -98,15 +121,17 @@
 Name	Motif	Specificity	PSI-MS Accession	PSI-MS Name	Cleavage Modification
 Arg-C	R|	full	MS:1001303	Arg-C	
 Asp-N	|D	full	MS:1001304	Asp-N	
-chymotrypsin|P	F[P]|,W[P]|,Y[P]|,L[P]|	full	MS:1001306	Chymotrypsin	
+chymotrypsin	F[P]|,W[P]|,Y[P]|,L[P]|	full	MS:1001306	Chymotrypsin	
+chymotrypsin/P	F|,W|,Y|,L|	full
```
acesnik, line 125: This row and `Lys-C/P` carry no accession, while `trypsin/P` carries `MS:1001313` — three new rows, two conventions. Both writers emit the field unconditionally, so a `chymotrypsin/P` search yields an empty `accession` value, which I'd expect a `cvParam` schema to reject. `MS:1001307` and `MS:1001310` are the `/P` terms for these two if you want them filled.

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
trishorts: Fixed. `trypsin` now carries `MS:1001251` (Trypsin) and `trypsin/P` carries `MS:1001313` (Trypsin/P), so each row's accession matches its own motif. Verified both terms against the pinned PSI-MS CV rather than from memory: `MS:1001251` = Trypsin, `MS:1001313` = Trypsin/P.

`TestSdrfBuilder.WritesEveryColumnAsAControlledVocabularyTerm` was the tripwire pinned for exactly this moment — it asserted `AC=MS:1001313` with a note saying it should become `MS:1001251` once the tsv was corrected. That has fired and is updated, with the comment rewritten as a regression guard rather than a known-wrong pin.

mzLib/Proteomics/ProteolyticDigestion/proteases.tsv
trishorts: Half fixed, half deliberately left blank — and the second half is worth recording.

`Lys-C/P` now carries `MS:1001310`, which is correct.

`chymotrypsin/P` cannot: **`MS:1001307` is `CNBr`, not `Chymotrypsin/P`.** It is already used four rows above in this same file for the CNBr entry, so filling it in here would have given two different proteases one accession — the exact defect you flagged on the trypsin row. Checked the pinned PSI-MS CV: the only two `/P` cleavage-agent terms that exist are `MS:1001310` (Lys-C/P) and `MS:1001313` (Trypsin/P). There is no Chymotrypsin/P term to point at.

On the unconditional emission: `SdrfBuilder.CleavageAgentCell` already handles a blank accession on purpose — it writes `NT=` alone, and its comment names the nine MetaMorpheus proteases that have none. So SDRF output is schema-clean. `MzIdentMLWriter` does emit an empty `accession`, but it already does so on master for elastase, Glu-C, Glu-C (with asp), Lys-N, tryptophan oxidation and peptidomics; `chymotrypsin/P` joins an existing set rather than opening a new case. That is a MetaMorpheus-side fix and out of this PR's scope — I'll file it there.

trishorts: note: filed https://github.com/smith-chem-wisc/MetaMorpheus/issues/2785 for the mzIdentML empty-accession half.

PR #1190, opened 2026-08-22
Move the actions off Node 20

acesnik:
🤖 GitHub is forcing our Node 20 actions onto Node 24 and warning on every job. Each version below was checked against the action's own `action.yml` rather than assumed:

| action | | runtime |
|---|---|---|
| `actions/checkout` | v4 → **v5** | node20 → node24 |
| `actions/setup-dotnet` | v4 → **v5** | node20 → node24 |
| `actions/upload-artifact` | v4 → **v6** | node20 → node24 |
| `nuget/setup-nuget` | v1 → **v4** | node16 → node24 |
| `actions/checkout` (release.yml) | v2 → **v5** | |

Two of those were not in the warnings GitHub displayed, because it lists only the first few annotations per job: `upload-artifact` and `setup-nuget` are Node 20 and Node 16 respectively, and would have surfaced next. Note `upload-artifact@v5` is **still** node20 — v6 is the first release that is not.

The MetaMorpheus clone step also used `actions/checkout@master`, unpinned. It is now `v5` like the rest; a floating ref to a third-party action is worth removing on its own account.

**`codecov/codecov-action@v4` is deliberately left alone.** It is Node 20, so it will keep warning. v5 is a composite action with renamed inputs, and the current step has `files` and `verbose` under `env:` rather than `with:` — so they are not being passed as inputs at all today. Bumping means fixing that and re-confirming coverage still uploads, which is its own change rather than part of a version bump.

`release.yml` is not exercised by pull-request CI, so those two bumps are unverified until a release tag is pushed. The alternative was leaving `setup-dotnet@v1` in place.

All three workflow files still parse, and the three job names are unchanged (`build`, `external-service-tests`, `integration`), so branch protection is unaffected.

This does not touch the C# compiler warnings, which are a separate and much larger piece of work: a clean solution build produces 893 unique warnings, roughly 850 of them from the half-finished nullable-reference-type migration. Scoping that is its own conversation.


trishorts (approved):
Checked each bump against the action's own `action.yml` rather than the release notes alone:

- `setup-nuget` v1 → v4 — `action.yml` is byte-identical apart from `using:`, so `nuget-version: '6.x'` is safe across the biggest jump.
- `upload-artifact` v4 → v6 — input sets are identical (all 10 names), so `compression-level: 0` still applies.
- All five majors' release notes flag **only** the Node runtime; no input or behavior breaks.
- `DOTNET_VERSION` is `10.0.x`, well clear of the older runtimes `setup-dotnet@v5` dropped in #647.

`dotnet.yml` is proven by this run, not just parsed — the integration job executed both `checkout@v5`, the rewritten MetaMorpheus clone, and both `upload-artifact@v6` uploads, all green. `release.yml`'s input contracts are provably unchanged, so the half CI can't reach is low risk.

`check-thermo-rawfilereader.yml` has no `uses:` steps at all, so this covers every action reference in the repo.

Non-blocking, take or leave:

1. No `.github/dependabot.yml` exists — a `github-actions` ecosystem entry would stop this recurring every few months.
2. The codecov `env:`/`with:` deferral is well-reasoned, but a deferral with nothing tracking it evaporates; worth an issue.
3. These are the oldest node24 majors rather than the newest (`checkout` v7, `setup-dotnet` v6, `upload-artifact` v7 all exist and are node24) — fine as a floor, just noting it's a choice.

trishorts:
Filed the two non-blocking items from my review so they don't evaporate — neither is a gate on merging this:

- #1195 — codecov-action. Confirmed your read: `files` and `verbose` sit under `env:` and are never passed, so the glob has been inert and coverage only uploads because the action falls back to searching the workspace. One correction to the reasoning for deferring it, in your favour: `files`, `verbose`, `token` and `fail_ci_if_error` all still exist under those exact names in v7, so the input-rename risk is smaller than the v5 notes suggest. It is still its own change — passing `files:` for the first time narrows the upload and could move the reported percentage.
- #1196 — a `github-actions` Dependabot config, so this particular bit of toil doesn't come back on the next runtime deprecation.

Worth noting on the merge: this branch is 3 commits behind master and #1189 edited `dotnet.yml` in an adjacent region. It's still `MERGEABLE`, so git handles it — but an update-branch would dismiss the approval, so merge as-is rather than syncing first.

pcruzparri (approved):

PR #1191, opened 2026-08-22
Annotate the four optional parameters that accept null

acesnik:
🤖 `MassSpectrometry` has nullable disabled at the project level, but `IndexingEngine.cs` and `IsoDecDeconvolutionParameters.cs` each carry their own `#nullable enable`. Four signatures in those two files contradicted it by defaulting a non-nullable reference type to `null` — the source of the CS8625 warnings.

| | |
|---|---|
| `IsoDecDeconvolutionParameters` | `float[] mzWindow = null` → `float[]?` |
| `IndexingEngine.GetXic` | `Dictionary<..> matchedPeaks` → `Dictionary<..>?` |
| `IndexingEngine.GetXicByScanIndex` | same |
| the `matchedPeaks` value type | `ExtractedIonChromatogram` → `ExtractedIonChromatogram?` |

**Annotations only.** They are erased at compile time, so behaviour, output and IL are unchanged. This fixes what the signatures claim, not what they do — worth being plain about, since it buys no correctness on its own.

The value type is the one that says something a reader could not previously see. `GetAllXics` uses `matchedPeaks` as a set with an optional payload: a peak that formed an XIC maps to it, and a peak that was examined without forming one maps to `null` so it is not revisited. The value is only ever `ContainsKey`-tested, never dereferenced. The parameter doc now says so.

`IFlashLfqIndexingEngine` declares the same parameter and is deliberately left alone: FlashLFQ has nullable disabled with no per-file opt-in, so annotating it there produces an inert `?` and a fresh CS8632 — one warning traded for another. I tried it and measured before reverting.

884 → 880 warnings, none new anywhere, solution builds clean.


mzLib/MassSpectrometry/PeakIndexing/IndexingEngine.cs
```diff
@@ -82,10 +82,11 @@ public virtual bool IndexPeaks(MsDataScan[] scanArray)
...
         /// <param name="maxPeakHalfWidth"> the maximum distance from the apex RT of the XIC to both start RT and end RT </param>
         /// <param name="charge"> an optional parameter used only for IIndexedMass and massIndexingEngine; must be null for mz peak indexing </param>
-        /// <param name="matchedPeaks"> the dictionary that stores all the peaks already matched to an xic </param>
+        /// <param name="matchedPeaks"> the peaks already considered; a null value marks a peak that was
+        /// examined but did not form an xic, so it is not revisited </param>
         /// <returns> A list of IIndexedPeak objects, ordered by retention time </returns>
         public List<IIndexedPeak> GetXic(double m, double retentionTime, Tolerance ppmTolerance,
-            int missedScansAllowed, double maxPeakHalfWidth = double.MaxValue, int? charge = null, Dictionary<IIndexedPeak, ExtractedIonChromatogram> matchedPeaks = null)
+            int missedScansAllowed, double maxPeakHalfWidth = double.MaxValue, int? charge = null, Dictionary<IIndexedPeak, ExtractedIonChromatogram?>? matchedPeaks = null)
```
trishorts, line 89: `Dictionary<K,V>` is invariant in nullability, so this one isn't erased the way the other three are. A caller in a `#nullable enable` context still holding the old type now gets a warning it didn't get before:

```
warning CS8620: Argument of type 'Dictionary<IIndexedPeak, ExtractedIonChromatogram>' cannot be used
for parameter 'matchedPeaks' of type 'Dictionary<IIndexedPeak, ExtractedIonChromatogram?>' ...
```

Warning-level, and in-tree nothing but `GetAllXics` passes this, so I'd keep the change. Just worth amending "behaviour, output and IL are unchanged" in the description — the IL is, but the source contract does move for nullable-enabled consumers.

trishorts (approved):
Verified locally rather than by inspection:

- `ExtractedIonChromatogram` is a `class`, so the `?` is annotation-only — no `Nullable<T>`, no boxing. The IL claim holds.
- Built the parent commit (`8f4bc451`) and this branch in separate trees: 894 → 890 warnings, and the four removed are exactly the four CS8625 sites. The eight that *look* new are the same warnings shifted +2 lines by the added doc lines. Nothing new anywhere — the description's count checks out.
- The new doc comment is accurate: the value is only `ContainsKey`-tested (lines 180, 207), never dereferenced, and line 223 `Add(peak, null)` is the examined-but-rejected case.
- Leaving `IFlashLfqIndexingEngine` alone is the right call — FlashLFQ.csproj sets no `Nullable` and the file has no `#nullable enable`, so annotating it would emit CS8632 exactly as you say.
- Build clean, 71 XIC/indexing/IsoDec tests pass.

One inline note on the dictionary annotation, non-blocking.

Separately, worth knowing: #958 promotes this same dictionary to public API as `out Dictionary<IIndexedPeak, ExtractedIonChromatogram>` — non-nullable — documented as "every peak that ended up in an XIC, mapped to that XIC". This PR establishes that isn't true: `Add(peak, null)` (which #958 leaves intact) also puts examined-but-rejected peaks in there, so its `matchedPeaks[peak]` can hand a caller `null`. Merging this first makes that visible.

nbollis (approved):

PR #1192, opened 2026-08-22
Clear the nineteen compiler warnings that meant something, two of them defects

acesnik:
🤖 The nullable warnings are the bulk of the build noise but almost none of them mean anything. Nineteen do. Seven are fixed in the first commit, two of those being defects rather than annotations; the second commit removes twelve declarations that nothing reads.

## Seven warnings that meant something

**`IntensityNormalizationEngine` was dropping a constructor argument.** It takes `quantifyAmbiguousPeptides`, never stored it, so the field was permanently `false` and all four `CalculatePeptideResults` calls ignored what the caller asked for. Now assigned.

> **Output:** no in-repo caller passes the parameter — both take the default `false`, which is what the field already was — so nothing in this repository quantifies differently. It is a public constructor in a published package, so an external caller passing `true` was silently getting `false` and will now get `true`. That is the intended behaviour, but it is a change for them.

**`SpectralSimilarity._keepAllTheoreticalPeaks`** is a different shape of the same warning: no constructor takes it, so the field only ever held `false` and `&& !_keepAllTheoreticalPeaks` was always true. Field removed and the condition in `SpectralEntropy()` reduced accordingly, which is exactly equivalent. The two parameters its doc comment described do not exist on the method either.

**`FakeMsDataFile`** shadows three `MsDataFile` members without saying so. They are now `new`, which changes nothing — it states the shadowing that was already happening. Deliberately **not** `override`: `GetClosestOneBasedSpectrumNumber` returns a binary-search insertion point where the base returns the nearest scan, so overriding would push the double's answer onto every caller. A comment records the divergence. Whether the double should exist at all is a separate question with test consequences.

> Since this branch was opened, #1163 has rewritten the base `GetClosestOneBasedSpectrumNumber` to binary search internally. The divergence still stands and the comment is still accurate — the base converts its result back through `OneBasedScanNumber` and returns the nearest scan, the double returns a raw insertion point. The distinction is insertion point versus nearest scan, not linear versus binary. #1163 records the same hiding trap from the other side.

**`CrossClassComparisonTests`** compared a null variable with itself to exercise the `==` and `!=` overloads on `(null, null)`. The assertions are worth keeping, so a second null local replaces the self-comparison; the operator receives identical arguments.

## Twelve unused declarations

Eight locals, two fields, one duplicate `using`, and one local function, none of them read anywhere. All are pure declarations or constant initializers, so removing them cannot change behaviour.

One is not a removal: `SSRCalc3` has a `break` guarded by `if (!DUPLICATE_ORIGINAL_CODE)`, and the constant is `true`, so the compiler calls the `break` unreachable. Deleting it would silently break the const if anyone ever flips it, so it is suppressed with `#pragma warning disable CS0162` and a comment instead.

## Effect

Solution-wide `dotnet build mzLib.sln -c Debug -t:Rebuild`, this branch against master `ef537bb6`: **1027 → 1008 warnings**, with **no warning code appearing that master did not already have**. The whole delta is removals:

| Code | Meaning | master | here |
|---|---|---|---|
| CS0105 | duplicate `using` | 1 | 0 |
| CS0114 | member hides inherited member | 3 | 0 |
| CS0162 | unreachable code | 1 | 0 |
| CS0169 | field never used | 1 | 0 |
| CS0219 | value assigned but never used | 9 | 1 |
| CS0649 | field never assigned | 2 | 0 |
| CS1718 | comparison made to same variable | 2 | 0 |
| CS8321 | local function never used | 1 | 0 |

## Testing

Full suite on the merged head, same runsettings the required CI job uses: **5642 passed, 0 failed, 29 skipped**. `TestSpectralSimilarity`, which covers `SpectralEntropy()`, and `CrossClassComparisonTests` both pass.

A third commit adds one test. Simplifying the `SpectralEntropy()` guard left it as the only partially covered line in the patch: nothing in the suite had ever asked for the entropy of a similarity built without spectrum-sum normalization and without all peaks, so the `return null` arm had never run. The new test asserts all three cases — neither condition met, all peaks kept, and spectrum sum — which pins the simplified condition to the behaviour it replaced.


trishorts (commented):
Reviewed the whole diff. Every change is behaviour-neutral or correct, and I checked that rather than taking the description's word for it — full suite green on the merged head (5642 passed, 0 failed), 1027 -> 1008 warnings with no new code appearing.

Four notes inline. None is a defect this PR introduces; the first is a "which did you mean" I'd like answered before merge, the rest are notes for the record.

Things I specifically checked and found clean:

- The `SpectralEntropy` guard simplification is exactly equivalent — CS0649 proves the field was never assigned, so `!_keepAllTheoreticalPeaks` was always true.
- All twelve removals are pure. I checked `BuildSyntheticSpectrum` in particular, where the removed `proton`/`isoDelta` consts could have signalled a broken helper: it uses `.ToMz(charge)` and Averagine's own masses, so they really were leftovers.
- The SSRCalc pragma is the right call — `DUPLICATE_ORIGINAL_CODE` is `true` at :593, so the `break` at :1022 is genuinely unreachable, and suppressing beats deleting a guard that should survive flipping the const.
- The three `new`s change no dispatch. `GetMsScansInTimeRange` at MsDataFile.cs:128 calls `GetClosestOneBasedSpectrumNumber` virtually and gets the base, which is the stated intent.

mzLib/FlashLFQ/IntensityNormalizationEngine.cs
```diff
@@ -23,6 +23,7 @@ public IntensityNormalizationEngine(FlashLfqResults results, bool integrate, boo
             this.integrate = integrate;
             this.silent = silent;
             this.maxThreads = maxThreads;
+            this.quantifyAmbiguousPeptides = quantifyAmbiguousPeptides;
```
trishorts, line 26: Assigning the field is right, but it leaves the actual inconsistency in place.

Two adjacent lines in `FlashLfqEngine`:

```
:298  new IntensityNormalizationEngine(_results, FlashParams.Integrate, FlashParams.Silent, FlashParams.MaxThreads)
:302  _results.CalculatePeptideResults(FlashParams.QuantifyAmbiguousPeptides);
```

`FlashParams.QuantifyAmbiguousPeptides` exists and reaches the same method four lines later — just not this engine, which calls `CalculatePeptideResults` four times itself. The parameter has been unthreaded since #700.

So the net result is a public knob no in-repo caller sets, nothing tests, and whose only effect is on external callers, while :298 still silently uses `false`.

Two coherent endings: **delete** the parameter, if normalization is meant to run on unambiguous peptides only (also kills the CS0649 and removes the API trap), or **pass `FlashParams.QuantifyAmbiguousPeptides` at :298** with a test. Wiring the field and stopping is the one outcome nobody chose deliberately. Which did you mean?

mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
```diff
@@ -679,7 +679,6 @@ private LibrarySpectrum ReadLibrarySpectrum_ms2pip(StreamReader reader, bool onl
                     if (ind > 0)
                     {
                         bool readingModName = false;
```
trishorts, line 681: `bracketCount` wasn't dead weight — it was the surviving half of an incomplete copy-paste, and removing it takes the last flag off a real bug here.

Compare the two mod parsers:

- **:419-441** (MSP reader): `readingModName` set `true` on `[`, back to `false` when `bracketCount` returns to 0.
- **:681-690** (this one, `ReadLibrarySpectrum_ms2pip`): same comment *"be careful about spaces! mod names can have spaces in them"*, same `if (line[i] == ' ' && !readingModName) break;` — but the `[`/`]` handling was never copied.

Assignments to `readingModName` exist only at :431 and :439, both in the first parser. So here it is `false` forever, the guard is always true, and the loop breaks at the first space, truncating `Mods=`.

The removal itself is correct and behaviour-neutral — no objection. Filed separately so it isn't lost. I have not confirmed that real ms2pip files carry spaced mod names; the comment asserts it and the sibling reader implements it, which is suggestive, not proof.

mzLib/Test/FileReadingTests/FakeMsDataFile.cs
```diff
@@ -63,12 +66,12 @@ public override void InitiateDynamicConnection()
...
+        public new IEnumerable<MsDataScan> GetMS1Scans()
         {
             throw new NotImplementedException();
         }
 
-        public MsDataScan GetOneBasedScan(int scanNumber)
+        public new MsDataScan GetOneBasedScan(int scanNumber)
         {
             return Scans[scanNumber - 1];
         }
```
trishorts, line 77: The comment above `GetClosestOneBasedSpectrumNumber` justifies that one well, and `new` is right there. These two are a different case, and I think both are better served by deletion than by `new`.

`GetMS1Scans()` — the base is `virtual` and **works** (MsDataFile.cs:88). This throws. So `MsDataFile`-typed callers work and `FakeMsDataFile`-typed callers throw, which is backwards. Deleting it makes the fake strictly more capable.

`GetOneBasedScan(int)` — this body is `Scans[scanNumber - 1]`; the base is the same after a load guard (MsDataFile.cs:102), so they agree whenever the fake has scans. Deletion isn't free though: for an empty fake the base reaches `LoadAllStaticData()` -> `NotImplementedException`, where today you get `IndexOutOfRange`.

Pre-existing either way — but `new` converts "the compiler is telling you to decide" into "decided", so it belongs in this PR's conversation rather than after it.

mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
```diff
@@ -1,4 +1,4 @@
-using Chemistry;
+﻿using Chemistry;
```
trishorts, line 1: Nit: this adds a UTF-8 BOM. Master `ef537bb6` starts `75 73 69`, this head starts `EF BB BF`. Unrelated to the change — worth dropping to keep the file's first line out of the diff.

mzLib/Readers/SpectralLibrary/SpectralLibrary.cs
trishorts: Filed as #1200 so it isn't lost with the breadcrumb. Nothing needed in this PR — the removal here is correct.

trishorts (dismissed):
Approving. The diff does what it says: every change is behaviour-neutral or a real fix, verified against the merged head rather than from the description — full suite 5642 passed / 0 failed, 1027 -> 1008 warnings with no new warning code appearing.

Two disclosures rather than objections:

- I pushed the third commit here (the `SpectralEntropy` guard test that closed the patch-coverage gap), so this approval covers a commit I wrote. Worth a second reviewer's eye on that one specifically.
- The question in my `IntensityNormalizationEngine` comment is still open — delete the parameter, or thread `FlashParams.QuantifyAmbiguousPeptides` into `FlashLfqEngine:298` with a test. It does not block this PR: in-repo behaviour is unchanged either way, and the field assignment is strictly more correct than dropping the argument. Happy for it to be a follow-up.

#1200 tracks the ms2pip parser bug that the `bracketCount` removal made compiler-invisible. Nothing needed here for it.

trishorts (approved):

nbollis (approved):

PR #1193, opened 2026-08-24
Deliver quantification results to the entities, and implement QuantificationWriter

trishorts:
Addresses the delivery half of #996. Companion PR: SampleCollapseStrategy (harvested from #1000).

## The problem

`QuantificationEngine.Run` computed a `QuantMatrix` and returned it. Nothing ever wrote those values onto `IBioPolymerGroup.IntensitiesBySample`, so a large body of already-merged, already-tested output machinery could not be fed by a run:

- `Omics/BioPolymerGroup/SampleGroupResult.cs`
- `BioPolymerGroup.PopulateSampleGroupResults()` — including its per-channel isobaric branch
- `BioPolymerGroup.ToString()` / `GetTabSeparatedHeader()`

Separately, `QuantificationWriter`'s three methods all threw `NotImplementedException` while `WriteRawInformation` / `WritePeptideInformation` / `WriteProteinInformation` all defaulted to `true` — so a default-configured `Run()` threw. `QuantificationTests` works around this by setting all three to `false` immediately after `GetSimpleParameters()` at three separate sites.

## What this does

**`Omics/IHasSampleIntensities.cs` (new)** carries `SamplesForQuantification` and `IntensitiesBySample`. `IBioPolymerGroup` now inherits it and the two members move up unchanged, so every existing implementer still satisfies the interface.

They are on their own interface rather than on `IBioPolymerWithSetMods` deliberately: `PeptideWithSetModifications` is `[Serializable]` and serialized in bulk during indexing, so adding stored quantification state to it would enlarge every indexed peptide.

**`QuantificationEngine.OverwriteSampleIntensities`** writes the final matrix values back onto any row key that opts in. It is destructive by design — hence the name — and that is safe because `WriteRawData` captures the PSM-level values *before* any normalization or roll-up, so re-processing under different strategies never depends on what is overwritten here. Zero cells are omitted: the matrix uses `0` for "not observed", and storing it would make `SampleGroupResult.HasIntensityData` meaningless.

**`QuantificationResults`** was `{ string Summary; // TODO }`. It now carries `Success`, `Samples` (column order), `PeptideIntensities`, `ProteinIntensities`, `WrittenFiles`, and `OutputDirectory`. Peptide values ride here rather than on the peptide objects, for the serialization reason above.

**`QuantificationWriter`** writes `RawQuantification.tsv` (the reloadable PSM-level snapshot #996 calls for), `PeptideQuantification.tsv`, and `ProteinGroupQuantification.tsv`, each returning the path written. Repeated sample labels are disambiguated so a malformed experimental design produces an awkward header rather than silently overwritten columns.

## Where output goes

**The three `Write*Information` flags stay `true` by default.** `RawQuantification.tsv` is what makes a search re-quantifiable under different strategies without re-searching, so it has to be something you opt *out* of.

**`OutputDirectory` is optional.** When writing is on and the caller set no directory, the engine derives one from the spectral matches' `FullFilePath`s and writes beside the data. The directory actually used comes back on `QuantificationResults.OutputDirectory`, so a caller who relied on the default can find the files. An explicit `OutputDirectory` always wins.

Three rules in the derivation:

- **Only absolute paths count.** A bare file name or relative path resolves against the process's working directory — the exact ambiguity this default exists to remove — so those matches are ignored rather than silently anchored to wherever the caller happened to be.
- **Several directories collapse to their nearest common ancestor.** A search over `data/fraction1` and `data/fraction2` writes to `data`. Comparison is per path segment, so `C:\dataset` and `C:\dataset2` are siblings rather than one being a prefix of the other.
- **A root-only ancestor is rejected.** Files on unrelated volumes, or in unrelated top-level folders, have no shared home, and `C:\` is not "beside the data".

`ValidateEngine` fails only when there is neither a set directory nor a derivable one, and names the fix:

```
QuantificationEngine Error: OutputDirectory is not set and no default could be derived,
because the spectral matches' file paths are missing, relative, or span unrelated
directories. Set OutputDirectory, or turn WriteRawInformation, WritePeptideInformation
and WriteProteinInformation off to quantify without writing files.
```

There is deliberately **no fall back to the working directory**. That is how output ends up somewhere nobody looks.

### Note for the caller that comes next

MetaMorpheus's `PostSearchAnalysisTask` should pass its **task output directory explicitly** rather than relying on this default. If it does not, quantification files land beside the user's raw data instead of in the task folder. The beside-the-data fallback is for FlashLFQ-style callers that hold real file paths and no output folder of their own.

`QuantificationParameters.GetSimpleParameters()` is test- and development-only — hence `internal`, and it now says so in its doc comment. It keeps writing off, which matters more with the fallback in place: a write-enabled run from it would otherwise leave files next to whatever test data the test pointed at.

## Blast radius

Nothing outside `Test/` and `Development/` constructs a `QuantificationEngine`; MetaMorpheus does not reference the `Quantification` namespace at all; `IBioPolymerGroup` has exactly one implementer. No production behaviour changes.

## Tests

`Test/Quantification/QuantificationDeliveryTests.cs`, 21 tests — write-back correctness against the matrix, unobserved-sample omission, the results tables, failure results, the three output files and their contents, column-label disambiguation, the write-flag defaults, output landing beside the data with nothing configured, an explicit directory winning over the derived one, and `TryGetSourceFileDirectory` over bare names, relative paths, one directory, sibling directories and a root-only ancestor.

989 tests across `Test.Quantification`, `Test.Omics` and `FlashLFQ` pass on the current head, and the solution builds with 0 errors.


mzLib/Quantification/QuantificationEngine.cs
```diff
@@ -60,41 +60,113 @@ public QuantificationResults Run(out QuantMatrix<IBioPolymerGroup> proteinMatrix
...
 
-        Task peptideWriter = null;
+        Task<string> peptideWriter = null;
         if (Parameters.WritePeptideInformation)
         {
-            peptideWriter = Task.Run(async () =>
-            {
-                QuantificationWriter.WritePeptideMatrix(peptideMatrix, Parameters.OutputDirectory);
-            });
+            peptideWriter = Task.Run(() =>
```
Alexander-Sol, line 85: This task should remain async

mzLib/Quantification/QuantificationEngine.cs
```diff
@@ -60,41 +60,113 @@ public QuantificationResults Run(out QuantMatrix<IBioPolymerGroup> proteinMatrix
...
+    /// setting both <see cref="IHasSampleIntensities.SamplesForQuantification"/> (the column order) and
+    /// <see cref="IHasSampleIntensities.IntensitiesBySample"/> (the values).
+    /// </summary>
+    /// <remarks>
+    /// Row keys that do not implement the interface are skipped, so this is safe to call on a peptide
+    /// matrix even though <see cref="IBioPolymerWithSetMods"/> does not carry quantification state.
+    /// Zero-valued cells are omitted: the matrix uses 0 to mean "not observed in this sample", and
+    /// omitting them keeps <c>SampleGroupResult.HasIntensityData</c> meaningful.
+    /// </remarks>
+    internal static void ApplySampleIntensities<T>(QuantMatrix<T> matrix) where T : IEquatable<T>
```
Alexander-Sol: This makes me nervous. The name is ambiguous, "OverwriteBiopolymerIntensities" would be more clear.

The main issue is that overwriting things at the peptide level prevents us from re-processing the data later. If we keep the raw data, we can try different sample processing methods. If we overwrite it, that information is lost. I like the idea of spitting out a separate quant file and keeping raw values in the peptide.

As i'm typing this, I realize that keeping raw values in the PSM output should be sufficient. So, i'm cool with this change. But I would still prefer a more specific name for the function

mzLib/Quantification/QuantificationParameters.cs
```diff
@@ -37,12 +37,20 @@ public class QuantificationParameters
         /// <summary>
         /// If true, the quantification engine will write raw quantification information to disk.
         /// This enables re-analysis later using different normalization or roll-up strategies without re-processing the raw data.
+        /// Off by default: writing is a side effect, and <see cref="OutputDirectory"/> defaults to the
+        /// working directory, so callers opt in rather than discover files they did not ask for.
```
Alexander-Sol: I'm not sure what this means. "writing is a side effect" implies that the raw information will always be written. Is this true? I don't know the rationale behind these changes

Alexander-Sol (commented):
Questions on this one

mzLib/Quantification/QuantificationEngine.cs
trishorts: I think this one is a false alarm from the diff — the task is still async. It just lost a keyword that was doing nothing.

Before:

```csharp
rawWriter = Task.Run(async () =>
{
    QuantificationWriter.WriteRawData(SpectralMatches, Parameters.OutputDirectory);
});
```

After:

```csharp
rawWriter = Task.Run(() =>
    QuantificationWriter.WriteRawData(SpectralMatches, Parameters.OutputDirectory));
```

`WriteRawData` is synchronous, so the `async` lambda had no `await` in it — the compiler warns CS1998 ("this async method lacks 'await' operators and will run synchronously") and builds a state machine for nothing. Both forms hand the work to `Task.Run`, so both run on a thread-pool thread and both still overlap with `RunPeptideQuant`. The only real change is that the lambda now returns the path it wrote, so `QuantificationResults.WrittenFiles` can report it.

That said, your instinct is pointing at something real: this was never *truly* async — it is synchronous `StreamWriter` I/O on a pool thread either way. If you want genuine async I/O, the change is `WriteLineAsync` inside the writer and `Task<string> WriteRawDataAsync(...)` on its surface. Happy to do that here or leave it for when the file sizes justify it — say which you'd prefer.

mzLib/Quantification/QuantificationEngine.cs
trishorts: Renamed to `OverwriteSampleIntensities` — you're right that the old name undersold what it does.

I went with "Sample" rather than "Biopolymer" only because the method is generic over anything implementing `IHasSampleIntensities`, and the thing it actually writes to today is an `IBioPolymerGroup` — a group rather than a biopolymer. The verb was the important half of your suggestion and that's kept. Say the word if you'd rather have your exact name.

The remark now states the destructiveness outright instead of leaving it implied, and records the conclusion you reached mid-comment, since it is the load-bearing reason this is safe:

> This is destructive by design — hence "Overwrite". [...] The pre-normalization PSM values are not lost: `QuantificationWriter.WriteRawData` records them before any normalization or roll-up runs, which is what makes re-processing with different strategies possible without re-searching.

That ordering is deliberate in `Run()` — the raw write is kicked off before `RunPeptideQuant`, so what lands in `RawQuantification.tsv` is what came in, untouched by any strategy.

Your point also changed the third comment's answer: since the raw file is what buys us re-processing, defaulting it off was wrong. See that thread.

mzLib/Quantification/QuantificationParameters.cs
trishorts: Fair — that comment explained nothing, and pushing on it made me change the behaviour rather than just the wording. Reverted the flip and gated it instead.

**Why they were flipped.** On master the three flags default to `true` while all three `QuantificationWriter` methods `throw new NotImplementedException()`, so any default-configured `Run()` throws. `QuantificationTests` works around it by setting all three to `false` right after `GetSimpleParameters()` — lines 296-298, 425-427 and 628-630. Once this PR implements the writer that trap is gone, but a new one appears: `OutputDirectory` also defaults to empty, so `true` by default writes three files into the working directory of whatever called it. Flipping to `false` was the blunt fix.

**Why that was the wrong fix.** Your other comment treats the raw file as the thing that lets us re-process later, and I agree. Making it opt-in means the common path silently gives that up.

**What it does now.** Defaults stay `true`, and `ValidateEngine` rejects the combination that was actually the problem:

```
QuantificationEngine Error: OutputDirectory must be set when WriteRawInformation,
WritePeptideInformation or WriteProteinInformation is enabled. Set an output directory,
or turn the write flags off to quantify without writing files.
```

So a caller who forgets the output directory is told, in the same way the engine already reports a null experimental design or an empty PSM list, rather than discovering files in the working directory. `GetSimpleParameters` keeps writing off because it supplies no `OutputDirectory`.

The doc comments now give the rationale rather than asserting it. Two tests cover it: one that a write-enabled run with no output directory comes back unsuccessful with that message and without touching the entities, and one that quantifying with writing off needs no output directory at all.

mzLib/Quantification/QuantificationEngine.cs
Alexander-Sol: Good catch! Let's leave it for now, I can create an issue so that we don't forget

mzLib/Quantification/QuantificationParameters.cs
```diff
@@ -56,9 +70,10 @@ internal static QuantificationParameters GetSimpleParameters()
                 ProteinNormalizationStrategy = new NoNormalization(),
                 OutputDirectory = string.Empty,
                 UseSharedPeptidesForProteinQuant = false,
-                WriteRawInformation = true,
-                WritePeptideInformation = true,
-                WriteProteinInformation = true
+                // No OutputDirectory, so writing must stay off or the engine will reject the run.
```
Alexander-Sol: Can we add a default output directory? Maybe parse the filepath for source files and use that to determine the output directory? Then we can set the "Write" bools == true and allow for the default ouput directory to be overwritten

Alexander-Sol: Are the "SimpleParameters" just used for internal testing? If so, this is fine. Just add a comment that states that

Alexander-Sol (dismissed):
If the GetSimpleParameters() method is only used for internal testing, then I'm ready to approve

mzLib/Quantification/QuantificationParameters.cs
trishorts: Both done — pushed.

**Yes, `GetSimpleParameters` is test-only.** It is `internal`, and its five call sites are all in `Test` and `Development`. It now says so:

```csharp
/// A no-op configuration — no normalization, no collapsing, sum roll-ups — for the test and
/// development projects only. It is deliberately internal: production callers should
/// construct a QuantificationParameters and choose strategies for their data.
///
/// Writing is switched off so that a test run leaves nothing behind. That matters more than it
/// looks: with no OutputDirectory, a write-enabled run would fall back to writing
/// beside whatever data files the test pointed at.
```

That last paragraph is because of your other point, which I took:

**The default output directory is in.** `OutputDirectory` is optional again. When writing is on and the caller set nothing, the engine derives a directory from the spectral matches' `FullFilePath`s and writes beside the data. An explicit `OutputDirectory` always wins, and the directory actually used comes back on `QuantificationResults.OutputDirectory` so a caller who relied on the default can find the files.

Three rules in the derivation, each of which I think earns its keep:

- **Only absolute paths count.** A bare file name or a relative path resolves against the process's working directory — which is the exact ambiguity the default exists to remove — so those matches are ignored rather than silently anchored to wherever the caller happened to be.
- **Several directories collapse to their nearest common ancestor.** A search over `data/fraction1` and `data/fraction2` writes to `data`. Comparison is per path segment, so `C:\dataset` and `C:\dataset2` are siblings rather than one being a prefix of the other.
- **A root-only ancestor is rejected.** Files on unrelated volumes, or in unrelated top-level folders, have no shared home, and `C:\` is not "beside the data".

Validation now fires only when there is neither a set directory nor a derivable one, and says which:

```
QuantificationEngine Error: OutputDirectory is not set and no default could be derived,
because the spectral matches' file paths are missing, relative, or span unrelated
directories. Set OutputDirectory, or turn WriteRawInformation, WritePeptideInformation
and WriteProteinInformation off to quantify without writing files.
```

Four tests: output lands beside the data with nothing configured, an explicit directory wins over the derived one and leaves the data folder untouched, and `TryGetSourceFileDirectory` is covered directly over bare names, relative paths, one directory, sibling directories, and a root-only ancestor.

One judgement call worth flagging: I did **not** make the fallback the working directory when nothing can be derived. Falling back there is how output ends up somewhere nobody looks, and MetaMorpheus will always pass a task output directory anyway — the fallback is for FlashLFQ-style callers holding real file paths. Say the word if you'd rather it just wrote to the CWD instead of refusing.

trishorts:
Both of your round-2 points are pushed, and I've rewritten the PR description, which had gone stale in a way that mattered — it still said *"the three `Write*Information` defaults flip to `false`"*, which is now the exact opposite of what the PR does. It also still named `ApplySampleIntensities`. Worth a re-read of the **Where output goes** section rather than trusting what was there before.

Three things that aren't in the inline thread:

**The next caller has to opt out of the new default.** MetaMorpheus's `PostSearchAnalysisTask` should pass its task output directory *explicitly*. If it just leaves `OutputDirectory` empty, quant files land beside the user's raw data instead of in the task folder — which is right for a FlashLFQ-style caller holding file paths and no output folder, and wrong for a MetaMorpheus task that has one. I've written that down so it doesn't get lost when #1193 unblocks that work.

**There is deliberately no fall back to the working directory.** When nothing can be derived, the run is rejected with a message naming the fix. Falling back to the CWD is how output ends up somewhere nobody looks. Easy to change if you disagree — it's one branch in `ValidateEngine`.

**Verification.** `QuantificationDeliveryTests` is 21 tests; 989 pass across `Test.Quantification`, `Test.Omics` and `FlashLFQ` on the current head, and the solution builds with 0 errors. The branch also picked up the master merge that had landed on it.

On the async point — happy for you to file that issue; if you'd rather I did, say so and I'll open it with the CS1998/`WriteLineAsync` detail from that thread so it isn't lost.

nbollis (approved):

trishorts:
@Alexander-Sol — this is ready for another look when you have a minute. Your approval got dismissed automatically when I pushed the round-2 changes, so it needs a fresh one.

**What changed since you approved**, both from your two comments:

- `OutputDirectory` is optional again. With writing on and nothing set, the engine derives a directory from the spectral matches' `FullFilePath`s and writes beside the data, reporting it back on `QuantificationResults.OutputDirectory`. Only absolute paths count, sibling directories collapse to their nearest common ancestor, and a root-only ancestor is refused. An explicit `OutputDirectory` always wins.
- `GetSimpleParameters` now says it is test- and development-only, which it is — `internal`, five call sites, all in `Test` and `Development`.

I also rewrote the PR description, which had gone stale in a way that mattered: it still said *"the three `Write*Information` defaults flip to `false`"*, the exact opposite of what the PR now does. The **Where output goes** section is the part worth re-reading.

**Since then:** #1194 is merged, and I've updated this branch from master. That merge was not purely textual — your collapse split changed the `CollapseSamples` signature this engine calls — so I verified it locally rather than trusting the merge: it resolved to the two-argument form with `CollapseAggregationStrategy`, the solution builds clean, and 1024 tests pass across `Test.Quantification`, `Test.Omics` and `FlashLFQ` on the merged head. CI is re-running now.

**One judgement call worth your eye before you approve:** when no directory can be derived, the run is rejected with a message naming the fix, rather than falling back to the working directory. I think falling back to the CWD is how output ends up somewhere nobody looks, but it is one branch in `ValidateEngine` if you disagree.

And a heads-up on what this unblocks: MetaMorpheus now calls the engine for TMT searches, so a TMT search produces per-channel peptide and protein tables instead of stopping at reporter-ion columns in the .psmtsv. That work needs an mzLib release carrying this PR and #1194, so this is the gate.

Alexander-Sol (approved):

PR #1194, opened 2026-08-24
Add SampleCollapseStrategy: collapse one sample dimension at a time

trishorts:
Harvests the one piece of #1000 that never reached master. Companion PR: #1193.

## The problem

`SumCollapse` and `MeanCollapse` both group by a single hardcoded key:

```csharp
ColumnKeys.GroupBy(s => s.Condition + '_' + s.BiologicalReplicate)
```

so they collapse fraction *and* technical replicate *and* everything else in one step. There is no way to sum fractions while keeping technical replicates — the usual bottom-up recipe.

## What this does

Collapsing is two independent questions, and this PR gives each one its own strategy interface:

- **`ICollapseStrategy` decides *which* samples are combined.** `NoCollapse`, `CollapseFractions`, `CollapseTechnicalReplicates`, `CollapseBiologicalReplicates`, `CollapseFractionsAndTechnicalReplicates`.
- **`IAggregationStrategy` decides *how* a group is reduced to one value.** `SumAggregation`, `MeanAggregation`, `MedianAggregation`, `MaxAggregation`.

```csharp
public interface IAggregationStrategy
{
    string Name { get; }
    double Aggregate(ReadOnlySpan<double> values);
}

public interface ICollapseStrategy
{
    string Name { get; }
    QuantMatrix<T> CollapseSamples<T>(QuantMatrix<T> m, IAggregationStrategy aggregation)
        where T : IEquatable<T>;
}
```

`QuantificationParameters` carries both — `CollapseStrategy` and `CollapseAggregationStrategy` — so reading the parameters still fully describes how the data was processed. The aggregation is a *per-stage* parameter rather than one shared setting, because summing fractions while taking a median at protein roll-up is a reasonable thing to want.

Composing the recipe the old types could not express:

```csharp
var summed = new CollapseFractions().CollapseSamples(matrix, new SumAggregation());
var final  = new CollapseTechnicalReplicates().CollapseSamples(summed, new MedianAggregation());
```

`CollapseStrategyBase` holds the grouping, the aggregation loop and the collapsed-column identity; a subclass declares only which of the three sample dimensions it collapses. `Aggregate` takes a `ReadOnlySpan<double>` so the base can pass a slice of a pooled buffer without copying.

### Why not one parameterized strategy

The first version of this PR was a single `SampleCollapseStrategy` parameterized on `CollapseDimension x AggregationType`. @Alexander-Sol's review rejected it, correctly: one type that can do twelve different things depending on its constructor arguments means `QuantificationParameters` no longer tells you how the data was processed, and every other strategy in the project (`SumRollUp`, `MedianRollUp`, `GlobalMedianNormalization`) encodes exactly one behaviour per type. **One behaviour per strategy type** is the rule this PR now follows.

### Breaking: `SumCollapse` and `MeanCollapse` are removed

Both grouped by condition and biological replicate, so both are now `CollapseFractionsAndTechnicalReplicates` paired with `SumAggregation` or `MeanAggregation`. Their tests are rewritten that way and still assert the same numbers.

They are removed rather than kept as thin wrappers because a wrapper would reintroduce exactly what this PR fixes — a type whose *name* asserts an aggregation, while its interface takes one as an argument.

**Blast radius:** nothing outside `Test/` and `Development/` constructs a collapse strategy, and MetaMorpheus does not reference the `Quantification` namespace at all. No production behaviour changes.

## Changes from the version in #1000

- **An `IsobaricQuantSampleInfo` now collapses to an `IsobaricQuantSampleInfo`.** The original returned a `SpectraFileInfo` for every input, discarding the channel label, plex, reporter m/z and reference-channel flag — which would silently disable `ReferenceChannelNormalization` for every step downstream. This is the one substantive bug fix.
- Aggregation uses pooled buffers rather than a `List<double>` per cell, and a private `Median` helper rather than `MathNet.Numerics.Statistics`, matching `MedianRollUp` and `GlobalMedianNormalization`.
- Grouping preserves first-appearance column order, so output is deterministic.

## The follow-up this opens up, not taken here

`SumRollUp` and `MedianRollUp` are also one algorithm differing only in how a set of values is reduced. They could collapse into a single `RollUpStrategy` taking an `IAggregationStrategy`, which would make the four aggregation implementations serve both families instead of one. Left out to keep this PR reviewable, but it is where the split actually pays off.

## Tests

- `AggregationStrategyTests` (14) — each of the four aggregations, even- and odd-count median, median agreeing across the `stackalloc` boundary (2/63/64/65/500 values), empty span, single value, that the input buffer is not mutated, and distinct names.
- `CollapseStrategyTests` (13) — each collapse dimension and what it deliberately keeps apart, every pairing with an aggregation, isobaric channel identity preserved and label-free collapsing to `SpectraFileInfo`, sum-fractions-then-median-technical-replicates composed, first-appearance column ordering, the singleton and no-op cases, and a null aggregation throwing.
- `QuantMatrixTests` also comes from #1000, rewritten against `MockSpectralMatch` — `BaseSpectralMatch` has since become abstract, so the original would not compile — and extended to the argument-validation paths, which had no coverage.

CI is green on the head commit: build, integration, external-service tests, and both codecov gates.


Alexander-Sol (changes requested):
I like the motivation behind this PR and what it accomplishes. However, the implementation is quite different and doesn't really conform to the strategy pattern. Currently, looking at the qunatification parameters is enough to fully describe how the data was processed. The public field CollapseStrategy holds the strategy used. Here, one Strategy could do 12 different things depending on the combination of aggregation  method and collapse dimension. 

One fix is to split this into two strategies. ICollapseStrategy could just describe which samples are combined (fraction only, fraction + tech rep). Then, IAggregationStrategy could be added. That would keep the design modular and enable all the behavior  that was implemented here. Happy to discuss this further in person

trishorts:
You're right, and I've done the split. Pushed.

The critique lands: `SampleCollapseStrategy` could do twelve different things depending on its constructor arguments, so `QuantificationParameters` no longer told you how the data was processed — and every other strategy in the project (`SumRollUp`, `MedianRollUp`, `GlobalMedianNormalization`) encodes exactly one behaviour per type. Mine was the odd one out.

**What changed**

`ICollapseStrategy` now decides only *which* samples are combined. The new `IAggregationStrategy` decides *how* the grouped values are reduced to one. `QuantificationParameters` carries both, so reading it still fully describes the run.

```csharp
public interface IAggregationStrategy
{
    string Name { get; }
    double Aggregate(ReadOnlySpan<double> values);
}
// SumAggregation, MeanAggregation, MedianAggregation, MaxAggregation

public interface ICollapseStrategy
{
    string Name { get; }
    QuantMatrix<T> CollapseSamples<T>(QuantMatrix<T> m, IAggregationStrategy aggregation)
        where T : IEquatable<T>;
}
// NoCollapse, CollapseFractions, CollapseTechnicalReplicates,
// CollapseBiologicalReplicates, CollapseFractionsAndTechnicalReplicates
```

`CollapseStrategyBase` holds the grouping, the aggregation loop and the collapsed-column identity; a subclass declares only which of the three sample dimensions it collapses. `Aggregate` takes a `ReadOnlySpan<double>` so the base can pass a slice of a pooled buffer without copying.

**One decision I made that you may want to revisit**

`SumCollapse` and `MeanCollapse` are removed rather than kept. Both grouped by condition and biological replicate, so they are now `CollapseFractionsAndTechnicalReplicates` paired with `SumAggregation` or `MeanAggregation`; their tests are rewritten that way and still assert the same numbers. Keeping them felt like it would reintroduce the exact problem this fixes — a type whose *name* asserts an aggregation, while its interface now takes one as an argument. But they are on master, so if you'd rather they stayed as thin wrappers, say so and I'll add them back.

**The bigger prize, not taken here**

`SumRollUp` and `MedianRollUp` are also one algorithm differing only in how a set of values is reduced. They could collapse into a single `RollUpStrategy` taking an `IAggregationStrategy`, which would make the aggregation implementations serve both families instead of one. I left it out to keep this PR reviewable, but it is where the split actually pays off. Related: should the aggregation be one shared parameter or one per stage? I lean per-stage — summing fractions while taking a median at protein roll-up is a reasonable thing to want.

Still very much up for talking this through in person — the roll-up question and the `SumCollapse` call are both worth more than a comment thread.

Alexander-Sol (approved):
This looks great

nbollis (approved):

PR #1201, opened 2026-08-25
Add .github/dependabot.yml for the github-actions ecosystem

### LINKED ISSUE #1196: Add .github/dependabot.yml for the github-actions ecosystem
Split out of #1190, which moved four actions off Node 20 by hand.

That PR was pure toil: GitHub started force-migrating our Node 20 actions to Node 24 and warning on every job, and somebody had to go read each action's `action.yml` to find the first major that was actually node24. Two of the four (`upload-artifact`, `nuget/setup-nuget`) were not even in the warnings GitHub displayed, because it only annotates the first few per job — they would have surfaced on the next runtime deprecation instead. One step was still on `actions/checkout@master`, an unpinned floating ref to a third-party action.

None of that needed a human. Dependabot does it on a schedule and opens the PR itself.

There is no `.github/dependabot.yml` in the repo today. Minimal version:

```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
```

`directory: "/"` is correct for the actions ecosystem — it means `.github/workflows`, not the repo root, and it picks up all three workflow files including `release.yml`, which PR CI never exercises and which is therefore the most likely place for a stale pin to rot unnoticed.

Notes:

- Monthly rather than weekly. Actions majors do not move often and weekly would be noise.
- This covers the workflow pins only. Whether to also let Dependabot near the NuGet ecosystem is a much bigger conversation (it would open PRs against pinned dependencies that exist for compatibility reasons, e.g. the OpenMcdf 2.x pin that `check-thermo-rawfilereader.yml` watches) and should not be bundled in.
- Dependabot PRs will need the same review as #1190 did: a major bump can change an action's input contract, and green CI on `dotnet.yml` says nothing about `release.yml`.

trishorts:
Closes #1196. Adds the `.github/dependabot.yml` that #1196 specifies, verbatim in config terms — `github-actions`, `directory: "/"`, monthly — with the reasoning written into the file, since that is how the workflows in this repo are commented.

**What it will actually find.** Dependabot scans all three workflows. `check-thermo-rawfilereader.yml` has no `uses:` at all by design ("No checkout: this job only talks to the GitHub API"), so the pins live in the other two:

| Workflow | Pin |
|---|---|
| `dotnet.yml:19,72,93` | `actions/setup-dotnet@v4` |
| `dotnet.yml:50` | `codecov/codecov-action@v4` |
| `dotnet.yml:145,151` | `actions/upload-artifact@v4` |
| `release.yml:18` | `actions/setup-dotnet@v1` |
| `release.yml:22` | `nuget/setup-nuget@v1` |

`release.yml` is the point of the exercise: PR CI never runs it, and both of its pins are three and four majors behind respectively. Nothing in the repo would have told us.

**One thing this does not fix.** `dotnet.yml:115` is `actions/checkout@master` — a floating branch ref, not a version. Dependabot has no version to compare and will leave it alone, so it still needs the hand fix in #1190. Worth saying out loud so nobody reads a green Dependabot as coverage of that line.

**Relationship to #1190.** Complementary, not overlapping. #1190 fixes today's pins; this stops the next round from needing a human. No file conflict — #1190 touches the workflows, this adds a new file. Either can merge first.

**Not included:** the `nuget` ecosystem, deliberately. Several dependencies here are pinned for compatibility rather than lag — the OpenMcdf 2.x pin that `check-thermo-rawfilereader.yml` exists to watch, for one — and Dependabot would open PRs against them. #1196 calls that a separate conversation and I agree.

**Testing.** None to run: no code, and no CI job reads this file. Validated that it parses and that the four fields match the schema Dependabot expects. The real proof is the first PR it opens, which will be within a month of merge.


pcruzparri (approved):

nbollis (approved):

acesnik (approved):
🤖 Approving. The config is right and matches #1196 in config terms — the points below are about sequencing and expectations, not the file.

**Verified against the head commit.** Parses to exactly the four fields Dependabot expects. `check-thermo-rawfilereader.yml` really does have no `uses:`, so the inventory is the other two workflows. Every line number in the table matches. `directory: "/"` is correct for this ecosystem. No conflict with #1190 — new file versus workflow edits.

**Merge #1190 first.** "Either can merge first" is the one claim I'd push back on. Dependabot runs an initial check as soon as the config lands, not on the next monthly tick, so if this merges first it will open PRs bumping the same pins #1190 is bumping by hand, while #1190 is still open. Both are approved and mergeable, so the fix is just ordering.

**The table misses one.** `release.yml:16` is `actions/checkout@v2` — a version pin, three majors back, so Dependabot would find it. #1190 fixes it, so this is a body nit rather than a gap.

**What the first run will actually propose,** checked just now against each action's latest release, assuming #1190 has landed:

| Pin after #1190 | Latest |
|---|---|
| `actions/checkout@v5` | v7.0.1 |
| `actions/setup-dotnet@v5` | v6.0.0 |
| `actions/upload-artifact@v6` | v7.0.1 |
| `codecov/codecov-action@v4` (untouched by #1190) | v7.0.0 |
| `nuget/setup-nuget@v4` | v4.0 ✓ |

Four of five still behind. Not a knock on #1190, which picked the first major that dropped Node 20 — that was the job. But it is this PR's argument stated in numbers: hand-pinning to "first sufficient major" leaves a standing gap that nothing in the repo reports.

Two consequences worth deciding rather than discovering:

- The first run opens roughly four PRs, each a full Windows build/test/coverage cycle, against a default `open-pull-requests-limit` of 5. If that burst is unwelcome, one `groups:` block collapses it into a single PR. Optional in my view — steady state after the catch-up is a handful a year.
- Half of what Dependabot will touch lives in `release.yml`, which PR CI never runs, so a green check on those PRs proves nothing about the release path. #1196 already says this; it's worth keeping in front of whoever merges them, and it's why I'd keep `nuget/setup-nuget` out of any grouping.

**Not verified:** whether `CODECOV_TOKEN` is mirrored into Dependabot secrets. Repository secrets aren't exposed to Dependabot-triggered runs, so the upload step will go tokenless on these PRs. Low stakes — there's no `fail_ci_if_error` and coverage is unchanged on a workflow-only PR — but a tokenless-upload warning there is expected, not a regression.

PR #1203, opened 2026-08-25
Drop the dead intersectsAppliedRegionIncompletely flag

### LINKED ISSUE #1178: ApplySingleVariant computes intersectsAppliedRegionIncompletely and never uses it, and two test comments describe it as live
Found while reviewing #1154. Pre-existing on master (`smith/master @ 5ba13155`), not introduced by that PR — filing so it does not get lost, and because #1154 and #968 both rewrite the region around it.

`VariantApplication.ApplySingleVariant` computes a flag and never reads it:

```csharp
// If an already-applied variation partially overlaps the current edit, use the consensus tail to avoid index corruption
bool intersectsAppliedRegionIncompletely =
    protein.AppliedSequenceVariations != null
    && protein.AppliedSequenceVariations.Any(x => variantGettingApplied.Intersects(x) && !variantGettingApplied.Includes(x));
```
— `VariantApplication.cs:257-260`

`intersectsAppliedRegionIncompletely` appears exactly once in the whole source tree (`git grep` over `smith/master` returns one hit, this assignment). Nothing branches on it. The line immediately below picks its tail unconditionally:

```csharp
string seqAfter = protein.BaseSequence.Length - afterIdx <= 0 ? "" : protein.BaseSequence.Substring(afterIdx);
```

## Why this is worth more than a compiler warning

The comment above it promises behaviour the code does not perform — "use the consensus tail" — and that promise is repeated as fact in two places in the test suite:

```
///    The overlap causes <c>intersectsAppliedRegionIncompletely</c> to fire inside
///    <see cref="ApplySingleVariant"/>. The production fix ensures the intermediate
///    tail ("AIDE") is used rather than the consensus tail ("TIDE") ...
```
— `Test/DatabaseTests/TestVariantProtein.cs:647-650`, and again at `:716-719`

Reading those, the flag looks load-bearing. It is not. What actually fixed the bug the test pins was making `seqAfter` unconditionally use `protein.BaseSequence` — at which point the flag became redundant and was left behind. The behaviour is correct today; the description of how it is achieved is wrong in three places.

The risk is not the dead local, it is that the surrounding code is under active change (#1154 rewrites the applied-variation bookkeeping directly below this; #968 rewrites the same method), and a reader who trusts the comments could reintroduce a branch on the flag and restore the consensus-tail bug the test exists to prevent.

## Suggested fix

Delete the assignment and its comment, and correct the two test comments to describe what the code does — the tail comes from `protein.BaseSequence` unconditionally, which is what makes the overlap case work. `OverlappingVariants_NullVcf_ProducesCorrectCombinedSequence` keeps pinning the behaviour either way.

Worth sequencing after #1154 and #968 land, since both touch this method and a trivial deletion is not worth a conflict.

trishorts:
> **Project of origin:** `variants`

Closes #1178.

## The flag was never read

`ApplySingleVariant` computed this and nothing branched on it — it appeared exactly once in the whole tree, at its own assignment:

```csharp
// If an already-applied variation partially overlaps the current edit, use the consensus tail to avoid index corruption
bool intersectsAppliedRegionIncompletely =
    protein.AppliedSequenceVariations != null
    && protein.AppliedSequenceVariations.Any(x => variantGettingApplied.Intersects(x) && !variantGettingApplied.Includes(x));
```

The comment promises "use the consensus tail". The line below it does not:

```csharp
string seqAfter = protein.BaseSequence.Length - afterIdx <= 0 ? "" : protein.BaseSequence.Substring(afterIdx);
```

What actually fixed the overlapping-variant bug was making that tail unconditional. At that point the flag became redundant and was left behind, and the description of how the fix works stayed.

## Two test comments described it as live

`TestVariantProtein.cs:647` and `:716` both stated the flag fires and that this is the production fix. Reading them, it looks load-bearing. Both now describe what the code does.

## Why a comment replaces it rather than nothing

Variations are applied in descending position order (`VariantApplication.cs:130`), so `BaseSequence` already carries every edit downstream of the current one. Branching back onto `ConsensusVariant.BaseSequence` reverts them — that is the `MPEAAIDE` → `MPEATIDE` regression `OverlappingVariants_NullVcf_ProducesCorrectCombinedSequence` exists to pin. The replacement comment says that, and names the test, so the branch is not re-derived from first principles by the next reader.

## Sequencing — one correction to the issue

The issue suggested landing this after #968, since #968 rewrites the same method. That is backwards. #968's merge-base is `e12f8004`, which predates the fix, and on that branch the flag is **live**:

```csharp
if (intersectsAppliedRegionIncompletely)
{
    // Tail from consensus (not the possibly already-mutated BaseSequence)
    seqAfter = protein.BaseSequence.Length - afterIdx <= 0 ? "" : protein.ConsensusVariant.BaseSequence.Substring(afterIdx);
}
```

So #968 does not merely touch this region — as it stands it reintroduces the exact bug the test pins. Sequencing after #968 means merging that and deleting the flag afterward. Landing this first makes the conflict surface during #968's rebase, where someone has to look at it. Commented there.

(#968 also rewrote `TestVariantProtein.cs` enough to drop both comments, so the test half of this only conflicts if that branch is rebased.)

## Verification

No behavior change — a dead local and three comments. `dotnet test --filter TestVariantProtein`: 47 passed, 0 failed, including `OverlappingVariants_NullVcf_ProducesCorrectCombinedSequence`.


acesnik:
🤖 Verified the flag is dead — one occurrence in `VariantApplication.cs` (its own assignment) and two in test comments — and the descending-order premise holds at `VariantApplication.cs:130` (`OrderByDescending(v => v.OneBasedBeginPosition)`).

Your correction to the issue's sequencing is right, and understated. I looked at #968:

- the flag is live there, assigned at `:276` and branched on at `:282`
- that branch takes `protein.ConsensusVariant.BaseSequence.Substring(afterIdx)` — the reverting tail
- `OverlappingVariants_NullVcf_ProducesCorrectCombinedSequence` is **not present on that branch at all**
- its merge-base is `e12f8004` (2026-02-11); master is 109 commits ahead

So landing #968 first would not merely reintroduce the branch — it would reintroduce it with the test that catches it already gone, and nothing would fail. That makes landing this one first not just tidier but the difference between the conflict being visible and not.

Worth someone restoring that test as part of #968's rebase rather than relying on the conflict surfacing in the right place.

acesnik (approved):

pcruzparri (approved):

nbollis (approved):

PR #1204, opened 2026-08-25
Refuse a spectra extension the mzML adapter cannot write, instead of doubling it

### LINKED ISSUE #1181: MsDataFileToResultFileAdapter.WriteResults doubles the extension: out.mgf becomes out.mgf.mzML
🤖 ```csharp
var adapter = FileReader.ReadFile<MsDataFileToResultFileAdapter>("input.mgf");
adapter.LoadResults();
adapter.WriteResults("output.mgf");
```

**Expected:** a file at `output.mgf`, or a clear refusal.
**Actual:** a file at `output.mgf.mzML`, containing mzML.

## Cause

`Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs`:

```csharp
public void WriteResults(string outputPath)
{
    if (!outputPath.EndsWith(".mzml", StringComparison.InvariantCultureIgnoreCase))
        outputPath += SupportedFileType.MzML.GetFileExtension();
    _dataFile.ExportAsMzML(outputPath, true);
}
```

The guard asks "is this already `.mzml`", not "does this have an extension", so it appends to any other extension rather than only to a bare path. `TestWritingNoExtension` covers the bare-path case (`"a"` → `"a.mzML"`), which is why this hasn't surfaced.

Not MGF-specific: the adapter is the `IResultFile` type for `ThermoRaw`, `MzML`, `Mgf`, `BrukerD`, `BrukerTimsTof`, `Ms1Align` and `Ms2Align` (`Readers/Util/SupportedFileTypes.cs:245-251`), so `output.raw` becomes `output.raw.mzML` today.

## To be clear about what is and isn't intended

Writing **mzML regardless of what was read is deliberate** — this adapter is a converter, and `TestWritingNoExtension` pins the extension-appending behaviour. This issue is only about appending a second extension to a path that already has one. `Path.ChangeExtension` would be the usual fix, though it silently rewrites the caller's chosen name, so refusing a mismatched extension may be the better behaviour.

## Related design question, worth deciding separately

Now that `MsDataFile.ExportAsMgf` exists (#1165), `SupportedFileType.Mgf` resolving to this adapter makes it look like MGF write support that isn't there. Whether `WriteResults` should honour the requested type instead of always emitting mzML is a design decision affecting every caller, not a bug — I'd rather it were decided than patched.

`SpectralAveraging/Algorithms/AveragedSpectraWriter.cs:32-39` is a smaller instance of the same question: it throws `"Output averaged scans type not implemented"` on any unhandled output type. Adding MGF there also implies a view on whether averaged MS1 spectra should be offered in a format that loses `PEPMASS` and `CHARGE`.

trishorts:
> **Project of origin:** none — standalone fix (issue #1181)

Closes #1181.

`MsDataFileToResultFileAdapter` writes mzML whatever it read, but its guard asked *"is this already `.mzml`"* rather than *"does this have an extension"*, so it appended to any other extension instead of only to a bare path. `WriteResults("output.mgf")` produced `output.mgf.mzML` — an mzML file wearing two extensions, neither of which the caller chose. Not MGF-specific: the adapter is the `IResultFile` type for `ThermoRaw`, `MzML`, `Mgf`, `BrukerD`, `BrukerTimsTof`, `Ms1Align` and `Ms2Align`, so `output.raw` doubled too. `TestWritingNoExtension` covers only the bare-path case, which is why this stood.

**Refuse rather than rename.** `Path.ChangeExtension` is the usual fix, but it silently rewrites the caller's chosen name and hands back a format they did not ask for. Now that `ExportAsMgf` exists (#1165), a caller writing `output.mgf` most likely wants MGF, and the honest answer is that this writer cannot give it to them — not to quietly produce `output.mzML` instead. #1181 leans the same way.

**Refusal is scoped, not blanket.** Only an extension that names a spectra format the adapter *reads* is refused. Everything else — a bare path, and a dotted sample name like `gradient_1.5uL`, whose `Path.GetExtension` is `.5uL` — still gets `.mzML` appended. Refusing on "has any extension at all" would have broken the second case, which is a real naming habit and not a format request.

| Input | Before | After |
|---|---|---|
| `out` | `out.mzML` | `out.mzML` (unchanged) |
| `out.mzML` | `out.mzML` | `out.mzML` (unchanged) |
| `gradient_1.5uL` | `gradient_1.5uL.mzML` | `gradient_1.5uL.mzML` (unchanged) |
| `out.mgf` | `out.mgf.mzML` | `ArgumentException` |
| `out.raw` | `out.raw.mzML` | `ArgumentException` |

**Anti-drift.** The refusal set is spelled by enum member — so it stays greppable against `GetResultFileType` — but takes its extension *strings* from `GetFileExtension`, so no literal here can go stale (`Path.GetExtension` is what reduces `_ms1.msalign` to the `.msalign` a caller's path would end in). The test derives the same set independently from `GetResultFileType`, so a spectra format added to the adapter later is covered without anyone remembering to update either list.

**Blast radius: none.** Nothing outside the tests calls this method. The only non-test `WriteResults` call in the repo is `FragmentIntensityModel.cs:816`, which is `SpectralLibrary`, and MetaMorpheus's five call sites are all FlashLFQ's unrelated `WriteResults` overload. Whether this is a breaking change for an outside caller depends on whether anyone was relying on the doubled name, which would mean relying on a file whose extension lies about its contents.

**Testing.** `TestWritingRefusesTheOtherSpectraFormatsItReads` fails on master and passes here; the two `TestWritingAppendsToNamesThatAreNotSpectraFormats` cases and the existing `TestWritingNoExtension` pass both before and after, which is the point — they pin what must *not* change. Verified by reverting the production hunk and re-running: exactly one of the four fails. Whole `FileReadingTests` namespace green, 1028/1028.

**Deliberately not addressed.** Whether `WriteResults` should honour the requested format instead of always emitting mzML. #1181 calls that a design decision affecting every caller and asks for it to be decided rather than patched, and I agree — as is the related `AveragedSpectraWriter` question it raises.

**One observation, not fixed here.** Every `ResultFile<T>` sibling has the same shape of guard — `if (!CanRead(outputPath)) outputPath += FileType.GetFileExtension();` — and `CanRead` is likewise an *EndsWith* test, so e.g. `MsFraggerPsmFile.WriteResults("out.csv")` writes `out.csv.tsv`. It is a milder version of the same defect (those callers are less likely to hand over a rival format's extension), and changing eleven writers' contracts is not something to bundle into a fix for this one. Worth its own issue if others agree it is a defect rather than a convention.


mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
nbollis, line 1: I believe a better solution would be to check the extension of the path, if it is mzml or mgf, write that format. If it is not either of those two, then throw.

nbollis (changes requested):

mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
trishorts: Done in `7a91ec0c` — you were right, and the original PR was reasoning from a stale fact.

Refusing `.mgf` was the correct answer only while there was no MGF writer. #1165 added `MsDataFile.ExportAsMgf` on 2026-08-24, four days before this branch, so a caller asking for MGF can simply be given MGF. `Mgf` comes out of `UnwritableSpectraExtensions` and gets its own branch.

Three outcomes now, and the middle one is why this still is not just `Path.ChangeExtension`:

- a format the adapter can write (mzML, and MGF since #1165) is written as asked;
- a spectra format it can only **read** is refused, because handing back mzML under the caller's chosen name is worse than saying no;
- anything else has `.mzML` appended.

That third case is the one place I departed from your comment as literally worded. "If it is not either of those two, then throw" would also throw on `gradient_1.5uL` — `Path.GetExtension` is `.5uL`, a decimal in a sample name rather than a format request — and on any bare path, which `TestWritingNoExtension` already pins. So the refusal stays scoped to extensions that actually name a spectra format the adapter reads. Say the word if you would rather it were stricter than that.

The refusal test derived its list from `GetResultFileType` so a newly served format would be covered automatically; it now excludes MGF as well as mzML, which is the same mechanism in reverse. Added the positive case too: writing `.mgf` produces a file under the caller's name whose contents begin with `BEGIN IONS`, so it is MGF and not mzML wearing an MGF name.

13 green in the fixture.

trishorts:
@nbollis this is ready for another look — your comment is implemented, not argued with, and it needs your changes-requested lifted to move.

`7a91ec0c` does what you asked: the extension decides the format. mzML and MGF are each written as asked (MGF became possible when #1165 added `MsDataFile.ExportAsMgf` on 08-24, four days before this branch — refusing it was reasoning from a stale fact), and a spectra format the adapter can only read is refused rather than handed back as mzML under the caller's name.

The one place I did not follow the comment literally is the fallback. "If it is not either of those two, then throw" would also throw on `gradient_1.5uL`, where `Path.GetExtension` is `.5uL` — a decimal in a sample name, not a format request — and on any bare path, which `TestWritingNoExtension` pins. So anything that is not a recognised spectra extension still gets `.mzML` appended, and only extensions that actually name a spectra format the adapter reads are refused. Happy to make it stricter if you'd rather.

Positive coverage added too: writing `.mgf` produces a file under the caller's name whose contents begin with `BEGIN IONS`, so it is real MGF and not mzML wearing an MGF name. 13 green in the fixture, and all checks pass.

mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
```diff
@@ -40,11 +67,71 @@ public void LoadResults()
...
+        /// silent rename only implied, and a caller who genuinely wants MGF can now call
+        /// <see cref="MsDataFileExtensions.ExportAsMgf"/> directly.
+        /// </remarks>
+        /// <param name="outputPath">
+        /// Destination path. Must end in ".mzML", or carry no spectra-format extension, in which case
+        /// ".mzML" is appended.
+        /// </param>
+        /// <exception cref="ArgumentException">
+        /// <paramref name="outputPath"/> names a spectra format this adapter cannot write.
+        /// </exception>
```
acesnik: This whole block, from the `<summary>` above down to here, describes the pre-`7a91ec0c` design and should go. It says the adapter "writes only mzML" and that a caller wanting MGF should call `ExportAsMgf` directly, which is what the PR changed. The block starting on the next line is the accurate one.

It ships, too: building `Readers` at this head produces a `Readers.xml` whose `WriteResults` member has two `<summary>` and two `<remarks>` elements, the first being the stale one, and `mzLib.nuspec` puts that file in the package. The build is clean, so no warning will catch it.

mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
```diff
@@ -40,11 +67,71 @@ public void LoadResults()
...
+            }
+
+            if (UnwritableSpectraExtensions.Contains(extension))
+                throw new ArgumentException(
+                    $"This writer handles {MzMlExtension} and {MgfExtension}, so it cannot write " +
+                    $"'{Path.GetFileName(outputPath)}'. Pass a path ending in one of those, or one with " +
+                    $"no extension and {MzMlExtension} will be appended.",
+                    nameof(outputPath));
+
+            _dataFile.ExportAsMzML(outputPath + MzMlExtension, true);
```
acesnik, line 122: Appends rather than replaces, so `data.txt` becomes `data.txt.mzML`. That is the documented intent and the reason `gradient_1.5uL` survives, so no change requested — but the `ArgumentException` message two lines up only describes the no-extension case ("or one with no extension and .mzML will be appended"), and the unrecognised-extension case lands here silently. Worth a word in the message.

acesnik (commented):
🤖 The reframing in `7a91ec0c` is the right one, and it is a better PR than the one that was opened: "what did the caller ask for" rather than "is this already mzML" is the question that makes `WriteResults("output.mgf")` producing `output.mgf.mzML` obviously wrong rather than merely untidy. Answering @nbollis by implementing the suggestion and then explaining the single place you departed from it is also the most useful shape a reply can take.

That departure is correct, and I checked it rather than taking it: `Path.GetExtension("gradient_1.5uL")` really is `".5uL"`, so a literal "throw on anything that is not mzML or MGF" would reject a dotted sample name and every bare path. Scoping the refusal to extensions that actually name a spectra format is the right line.

I also checked the pieces that would have been quiet failures, and they hold:

The refusal set is exactly complete. `GetResultFileType` maps seven members to this adapter — ThermoRaw, MzML, Mgf, BrukerD, BrukerTimsTof, Ms1Align, Ms2Align — and `UnwritableSpectraExtensions` is those seven minus the two you can write. Deriving the strings from `GetFileExtension` rather than writing literals is what will keep that true.

`Path.GetExtension` does the right thing on every input in that set, which is not obvious for two of them. `.d` maps to `".d"` rather than to the empty string — had it mapped to `""`, the set would have contained `""` and every bare path would have been refused, breaking `TestWritingNoExtension`. And `_ms1.msalign` and `_ms2.msalign` both reduce to `".msalign"`, so the `HashSet` folds them together as the remarks say.

`ExportAsMgf` is real (`MsDataFileExtensions.cs:52`, from #1165) and currently writes every scan. Worth knowing that #1227, which is mine and still open, changes `WriteMgf` to take `includeMs1Scans` — it defaults to `true`, so it preserves what this PR does, but the two PRs are editing the same writer.

**One thing to fix, and it is why I am commenting rather than approving.** `WriteResults` has two complete XML doc blocks stacked on it, and the first one describes the design this PR replaced. It says the adapter "writes only mzML", that a different spectra format is refused, and that "a caller who genuinely wants MGF can now call `ExportAsMgf` directly" — the opposite of what the method now does. The second block, the one with the three outcomes, is the accurate one.

This is not cosmetic, because it ships. `mzLib.nuspec` puts an `.xml` beside each assembly, and building `Readers` at this head emits a `Readers.xml` whose `WriteResults` member carries **two `<summary>` and two `<remarks>` elements**, the first reading "Writes the loaded scans to outputPath as mzML, whatever format was read." A consumer's IntelliSense gets that sentence. There is no compiler warning — the build is clean — so nothing else will catch it, which makes it the same shape as the CS1587 problem on #1255 except quieter. Deleting the first block is the whole fix, and I will approve on sight.

Unrelated and pre-existing, so not a request: `_dataFile` is assigned only in `LoadResults`, and every other member dereferences it, so `WriteResults` before `LoadResults` is a `NullReferenceException`. Your new guard is careful about the argument and silent about the state; if you are touching the method anyway, that is a cheap sentence.

Not verified by me: I did not run the fixture — `Test.csproj` is `net10.0-windows` via `mzPlot` and will not run here — so your 13 green is CI's to confirm. Everything above came from reading the head and building `Readers`.

acesnik (dismissed):
🤖 Approving. The one thing I asked for was the stale doc block, and rather than leave a two-line deletion sitting in a queue I have pushed it as `bcf6e6a0`.

What it does: removes the first of the two XML doc blocks on `WriteResults`, the one describing the pre-`7a91ec0c` design where MGF was refused and callers were pointed at `ExportAsMgf`. That block also held the only `<param>` and `<exception>` tags, so both are restored against the behaviour the method actually has instead of disappearing with it. `Readers.xml` now carries one `<summary>`, one `<remarks>`, one `<param>` and one `<exception>`, and the summary is the accurate one; `Readers` builds with 0 errors and no new warnings.

Everything else I checked stands as it was, and the substance of this PR is right. The refusal set is exactly the seven adapter-served types minus the two you can write, derived from `GetFileExtension` so no literal can rot. `Path.GetExtension(".d")` is `".d"` rather than `""`, which is the difference between this working and every bare path being refused. And your reason for departing from @nbollis's literal ask holds — `Path.GetExtension("gradient_1.5uL")` really is `".5uL"`.

Two things I left alone deliberately. The `ArgumentException` message still describes only the no-extension fallback, so the unrecognised-extension case lands there unmentioned — that is wording in your voice, not mine to pick. And `_dataFile` is still assigned only in `LoadResults`, so `WriteResults` before it is an NRE; pre-existing across every member of the class and not this PR's to carry.

One small correction to the record: the CLI prompt and #2780's GUI prompt are not word for word — yours is "Continue without an experimental design? y/n" and #2780's is "Do you wish to continue without an experimental design?". They agree on the decision and on the phrase that matters, which is what counts, but nothing mechanically enforces the alignment. A shared const would, if you want the "cannot drift" property you described.

Still CI's to confirm rather than mine: `Test.csproj` is `net10.0-windows` via `mzPlot`, so I have not run the fixture.

mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
trishorts: Already gone — you fixed this yourself in `bcf6e6a0` ("docs(readers): drop the superseded WriteResults doc block"), pushed a few minutes after the comment. Confirmed at the current head: `WriteResults` carries one `<summary>`, one `<remarks>` and one `<param>`, all describing the post-`7a91ec0c` behaviour, so the duplicated elements in the shipped `Readers.xml` are gone with it. Resolving.

mzLib/Readers/ExternalResults/ResultFiles/MsDataFileToResultFileAdapter.cs
trishorts: Word added. The message now covers both branches of the fallback and names the appending explicitly, since that is the part a caller cannot infer:

> This writer handles .mzML and .mgf, so it cannot write 'run.raw'. Pass a path ending in one of those, or one whose extension names no spectra format at all -- including no extension -- and .mzML is appended to what you passed rather than replacing it, so 'sample.txt' is written as 'sample.txt.mzML'.

Behaviour unchanged, as you said it should be — `gradient_1.5uL` still survives, and that is now the documented reason rather than an accident the message hid.

acesnik (approved):
🤖 Approving. `b45114de` is the only commit since, and it is message text only — the `ArgumentException` now covers both fallback branches and names the appending rather than implying a replacement, so `'sample.txt'` is written as `'sample.txt.mzML'` (`MsDataFileToResultFileAdapter.cs:114-120`). Behaviour unchanged, as you said it should be, and `WriteResults` still carries one of each doc element.

One thing that costs no push, in the place a reader looks first: the PR body still describes the design `7a91ec0c` replaced. The table says `out.mgf` → `ArgumentException`, and "Refuse rather than rename" has a caller writing `output.mgf` being told this writer cannot give them MGF — but `.mgf` has had its own branch since `7a91ec0c` (`MsDataFileToResultFileAdapter.cs:107-111`), and `TestWritingMgfWritesMgfRatherThanMzml` pins it by asserting the output begins with `BEGIN IONS`. That test is also absent from the Testing paragraph, which counts four cases where there are now five. Same staleness as the doc block, one layer up.

pcruzparri (approved):

nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve with nits** — correct, tested, does exactly what the title claims.

**Summary:** Replaces `if (!outputPath.EndsWith(".mzml")) outputPath += ".mzML"` (which turned `output.mgf` → `output.mgf.mzML`) with extension-driven dispatch: `.mzML`→mzML, `.mgf`→MGF, read-only spectra extensions throw `ArgumentException`, everything else still gets `.mzML` appended.

**Findings (none blocking):**
1. *(minor)* The production `UnwritableSpectraExtensions` set is a hand-maintained literal — only the test derives its set from `GetResultFileType`. If a new read-only spectra format is routed to the adapter later, production falls through to the append branch and re-introduces this exact doubling bug (the test would fail, so it's fail-closed, but the description's "covered without anyone remembering to update either list" oversells it). Consider deriving the production set from `GetResultFileType` too.
2. *(minor)* Silent-wrong → throw is a real breaking change: anyone who passed `x.raw` relying on `x.raw.mzML` output now gets a runtime `ArgumentException`. Blast radius looks small (only non-test caller is a different `IResultFile` type) but worth a release-note call-out.
3. *(nit)* `ExportAsMgf` uses default `includeMs1Scans = true`; MGF is conventionally MS2-centric — pass `false` if MS1-in-MGF isn't intended.
4. *(nit)* `WriteResults` still NREs if called before `LoadResults` (pre-existing, not a regression; refusal branch correctly throws before touching `_dataFile`).
5. *(positive)* `Path.GetExtension` is genuinely more robust than `EndsWith` for directory-dot cases.

Tests are solid (refusal with derived set, MGF write, both append cases; uppercase covered by `InvariantCultureIgnoreCase`).

PR #1208, opened 2026-08-26
Ship an IExperimentalDesign implementation, instead of only test doubles

trishorts:
`IExperimentalDesign` has no implementation that ships. All four in the repo are test scaffolding:

```
Development/QuantificationDevelopment/TestHelpers/SpikeInExperimentalDesign.cs
Test/Quantification/QuantificationDeliveryTests.cs   (private TestExperimentalDesign)
Test/Quantification/QuantificationTests.cs           (private TestExperimentalDesign)
Test/Quantification/TmtSpikeInTests.cs               (private TmtTestExperimentalDesign)
```

MetaMorpheus's `TmtExperimentalDesign.ToMzLibDesign` is the only real one anywhere, and it lives in MetaMorpheus. So anyone wanting to run `QuantificationEngine` has to implement the interface before they can call anything — and FlashLFQ, which is the next consumer, has nothing to hand it at all.

### What it does

`SampleExperimentalDesign` covers both modalities: label-free designs hold one `ISampleInfo` per file, isobaric designs hold one per channel.

Two things it fixes rather than merely wraps:

**Files are keyed the way the engine looks them up.** `PivotByFile` reduces each spectral match's `FullFilePath` with `Path.GetFileName`, so the key is a name *with its extension*, not a path. `Add` accepts either and stores the name, so a design built from full paths resolves instead of throwing `KeyNotFoundException` on every file. (`SpikeInExperimentalDesign` keys by name *without* extension, which is one of the reasons it only works with its own fixture.)

**Lookups are case-insensitive.** A design naming `Sample1.raw` should match data at `sample1.raw`; under the default ordinal comparer it does not. A case-only duplicate is then rejected at `Add` rather than one of the two being silently dropped.

### Why validation is at the door

The failure it prevents is silent. Intensities map to samples by position, so a null channel that gets skipped shifts every later channel by one and quietly reports the wrong sample's abundance — a plausible number, in the wrong column, with nothing to notice. `Add` refuses a null sample, an empty sample array, and an empty file name.

### The two factories

- `FromSamples` groups a flat sequence by file, preserving input order within each file — which is the order the engine aligns intensities to. The caller still owns that order; for MetaMorpheus it is ascending reporter m/z.
- `LabelFree` takes `SpectraFileInfo` objects, which already implement `ISampleInfo`, and refuses a repeated file. Label-free measures a file once, so a repeat is a caller mistake rather than a second channel.

### Named `SampleExperimentalDesign`, not `ExperimentalDesign`

Because MetaMorpheus already has one. `EngineLayer.ExperimentalDesign` is the reader for its `ExperimentalDesign.tsv`, and TaskLayer files import both namespaces — so the plain name made six of them fail to compile with `CS0104: ambiguous reference`. The integration job caught it. Named for the samples it holds instead, with a comment on the class saying why.

### Scope

Additive. Nothing existing changes, no behaviour moves. 23 new tests — every throw in the class has a test that reaches it; `Test.Quantification` + `Test.Omics` are green (928 passed).

This is the first of the pieces still standing between the merged quantification engine and the LFQ half of #996 — it is what FlashLFQ will need in order to describe an experiment to the engine at all.

### DIFF
--- a/mzLib/MassSpectrometry/ExperimentalDesign/SampleExperimentalDesign.cs
+++ b/mzLib/MassSpectrometry/ExperimentalDesign/SampleExperimentalDesign.cs
@@ -11,6 +11,11 @@ namespace MassSpectrometry
     /// channel, and the per-file order is the contract that aligns them with
     /// <c>ISpectralMatch.Intensities</c>.
     ///
+    /// Named for the samples it holds rather than simply "ExperimentalDesign", because MetaMorpheus
+    /// already has an <c>EngineLayer.ExperimentalDesign</c> -- the reader for its ExperimentalDesign.tsv
+    /// -- and files there import both namespaces. An unqualified "ExperimentalDesign" in
+    /// <c>MassSpectrometry</c> makes every one of those references ambiguous.
+    ///
     /// Before this type, the only implementations of the interface lived in the Development project and
     /// in test fixtures, so every consumer outside MetaMorpheus had to write its own.
     ///
@@ -23,7 +28,7 @@ namespace MassSpectrometry
     /// <c>sample1.raw</c>, which it would not under the default ordinal comparer — and a case-only
     /// collision is rejected at <see cref="Add"/> rather than silently keeping one of the two.
     /// </summary>
-    public class ExperimentalDesign : IExperimentalDesign
+    public class SampleExperimentalDesign : IExperimentalDesign
     {
         /// <inheritdoc />
         public Dictionary<string, ISampleInfo[]> FileNameSampleInfoDictionary { get; }
@@ -31,7 +36,7 @@ namespace MassSpectrometry
         /// <summary>
         /// Creates an empty design. Populate it with <see cref="Add"/>.
         /// </summary>
-        public ExperimentalDesign()
+        public SampleExperimentalDesign()
         {
             FileNameSampleInfoDictionary =
                 new Dictionary<string, ISampleInfo[]>(StringComparer.OrdinalIgnoreCase);
@@ -102,14 +107,14 @@ namespace MassSpectrometry
         /// them (ascending reporter m/z, for MetaMorpheus).
         /// </summary>
         /// <exception cref="ArgumentException">A sample is null or names no file.</exception>
-        public static ExperimentalDesign FromSamples(IEnumerable<ISampleInfo> samples)
+        public static SampleExperimentalDesign FromSamples(IEnumerable<ISampleInfo> samples)
         {
             if (samples == null)
             {
                 throw new ArgumentNullException(nameof(samples));
             }
 
-            var design = new ExperimentalDesign();
+            var design = new SampleExperimentalDesign();
 
             var byFile = samples
                 .Select((sample, index) => (sample, index))
@@ -147,14 +152,14 @@ namespace MassSpectrometry
         /// A file is null, names no path, or appears twice. Label-free measures a file once; a repeat is
         /// a caller mistake rather than a second channel.
         /// </exception>
-        public static ExperimentalDesign LabelFree(IEnumerable<SpectraFileInfo> files)
+        public static SampleExperimentalDesign LabelFree(IEnumerable<SpectraFileInfo> files)
         {
             if (files == null)
             {
                 throw new ArgumentNullException(nameof(files));
             }
 
-            var design = new ExperimentalDesign();
+            var design = new SampleExperimentalDesign();
 
             foreach (var file in files)
             {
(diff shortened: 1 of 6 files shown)


Alexander-Sol (dismissed):

Alexander-Sol (dismissed):

mzLib/MassSpectrometry/ExperimentalDesign/SampleExperimentalDesign.cs
```diff
@@ -0,0 +1,183 @@
...
+
+                    string path = t.sample.FullFilePathWithExtension;
+                    if (string.IsNullOrWhiteSpace(path))
+                    {
+                        throw new ArgumentException(
+                            $"The sample at index {t.index} names no file, so it cannot be grouped into a design.",
+                            nameof(samples));
+                    }
+
+                    return Path.GetFileName(path);
```
nbollis, line 146: [synthesis] Design construction collapses different files that share the same basename

`FromSamples` groups by `Path.GetFileName(FullFilePathWithExtension)`, so files from different directories or case-only-different paths can merge into one design entry and silently misalign later quantification that still keys by full path.

Suggestion: Build the design map from a stable full-path identity, or validate and reject basename collisions explicitly.

nbollis (commented):
## Review Summary

- 1 major issue in `mzLib/MassSpectrometry/ExperimentalDesign/SampleExperimentalDesign.cs`.
- No summary-only findings.

This comment is about a silent file-identity collision that can misalign downstream quantification.

mzLib/MassSpectrometry/ExperimentalDesign/SampleExperimentalDesign.cs
trishorts: Fixed in `442c1c84` — though not the way suggested, because keying on full paths is not available here.

`IExperimentalDesign` is a `Dictionary<fileName, ISampleInfo[]>`, and `QuantificationEngine.PivotByFile` resolves entries with `Path.GetFileName(spectralMatch.FullFilePath)`. Key the design on full paths and every lookup misses.

So the collision is rejected where the information still exists rather than resolved later where it does not: a group whose samples carry more than one distinct `FullFilePathWithExtension` throws, naming both paths. That also makes `FromSamples` consistent with `Add`, which already refused a second entry for a file.

Channels of one run share a path, so the isobaric case — the reason grouping exists at all — is untouched, and there is a test pinning that. Case-only differences still compare equal, matching the lookup the engine does.

nbollis (approved):

trishorts:
@acesnik — could I ask you for a second approval on this one? It is at 1/2 with nbollis.

Alex approved it twice today and both were dismissed by pushes answering review comments, so this is a re-approval of code he has already been through rather than a cold read.

**Worth your eye specifically**, since it landed in your area: this originally added `MassSpectrometry.ExperimentalDesign`, which collided with `EngineLayer.ExperimentalDesign` and broke six MetaMorpheus TaskLayer files with `CS0104`. The `integration` job caught it. It is now `SampleExperimentalDesign`, and no type named exactly `ExperimentalDesign` remains anywhere in mzLib.

It is #1 in the merge order for the quant batch, so it is the one holding up the rest.

acesnik:
🤖 Keying claim verified exactly — `PivotByFile` does `Path.GetFileName(filePath)` then `TryGetValue` (`QuantificationEngine.cs:402`), so name-with-extension is what the engine wants and `Add` matches it. Confirmed too that all four existing implementations are test scaffolding.

**One gap worth closing while this is the first shipping implementation.** Case-insensitivity is a property of *your dictionary*, not of the contract. `PivotByFile` inherits whatever comparer the design was built with, and `IExperimentalDesign` never says the dictionary must be case-insensitive — so the next implementor gets ordinal matching and a `KeyNotFoundException` on a case mismatch, with this class sitting beside it working fine. The two `OrdinalIgnoreCase` uses already in `QuantificationEngine.cs` are in unrelated path helpers, so nothing upstream normalizes it.

Either state the requirement on `IExperimentalDesign`, or have `PivotByFile` normalize rather than trust the dictionary it is handed. The second is the one that cannot be got wrong by a future implementor.

Two smaller things:

- `FromSamples` throws from inside a `GroupBy` key selector, so those `ArgumentException`s surface during the `foreach` rather than at the call. Behaviour is fine; it just reads eager and isn't.
- `Path.GetFileName` on a Windows-style path under Linux returns the whole string, since a backslash is an ordinary filename character there. Both sides do the same thing so it is not a mismatch — but a design authored with Windows paths keys by the full path on Linux, which is worth knowing before someone debugs it.

And a note for the description rather than the code: MetaMorpheus master has no `IExperimentalDesign` references at all, so `TmtExperimentalDesign.ToMzLibDesign` is on an unmerged branch. "The only real one anywhere" sends a reader looking for something that is not on master yet.

acesnik (approved):

PR #1209, opened 2026-08-26
Implement GetAllPeptideToProteinMap, so shared-peptide protein quant can run

trishorts:
`QuantificationParameters.UseSharedPeptidesForProteinQuant` is a public settable bool. Setting it `true` made `QuantificationEngine.Run()` throw `NotImplementedException` from `RunProteinQuant`:

```csharp
var proteinMap = Parameters.UseSharedPeptidesForProteinQuant
    ? GetAllPeptideToProteinMap(peptideMatrixNorm)   // throw new NotImplementedException();
    : GetUniquePeptideToProteinMap(peptideMatrixNorm, BioPolymerGroups);
```

Settable and unusable, with nothing saying so until a search had already run. It was the only `TODO` left in the Quantification project.

### What the map does

It differs from `GetUniquePeptideToProteinMap` in exactly one way: a shared peptide belongs to more than one protein group, so its row index appears in more than one list, and its intensity contributes to every group it was assigned to. Everything else matches the sibling — an entry for every group, and an empty list rather than a missing key for a group whose peptides aren't in the matrix, because the roll-up strategy iterates the map either way.

### Two details worth naming

**Row indices are sorted.** `AllBioPolymersWithSetMods` is a `HashSet` and its enumeration order is not guaranteed stable, so an unsorted list would let roll-up results depend on set ordering. That's the same class of irreproducibility as #1155.

**The matrix rows are indexed once** into a dictionary rather than rescanned per protein group, so the cost is linear in peptides plus group memberships rather than their product.

### One signature change

Made `static` and given the protein-group list, matching `GetUniquePeptideToProteinMap` immediately above it. Changing a public signature is normally a break — but the method threw unconditionally, so no caller can exist.

### Tests

7 new, including one that runs the engine end to end with the flag on: the call that used to throw. The contrast test asserts both maps on the same input, so the shared row being absent from one and present twice in the other is visible in a single assertion block.

`Test.Quantification`, `Test.Omics` and `FlashLFQ` are green — 1031 passed.

### DIFF
--- a/mzLib/Test/Quantification/SharedPeptideProteinMapTests.cs
+++ b/mzLib/Test/Quantification/SharedPeptideProteinMapTests.cs
@@ -1,0 +1,239 @@
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using MassSpectrometry;
+using NUnit.Framework;
+using Omics;
+using Omics.BioPolymerGroup;
+using Omics.Digestion;
+using Omics.Modifications;
+using Proteomics;
+using Proteomics.ProteolyticDigestion;
+using Quantification;
+using Omics.SpectralMatch;
+using Test.Omics;
+
+namespace Test.Quantification
+{
+    /// <summary>
+    /// Tests for QuantificationEngine.GetAllPeptideToProteinMap -- the shared-peptide half of
+    /// protein roll-up, which threw NotImplementedException until now, so
+    /// UseSharedPeptidesForProteinQuant = true could not run at all.
+    /// </summary>
+    [TestFixture]
+    [ExcludeFromCodeCoverage]
+    public class SharedPeptideProteinMapTests
+    {
+        private class TestExperimentalDesign : IExperimentalDesign
+        {
+            public Dictionary<string, ISampleInfo[]> FileNameSampleInfoDictionary { get; }
+
+            public TestExperimentalDesign(Dictionary<string, ISampleInfo[]> dict)
+            {
+                FileNameSampleInfoDictionary = dict;
+            }
+        }
+
+        private static IBioPolymerWithSetMods Peptide(string sequence, string accession)
+        {
+            var protein = new Protein(sequence, accession);
+            var digestionParams = new DigestionParams(maxMissedCleavages: 0, minPeptideLength: 5);
+            return protein.Digest(digestionParams, new List<Modification>(), new List<Modification>()).First();
+        }
+
+        private static BioPolymerGroup Group(
+            string accession,
+            IEnumerable<IBioPolymerWithSetMods> all,
+            IEnumerable<IBioPolymerWithSetMods> unique)
+        {
+            return new BioPolymerGroup(
+                new HashSet<IBioPolymer> { new Protein("MPEPTIDEK", accession) },
+                new HashSet<IBioPolymerWithSetMods>(all),
+                new HashSet<IBioPolymerWithSetMods>(unique));
+        }
+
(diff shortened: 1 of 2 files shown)


Alexander-Sol (dismissed):

nbollis (commented):
## Review Summary

- 1 minor issue in `mzLib/Test/Quantification/SharedPeptideProteinMapTests.cs`.
- No summary-only findings.

The implementation risk here is mostly in the proof: the current end-to-end assertion can pass under a symmetric failure mode.

mzLib/Test/Quantification/SharedPeptideProteinMapTests.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(result.Success, Is.True, result.Summary);
+
+                // Both groups quantify. G1 gets pepA + pepShared, G2 gets pepShared + pepC, so with
+                // equal peptide intensities the shared peptide lands in both totals.
+                Assert.That(g1.IntensitiesBySample, Is.Not.Null.And.Not.Empty);
+                Assert.That(g2.IntensitiesBySample, Is.Not.Null.And.Not.Empty);
+                Assert.That(g1.IntensitiesBySample.Values.Sum(), Is.EqualTo(g2.IntensitiesBySample.Values.Sum()));
```
nbollis: [synthesis] End-to-end assertion does not actually prove shared-peptide inclusion

The final equality check passes even if shared peptides are dropped from both groups, because all peptide intensities are identical. The test therefore cannot distinguish the advertised shared-peptide behavior from a symmetric failure mode.

Suggestion: Use heterogeneous peptide intensities and assert the exact protein totals contributed by unique and shared rows.

trishorts: Fixed in `597ffb21`. You are right that it could not fail — every peptide carried the same intensity, so comparing the two groups to each other passed just as happily if the shared peptide were dropped from both.

Peptides now carry distinct intensities (100/200, 30/60, 7/14) and the test asserts exact per-channel totals: G1 = 130/260, G2 = 37/74.

I also added the companion run with the flag off, which produces 100/200 and 7/14 from the same input — every value different. The pair distinguishes the two behaviours rather than merely agreeing with each other.

nbollis (approved):

trishorts:
@acesnik — second approval wanted here too, if you have a moment. 1/2 with nbollis.

Small and self-contained: `GetAllPeptideToProteinMap` threw `NotImplementedException`, so `UseSharedPeptidesForProteinQuant = true` was a settable, unusable flag. Now implemented, with sorted row indices so roll-up does not depend on `HashSet` enumeration order — the same reproducibility class as #1155.

nbollis caught that my end-to-end test could pass under a symmetric failure; it now asserts exact per-channel totals plus a contrasting run with the flag off.

acesnik:
🤖 Matches its sibling, and sorting the row indices is the right instinct — `AllBioPolymersWithSetMods` being a `HashSet` makes unsorted output exactly the #1155 class of irreproducibility, where the answer depends on set ordering rather than on the data.

The static-plus-parameter signature change is defensible precisely because the old body threw unconditionally, so no working caller can exist to break.

One small thing worth a line: `rowIndexByPeptide[peptideMatrix.RowKeys[i]] = i` lets a duplicate row key silently take the later index. If `RowKeys` is unique by construction, saying so — or using `TryAdd` — keeps it that way when someone changes how the matrix is built.

acesnik (approved):

PR #1210, opened 2026-08-26
Exclude ambiguous spectral matches from quantification, and say how many

trishorts:
`GetPsmToPeptideMap` did this:

```csharp
var peptide = sm.GetIdentifiedBioPolymersWithSetMods().First(); // Assumes unambiguous mapping
```

The assumption is not checked anywhere. A match that identified several biopolymers was attributed to whichever one came first — a plausible intensity recorded against a peptide that may not be the one measured, with nothing to indicate it happened. #996's pipeline specifies an unambiguous filter at exactly this point; this is it.

MetaMorpheus avoids the bug by filtering at its own call site with `includeAmbiguous: false`, so it has been exercising the safe path all along. No other consumer of the engine gets that for free.

### What changes

**A spectral match is quantified only when it identifies exactly one biopolymer.** `Take(2)` distinguishes "exactly one" from "more than one" without walking a long ambiguity list.

That also removes a crash: `.First()` on a match carrying no identification threw `InvalidOperationException` from inside the engine.

**`QuantificationResults.AmbiguousSpectralMatchesExcluded` reports how many were dropped**, and the count is echoed in `Summary` when it isn't zero:

> Quantification completed successfully. 2 spectral match(es) were excluded because they did not identify exactly one biopolymer.

Dropping them is correct. Dropping them *silently* would leave a result set smaller than the search with no stated cause — the kind of gap that gets noticed a paper later. `Summary` is unchanged when the count is zero, so existing assertions on that exact string still hold.

### Tests

5 new. The first asserts the ambiguous row lands on *neither* candidate rather than on the first one; the rest cover the no-identification case, the unaffected happy path, and both branches of the summary text.

`Test.Quantification`, `Test.Omics` and `FlashLFQ` are green — 1029 passed.

### DIFF
--- a/mzLib/Quantification/QuantificationEngine.cs
+++ b/mzLib/Quantification/QuantificationEngine.cs
@@ -103,10 +103,19 @@ public class QuantificationEngine
         OverwriteSampleIntensities(peptideMatrix);
         OverwriteSampleIntensities(proteinMatrix);
 
+        // Report the matches the peptide map had to drop, so an unexpectedly small result set has a
+        // stated cause rather than being silently smaller than the search.
+        int ambiguousExcluded = SpectralMatches
+            .Count(sm => sm.Intensities != null && sm.GetIdentifiedBioPolymersWithSetMods().Take(2).Count() != 1);
+
         return new QuantificationResults
         {
-            Summary = "Quantification completed successfully.",
+            Summary = ambiguousExcluded == 0
+                ? "Quantification completed successfully."
+                : $"Quantification completed successfully. {ambiguousExcluded} spectral match(es) were " +
+                  "excluded because they did not identify exactly one biopolymer.",
             Success = true,
+            AmbiguousSpectralMatchesExcluded = ambiguousExcluded,
             Samples = proteinMatrix.ColumnKeys.ToList(),
             PeptideIntensities = BuildIntensityTable(peptideMatrix),
             ProteinIntensities = BuildIntensityTable(proteinMatrix),
@@ -491,8 +500,11 @@ public class QuantificationEngine
     /// Creates a mapping from each specified modified biopolymer to a list of indices that identify the position of corresponding
     /// spectral matches in the smMatrix
     /// </summary>
-    /// <remarks>The mapping assumes that each PM in the matrix is associated with a single, unambiguous
-    /// modified biopolymer. Biopolymers not present in the input list are ignored.</remarks>
+    /// <remarks>A spectral match is quantified only when it identifies exactly one modified biopolymer.
+    /// An ambiguous match -- one that identifies several -- is excluded rather than attributed to whichever
+    /// biopolymer happens to be enumerated first, and a match that identifies none is excluded rather than
+    /// throwing. Biopolymers not present in the input list are ignored.
+    /// <see cref="QuantificationResults.AmbiguousSpectralMatchesExcluded"/> reports how many were dropped.</remarks>
     /// <param name="smMatrix">The matrix containing spectrum matches to be mapped to their corresponding modified bioPolymer.</param>
     /// <param name="modifiedBioPolymers">The list of modified bioPolymers for which to generate the mapping.
     /// Only SMs corresponding to these bioPolymers are included in the result.</param>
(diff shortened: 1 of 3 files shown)


Alexander-Sol (dismissed):
Minor nitpick: The AmbiguousMatchFilterTests should go in Test/FlashLFQ, not Test/Quantification.

mzLib/Quantification/QuantificationEngine.cs
```diff
@@ -508,7 +520,16 @@ public static Dictionary<IBioPolymerWithSetMods, List<int>> GetPsmToPeptideMap(Q
         for (int i = 0; i < smMatrix.RowKeys.Count; i++)
         {
```
nbollis, line 550: [synthesis] Arity check counts raw entries instead of distinct identified biopolymers

`Take(2).Count() != 1` treats the raw enumeration length as the ambiguity test. A match like `[pepA, pepA]` or `[pepA, null]` is now classified as ambiguous and dropped even though it still identifies one peptide in substance, so valid matches can disappear from quantification.

Suggestion: Count distinct non-null identified biopolymers before applying the exact-one test, and use that same predicate for both filtering and reporting.

nbollis (commented):
## Review Summary

- 1 major issue in `mzLib/Quantification/QuantificationEngine.cs`.
- No summary-only findings.

The reported exclusion count and actual filter both currently depend on raw enumeration length rather than distinct identified biopolymers.

mzLib/Quantification/QuantificationEngine.cs
trishorts: Fixed in `8dffb594`, and you were right that it was losing data — a match naming the same peptide once per protein it maps to is a normal shape, and the raw count discarded it as ambiguous.

Now judged on distinct non-null identifications. `Take(2)` still short-circuits, so a long ambiguity list is not enumerated.

On using the same predicate for both: the reported count was a second copy of the expression, free to drift from the filter it described. Both now call `SingleIdentifiedBioPolymerOrNull`, and a test asserts the arithmetic ties — kept + excluded == total.

nbollis (commented):
Is this necessary as we already do this filtering in MetaMorpheus? 

If so, I will approve after CI/CD passes

trishorts:
Fair question, and partly the answer is no — for MetaMorpheus specifically this is redundant today. `PostSearchAnalysisTask` already passes `includeAmbiguous: false`, so those matches never reach the engine.

Three reasons I still think it belongs here:

**1. The old code crashed, it did not merely fail to filter.** `GetIdentifiedBioPolymersWithSetMods().First()` on a match carrying no identification throws `InvalidOperationException` from inside `QuantificationEngine`. MetaMorpheus's filter screens ambiguity, not emptiness, so that path was reachable and fatal rather than merely unfiltered.

**2. The failure it prevents is a wrong answer, not a missing one.** `.First()` attributed an ambiguous match to whichever biopolymer the enumeration happened to yield first — a plausible intensity recorded against a peptide that may not be the one measured. That is the shape of bug that survives a long time, because nothing looks wrong downstream.

**3. MetaMorpheus is not the only consumer, and its filter is a call-site convention.** FlashLFQ is next through this engine and has no equivalent filter; `Development/PsmTsvQuantAdapter` has none either. A call-site convention also only holds until someone adds a second call site — this is the API boundary, and the check is one `Take(2)`.

The reporting half is new value regardless of who filters: `QuantificationResults.AmbiguousSpectralMatchesExcluded` and the summary line mean a result set smaller than the search has a stated cause. For MetaMorpheus it will read 0, which is the correct and useful answer.

On CI: `build`, `integration` and both codecov checks are green. `external-service-tests` failed on `HttpIOException: The response ended prematurely` — a live download dropping mid-transfer, unrelated to this change and not a required check. Happy to re-run it if you would rather see it green before approving.

trishorts:
@acesnik — this one needs both approvals rather than one, so it is the bigger ask. Currently 0/2.

It fixes a silent wrong-answer path: `GetPsmToPeptideMap` took `GetIdentifiedBioPolymersWithSetMods().First()` under a comment reading "Assumes unambiguous mapping", so an ambiguous match was attributed to whichever biopolymer happened to come first. It also removes a crash — `.First()` threw `InvalidOperationException` on a match carrying no identification.

nbollis then caught that judging on raw enumeration length discarded good data: a match naming the same peptide once per protein it maps to counted as two and was dropped as ambiguous. It is now judged on distinct non-null identifications, with the filter and the reported count running off one shared predicate so they cannot drift.

His remaining question was whether this is needed at all given MetaMorpheus already filters at its call site. Answered above, in short: MetaMorpheus does, FlashLFQ does not and is next through this engine, and the crash was reachable regardless of ambiguity filtering.

CI is green except `external-service-tests`, which fails on a live PRIDE download (`HttpIOException: The response ended prematurely`) and is not a required check.

acesnik (approved):

trishorts:
@nbollis — CI is green, including the one that was failing.

```
build                    pass   9m14s
external-service-tests   pass   3m54s
integration              running
```

`external-service-tests` was the only red on this PR, and it was the live PRIDE download (`HttpIOException: The response ended prematurely`) rather than anything in the change. It has now passed on its own, so there is nothing to wave through.

That is your stated condition met. Your question about whether the filter is needed given MetaMorpheus already filters at its call site is answered above — short version: MetaMorpheus does, FlashLFQ is next through this engine and does not, and the `.First()` crash on a match carrying no identification was reachable regardless of ambiguity filtering.

At 1/2 with acesnik. This is the last quant PR in the batch apart from #1180, and an mzLib release is waiting on both.

PR #1211, opened 2026-08-26
Collapse the roll-up strategies onto one aggregation-driven implementation

trishorts:
Follow-up to the review of #1194, which split sample collapsing into a grouping strategy and an aggregation strategy and noted the roll-ups could go the same way.

Which rows get combined during roll-up is not a strategy decision — the caller's map already says so. The only behaviour a roll-up strategy carries is *how* the values are combined, and that is exactly what `IAggregationStrategy` names. `SumRollUp` and `MedianRollUp` were therefore two copies of one loop differing in a single line.

### What it does

`AggregatingRollUp` is that loop, an **abstract** base taking an `IAggregationStrategy`. `SumRollUp` and `MedianRollUp` become two-line subclasses naming their aggregation, so **no caller changes**.

Abstract rather than concrete (per review): there is no default way to combine intensities, so a roll-up has to be named by a concrete class. That is what keeps `QuantificationParameters` a record of what was actually done to compute the intensities, rather than "some aggregation" — the same reason the strategies were split in the first place. Adding an estimator is now a two-line subclass instead of a fresh copy of the loop; the tests define a `MaxRollUp` to show it, and #1212 adds `Top3RollUp` for real.

Zeros are excluded before aggregating, which is what `MedianRollUp` already did. The matrix uses `0` for "not observed", so a row never measured in a sample should not pull a median down. Excluding zeros does not change a sum, which is why both named strategies keep their previous results — the 69 existing quantification tests pass untouched.

The exclusion is done by the roll-up, **not** by the aggregation: `IAggregationStrategy.Aggregate` is specified to include zeros and leave the choice to its caller. So the same aggregation passed as `QuantificationParameters.CollapseAggregationStrategy` will see zeros and behave differently there, and an aggregation whose meaning depends on the exclusion is not recommended in that role. Now stated on the class.

### It also fixes a live bug

`SumRollUp` accumulated with `+=` into an array straight from `ArrayPool<double>.Shared`:

```csharp
double[] summedValues = ArrayPool.Rent(matrix.ColumnCount);   // NOT zeroed
foreach (int lowIndex in lowIndices)
    for (int col = 0; col < matrix.ColumnCount; col++)
        summedValues[col] += matrix.Matrix[lowIndex, col];
```

`Rent` does not zero what it hands back. `SumRollUp` returned *its own* buffers cleared — but `MedianAggregation` and the collapse strategies share that pool and return theirs uncleared:

```csharp
ArrayPool<double>.Shared.Return(rented);   // MedianAggregation -- no clearArray
```

So a sum could pick up another strategy's leftovers, silently, as inflated intensities. It needs a median or a collapse to have run first in the same process, which is why the existing tests don't catch it.

The new loop writes every column before reading it, and bounds the gather buffer by count rather than by length. `SumIsCorrectEvenWhenTheSharedPoolHandsBackADirtyBuffer` poisons the shared pool the way `MedianAggregation` does — **it fails against the old `SumRollUp` (verified) and passes now.**

### One small removal

`SumRollUp`'s `public ArrayPool<double> ArrayPool` property, which exposed an implementation detail no caller has a use for.

### Tests

12. `Test.Quantification` green at 80.

### DIFF
--- a/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
+++ b/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
@@ -1,0 +1,81 @@
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using Quantification.Interfaces;
+
+namespace Quantification.Strategies
+{
+    /// <summary>
+    /// Rolls lower-level rows up to higher-level rows by combining each column with an
+    /// <see cref="IAggregationStrategy"/>.
+    ///
+    /// Which rows are combined is not a strategy decision here -- the caller's map already says so --
+    /// so the only behaviour a roll-up carries is how the values are combined, and that is exactly
+    /// what IAggregationStrategy names. Splitting it out is the same separation that
+    /// <see cref="ICollapseStrategy"/> and IAggregationStrategy make for sample collapsing, and it
+    /// keeps <see cref="QuantificationParameters"/> readable as a full description of the processing:
+    /// <see cref="SumRollUp"/> still means sum and nothing else.
+    ///
+    /// Zeros are excluded before aggregating. The matrix uses 0 for "not observed", so a row that was
+    /// never measured in a sample should not pull a median down or a mean toward zero. Summing is
+    /// unaffected by the exclusion, which is why <see cref="SumRollUp"/> and <see cref="MedianRollUp"/>
+    /// both keep their previous results.
+    /// </summary>
+    public class AggregatingRollUp : IRollUpStrategy
+    {
+        private readonly IAggregationStrategy _aggregation;
+
+        public AggregatingRollUp(IAggregationStrategy aggregation)
+        {
+            _aggregation = aggregation ?? throw new ArgumentNullException(nameof(aggregation));
+        }
+
+        public string Name => _aggregation.Name + " Roll-Up";
+
+        public QuantMatrix<THigh> RollUp<TLow, THigh>(QuantMatrix<TLow> matrix, Dictionary<THigh, List<int>> map)
+            where TLow : IEquatable<TLow>
+            where THigh : IEquatable<THigh>
+        {
+            var rolledUpMatrix = new QuantMatrix<THigh>(map.Keys, matrix.ColumnKeys, matrix.ExperimentalDesign);
+
+            var pool = ArrayPool<double>.Shared;
+            double[] aggregated = pool.Rent(matrix.ColumnCount);
+            double[] observed = pool.Rent(Math.Max(1, matrix.RowCount));
+
+            try
+            {
+                foreach (var kvp in map)
+                {
+                    List<int> lowIndices = kvp.Value;
+
+                    for (int col = 0; col < matrix.ColumnCount; col++)
+                    {
+                        // Gather the observed values for this column. A rented buffer is not zeroed,
+                        // so the count -- not the buffer length -- bounds what the aggregator sees.
+                        int observedCount = 0;
+                        foreach (int lowIndex in lowIndices)
+                        {
+                            double value = matrix.Matrix[lowIndex, col];
+                            if (value > 0)
+                            {
+                                observed[observedCount++] = value;
+                            }
+                        }
+
+                        aggregated[col] = _aggregation.Aggregate(
+                            new ReadOnlySpan<double>(observed, 0, observedCount));
+                    }
+
+                    rolledUpMatrix.SetRow(kvp.Key, aggregated);
+                }
+            }
+            finally
+            {
+                pool.Return(aggregated, clearArray: true);
+                pool.Return(observed, clearArray: true);
+            }
+
+            return rolledUpMatrix;
+        }
+    }
+}
--- a/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
+++ b/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
@@ -21,10 +21,16 @@ namespace Quantification.Strategies
     /// <see cref="QuantificationParameters"/> says what was actually done to compute the intensities.
     /// Adding one is a two-line subclass naming its aggregation.
     ///
-    /// Zeros are excluded before aggregating. The matrix uses 0 for "not observed", so a row that was
-    /// never measured in a sample should not pull a median down or a mean toward zero. Summing is
-    /// unaffected by the exclusion, which is why <see cref="SumRollUp"/> and <see cref="MedianRollUp"/>
-    /// both keep their previous results.
+    /// Zeros are excluded before aggregating -- exactly the zero sentinel, not everything that is not
+    /// positive. The matrix uses 0 for "not observed", so a row never measured in a sample should not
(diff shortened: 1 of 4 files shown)


Alexander-Sol (changes requested):
See my comments on #1212. AggregatingRollUp should be abstract

trishorts:
Both done in `a165719`.

`AggregatingRollUp` is abstract with a protected constructor. You're right that there's no default aggregation to fall back on, and a concrete class per behaviour is the thing that makes `QuantificationParameters` readable as a record of what was done — same argument as the collapse split in #1194. Adding an estimator stays a two-line subclass; the tests define a `MaxRollUp` at the top of the file to show that, in place of constructing the base directly.

On the comment: I've said explicitly that the zero-exclusion happens in the roll-up and **not** in the aggregation. `IAggregationStrategy.Aggregate` is documented as including zeros and leaving the choice to its caller, so the same aggregation passed as `CollapseAggregationStrategy` sees zeros and behaves differently there — and an aggregation whose meaning depends on the exclusion isn't recommended in that role. #1212 carries the sharper version of that on `Top3Aggregation` itself, since Top3 is the clearest case.

Rewrote the PR description too — it still claimed you could construct `AggregatingRollUp` directly, which your review removed. 80 quantification tests green.

Alexander-Sol (dismissed):

nbollis (commented):
## Review Summary

- 2 major issues in `mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs`.
- No summary-only findings.

Both comments are on the shared roll-up abstraction and affect public behavior or API robustness at the new aggregation-driven layer.

mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                    for (int col = 0; col < matrix.ColumnCount; col++)
+                    {
+                        // Gather the observed values for this column. A rented buffer is not zeroed,
+                        // so the count -- not the buffer length -- bounds what the aggregator sees.
+                        int observedCount = 0;
+                        foreach (int lowIndex in lowIndices)
+                        {
+                            double value = matrix.Matrix[lowIndex, col];
+                            if (value > 0)
```
nbollis: [synthesis] Shared `value > 0` filter changes SumRollUp behavior while docs claim summing is unaffected

The new shared loop excludes every non-positive value before aggregation. That is harmless for zeros, but it silently drops negative matrix values that the old SumRollUp would have summed, so the doc claim that summing is unaffected is only true for non-negative inputs.

Suggestion: If only the zero sentinel should be excluded, filter `value != 0`; otherwise document that all non-positive values are treated as unobserved and add a parity test.

mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public QuantMatrix<THigh> RollUp<TLow, THigh>(QuantMatrix<TLow> matrix, Dictionary<THigh, List<int>> map)
+            where TLow : IEquatable<TLow>
+            where THigh : IEquatable<THigh>
+        {
+            var rolledUpMatrix = new QuantMatrix<THigh>(map.Keys, matrix.ColumnKeys, matrix.ExperimentalDesign);
+
+            var pool = ArrayPool<double>.Shared;
+            double[] aggregated = pool.Rent(matrix.ColumnCount);
+            double[] observed = pool.Rent(Math.Max(1, matrix.RowCount));
```
nbollis: [synthesis] Pooled `observed` buffer can overflow on oversized group lists

The staging buffer is rented using `matrix.RowCount`, but writes are driven by `lowIndices.Count`. A caller-supplied group with duplicate indices or any list longer than the row count can drive `observedCount` past the rented length and throw mid-roll-up.

Suggestion: Size the buffer to the largest group size or validate group lists before writing into the pooled array.

trishorts: Fixed in `32a0e4a4`. The buffer is sized by the largest group now, not by row count.

The test drives 40 indices through a 2-row matrix, which overran the pool's 16-element minimum bucket before. Also treats a null index list as empty rather than throwing.

mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
trishorts: Fixed in `32a0e4a4` — the doc claim was the part that was wrong, so I made the code match what it said rather than the other way round.

Only the zero sentinel is excluded now. `0` means "not observed"; a negative is data, and dropping it would silently change a sum. So `SumRollUp` reproduces its old results for any input, and the comment is true as written.

This does change `MedianRollUp` where a value is negative: its old `v > 0` treated a negative as missing, conflating "not measured" with "measured, and less than zero". Intensities are non-negative so no existing result moves, but I said so on the class rather than leaving it implicit, and there is a parity test covering both strategies with a negative in the input.

nbollis (approved):

trishorts:
@acesnik — could you take the second approval on this one? 1/2 with nbollis.

This is the follow-up Alex asked for when he reviewed #1194 — collapsing `SumRollUp`/`MedianRollUp` onto one aggregation-driven base. No caller changes.

**It also fixes a live bug**, which is the part worth reading: `SumRollUp` accumulated with `+=` into an `ArrayPool<double>.Shared` buffer, and `Rent` does not zero what it hands back. It returned its own buffers cleared, but `MedianAggregation` and the collapse strategies share that pool and return theirs uncleared — so a sum could absorb another strategy's leftovers as inflated intensity. The regression test poisons the pool the way `MedianAggregation` does; it fails against the old code and passes now.

acesnik:
🤖 The collapse is clearly right and `abstract` addresses Alexander-Sol's point. The `ArrayPool` fix is the part that should lead the description, though, and it is broader than the description says.

Verified the whole chain on master:

- `SumRollUp.ArrayPool => ArrayPool<double>.Shared` — hardcoded, `:17`
- `Rent(matrix.ColumnCount)` at `:31` does not zero what it returns
- `summedValues[col] += ...` at `:37` reads before writing
- `MedianAggregation.cs:41` is the **only** `Return(` in the Strategies tree missing `clearArray: true` — `SumRollUp`, `MedianRollUp` and both `CollapseStrategyBase` returns all pass it

The bit worth changing: the description says this needs "a median or a collapse to have run first in the same process". `ArrayPool<double>.Shared` is process-wide and shared with every other consumer in that process — MetaMorpheus, the BCL, anything. So any code anywhere that rents `double[]` and returns it uncleared can poison a subsequent sum, and `MedianAggregation` is one instance rather than the precondition.

That makes it a live silent-inflation bug on master rather than a latent one reachable only through mzLib's own strategies, and since it changes reported protein intensities it is worth stating that way prominently rather than as a subsection. Anyone who has summed protein intensities in a process that also did anything else with pooled doubles has a reason to re-check.

`SumIsCorrectEvenWhenTheSharedPoolHandsBackADirtyBuffer` is the right test and the right name for it.

acesnik (approved):

PR #1212, opened 2026-08-26
Add Top3, the peptide-to-protein estimator that does not scale with peptide count

trishorts:
> **Stacked on #1211.** That is what makes this small: with roll-up an abstract base driven by an `IAggregationStrategy`, Top3 is an aggregation plus a two-line roll-up rather than a fourth copy of the roll-up loop. Merge #1211 first and this diff is two small files plus tests.

Summing peptides to a protein makes the protein's intensity depend on how many of its peptides were identified — a property of the search as much as of the sample. Averaging every peptide removes that bias but lets one poorly-ionizing peptide drag the estimate down.

Top3 — the average of the three most intense observed peptides ([Silva et al. 2006](https://doi.org/10.1074/mcp.M500230-MCP200)) — is the usual answer, and is the basis for comparing abundance between *different proteins* rather than one protein across samples. It is one of the two protein strategies PR #985 proposed for the LFQ half of #996; the other, median polish, is now issue #1214.

### Two judgement calls

**Fewer than three observations average what is there**, rather than dividing by three. Dividing by three would report a protein seen twice as *smaller* than the same protein seen three times — precisely the bias Top3 exists to remove.

**Not suitable as a `CollapseAggregationStrategy`**, and now documented as such on the class (per review). Top3 estimates a protein's abundance from its peptides, so it means something only when the values being combined are peptides of one protein. Collapsing fractions or technical replicates combines repeated measurements of the *same* thing, where taking the three largest is a bias rather than an estimate — and the collapse path does not exclude zeros first, so unobserved samples would count as values. It is meant to be used through `Top3RollUp`.

### Implementation notes

The top three are tracked in a single pass rather than by sorting the span. The running maxima start at `double.NegativeInfinity` rather than `0`, because the caller's buffer comes from an array pool that does not zero what it hands back.

### Tests

9, including the one that makes the case: the same per-peptide abundance rolled up for a three-peptide and a six-peptide protein.

| | 3 peptides @ 100 | 6 peptides @ 100 |
|---|---|---|
| `Top3RollUp` | 100 | 100 |
| `SumRollUp` | 300 | 600 |

`Test.Quantification` and `Test.Omics` are green — 925 passed.

### DIFF
--- a/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
+++ b/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
@@ -1,0 +1,81 @@
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using Quantification.Interfaces;
+
+namespace Quantification.Strategies
+{
+    /// <summary>
+    /// Rolls lower-level rows up to higher-level rows by combining each column with an
+    /// <see cref="IAggregationStrategy"/>.
+    ///
+    /// Which rows are combined is not a strategy decision here -- the caller's map already says so --
+    /// so the only behaviour a roll-up carries is how the values are combined, and that is exactly
+    /// what IAggregationStrategy names. Splitting it out is the same separation that
+    /// <see cref="ICollapseStrategy"/> and IAggregationStrategy make for sample collapsing, and it
+    /// keeps <see cref="QuantificationParameters"/> readable as a full description of the processing:
+    /// <see cref="SumRollUp"/> still means sum and nothing else.
+    ///
+    /// Zeros are excluded before aggregating. The matrix uses 0 for "not observed", so a row that was
+    /// never measured in a sample should not pull a median down or a mean toward zero. Summing is
+    /// unaffected by the exclusion, which is why <see cref="SumRollUp"/> and <see cref="MedianRollUp"/>
+    /// both keep their previous results.
+    /// </summary>
+    public class AggregatingRollUp : IRollUpStrategy
+    {
+        private readonly IAggregationStrategy _aggregation;
+
+        public AggregatingRollUp(IAggregationStrategy aggregation)
+        {
+            _aggregation = aggregation ?? throw new ArgumentNullException(nameof(aggregation));
+        }
+
+        public string Name => _aggregation.Name + " Roll-Up";
+
+        public QuantMatrix<THigh> RollUp<TLow, THigh>(QuantMatrix<TLow> matrix, Dictionary<THigh, List<int>> map)
+            where TLow : IEquatable<TLow>
+            where THigh : IEquatable<THigh>
+        {
+            var rolledUpMatrix = new QuantMatrix<THigh>(map.Keys, matrix.ColumnKeys, matrix.ExperimentalDesign);
+
+            var pool = ArrayPool<double>.Shared;
+            double[] aggregated = pool.Rent(matrix.ColumnCount);
+            double[] observed = pool.Rent(Math.Max(1, matrix.RowCount));
+
+            try
+            {
+                foreach (var kvp in map)
+                {
+                    List<int> lowIndices = kvp.Value;
+
+                    for (int col = 0; col < matrix.ColumnCount; col++)
+                    {
+                        // Gather the observed values for this column. A rented buffer is not zeroed,
+                        // so the count -- not the buffer length -- bounds what the aggregator sees.
+                        int observedCount = 0;
+                        foreach (int lowIndex in lowIndices)
+                        {
+                            double value = matrix.Matrix[lowIndex, col];
+                            if (value > 0)
+                            {
+                                observed[observedCount++] = value;
+                            }
+                        }
+
+                        aggregated[col] = _aggregation.Aggregate(
+                            new ReadOnlySpan<double>(observed, 0, observedCount));
+                    }
+
+                    rolledUpMatrix.SetRow(kvp.Key, aggregated);
+                }
+            }
+            finally
+            {
+                pool.Return(aggregated, clearArray: true);
+                pool.Return(observed, clearArray: true);
+            }
+
+            return rolledUpMatrix;
+        }
+    }
+}
--- a/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
+++ b/mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
@@ -21,10 +21,16 @@ namespace Quantification.Strategies
     /// <see cref="QuantificationParameters"/> says what was actually done to compute the intensities.
     /// Adding one is a two-line subclass naming its aggregation.
     ///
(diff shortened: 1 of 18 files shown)


Alexander-Sol (changes requested):
I love this, great change. However, AggregatingRollUp should be an abstract class, as their is no default aggregation behavior and it is must have a concrete class associated with it to meaningfully record what was done to calculate protein intensities.

mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// <see cref="IAggregationStrategy"/>.
+    ///
+    /// Which rows are combined is not a strategy decision here -- the caller's map already says so --
+    /// so the only behaviour a roll-up carries is how the values are combined, and that is exactly
+    /// what IAggregationStrategy names. Splitting it out is the same separation that
+    /// <see cref="ICollapseStrategy"/> and IAggregationStrategy make for sample collapsing, and it
+    /// keeps <see cref="QuantificationParameters"/> readable as a full description of the processing:
+    /// <see cref="SumRollUp"/> still means sum and nothing else.
+    ///
+    /// Zeros are excluded before aggregating. The matrix uses 0 for "not observed", so a row that was
```
Alexander-Sol: Can you edit this comment to say this method is not recommended for use as an ICollapseAggregationStrategy?

trishorts:
Thanks — both addressed.

The abstract change lives in #1211 (`a165719`) and is merged into this branch; `Top3RollUp` already satisfied it.

Your comment about the collapse aggregation strategy is on `Top3Aggregation` now, and it's the case that most deserved it. Top3 estimates a protein's abundance *from its peptides*, so it only means anything when the values being combined are peptides of one protein. Collapsing fractions or technical replicates combines repeated measurements of the same thing, where taking the three largest is a bias rather than an estimate — and the collapse path doesn't exclude zeros first, so unobserved samples would be counted as values too. Documented as not recommended in that role, with `Top3RollUp` named as the way in.

The general form of the same point is on `AggregatingRollUp` in #1211: the zero-exclusion is done by the roll-up, not by the aggregation, so any aggregation whose meaning depends on it changes behaviour when used to collapse.

Descriptions on both PRs rewritten. 925 tests green across `Test.Quantification` and `Test.Omics`.

mzLib/Quantification/Strategies/RollUp/AggregatingRollUp.cs
trishorts: Done — and split across both classes, since the point applies at two levels.

On `AggregatingRollUp` (in #1211, commit `a165719`) I documented the general form: the zero-exclusion happens **here, in the roll-up**, not in the aggregation. `IAggregationStrategy.Aggregate` is specified to include zeros and leave that choice to its caller, so the same aggregation passed as `QuantificationParameters.CollapseAggregationStrategy` sees zeros and behaves differently there. Any aggregation whose meaning depends on the exclusion is therefore not recommended in that role.

On `Top3Aggregation` (this PR) I put the specific version, since Top3 is the sharpest case: it estimates a protein's abundance *from its peptides*, so it only means anything when the values being combined are peptides of one protein. Collapsing fractions or technical replicates combines repeated measurements of the same thing, where taking the three largest is a bias rather than an estimate — and the collapse path does not exclude zeros first, so unobserved samples would count as values too. It names `Top3RollUp` as the way in.

Alexander-Sol (dismissed):

mzLib/Test/Quantification/Top3AggregationTests.cs
```diff
@@ -0,0 +1,168 @@
...
+            };
+            var design = new TestExperimentalDesign(
+                new Dictionary<string, ISampleInfo[]> { [File] = columns.ToArray() });
+
+            // Six peptides, all at 100. Rows 0-2 are one protein, rows 0-5 the other.
+            var rows = Enumerable.Range(0, 6).Select(i => new Key($"pep{i}")).ToList();
+            var matrix = new QuantMatrix<Key>(rows, columns, design);
+            foreach (var row in rows)
+            {
+                matrix.SetRow(row, new[] { 100.0 });
```
nbollis, line 127: [synthesis] Roll-up-level tests do not prove Top3 wiring through the shared loop

The roll-up tests use only uniform or two-observation inputs, so Mean or Median aggregation would produce the same arithmetic results. The exact Top3 estimator is therefore pinned mostly by the Name string and direct aggregation tests, not by an end-to-end roll-up case with heterogeneous peptide intensities.

Suggestion: Add a roll-up test with heterogeneous peptide values, e.g. `{100, 80, 60, 5, 1}` rolling up to `80`, so Mean and Median would fail.

nbollis (commented):
## Review Summary

- 1 suggestion in `mzLib/Test/Quantification/Top3AggregationTests.cs`.
- No summary-only findings.

This is a test-strength comment: the current roll-up integration cases do not distinguish Top3 from Mean or Median end-to-end through the shared loop.

mzLib/Test/Quantification/Top3AggregationTests.cs
trishorts: Fixed in `5804ab39`, with the input you suggested.

A roll-up over `{100, 80, 60, 5, 1}` separates all four estimators — Top3 = 80, Mean = 49.2, Median = 60, Sum = 246 — and the test asserts all of them, so a mis-wiring reads as a visibly different number rather than a coincidental match.

Mean and Median have no named roll-up to call, so the test defines a subclass of `AggregatingRollUp` to drive them. That is the two-line cost the abstract base was meant to have.

nbollis (approved):

trishorts:
@acesnik — second approval please, if you can. 1/2 with nbollis.

**Stacked on #1211** — merge that first and this shrinks to two small files plus tests. Adds Top3 (Silva et al. 2006) as an `IAggregationStrategy` plus a two-line named roll-up, which is the shape #1211 was meant to make cheap.

One judgement call worth checking: fewer than three observed peptides average what is there rather than dividing by three. Dividing by three would report a protein seen twice as smaller than the same protein seen three times, which is the bias Top3 exists to remove.

acesnik:
🤖 Both review points are addressed — `AggregatingRollUp` is abstract, and the collapse-strategy caveat is documented. The Silva 2006 citation with the reasoning for why Top3 rather than sum or mean is the kind of comment that survives a rewrite, and worth having.

One process point rather than a code one, which is why I have not approved yet: this branch is a git descendant of `quant/rollup-aggregation`, so its diff against master contains all of #1211's collapse, including the `ArrayPool` fix. Reviewing it now means reviewing that collapse a second time inside a larger diff, and re-reviewing it after #1211 merges when the diff shrinks.

Merging #1211 first reduces this to the Top3 addition alone, which is the review actually worth having here. Happy to approve as soon as that lands — flagging it so it does not look like this is waiting on anything of yours.

PR #802, opened 2024-09-20
Changes to MBR within FlashLFQ

Alexander-Sol:
Major changes to MBR within FlashLFQ.

### Peak Decoys
When performing MBR, we attempt to locate a decoy peak for each acceptor peak. Acceptor peaks are located by predicting a retention time and looking for the theoretical peptide mass at that retention time. Decoy peaks are located by choosing a "random" retention time and looking for the theoretical peptide mass at the random retention time. 

### PEP Calculation for MBR Peak
All MBR peaks are fed into a binary classifier (similar to the one used within MetaMorpheus) and semi-supervised learning is performed to calculate a PEP for each peak.

### MBR Q-Values
For each file, Q-Values are calculated for each MBR peak by ranking them in PEP order and performing target-decoy competition

---

This pull request increases both the accuracy and precision of FlashLFQ. This was tested by analyzing a spike-in dataset, originally published by [Shen et al.](https://www.pnas.org/doi/10.1073/pnas.1800541115?url_ver=Z39.88-2003&rfr_id=ori%3Arid%3Acrossref.org&rfr_dat=cr_pub++0pubmed). In this dataset, e. coli peptides were spiked into human peptides at 5 different ratios (1x, 1.5x, 2x, 2.5x, 3x). There were four replicates for each condition/ratio.

These files were analyzed with the previous version of FlashLFQ (with MBR and normalization enabled, denoted as v1.0 below) and the new version (with MBR and normalization enabled, 1% PIP FDR, 0.2% Donor FDR, denoted as PIP-ECHO below) . These results show that the new version of FlashLFQ has increased quantitative accuracy ( the average fold change is closer to the true fold change) and precision (smaller variance and CV).

Spike-In Comparison

| Comparison   | v1.0 Avg FC | PIP-ECHO Avg. FC | Empirical FC |
| :---------------- | ------: | ----: |------: | 
| 1x vs 1.5x         | 0.484 | 0.497 | 0.585 | 
| 1x vs 2x           | 0.857 | 0.871 | 1 | 
|1x vs 2.5x         | 1.16 | 1.16 | 1.32 | 
| 1x vs 3x           |  1.48 | 1.48 | 1.58 | 

| Comparison   |  v1.0 CV | PIP-ECHO CV | 
| :---------------- | ----: |------: | 
| 1x vs 1.5x         |  0.854 | 0.619 | 
| 1x vs 2x           |  0.544 | 0.405 | 
|1x vs 2.5x         | 0.436 | 0.324 | 
[...]

PR #1216, opened 2026-08-26
Changed StdDev of RT Error Distribution for MBR

Alexander-Sol:
Previously, the distribution used to score the retention time of MBR hits had it's std.dev hard-coded to 1. This PR sets the std.dev by calculating the rt deviation between MS2 identified peaks in the donor and acceptor runs. This will allow RT difference to better discriminate between true and false MBR transfers

Additionally, changed two tests that were impacted by this change.

### Note on `RetentionTimeCalibDataPoint.RtDiff` sign convention

This PR also flips the sign of the public `RtDiff` property from `acceptor - donor` to `donor - acceptor`. This is deliberate: the old `RtDiff` disagreed with what `AddRtPredErrorDistribution` was actually fed (`FlashLfqEngine` built its own `donor - acceptor` list), so the flip is what makes the new `RetentionTimeCalibrationCurve` type substitutable for that list, and it matches `PredictRetentionTime`'s local computation. The quantity is now defined in exactly one place. No external consumer (including MetaMorpheus) reads `RtDiff`.

### Review-response cleanup (in response to @trishorts)

- Added a test that drives `RtStandardDeviationMin` (the floor): tightly-aligned anchors are held at ~0.55 `RtScore` instead of collapsing to `_minScore`.
- Removed the unreachable `RtStandardDeviationDefault` branch; the constant now names the `1` used as the actual default distribution (in `AddRtPredErrorDistribution` and the `ScoreMbr` fallback).
- Removed the dead `anchorPeptideRtDiffs` local left behind by the refactor; `PredictRetentionTime` now reads `RtDiff` directly instead of recomputing it.
- `RtDiff` kept `readonly`; tests use a new `internal` constructor instead of a `protected set`.
- Standardized the robust-sigma divisor to `IQR / 1.36` (matching the three other uses in the file).
- Dropped the unused usings (`Easy.Common.Extensions`, and `MathNet.Numerics.Statistics` in `MzLibUtil/ClassExtensions.cs`), fixed the stale `<param>` doc, and reverted the unrelated `mzLib.sln` VS-version rewrite.


mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -285,7 +285,7 @@ internal void AddRtPredErrorDistribution(SpectraFileInfo donorFile, List<double>
                         // However, changing this now would change MBR scores in all existing tests that use MBR
                         // I'm not sure what the overall effect of fixing this is, and plan to carefully evaluate it 
                         // in the near future and then fix this issue.
-                        rtPredictionErrorDist = new Normal(medianRtError, 1);
+                        rtPredictionErrorDist = new Normal(medianRtError, sigma);
```
trishorts: **This is the main thing I'd like resolved.**

`sigma` is the spread of `rtPredictionErrors`, and each of those is `avgDiff - anchorPeptideRtDiffs[i]` where `avgDiff` is the mean of the **+/-3 neighbours in list order** (lines 256-268 above). The method's own comment says the point is "using neighboring anchor peptides to predict the Rt shift" - that premise requires the list to be sorted by retention time.

Tracing where the list comes from, I don't think it is:

- `FlashLfqEngine.cs:619` appends to `anchorPeptideRtDiffs` while iterating `acceptorFileAllMsmsPeaks`, which is a `Dictionary<string, List<ChromatographicPeak>>` **keyed by modified sequence** (built at `:588`).
- That dictionary's order traces back to `_results.Peaks[acceptor]`, which is filled at `:551` from an array indexed by `ms2IdsForThisFile` (`:479`) - i.e. the order of `_allIdentifications`, which is PSM order from the caller, not RT.
- Tellingly, `:627` *does* sort - `rtCalibrationCurve.OrderBy(p => p.DonorFilePeak.Apex.IndexedPeak.RetentionTime)` - because `PredictRetentionTime` binary-searches it at `:739`. But that sort happens **after** `:624` has already handed the unsorted `anchorPeptideRtDiffs` to this method.

If that's right, the "neighbours" being averaged are arbitrary peptides from anywhere in the gradient, so this spread is gradient-wide peptide-to-peptide RT-diff scatter, not local prediction error.

Why it didn't matter before and does now: with `sigma` pinned to 1, the only thing consumed from these errors was `medianRtError` - a median, which is fairly insensitive to ordering here and sits near 0 either way. This PR promotes an **order-dependent** statistic to control the score.

Sorting `anchorPeptideRtDiffs` by donor RT before `:624` (or building it from the already-sorted calibration curve) would be a small change and would make `sigma` mean what the doc comment says it means. Am I missing a sort somewhere?

trishorts: **No lower bound on `sigma` - and it fails hardest on the best data.**

The guard on line 279 rejects `stdDevRtError <= 0`, but not a `sigma` that is positive and tiny. `CalculateScore` (line 331) returns `2 * Phi(-|x - mu| / sigma)` floored at `_minScore = 3e-7`, so as `sigma` shrinks the score falls off a cliff.

I simulated this method with anchor diffs that agree to within a given jitter (n=3, probing a genuinely good transfer with a 0.01 min / 0.6 s RT error):

```
anchor jitter    sigma       RtScore old(sigma=1)  RtScore new       MbrScore factor
+/- 0.01  min    4.07e-03    0.9929                2.84e-02          x0.49
+/- 0.001 min    5.06e-04    0.9920                3.00e-07 (floor)  x0.05
+/- 0.0001 min   3.16e-05    0.9920                3.00e-07 (floor)  x0.05
```

Because the final score is the geometric mean of five components, flooring one of them multiplies the MBR score by `(3e-7)^(1/5)` ~ 0.05 - a ~20x crush applied uniformly to **every** acceptor peak for that donor file. There's no exception and no warning; MBR just quietly stops producing competitive scores for the best-aligned file pairs.

The second-order effect is the one that actually defeats the PR's goal: once every candidate floors at `_minScore`, `RtScore` is a **constant**, so RT contributes nothing to choosing between candidate peaks at `FlashLfqEngine.cs:1034`, `:1230` and `:1376`. And since the distribution is per-donor-file, a tightly-aligned donor gets systematically crushed scores relative to a loosely-aligned one, which matters where scores are ranked across donor groups (`DonorGroup.BestTargetMbrScore`, `PepAnalysisEngine.cs:67`, `:193`).

Suggestion: floor it at something physically meaningful rather than leaving it unbounded, e.g.

```csharp
double sigma = ... ;
sigma = Math.Max(sigma, minRtSigma); // instrument RT reproducibility, or a fraction of the MBR RT window
```

**Separately, on the estimator itself:** this pairs `medianRtError` (robust) with `StandardDeviation()` (not robust). RT prediction errors are heavy-tailed - a handful of bad alignments will inflate `sigma` and *widen* the distribution, which reduces discrimination. If you want a scale estimator consistent with the median, `1.4826 * MAD` is the usual pairing and it gives you the small-sigma problem in a more controlled form. Worth a look while you're in here.

mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -285,7 +285,7 @@ internal void AddRtPredErrorDistribution(SpectraFileInfo donorFile, List<double>
                         // However, changing this now would change MBR scores in all existing tests that use MBR
                         // I'm not sure what the overall effect of fixing this is, and plan to carefully evaluate it 
                         // in the near future and then fix this issue.
```
trishorts: This comment block (lines 283-287) is now stale and actively misleading - it tells the next reader that the bug is *unfixed* and that someone "plan[s] to carefully evaluate it in the near future and then fix this issue," on the very line that fixes it.

Please delete lines 283-287. If you want to keep the provenance, one line is enough:

```csharp
// sigma is the observed spread of local RT prediction errors; previously hard-coded to 1 (bug introduced in #802)
```

mzLib/Test/FlashLFQ/TestPipEcho.cs
```diff
@@ -124,10 +124,10 @@ public static void TestMbrScorer()
 
             acceptorPeak.MbrScore = scorer.ScoreMbr(acceptorPeak, donorPeak, predictedRt: 25.1);
 
-            Assert.That(acceptorPeak.MbrScore, Is.EqualTo(73.7).Within(0.1));
+            Assert.That(acceptorPeak.MbrScore, Is.EqualTo(61.5).Within(0.1));
             Assert.That(acceptorPeak.PpmScore, Is.EqualTo(1).Within(0.01));
             Assert.That(acceptorPeak.IntensityScore, Is.EqualTo(0.46).Within(0.01));
-            Assert.That(acceptorPeak.RtScore, Is.EqualTo(0.96).Within(0.01));
+            Assert.That(acceptorPeak.RtScore, Is.EqualTo(0.39).Within(0.01));
```
trishorts: Worth spelling out what these new golden values encode, because this test is a clean worked example of the `sigma` behaviour.

With the input on line 123 - `{0.5, 0.6, 0.5, 0.6, 0.5, 0.6, 0.5}`, `numberOfAnchorPeptides: 2` - the method produces prediction errors `[0.05, -0.05, 0.05]`, so `medianRtError = 0.05` and `sigma = 0.0577`. That's what moves `RtScore` from 0.96 to 0.39 and `MbrScore` from 73.7 to 61.5.

Two things follow:

1. The `sigma` here is an artifact of a synthetic strictly-alternating input, not of anything resembling real alignment error. Which is fine for a regression test, but it means the new numbers don't validate the change - they just record it.
2. It's the same mechanism as my comment on `MbrScorer.cs:288`, in miniature: a 0.05 min error against a 0.058 min `sigma`. Shrink the jitter in that input list and `RtScore` goes to `_minScore`.

Since this PR's whole payload is "`sigma` should come from the data," could you add one test that actually asserts that - e.g. call `AddRtPredErrorDistribution` twice with a tight and a loose set of anchor diffs and assert the resulting `RtScore` for the same peak is higher under the loose one? That's the behaviour the PR claims, and right now nothing checks it. It would also pin the degenerate-sigma case once you decide how to handle it.

mzLib/Test/FlashLFQ/TestFlashLFQ.cs
```diff
@@ -1577,7 +1577,7 @@ public static void RealDataMbrTest()
             Console.WriteLine("r1 PIP event count: " + f1r1MbrResults.Count);
             Console.WriteLine("r2 PIP event count: " + f1r2MbrResults.Count);
             Assert.AreEqual(141, f1r1MbrResults.Count);
-            Assert.AreEqual(78, f1r2MbrResults.Count);
+            Assert.AreEqual(77, f1r2MbrResults.Count);
```
trishorts: This is a golden-value update (78 -> 77) with no accompanying assertion about *why* the count changed - so if a future change to `sigma` moved it back to 78, or to 60, the test would just be re-baselined again.

One fewer MBR hit is also consistent with either reading of this PR: better discrimination correctly rejecting a false transfer, or the score collapse I describe on `MbrScorer.cs:288` starting to bite. Worth confirming which by checking whether the dropped peptide was a plausible transfer.

Not asking you to convert this into a results-quality test - just noting that between this file and `TestPipEcho.cs`, the PR currently re-baselines three numbers and adds no test of the new behaviour.

trishorts (commented):
Thanks for finally pulling the trigger on this one - the hard-coded `1` has been wrong since #802 and it's good to see it go.

I ran the numbers on this and I have two things I think need resolving before it merges, both about **what `sigma` is actually measuring**. Neither is a reason to abandon the change; the change is right in principle.

**1. The list `sigma` is derived from is not retention-time ordered.** `AddRtPredErrorDistribution` averages the +/-3 *neighbours* of each entry to estimate a local alignment error - that only means something if the list is sorted by RT. As far as I can trace it, `anchorPeptideRtDiffs` isn't. Details in the inline comment.

**2. There is no lower bound on `sigma`.** The existing guard catches `stdDev <= 0` but not "tiny". When the local alignment is *good*, `sigma` collapses and `CalculateScore` floors every acceptor peak at `_minScore` (3e-7), crushing the MBR score ~20x for the whole donor file - and, worse, making `RtScore` a constant, so it stops discriminating between candidate peaks at all (`FlashLfqEngine.cs:1034`, `:1230`, `:1376`). That's the opposite of what the PR is for. Simulation numbers inline.

**On evidence.** The PR body says this "will allow RT difference to better discriminate between true and false MBR transfers," but the only visible evidence is that `RealDataMbrTest` lost one hit (78 -> 77) and `TestMbrScorer`'s `RtScore` fell 0.96 -> 0.39. Those are consistent with better discrimination *and* with the collapse in (2) - they don't separate the two.

FlashLFQ already has exactly the instrument to settle it: the `RandomRt` decoy peaks and `MbrQValue` / PIP-ECHO machinery in `PepAnalysisEngine`. Could you post the target-vs-decoy MbrScore separation, or the number of MBR peptides at a fixed MBR q-value, before and after? That would turn this from a plausible fix into a demonstrated one, and it's the number I'd want in hand given this changes quantification for every downstream MetaMorpheus user.

Rest is minor: a stale TODO that now contradicts the code, plus one thing GitHub won't let me comment on inline.

---

**`mzLib/mzLib.sln` - unrelated churn.** (GitHub's API doesn't expose the `.sln` as a commentable file, so it's here rather than inline.) This looks like Visual Studio 18 auto-migration: it bumps `VisualStudioVersion` to `18.7.12002.237 oobstable`, reshuffles ~200 lines of `ProjectConfigurationPlatforms`, and adds a brand-new `TestAndRelease|Any CPU` solution configuration that didn't exist on master - in a PR whose payload is one line of MBR statistics.

I checked and I believe it's functionally inert: the workflows pin the configuration explicitly (`dotnet.yml:28`/`:88`, `release.yml:44`/`:49` all use `Release` + `/p:Platform="Any CPU"`), nothing references `TestAndRelease|Any CPU`, and - importantly - the `Development` project's deliberately-absent `Release|x64.Build.0` and `Release|Any CPU.Build.0` rows are preserved. (Those omissions are load-bearing: re-adding them makes `release.yml`'s solution-wide `dotnet test` try to build dev-only fixtures, and PR CI won't catch it.)

So no objection on correctness, but it buries the actual diff and would mask a real `.sln` change later. Could you drop it?

```
git checkout origin/master -- mzLib/mzLib.sln
```

nbollis (dismissed):

acesnik:
🤖 @Alexander-Sol — the change is clearly right. A scoring term whose width was hardcoded to a constant, in the component that decides whether a quantification value gets reported at all, is the kind of thing that goes unnoticed for years because nothing about it looks wrong until someone measures what the term is actually contributing. The length of the review above reflects what MBR touches rather than doubt about the direction.

Two things worth adding. The golden values in `RealDataMbrTest` don't reproduce off Windows: an unmodified master baseline here gives 140/77 where the test asserts 141/78, which is ML.NET drift against the `windows-latest` they were pinned on, and the test half-anticipates it at `TestFlashLFQ.cs:1575-1576`. So the 78 → 77 in your diff can't be read in either direction, and checking whether the dropped peptide was a plausible transfer may meet the same wall.

That leaves the separation measurement as the only thing that will settle this, which seems worth knowing before you build it — because a single number won't get you there. Both readings on the table, real discrimination and the collapse trishorts simulated, predict a wider target/decoy gap, so a median ratio or a count at a fixed q-value moves the same way under either and can't tell you which. A distribution separates them: a collapse bunches candidates against `_minScore` and holds them there, while discrimination spreads them across the range. A histogram of `RtScore` for targets and decoys, per donor file, before and after, answers his question and closes concern (2) in one figure.

Pin both arms to one machine and one cached index, given the drift above. Curious what the shape turns out to be.

acesnik (commented):
🤖 @Alexander-Sol — the change is clearly right. A scoring term whose width was hardcoded to a constant, in the component that decides whether a quantification value gets reported at all, is the kind of thing that goes unnoticed for years because nothing about it looks wrong until someone measures what the term is actually contributing. The length of the review above reflects what MBR touches rather than doubt about the direction.

Two things worth adding. The golden values in `RealDataMbrTest` don't reproduce off Windows: an unmodified master baseline here gives 140/77 where the test asserts 141/78, which is ML.NET drift against the `windows-latest` they were pinned on, and the test half-anticipates it in the comment above those assertions — "the counts and correlation values can be quite variable. Any change to ML.NET or the PEP Analysis engine will cause these to change". So the 78 → 77 in your diff can't be read in either direction, and checking whether the dropped peptide was a plausible transfer may meet the same wall.

That leaves the separation measurement as the only thing that will settle this, which seems worth knowing before you build it — because a single number won't get you there. Both readings on the table, real discrimination and the collapse trishorts simulated, predict a wider target/decoy gap, so a median ratio or a count at a fixed q-value moves the same way under either and can't tell you which. A distribution separates them: a collapse bunches candidates against `_minScore` and holds them there, while discrimination spreads them across the range. A histogram of `RtScore` for targets and decoys, per donor file, before and after, answers his question and closes concern (2) in one figure.

Pin both arms to one machine and one cached index, given the drift above. Curious what the shape turns out to be.

Alexander-Sol:
### Target/decoy score separation: before vs. after this change

To quantify the effect of the RT std.dev fix, I ran a new MBR target/decoy distribution test (`Test.FlashLFQ.MbrTargetDecoyScoreDistributionTest`) on the `RealDataMbrTest` sliced data, once with this PR's data-driven σ and once with the previous hard-coded σ = 1. The metric is a rank-based AUC — the probability that a randomly chosen target transfer outscores a randomly chosen decoy transfer (0.5 = no separation, 1.0 = perfect). All scored MBR transfers (targets + random-RT decoys + decoy-peptide + double-decoy) are pulled from `results.Peaks`.

| Score component | Previous (σ = 1) | This PR (data-driven σ) | Δ |
|---|---|---|---|
| **MbrScore (composite)** | **0.593** | **0.816** | **+0.223** |
| RtScore | 0.825 | 0.863 | +0.038 |
| IntensityScore | 0.703 | 0.637 | −0.066 |
| PpmScore | 0.579 | 0.568 | −0.011 |
| ScanCountScore | 0.787 | 0.785 | ≈0 |
| IsotopicDistributionScore | 0.303 | 0.369 | +0.066 |

The composite MBR score separation jumps from **0.59 → 0.82**. The mechanism is clearest in the RtScore medians:

| | target median | decoy median |
|---|---|---|
| Previous (σ = 1) | 0.9902 | 0.9645 |
| This PR | 0.4330 | 0.0010 |

With σ hard-coded to 1 minute, the RT score saturated near 1 for essentially every transfer, so it barely distinguished targets from decoys and contributed little to the composite — dragging the composite AUC down to 0.59. Setting σ from the observed donor/acceptor RT deviations lets the RT score actually discriminate (target 0.43 vs decoy 0.001), and that discriminating power flows into the composite. This is direct evidence for the PR's premise that a properly-scaled RT distribution better separates true from false MBR transfers.

Peak counts are essentially unchanged (238 → 237 total MBR transfers on this data; `RealDataMbrTest`'s reported MBR counts shift 141/78 → 142/77).

**Unrelated observation:** `IsotopicDistributionScore` is anti-discriminative (AUC < 0.5) both before and after this change — decoys systematically score *higher* on it. This is a pre-existing issue independent of this PR: a Pearson correlation over the minimum 2 observed isotopes is always exactly 1, so peaks with only 2 isotopes (disproportionately decoys/noise) receive the maximum isotopic score. Worth a separate follow-up.

Alexander-Sol:
This PR will be followed by additional PRs to address issues #1282  and #1283

nbollis (dismissed):

Alexander-Sol:
### Target/decoy separation on the full E. coli spike-in vignette (larger, production-scale dataset)

Following up on the [sliced-test-data comparison above](https://github.com/smith-chem-wisc/mzLib/pull/1216#issuecomment-5589530592): I re-ran the same before/after target/decoy score-separation analysis on the full E. coli spike-in vignette (two ~1.3 GB Thermo `.raw` files, `AllPSMs.psmtsv` → ~135k identifications, ~16.5k MBR transfers), which is far less noisy than the tiny sliced data. Metric is the same rank-based AUC — P(random target transfer outscores a random decoy transfer); 0.5 = no separation.

| Score component | Before (σ = 1) | This PR (data-driven σ) | Δ |
|---|---|---|---|
| **MbrScore (composite)** | 0.764 | **0.804** | **+0.040** |
| RtScore | 0.641 | 0.677 | +0.036 |
| ScanCountScore | 0.591 | 0.613 | +0.022 |
| IntensityScore | 0.637 | 0.652 | +0.015 |
| PpmScore | 0.749 | 0.762 | +0.013 |
| IsotopicDistributionScore | 0.421 | 0.406 | −0.015 |

Every component improves except the isotopic one, and the composite separation rises from **0.764 → 0.804**. As on the sliced data, the mechanism is visible in the RtScore medians:

| | target median | decoy median |
|---|---|---|
| Before (σ = 1) | 0.9278 | 0.8800 |
| This PR | 0.3123 | 0.0029 |

With σ hard-coded to 1 minute the RT score is saturated near 1 for essentially all transfers (target 0.93 vs decoy 0.88 — barely distinguishable), so it contributes little to the geometric-mean composite. The data-driven σ restores the RT score's dynamic range (0.31 vs 0.003), and that discriminating power flows into the composite.

Transfer counts are effectively unchanged (16,464 → 16,484 total; targets 15,198 → 15,146; decoys 1,266 → 1,338), so the improvement is a genuine gain in ranking, not a change in how many transfers are produced.

The effect here (+0.04 composite) is smaller than on the sliced data (+0.22) — expected, since the tiny sliced set was pathologically dragged down by the saturated RT score (composite 0.59), whereas the production-scale composite started at a healthy 0.76. Both datasets agree on the direction and mechanism, and the vignette is the more trustworthy number.

**Unrelated:** `IsotopicDistributionScore` remains anti-discriminative (AUC < 0.5) both before and after — decoys score higher on it. That is independent of this PR and is tracked separately in #1282.

mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -272,20 +278,16 @@ internal void AddRtPredErrorDistribution(SpectraFileInfo donorFile, List<double>
...
-                            : stdDevRtError;
-
-                        //TODO: This distribution should use the calculated sigma value, not 1 for all cases
-                        // apparently, this is a long-standing bug that was introduced in PR #802
-                        // However, changing this now would change MBR scores in all existing tests that use MBR
-                        // I'm not sure what the overall effect of fixing this is, and plan to carefully evaluate it 
-                        // in the near future and then fix this issue.
-                        rtPredictionErrorDist = new Normal(medianRtError, 1);
+                        if(double.IsNaN(stdDevRtError))
+                            stdDevRtError = RtStandardDeviationDefault;
```
trishorts: `RtStandardDeviationDefault` is dead, and the two places that actually implement the default it describes do not use it.

I checked whether this branch can fire. Once you reach line 281 you are guaranteed `rtPredictionErrors.Count >= 2` and every entry finite (they are filtered at line 272), and `InterquartileRange()` on that returns a finite number in every case -- including the fully degenerate one, where it returns **0, not NaN**. I ran it:

```
n=2  [0.05,-0.05]          IQR/1.35 = 7.41e-2   NaN=False
n=3  [0,0,0]               IQR/1.35 = 0         NaN=False
n=3  [1e-12,-1e-12,1e-12]  IQR/1.35 = 1.23e-12  NaN=False
n=3  [0.05,-0.05,0.05]     IQR/1.35 = 6.17e-2   NaN=False
```

So line 286 never executes -- which is also what Codecov's "1 Missing and 2 partials" on this file is pointing at.

Meanwhile the field's own comment says it is "a default StdDev for when a donor file has no anchor peptides", but that case is handled twice by a hard-coded literal instead: `new Normal(0, 1)` at line 249, and again in `ScoreMbr`'s missing-key fallback. So the constant is defined for a purpose it is not used for, and used at a site that cannot be reached.

Either drop lines 285-286 and use `RtStandardDeviationDefault` at line 249 and in the `ScoreMbr` fallback, or drop the field. I would prefer the former -- naming that `1` is worth doing, given it is the same hard-coded `1` this PR exists to remove from the other branch.

mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -272,20 +278,16 @@ internal void AddRtPredErrorDistribution(SpectraFileInfo donorFile, List<double>
...
+                        if(stdDevRtError < RtStandardDeviationMin)
+                            stdDevRtError = RtStandardDeviationMin;
```
trishorts, line 288: **This floor is the load-bearing part of the PR, and nothing tests it.**

It works -- I confirmed that, and it is why I am closing my old thread about the score collapse. But no test in the diff drives line 288. `TestMbrScorer`'s alternating `{0.5, 0.6, ...}` input yields sigma ~ 0.062, comfortably above the floor, and `TestMbrScorer_ScoreMbr_WithoutRtDistribution_ThrowsKeyButHandledBySetup` supplies 5 points with `numberOfAnchorPeptides: 2`, which produces a single prediction error and falls through the `Count >= 2` guard. So if someone deletes lines 287-288, the suite stays green and MBR silently regresses to the behaviour I described in August.

This is the tight-vs-loose test I asked for in my `TestPipEcho` thread, in its most useful form:

```csharp
// anchor diffs that agree to within a millisecond -> sigma is floored, not collapsed
scorer.AddRtPredErrorDistribution(donorFile, TightCurve(jitter: 1e-4), 2);
...
Assert.That(acceptorPeak.RtScore, Is.GreaterThan(0.1)); // would be 3e-7 without the floor
```

For calibration, here is what I measured on this branch, holding everything else fixed and varying only the anchor jitter:

| anchor jitter (min) | RtScore | MbrScore |
|---|---|---|
| 0.1 | 0.517 | 65.2 |
| 0.01 | 0.765 | 70.5 |
| 0.001 | 0.569 | 66.5 |
| 0.0001 | 0.551 | 66.0 |
| 0 (degenerate) | 0.549 | 66.0 |

The bottom three rows are all the floor doing its job; without it they would be `3e-7`.

mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -272,20 +278,16 @@ internal void AddRtPredErrorDistribution(SpectraFileInfo donorFile, List<double>
                 if (rtPredictionErrors.Count >= 2)
                 {
                     double medianRtError = rtPredictionErrors.Median();
-                    double stdDevRtError = rtPredictionErrors.StandardDeviation();
+                    double stdDevRtError = rtPredictionErrors.InterquartileRange() / 1.35; // Use IQR to estimate stddev for robustness. IQR/1.35 is a robust estimator of stddev for normal distributions, and is less sensitive to outliers than the standard deviation.
```
trishorts: Minor: this file already uses `InterquartileRange() / 1.36` in three other places -- lines 99, 120 and 202 -- so the new `/ 1.35` is the odd one out. (The exact constant is 1.349; either rounding is defensible, but one file should not carry both.)

Good change otherwise -- pairing the median with a robust scale estimator is what I was angling at, and it is what makes the floor a backstop rather than the primary defence.

mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -238,31 +243,32 @@ private Gamma GetIsotopicEnvelopeCorrDistribution()
         /// match-between-runs for the specified donor file
         /// </summary>
         /// <param name="anchorPeptideRtDiffs">List of retention time differences (doubles) calculated as donor file RT - acceptor file RT</param>
```
trishorts: Stale doc -- the compiler flags this one:

```
MbrScorer.cs(245,26): warning CS1572: XML comment has a param tag for 'anchorPeptideRtDiffs',
but there is no parameter by that name
```

The parameter is now `calibrationCurve`, and neither it nor `numberOfAnchorPeptides` is documented. The summary above it also still says "Takes in a list of retention time differences" -- worth updating both, since the whole point of the new type is that it carries the ordering guarantee the old `List<double>` could not.

mzLib/FlashLFQ/MBR/MbrScorer.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using Easy.Common.EasyComparer;
+using Easy.Common.Extensions;
```
trishorts: Unused. I deleted this line and rebuilt `FlashLFQ.csproj`: 0 errors. Nothing in the file resolves through `Easy.Common.Extensions` -- the `InterquartileRange` / `Median` / `StandardDeviation` extensions all come from `MathNet.Numerics.Statistics`, already imported on line 5.

mzLib/MzLibUtil/ClassExtensions.cs
```diff
@@ -20,6 +20,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
+using MathNet.Numerics.Statistics;
```
trishorts: This one line is the entire diff for this file, and it is unused -- `ClassExtensions.cs` references nothing from `MathNet.Numerics.Statistics`. I deleted it and rebuilt: 0 errors.

Suggest dropping `MzLibUtil/ClassExtensions.cs` from the PR entirely. Keeping MzLibUtil out of the changed-file list also keeps the blast radius of this PR honestly scoped to FlashLFQ.

trishorts (commented):
This is a much stronger PR than the one I reviewed on 8/26. The three things I flagged as needing resolution are resolved, and I verified each one on this branch rather than taking the diff's word for it:

- **Ordering.** `RetentionTimeCalibrationCurve` establishes the donor-RT order once, in its constructor, and both consumers -- `AddRtPredErrorDistribution` and the binary search in `PredictRetentionTime` -- now read the same ordered array. That is a better fix than the sort-before-the-call I suggested, and the two new ordering tests pin it.
- **The sigma collapse.** I re-ran my original simulation against this branch. With anchor diffs agreeing to within 1e-4 min -- the case that used to floor `RtScore` at `_minScore` and crush `MbrScore` ~20x -- `RtScore` is now 0.551 and `MbrScore` 66.0. Fully degenerate input (jitter 0) gives 0.549 / 66.0. `RtStandardDeviationMin` does what it needs to do.
- **Evidence.** The target/decoy AUC numbers on both the sliced data and the E. coli vignette are exactly the measurement I asked for, and they separate the two readings of the 78 -> 77 change. Thank you for running the vignette -- the +0.04 composite on production-scale data is the number I would cite.

I am resolving four of my five earlier threads on that basis. What is left below is all small: a piece of dead code the refactor left behind, an unreachable branch, two unused usings, and an unrelated solution-file rewrite. Nothing here blocks merge.

One residual from my old test thread, re-raised inline rather than left buried: `RtStandardDeviationMin` is the guard that makes this change safe, and no test drives it.

mzLib/FlashLFQ/FlashLfqEngine.cs
```diff
@@ -589,12 +589,12 @@ private void QuantifyMs2IdentifiedPeptides(SpectraFileInfo fileInfo)
...
         /// </summary>
-        private RetentionTimeCalibDataPoint[] GetRtCalSpline(SpectraFileInfo donor, SpectraFileInfo acceptor, MbrScorer scorer,
+        private RetentionTimeCalibrationCurve GetRtCalSpline(SpectraFileInfo donor, SpectraFileInfo acceptor, MbrScorer scorer,
             out List<ChromatographicPeak> donorFileBestMsmsPeaksOrderedByMass)
         {
             Dictionary<string, ChromatographicPeak> donorFileBestMsmsPeaks = new();
             Dictionary<string, ChromatographicPeak> acceptorFileBestMsmsPeaks = new();
-            List<RetentionTimeCalibDataPoint> rtCalibrationCurve = new();
+            List<RetentionTimeCalibDataPoint> calibrationDataPoints = new();
             List<double> anchorPeptideRtDiffs = new(); // anchor peptides are peptides that were MS2 detected in both the donor and acceptor runs
```
trishorts: Dead code left behind by the refactor. `anchorPeptideRtDiffs` is declared here and populated at lines 650-653, but since `AddRtPredErrorDistribution` now takes the calibration curve, nothing reads it. I removed both the declaration and the `if (donorFilePeak.ApexRetentionTime > 0 && ...)` block that fills it, rebuilt, and got 0 errors.

Worth deleting rather than leaving -- a list named "the anchor peptide RT diffs" sitting next to the calibration curve is exactly the thing that gets accidentally re-wired later.

One behavioural note while you are removing it, in case it was not deliberate: that block gated on `ApexRetentionTime > 0` on **both** peaks, and the new path filters only on `!double.IsNaN(x.RtDiff)`. `ApexRetentionTime` returns `-1` for a null apex (`ChromatographicPeak.cs:28`), so in principle the set feeding sigma is now slightly wider. In practice a null apex would have thrown in the `RetentionTimeCalibDataPoint` constructor one line earlier, so I believe this is a no-op -- just flagging that the filter changed shape, not only location.

mzLib/FlashLFQ/MBR/RetentionTimeCalibDataPoint.cs
```diff
@@ -15,19 +17,27 @@ public RetentionTimeCalibDataPoint(ChromatographicPeak donorFilePeak, Chromatogr
 
             if (donorFilePeak != null && acceptorFilePeak != null)
             {
-                RtDiff = acceptorFilePeak.Apex.IndexedPeak.RetentionTime - donorFilePeak.Apex.IndexedPeak.RetentionTime;
+                RtDiff = donorFilePeak.Apex.IndexedPeak.RetentionTime - acceptorFilePeak.Apex.IndexedPeak.RetentionTime;
```
trishorts, line 20: The sign convention flips here -- `RtDiff` was `acceptor - donor` and is now `donor - acceptor` -- and the PR body does not mention it.

I think it is right, and it is worth saying why so the next reader does not have to re-derive it: the old `RtDiff` disagreed with what `AddRtPredErrorDistribution` was actually fed (`FlashLfqEngine.cs` built its own `donor - acceptor` list), so this change is what makes the new type substitutable for that list. It also matches `PredictRetentionTime`'s local computation. I checked for outside consumers -- nothing in MetaMorpheus touches `RtDiff`, and inside mzLib it is only this class, the scorer, and the tests -- so the flip is contained. Please just add a line to the PR description; it is a public property changing meaning.

Related cleanup, now that the two agree: `FlashLfqEngine.cs:869-871` still recomputes this quantity inline --

```csharp
List<double> rtDiffs = nearbyCalibrationPoints
    .Select(p => p.DonorFilePeak.ApexRetentionTime - p.AcceptorFilePeak.ApexRetentionTime)
```

-- which is `p.RtDiff` by definition after this change. Two definitions of one quantity is how the sign got out of step in the first place. (They differ for a null apex: `-1` there versus the constructor's NaN. That difference is an argument for one definition, not two.)

mzLib/FlashLFQ/MBR/RetentionTimeCalibDataPoint.cs
```diff
@@ -1,12 +1,14 @@
...
+using System.Linq;
 
 namespace FlashLFQ
 {
     public class RetentionTimeCalibDataPoint : IComparable
     {
         public readonly ChromatographicPeak DonorFilePeak;
         public readonly ChromatographicPeak AcceptorFilePeak;
-        public readonly double RtDiff;
+        public double RtDiff { get; protected set; }
```
trishorts: `readonly` field to `{ get; protected set; }` is a wider change than it looks: it is a binary-breaking change for anything compiled against the old mzLib, and it makes `RtDiff` mutable after construction for any external subclass of a public type -- all so `MockRtCalibDataPoint` in `TestPipEcho.cs:70` can set it.

An `internal` constructor taking the diff directly would give the tests the same seam without widening the public contract:

```csharp
internal RetentionTimeCalibDataPoint(double rtDiff)   // test seam: no peaks, diff supplied directly
{
    RtDiff = rtDiff;
}
```

with `RtDiff` kept as `{ get; private set; }`, or back to `readonly`. Test already sees mzLib internals -- it constructs `MbrScorer` and calls `AddRtPredErrorDistribution`, both `internal`.

mzLib/mzLib.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.3.32825.248
+# Visual Studio Version 18
+VisualStudioVersion = 18.7.12002.237 oobstable
```
trishorts: Visual Studio 18 has rewritten the solution file -- version header, platform reordering, `Any CPU` configuration rows for every project, and a new `TestAndRelease|Any CPU` solution configuration. That is 117 added / 78 removed lines with no relationship to MBR scoring.

I checked the specific hazard, because a solution rewrite has bitten us here before (#1127): the `Development` project deliberately has **no** `Release|*.Build.0` row, so that the solution-wide `dotnet test` in release.yml does not try to build the dev-only fixtures. That omission survived -- `{04615EBF-...}` has `Release|Any CPU.ActiveCfg` and `Release|x64.ActiveCfg` but no `.Build.0` for either. So this is safe as written.

Still, I would revert `mzLib.sln`. Adding `MbrTargetDecoyScoreDevelopmentTests.cs` to an existing project does not require a solution change -- the `ProjectReference` addition in `Development.csproj` is the only project-file change this PR actually needs -- and an unrelated VS-version rewrite in a science PR is a merge-conflict magnet for every other open PR. Next time the omission might not survive.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved. The fix is better than the sort I proposed: `RetentionTimeCalibrationCurve` establishes the donor-RT order in its constructor, so `AddRtPredErrorDistribution` and the binary search in `PredictRetentionTime` read the same ordered array and the ordering is stated once instead of being an unwritten precondition at two call sites. `TestRetentionTimeCalibrationCurveOrdersByDonorRt` pins it, and the degenerate-point test pins the stability that the null-apex probe depends on. Nothing was missing on my end -- the list genuinely was not sorted.

trishorts: Resolved. I re-ran my simulation against this branch rather than reading the diff for it. Varying only the anchor jitter, everything else fixed:

| anchor jitter (min) | RtScore | MbrScore |
|---|---|---|
| 0.1 | 0.517 | 65.2 |
| 0.01 | 0.765 | 70.5 |
| 0.001 | 0.569 | 66.5 |
| 0.0001 | 0.551 | 66.0 |
| 0 (degenerate) | 0.549 | 66.0 |

The three tight rows were `3e-7` before; `RtStandardDeviationMin` holds them at ~0.55, so the ~20x uniform crush and the constant-`RtScore` failure mode are both gone. You also took the robust-estimator suggestion, which is the better ordering of defences -- IQR/1.35 keeps sigma honest and the floor is only a backstop.

Two small follow-ups from this are in my new review: the floor is untested, and `RtStandardDeviationDefault` is unreachable.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved -- the block is gone.

mzLib/Test/FlashLFQ/TestPipEcho.cs
trishorts: Leaving this one open, but the substance is mostly addressed: `MbrTargetDecoyScoreDistributionTest` plus the vignette numbers are a stronger answer to "does sigma-from-data actually discriminate" than the unit test I sketched, because they measure it at the results level.

What is still missing is the degenerate case I mentioned at the end -- nothing drives `RtStandardDeviationMin`. `TestMbrScorer`'s input gives sigma ~ 0.062 and the other scorer test falls through the `Count >= 2` guard, so deleting the floor leaves the suite green. Details and calibration numbers on `MbrScorer.cs:287-288` in the new review.

One note on the new AUC assertion: `Is.GreaterThan(0.5)` is "better than a coin flip", and you measured 0.816. A threshold nearer the measurement -- 0.7, say -- would actually catch a regression.

mzLib/Test/FlashLFQ/TestFlashLFQ.cs
trishorts: Resolved. Moving `RealDataMbrTest` into `MbrRealDataTests.cs` with `Is.EqualTo(142).Within(10)` / `Is.EqualTo(77).Within(10)` is the right call -- the counts stop being golden values that get quietly re-baselined, and the thing the PR actually claims is now asserted separately by the target/decoy test alongside it. The vignette numbers answered the underlying question about which direction the 78 -> 77 change came from.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Correcting myself, because I buried the signal: my review body said "nothing here blocks merge", and for this particular item that is not what I meant.

**This is the one thing I am holding my approval on.** Everything else I raised -- the dead `anchorPeptideRtDiffs` local, the unreachable `RtStandardDeviationDefault` branch, the two unused usings, the stale `<param>` doc, the `mzLib.sln` rewrite, the `protected set` -- is cleanup. Take it or leave it, argue me out of any of it, and I will still approve.

This one is different only because of what it guards. The floor is the whole reason the August version of this change was unsafe and this one is not: without it, a well-aligned donor/acceptor pair drives sigma toward zero, every acceptor peak floors at `_minScore`, and `RtScore` stops discriminating between candidates entirely. It works today -- I measured it -- but nothing in the suite would notice if it stopped working, and it is two lines that a future refactor would have no reason to preserve.

A single test closes it. Feed `AddRtPredErrorDistribution` anchor diffs that agree to ~1e-4 min and assert `RtScore` stays well above `_minScore` -- on this branch that value is 0.551, and without the floor it is 3e-7, so the assertion has a five-order-of-magnitude margin and will not be flaky.

Not asking for anything beyond that. Push it and I will approve.

Worth flagging on timing, since the approval is the point: mzLib master needs two approvals *and* an up-to-date branch, and any diff-changing push dismisses a standing approval. So the efficient order is for you to land this plus whichever of the cleanup items you want in one push, and for me to approve after -- an approval now would just be thrown away by your next commit.

mzLib/FlashLFQ/MBR/RetentionTimeCalibDataPoint.cs
```diff
@@ -15,19 +17,38 @@ public RetentionTimeCalibDataPoint(ChromatographicPeak donorFilePeak, Chromatogr
...
+        /// <summary>
+        /// Test seam: builds a data point carrying only an <see cref="RtDiff"/>, with no underlying peaks.
+        /// Lets tests exercise the prediction-error math in <see cref="MbrScorer.AddRtPredErrorDistribution"/>
+        /// without constructing full chromatographic peaks. Points built this way have a null donor peak, so
+        /// they sort before points that have one (see <see cref="CompareTo"/>).
+        /// </summary>
+        internal RetentionTimeCalibDataPoint(double rtDiff)
+        {
+            RtDiff = rtDiff;
+        }
```
trishorts, line 37: Non-blocking, and only worth doing if you touch the file again.

This constructor guarantees `DonorFilePeak == null`, which makes the unguarded `ToString()` below (lines 57-58) reachable in a way it was not before -- it dereferences both peaks. I ran it:

```csharp
var p = new RetentionTimeCalibDataPoint(0.5);
p.ToString();   // NullReferenceException
```

It only bites someone debugging -- a watch window, or an NUnit failure message that renders a collection of these -- so it is not worth a round trip on its own. Null-guarding the two RTs the way `CompareTo` now does would close it.

trishorts (approved):
Approving. Everything I raised is addressed, and I settled the load-bearing item by running it rather than reading the diff for it.

**The floor is now pinned.** I deleted lines 287-288 (`if(stdDevRtError < RtStandardDeviationMin) stdDevRtError = RtStandardDeviationMin;`) from `MbrScorer.cs`, rebuilt, and re-ran `Test.FlashLFQ`:

```
Failed TestAddRtPredErrorDistributionFloorsSigmaForTightlyAlignedAnchors
  Expected: 0.0167 +/- 1e-09
  But was:  0.00022058823529409334
Failed TestAddRtPredErrorDistributionSkipsNonFiniteErrorsAndFloorsStdDev
  Expected: 0.0167 +/- 1e-09
  But was:  0.0
Failed!  - Failed: 2, Passed: 155, Total: 157
```

Restored the two lines and the suite is 157/157 green. Two tests now fail if the floor is removed, from two different directions (a small-but-valid sigma and a fully degenerate one). That was the only thing I was holding approval on.

Also confirmed on this branch:

- `RtStandardDeviationDefault` is used at both sites that implement the default it names (`AddRtPredErrorDistribution` and the `ScoreMbr` fallback), and the unreachable NaN branch is gone.
- `IQR / 1.36`, matching lines 99, 120 and 202.
- CS1572 is gone -- clean `FlashLFQ.csproj` rebuild, 0 errors and no `MbrScorer.cs` warnings.
- `Easy.Common.Extensions` dropped, and `MzLibUtil/ClassExtensions.cs` is out of the changed-file list entirely, so the blast radius is FlashLFQ + Test + Development.
- `anchorPeptideRtDiffs` gone, and `PredictRetentionTime` reads `p.RtDiff`. I checked that substitution is exact rather than merely equivalent-looking: every point in `nearbyCalibrationPoints` has passed `AcceptorFilePeak != null && AcceptorFilePeak.ApexRetentionTime > 0` and has had its donor apex dereferenced, so the stored `RtDiff` is always the same finite number the inline expression produced. The null-apex divergence I flagged cannot be reached from here.
- `RtDiff` is back to `readonly` with an `internal` constructor for the test seam; the public contract is unchanged.
- `mzLib.sln` is byte-identical to master.
- The AUC assertion is `> 0.7` against a measured ~0.82.
- The `Development` harness is `[Explicit]` + `[ExcludeFromCodeCoverage]` with empty path defaults, so it cannot run in CI, and the reverted `.sln` still omits `Development`'s `Release|*.Build.0` rows -- the #1127 hazard stays clear.

One non-blocking nit inline. Branch is up to date with master (`6fc1c0f`) and all five checks are green, so this approval should stick rather than being dismissed by a required update.

mzLib/Test/FlashLFQ/TestPipEcho.cs
trishorts: Resolved. `TestAddRtPredErrorDistributionFloorsSigmaForTightlyAlignedAnchors` is the degenerate case I said was still missing, and the AUC assertion moved from "better than a coin flip" to `> 0.7` against a measured ~0.82. I verified the new test bites by deleting the floor and re-running -- details in my review.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved -- and you took the option I preferred. The unreachable NaN branch is gone, and `RtStandardDeviationDefault` now names the `1` at both sites that actually implement the default: the initial `new Normal(0, RtStandardDeviationDefault)` in `AddRtPredErrorDistribution` and the missing-key fallback in `ScoreMbr`. `TestAddRtPredErrorDistributionWithTooFewPointsUsesDefault` pins the first of those.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved, and this closes the one thing I was holding approval on. I did not take the test on faith -- I deleted lines 287-288, rebuilt, and re-ran `Test.FlashLFQ`:

```
Failed TestAddRtPredErrorDistributionFloorsSigmaForTightlyAlignedAnchors
  Expected: 0.0167 +/- 1e-09
  But was:  0.00022058823529409334
Failed TestAddRtPredErrorDistributionSkipsNonFiniteErrorsAndFloorsStdDev
  Expected: 0.0167 +/- 1e-09
  But was:  0.0
Failed!  - Failed: 2, Passed: 155, Total: 157
```

Restored, 157/157 green. Two tests fail if the floor is removed, from two directions -- a small-but-valid sigma and a fully degenerate one -- so a future refactor now has a reason to preserve it. Approving.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved -- `IQR / 1.36`, so the file carries one constant.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved. Both `<param>` tags are correct and documented, the summary describes the curve rather than a list of doubles, and a clean `FlashLFQ.csproj` rebuild has no CS1572 and no `MbrScorer.cs` warnings at all.

mzLib/FlashLFQ/MBR/MbrScorer.cs
trishorts: Resolved -- the using is gone and FlashLFQ builds with 0 errors.

mzLib/MzLibUtil/ClassExtensions.cs
trishorts: Resolved, and better than I asked for: `MzLibUtil/ClassExtensions.cs` is out of the changed-file list entirely, so the PR's blast radius is FlashLFQ + Test + Development.

mzLib/FlashLFQ/FlashLfqEngine.cs
trishorts: Resolved. The local and its populating block are gone, and `PredictRetentionTime` reads `p.RtDiff` instead of recomputing.

On the filter-shape note I raised alongside it: I checked, and the divergence cannot be reached from here. Every point in `nearbyCalibrationPoints` has already passed `AcceptorFilePeak != null && AcceptorFilePeak.ApexRetentionTime > 0` and has had its donor apex dereferenced in the same loop, so the stored `RtDiff` is always the identical finite number the inline expression produced. The substitution is exact, not merely equivalent-looking.

mzLib/FlashLFQ/MBR/RetentionTimeCalibDataPoint.cs
trishorts: Resolved. The PR description now has a section on the sign convention, and `FlashLfqEngine.cs` reads `p.RtDiff` rather than recomputing it -- one definition of the quantity, which was the point.

mzLib/FlashLFQ/MBR/RetentionTimeCalibDataPoint.cs
trishorts: Resolved. `RtDiff` is `readonly` again and the tests go through the new `internal` constructor, so the public contract is unchanged and nothing compiled against the old mzLib breaks. `MockRtCalibDataPoint` is gone with it.

mzLib/mzLib.sln
trishorts: Resolved -- `mzLib.sln` is byte-identical to master (`git diff smith/master HEAD -- mzLib/mzLib.sln` is empty), so `Development`'s missing `Release|*.Build.0` rows survive untouched and the #1127 hazard stays clear.

nbollis (approved):

PR #1217, opened 2026-08-26
Non-specific digestion for nucleic acids (SingleN/SingleC)

trishorts:
Adds `SingleN`/`SingleC` digestion for nucleic acids, so a non-specific search can run against an RNA database.

**5901 tests pass, 0 fail.** (Skip count varies ±2 between runs because the `ExternalService` tests classify-or-skip against live web services; nothing here touches them.)

## Why

`Rnase.GetUnmodifiedOligos` handled two modes and threw for the rest:

```csharp
return CleavageSpecificity switch
{
    CleavageSpecificity.None => TopDownDigestion(...),   // whole molecule
    CleavageSpecificity.Full => FullDigestion(...),
    _ => throw new ArgumentException(
        "Cleave Specificity not defined for Rna digestion, currently supports Full and None")
};
```

`Semi`, `SingleN` and `SingleC` — the three modes a non-specific search is built on — all threw. And every RNase in `rnases.tsv` was `full` except `top-down`, which is `none`. So there was no way to express "I don't know where this was cut", and MetaMorpheus's non-specific search had nothing to work with.

This came out of a downstream request: modern search over RNA now works ([MetaMorpheus#2759](https://github.com/smith-chem-wisc/MetaMorpheus/pull/2759)), and the obvious next question was non-specific search. That turned out to be blocked here rather than there, which is why this PR exists.

A naming trap worth stating, because it inverts the proteomics intuition: for a nucleic acid, **`CleavageSpecificity.None` means top-down whole molecule**, not "non-specific".

## What this adds

`SingleN` anchors the 5' end — one oligo per start position, running toward 3'. `SingleC` mirrors it. Two RNase entries, `singleN` and `singleC`, select them.

```
GUACUGA, minLength 3

  singleN →  GUACUGA, UACUGA, ACUGA, CUGA, UGA
  singleC →  GUA, GUAC, GUACU, GUACUG, GUACUGA
```

**The methods are named `SingleFivePrimeDigestion` / `SingleThreePrimeDigestion`.** The enum values have to stay `SingleN`/`SingleC` — they're shared with `Proteomics` — but N and C termini don't exist on a nucleic acid. The names say 5'/3' and the doc comments map back to the enum, so nobody has to hold the translation in their head. The RNase entries keep the `singleN`/`singleC` spelling to match the enum and, incidentally, to keep MetaMorpheus's `DigestionAgent.Name.Contains("single")` check working.

**Termini follow the existing rules.** `GetDigestedTermini` already encodes them: an oligo reaching an original terminus keeps it, otherwise it gets the digested default (5'-OH, 3'-phosphate). Both new paths go through it. This is the part that would otherwise shift every mass with no test failing on sequence alone, so it has a dedicated test rather than being covered incidentally.

## Two deliberate limits

Both are documented at the methods, not just here.

**Only the longest oligo per anchor position is emitted.** This matches `Protease.SingleN_Digestion`, and the reasoning is the same: a non-specific search scores every prefix as it walks the fragment index, so emitting the shorter oligos is redundant work rather than extra coverage.

This is the assumption most worth a second opinion — it is a faithful port of the proteomics semantics, made without knowing whether nucleic acid chemistry wants something different. **If RNA non-specific search needs every subsequence materialised, that loop is the line to change**, and the method says so.

**There is no specific-agent variant.** `Protease.SingleN_Digestion` has a second branch that honours a named protease's missed-cleavage rules, driven by `DigestionParams.SpecificProtease`. `RnaDigestionParams` has no `SpecificRnase`, so there is nothing to honour — a *semi*-specific nucleic acid search is not expressible yet. Faking it would have meant inventing a parameter, so the branch is simply absent and documented as absent.

## `Semi` is still unsupported, and still advertised

`rnases.tsv`'s header documents the valid specificities as `full, semi, none, SingleN, SingleC`, and `RnaseDictionary` parses all five. Before this PR, three of them loaded fine and then threw at digestion time — a custom RNase declared `SingleN` was accepted by the loader and crashed during search.

This PR closes that for `SingleN` and `SingleC`. **`Semi` remains a mode the file format promises and the code refuses.** Rather than leave it undocumented, `TestRnase_UnmodifiedOligos_Exception` now pins it — that test previously used `SingleC` as its "unsupported" example, which this PR makes supported, so it failed by succeeding. Repointing it at `Semi` preserves its actual intent: an unsupported specificity is refused rather than doing something silently wrong.

Either the loader should reject `semi`, or digestion should implement it. Out of scope here, but it should not stay invisible.

## Tests

New `Test/Transcriptomics/TestRnaseNonSpecificDigestion.cs` (8 tests):

- `SingleN_ProducesOneOligoPerStartPosition` / `SingleC_ProducesOneOligoPerEndPosition` — assert the **exact sequences**, not just counts, so the semantics are pinned rather than approximated.
- `SingleN_RespectsMaxLength` / `SingleC_RespectsMaxLength` — including the `int.MaxValue` overflow guard both paths need (`start + maxLength` goes negative otherwise; `Protease` has the same guard).
- `SingleN_TerminiFollowTheDigestionRules` — the 5'-OH / 3'-phosphate rules described above.
- `SingleN_CoversWhatFullDigestionFinds` — every start position a full RNase T1 digestion finds is also reachable non-specifically. This is the property that makes a non-specific search meaningful.
- `SingleN_AndSingleC_HonourMinLength`.

Changed: `TestRnase_UnmodifiedOligos_Exception`, as described above.

## Downstream

This unblocks, but does not itself deliver, non-specific RNA search. The remaining work is in MetaMorpheus, after this releases and the pin bumps:

- `NonSpecificEnzymeSearchEngine` is typed to `PeptideWithSetModifications` throughout.
- `GetVariableTerminalMods` finds terminal mods by `LocationRestriction.Contains("C-terminal")` / `"N-terminal"`. **RNA mods use `5'-terminal` / `3'-terminal`** — mzLib's own `GetModificationDictionaryFromFullSequence` checks for exactly those. So terminal-mod discovery would silently find nothing for RNA rather than failing loudly.
- MetaMorpheus#2759 currently refuses non-specific search on nucleic acids with an explicit message; that guard comes out at the end of this chain.


nbollis (approved):

acesnik:
🤖 Checked the parts that would fail silently rather than loudly, since that is where this kind of change hurts:

- **Terminus tuple order is right.** `GetDigestedTermini` is declared `(IHasChemicalFormula ThreePrime, IHasChemicalFormula FivePrime)` (`Rnase.cs:155`), both new paths destructure it in that order, and that matches all four existing call sites (`:46, :69, :81, :149`). A swap here would have shifted every mass with the sequence assertions still passing, so it was worth confirming rather than assuming.
- **Constructor argument order matches `FullDigestion`** — `fivePrimeTerminus, threePrimeTerminus` in the same positions as `:70-71`.
- **The overflow guard is a faithful port.** `Protease.SingleN_Digestion` has the identical shape: `maxTooBig` at `Protease.cs:493`, `ValidMinLength(Length - start + 1, min)` at `:497`, `Math.Min(Length, start + max - 1)` at `:502`. Your guard is also conservative in the right direction — it tests `Length + maxLength` where the largest expression is `Length + maxLength - 1`, so it fires slightly early rather than slightly late.

**On the assumption you asked for a second opinion on** — only the longest oligo per anchor position:

The proteomics justification is that a non-specific search scores every prefix as it walks the fragment index, so the shorter oligos are redundant. That is a property of `NonSpecificEnzymeSearchEngine`, not of the digestion — and per your own Downstream section that engine is proteomics-typed (15 `PeptideWithSetModifications` references in `NonSpecificEnzymeSearchEngine.cs`) and unported. So the assumption is not verifiable on the RNA path yet: it is correct **iff** the port preserves prefix scoring.

That makes this the wrong place to second-guess it, I think, and the right place to pin it the MetaMorpheus port — a test there asserting that a subsequence of an emitted oligo is still identified would catch a port that drops prefix scoring, which is the only way this choice becomes wrong. Worth carrying that sentence into #2759's follow-up rather than leaving it as an open question here.

Confirmed both downstream hazards you flagged, with lines for whoever picks them up:

- `NonSpecificEnzymeSearchEngine.cs:595` and `:613` both do `terminalStringToFind = ... ? "C-terminal" : "N-terminal"`, and mzLib's RNA restrictions are `"5'-terminal."` / `"3'-terminal."` (`DigestionProduct.cs:106, :145`). So terminal-mod discovery finds nothing for RNA and reports no error — the failure mode you describe, and the more dangerous of the two.

One inherited quirk, noted for completeness rather than as something to change: the min-length check tests the longest oligo available from the anchor, and the length is then clamped by `maxLength`, so `maxLength < minLength` emits oligos below `minLength`. `Protease` behaves identically, so this is consistent rather than new, and the configuration is nonsensical anyway.

Repointing `TestRnase_UnmodifiedOligos_Exception` at `Semi` was the right call — it keeps the test's intent instead of deleting a test that failed by succeeding, and it documents the loader/digestion mismatch instead of hiding it.

acesnik (approved):

PR #1219, opened 2026-08-26
Fix Extremum's equality contract: recursion, GetHashCode, and a null check

acesnik:
🤖 `Extremum.Equals(object)` recursed into itself until the stack ran out, and `Extremum` had no `GetHashCode`.

```csharp
public override bool Equals(Object obj) => Equals(this, (Extremum)obj);
```

Two arguments, so that binds to the static `object.Equals(object, object)`, whose last act is a virtual `objA.Equals(objB)` — back into this same override. Comparing two distinct, non-null `Extremum`s through any non-generic path recursed until the stack ran out.

Measured before the fix, in a console app so the crash was visible:

```
IEquatable<Extremum>.Equals(a, b)     : True
about to call ((object)a).Equals(b) ...
Stack overflow.
Repeated 174209 times:
   at FlashLFQ.IsoTracker.Extremum.Equals(System.Object)
```

A `StackOverflowException` cannot be caught and takes the process with it.

## Why it has gone unnoticed

Every path in use is generic. `Dictionary` and LINQ resolve `EqualityComparer<Extremum>.Default`, which picks `IEquatable<Extremum>` and reaches the typed overload directly. The recursion needs `object.Equals(a, b)`, a non-generic collection, or — the one that matters here — `ClassicAssert.AreEqual`, which this suite uses widely. A test that compared two `Extremum`s would have killed the test host rather than failing.

## The fix

`obj is Extremum other && Equals(other)`. Two incidental corrections come with that shape: a non-`Extremum` argument returns false instead of throwing `InvalidCastException` from the cast, and null returns false without relying on `object.Equals` to short-circuit first.

The typed `Equals(Extremum)` also dereferenced its argument before testing it, so `apex.Equals((Extremum)null)` threw. That is the overload every generic caller reaches, which makes it the more exposed of the two.

## `GetHashCode`

Overriding `Equals` without `GetHashCode` left the type on `object`'s reference hash, so equal Extrema bucketed apart. `XICGroups.SortExtrema` keys a `Dictionary<Extremum, List<Extremum>>` on it, which made its own `ContainsKey` dedup guard unreachable.

An earlier revision of this description claimed no `GetHashCode` could be correct, because equality is tolerant. That was wrong, and it is worth being explicit about since it was the stated reason for leaving the bug in place. Equality has three clauses and only the middle one is tolerant:

```csharp
this.Intensity == others.Intensity && Math.Abs(this - others) < 0.006 && this.Type == others.Type
```

`RetentionTime` is the only field the tolerance touches. `Intensity` and `Type` are exact, so equal Extrema always agree on them exactly, and `HashCode.Combine(Intensity, Type)` satisfies the contract. The price is collisions between same-intensity, same-type Extrema at unrelated retention times — permitted, and separated by the typed `Equals`.

## Does this move quantification output?

No, and measured rather than assumed, because it touches FlashLFQ results.

- Instrumented the dedup guard in `SortExtrema` and ran every IsoTracker PSM fixture through the engine with `isoTracker: true` — nine datasets unfiltered, plus the motif-filtered variant for `SearchTarget`. **125 dictionary key insertions, 0 collisions** under the new hash.
- `peaks.tsv`, `peptides.tsv`, `proteins.tsv` and the isobaric-peptide sets across those runs: **30 files, byte-identical** against the pre-change build. The dump was checked reproducible run-to-run first, so the zero is meaningful rather than an artefact of a stable-but-unexercised path.
- The 28 existing IsoTracker tests pass unchanged.

## Tests

Two regression tests in `TestIsoTracker`, both mutation-checked against the unfixed code:

- `ExtremumEqualsObjectDoesNotRecurse` — the four `Equals(object)` cases, plus `Equals((Extremum)null)`. Removing the null guard throws instead of returning false.
- `ExtremumEqualExtremaShareAHashCodeAndADictionarySlot` — equal and within-tolerance Extrema hash alike and hit the same `Dictionary` slot; unequal ones do not. Removing `GetHashCode` fails three of its assertions, including the `ContainsKey`.

## Still not fixed

The tolerance makes equality non-transitive: A≈B and B≈C without A≈C. That violates the `Equals` contract independently of hashing, so which Extrema a `Dictionary` collapses into a single key depends on insertion order. It is a property of the equality rather than of the hash, and the honest fix is for `Extremum` to stop being a dictionary key — an explicit grouping pass over the already-sorted list, which is a `XICGroups` rework rather than a change to this type.

Adjacent, same cause: `CompareTo` orders on `RetentionTime` alone while `Equals` also requires `Intensity` and `Type`, so `CompareTo(x) == 0` does not imply `Equals(x)`. `SortExtrema` sorts with the first and keys on the second.

Both are filed as #1225 rather than guessed at here.


mzLib/FlashLFQ/IsoTracker/Extremum.cs
```diff
@@ -67,9 +67,16 @@ public bool Equals(Extremum others)
...
         }
 
+        /// <summary>
+        /// Defers to the typed overload. The previous body was Equals(this, (Extremum)obj), which binds
+        /// to the static object.Equals(object, object); that ends in a virtual objA.Equals(objB) and
+        /// arrives straight back here, so any non-generic comparison of two distinct Extrema recursed
+        /// until the stack ran out. Generic callers never saw it, because Dictionary and LINQ resolve
+        /// IEquatable<Extremum> and reach the typed overload directly.
+        /// </summary>
         public override bool Equals(Object obj)
```
trishorts, line 83: **`Extremum` overrides `Equals(object)` and has no `GetHashCode()`. That is now the live half of the bug.**

There is no `GetHashCode` anywhere in the class — and the comment added just above says generic callers were safe because "Dictionary and LINQ resolve `IEquatable<Extremum>` and reach the typed overload directly." The first half is true, the conclusion is not: `Dictionary` resolves the *typed* `Equals` through `EqualityComparer<Extremum>.Default`, but it still buckets by `GetHashCode()`, which is `object`'s reference hash. Two Extrema that compare equal land in different buckets and never meet.

That is not hypothetical here — `IsoTracker/XICGroups.cs:109`:

```csharp
Dictionary<Extremum, List<Extremum>> group = new Dictionary<Extremum, List<Extremum>>();
```

So lookups in that dictionary miss for equal-but-distinct Extrema today. It fails quietly, as a group that should have merged and didn't, which is worse than the stack overflow this PR fixes — that one at least announced itself.

**The awkward part, and the reason I would not just bolt on a `GetHashCode`:** the typed equality is tolerance-based —

```csharp
Math.Abs(this - others) < 0.006
```

— and no `GetHashCode` can be correct against a tolerance. Two values 0.005 apart must hash identically, and hashing is not transitive-friendly: A≈B and B≈C does not give A≈C, so there is no consistent bucketing. Rounding to a 0.006 grid gets it wrong at every bucket boundary.

Which suggests the real answer is that `Extremum` should not be a dictionary key at all, and `XICGroups` wants an explicit grouping pass over a sorted list rather than hash lookup. That is more than this PR signed up for — but I would rather see it recorded as an issue than have the new comment assert that generic callers are fine, since it is the sentence someone will rely on next time.

Smaller, same area: the typed `Equals(Extremum others)` dereferences `others.Intensity` with no null check, so `apex.Equals((Extremum)null)` throws. `Equals(object)` is safe now because `is` excludes null, and the new test pins that — but the typed path is the one every generic caller takes.

trishorts (commented):
The diagnosis is right and the fix is the correct one — `obj is Extremum other && Equals(other)` also removes the `InvalidCastException` the old cast threw on a non-Extremum, which the new test covers.

One thing the PR walks up to and does not address, and I think it has to, because it undercuts the claim in the new comment. Detail inline.

mzLib/FlashLFQ/IsoTracker/Extremum.cs
acesnik: 🤖 Fixed in d4f32575 — `GetHashCode`, and the null check on the typed overload.

On the part you flagged as the reason not to just bolt one on: that argument was mine, from this PR's own body, and it is wrong. I wrote "it cannot be fixed by adding a `GetHashCode`, because equality is `Math.Abs(rtDiff) < 0.006` and so is not transitive," you took it at face value, and neither of us checked which field the tolerance is actually on.

Equality is three clauses:

```csharp
this.Intensity == others.Intensity && Math.Abs(this - others) < 0.006 && this.Type == others.Type
```

Only the middle one is tolerant, and it is the only one that reads `RetentionTime`. `Intensity` and `Type` are exact. So equal Extrema always agree exactly on those two, and hashing them alone satisfies the contract:

```csharp
public override int GetHashCode() => HashCode.Combine(Intensity, Type);
```

The cost is collisions between same-intensity, same-type Extrema at unrelated retention times, which is permitted and which the typed `Equals` then separates. Your general point — that no hash can bucket a tolerance — is right; it just does not bind here, because the tolerant field can be left out.

**What does not go away** is the non-transitivity itself. A≈B and B≈C without A≈C violates the `Equals` contract regardless of hashing, so which Extrema a `Dictionary` collapses into one key is insertion-order dependent. That is a property of the equality, not of the hash, and it is the real argument for your conclusion that `Extremum` should not be a dictionary key at all. I have written it into the doc comment so the next reader hits it, and I would open the issue you suggested for the `XICGroups` rework rather than attempt it here.

**One correction to the failure mode, though**, because it changes how urgent this is. `SortExtrema` never looks up by a *different* key — the dictionary is only ever `ContainsKey(currentExtremum)`, `Add(currentExtremum, …)`, then enumerated. So the sole hash-sensitive operation is that self-dedup guard, and "a group that should have merged and didn't" is not something this dictionary can express.

Measured rather than argued:

- Instrumented that guard and ran every IsoTracker PSM fixture through the engine with `isoTracker: true` — nine datasets unfiltered, plus the motif-filtered variant for `SearchTarget`. **125 key insertions, 0 collisions** under the new hash.
- `peaks.tsv`, `peptides.tsv`, `proteins.tsv` and the isobaric-peptide sets across those runs: **30 files, byte-identical** before and after. Confirmed the dump is reproducible run-to-run first, so the zero means something.
- The 28 existing IsoTracker tests pass unchanged.

So the guard was unreachable before and still does not fire; the fix is latent-correctness, not a behaviour change. Worth saying plainly since it touches quantification output.

Both new assertions are mutation-checked — dropping `GetHashCode` fails three of them including the `ContainsKey`, and dropping the null guard throws rather than returning false.

Separately, and not fixed here: `CompareTo` orders on `RetentionTime` alone while `Equals` also requires `Intensity` and `Type`, so `CompareTo(x) == 0` does not imply `Equals(x)`. `SortExtrema` sorts with the former and then keys on the latter. Probably belongs in the same issue as the rework.

trishorts: Verified `d4f32575`, and you're right on both counts.

`GetHashCode` is `HashCode.Combine(Intensity, Type)`, and those are exactly the two clauses of the typed `Equals` that are exact (`Extremum.cs:59-70`) — equal Extrema agree on both, so equal implies same hash. My "no hash can bucket a tolerance" was a statement about the equality in general that I applied without checking which field the tolerance was on. It's on `RetentionTime` only, excluding it is sound, and the collisions that buys are permitted and separated by the typed `Equals`. `Equals(object)` is now `obj is Extremum other && Equals(other)`, which also kills the `InvalidCastException` on a non-Extremum that the old cast would have thrown.

Your correction to the failure mode is the part that matters: `SortExtrema` only ever does `ContainsKey(currentExtremum)` / `Add(currentExtremum, …)` on the same instance, so the guard was unreachable rather than mis-merging, and 125 insertions / 0 collisions plus 30 byte-identical output files says so rather than argues it. Latent correctness, not a behaviour change — good that you said it plainly on a quantification path.

Resolving. #1225 covers what's left (non-transitive equality making dictionary-key collapse insertion-order dependent, and `CompareTo` ordering on `RetentionTime` alone while `Equals` needs three fields), so the doc comment isn't the only record.

trishorts (approved):
Approving at `d1fb701a`.

The equality contract is whole now, and each piece is verifiable rather than plausible:

- `Equals(object)` is `obj is Extremum other && Equals(other)`, which ends the recursion and also disposes of the `InvalidCastException` the old `(Extremum)obj` cast would have thrown on any non-Extremum. The test pins both, including the `"not an extremum"` case.
- `GetHashCode() => HashCode.Combine(Intensity, Type)` satisfies the contract because those are precisely the two clauses of the typed `Equals` that are exact. `RetentionTime` is the only tolerant one and it is excluded, so equal Extrema always hash equally. The collisions that buys — same intensity and type at unrelated retention times — are permitted, and the typed `Equals` separates them. `ExtremumEqualExtremaShareAHashCodeAndADictionarySlot` asserts both directions of that, which is the right shape for a hash test.
- The null guard is on the typed overload, which is the one every generic caller reaches, so `apex.Equals((Extremum)null)` returns false rather than throwing.

I was wrong in my first comment to argue against bolting on a `GetHashCode`, and your correction is the reason this is a two-line fix rather than an `XICGroups` rework. Worth restating what makes it safe to merge on a quantification path: the dictionary in `SortExtrema` only ever keys on the instance it just created, so the `ContainsKey` guard was unreachable rather than mis-merging, and 125 key insertions with 0 collisions plus 30 byte-identical output files is measurement rather than argument. Latent correctness, no behaviour change.

What is deliberately not fixed here is recorded in #1225 — non-transitive equality making which Extrema a `Dictionary` collapses insertion-order dependent, and `CompareTo` ordering on `RetentionTime` alone while `Equals` requires all three fields. Both are properties of the equality rather than of the hash, and both belong to that rework.

mzLib/FlashLFQ/IsoTracker/Extremum.cs
```diff
@@ -67,9 +72,33 @@ public bool Equals(Extremum others)
...
+        /// <summary>
+        /// Hashes the two fields equality compares exactly. RetentionTime is excluded deliberately: it is
+        /// compared within 0.006, and no hash can put two values that near each other in one bucket without
+        /// putting everything there. Excluding it keeps the contract -- equal Extrema hash equally -- at the
+        /// cost of collisions between same-intensity, same-type Extrema at unrelated retention times, which
+        /// the typed Equals then separates.
+        ///
+        /// The tolerance still makes equality non-transitive (A and B within 0.006, B and C within 0.006, A
+        /// and C not), so which Extrema a Dictionary treats as one key depends on insertion order. That is a
+        /// property of the equality, not of the hash, and it is why Extremum is a poor dictionary key at all.
```
trishorts, line 97: One thing worth adding to this paragraph, because "unreachable dedup guard" undersells what happens when it becomes reachable.

`XICGroups.cs:132` is:

```csharp
if (group.ContainsKey(currentExtremum))
{
    continue;
}

group.Add(currentExtremum, currentGroup);
```

`continue` skips the `Add`, so `currentGroup` is discarded whole -- it is not merged into the existing entry. And the return filters on `p.Value.Count() >= count_threshold * XICs.Count()`, so the surviving group is the only one the threshold ever sees. A hit therefore drops a set of shared extrema rather than deduplicating anything.

Which makes your "125 insertions, 0 collisions" load-bearing rather than reassuring background: the change is inert *because* the guard never fires, not because firing would be harmless. Worth one sentence saying so, here or on #1225, so nobody later reads the guard as a merge.

mzLib/FlashLFQ/IsoTracker/Extremum.cs
```diff
@@ -67,9 +72,33 @@ public bool Equals(Extremum others)
...
+        /// </summary>
+        public override int GetHashCode()
```
trishorts, line 99: A second thing worth recording for whoever picks up #1225, since you have the context now and it will be expensive to rediscover.

Within a single `SortExtrema` call the guard is decided almost entirely by `Intensity`, not by the tolerance. `FindSharedExtrema` builds `shared_min` and `shared_max` by filtering on `p.Type == ExtremumType.Minimum` / `Maximum` (`XICGroups.cs:82-83`), so every element in one call already agrees on `Type` and that clause is free. The elements come from `xic.Extrema` across different XICs, so they are independently interpolated -- and `Intensity` is compared with exact `==`. That is what holds the collision count at zero, more than the hash does.

It is not a guarantee, which is the useful half: any case where two XICs yield the same double would fire it -- a flat region, or zero-intensity extrema passing a zero `cutOff`. I have not tried to construct one, so treat that as the hypothesis it is rather than a finding.

The reframing it gives #1225 is worth having: the non-transitive equality is genuinely unsound, but it is currently unreachable for the same reason the guard is, so the rework there is about correctness rather than about a bug anyone is hitting.

trishorts (commented):
Read the diff directly (two files, 99 added lines). The core of it is right and I checked each load-bearing claim against the source rather than the description.

**Verified:**

- **The recursion is real and the diagnosis is exact.** Master's `Equals(object)` is `Equals(this, (Extremum)obj)`. There is no two-argument instance `Equals` on the type, so that binds to the static `object.Equals(object, object)`, whose last act is a virtual `objA.Equals(objB)` -- straight back into the override. Two distinct non-null Extrema recurse until the stack goes.
- **`GetHashCode` is contract-correct.** Equality's three clauses are `Intensity ==` (exact), `Math.Abs(this - others) < 0.006`, and `Type ==` (exact). `operator -` returns `RetentionTime - RetentionTime`, so the tolerance touches retention time and nothing else, and `HashCode.Combine(Intensity, Type)` covers exactly the two exact clauses. All three fields are `readonly`, so there is no mutate-after-insert hazard either. The earlier claim that no correct hash existed was indeed wrong, and the retraction in the body is the right way to handle it.
- **Both incidental fixes are real.** Master's typed `Equals` dereferences `others.Intensity` with no null check, and the old cast `(Extremum)obj` threw `InvalidCastException` for a non-`Extremum`.
- **The tests are sound, and their constructor argument order matches** `Extremum(double intensity, double retentionTime, ExtremumType type)`, so `new Extremum(100.0, 10.005, Maximum)` really is the within-tolerance case (0.005 < 0.006) and `new Extremum(101.0, 10.0, ...)` really is the differing-intensity one.
- **The fully-qualified `NUnit.Framework.Assert` in the new tests is necessary, not verbosity.** This file carries `using Assert = NUnit.Framework.Legacy.ClassicAssert;` and uses zero modern-API assertions anywhere else. Worth leaving a note if anyone is tempted to "tidy" it later.
- #1225 exists and says what the body says it says.

**Two notes on the `SortExtrema` guard this makes reachable**, inline below. Neither blocks -- your measurement stands and the outputs are byte-identical -- but the shape of the risk is worth writing down while it is fresh, because it is the part a future reader will get wrong.

Nothing else. All five checks pass; already approved, so this only wants the second.

nbollis (approved):

PR #1220, opened 2026-08-26
Declare that PeptideWithSetModifications.Parent shadows the base

acesnik:
🤖 Clears the CS0108 on `PeptideWithSetModifications.Parent`.

```csharp
// DigestionProduct.cs               [field: NonSerialized] public IBioPolymer Parent { get; protected set; }
// PeptideWithSetModifications.cs    public IBioPolymer Parent => Protein;
```

The derived property hides the base one without saying so. `override` is not available — the base is a plain auto-property, not virtual — so this adds `new`, which **changes nothing at all**. It states the shadowing that already happens, plus a comment recording where the two can disagree.

## They can disagree, and it looks deliberate

The base backing field is `[NonSerialized]`, so on a peptide read back from a serialized index it is null until `SetNonSerializedPeptideInfo` repopulates the protein, while `Protein` itself is restored. Members inherited from the base — `PreviousResidue`, `NextResidue`, `BaseSequence` — read the base field, so they see null where a caller reading `peptide.Parent` sees the protein.

That is a plausible reason for the property to exist, which is why this PR documents the split rather than removing it.

## What would be better, and why it is not here

Unifying the two — set the base field and drop this property, so every caller sees one value — is the better end state. It is not done here because it changes what base-typed callers observe and runs straight into the serialization behaviour above. That wants whoever owns `SetNonSerializedPeptideInfo` to confirm the intent first.

Solution builds clean and the warning is gone.

### DIFF
--- a/mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
+++ b/mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
@@ -220,7 +220,14 @@ namespace Proteomics.ProteolyticDigestion
         private string? _fullSequenceWithMassShifts;
         public string FullSequenceWithMassShifts => _fullSequenceWithMassShifts ??= this.FullSequenceWithMassShift();
 
-        public IBioPolymer Parent => Protein;
+        /// <summary>
+        /// Shadows <see cref="DigestionProduct.Parent"/> rather than overriding it, which is not
+        /// possible because the base property is not virtual. The two can disagree: the base backing
+        /// field is [NonSerialized], so it is null on a deserialized peptide until
+        /// SetNonSerializedPeptideInfo runs, while Protein is repopulated. Members inherited from the
+        /// base -- PreviousResidue, NextResidue, BaseSequence -- read the base field, not this.
+        /// </summary>
+        public new IBioPolymer Parent => Protein;
 
 
         /// <summary>


mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
```diff
@@ -220,7 +220,14 @@ public string SequenceWithChemicalFormulas
...
 
-        public IBioPolymer Parent => Protein;
+        /// <summary>
+        /// Shadows <see cref="DigestionProduct.Parent"/> rather than overriding it, which is not
+        /// possible because the base property is not virtual. The two can disagree: the base backing
+        /// field is [NonSerialized], so it is null on a deserialized peptide until
+        /// SetNonSerializedPeptideInfo runs, while Protein is repopulated. Members inherited from the
+        /// base -- PreviousResidue, NextResidue, BaseSequence -- read the base field, not this.
+        /// </summary>
+        public new IBioPolymer Parent => Protein;
```
trishorts, line 230: **The comment describes a divergence that can throw, and I would rather see it closed than described.**

The base members do not just read a different field — one of them is unguarded. `DigestionProduct` guards two of the three:

```csharp
public virtual char PreviousResidue => Parent is null ? '-' : ...   // :31
public virtual char NextResidue     => Parent is null ? '-' : ...   // :33
_baseSequence ??= Parent.BaseSequence.Substring(...)                 // :36  <- no guard
```

So on a deserialized peptide before `SetNonSerializedPeptideInfo` runs, `pwsm.Parent` returns a perfectly good `Protein` while `pwsm.BaseSequence` dereferences the null base field and throws. The shadowing is what makes that pair of behaviours possible to observe at once, and the window is real — that is the whole reason the comment exists.

The comment says overriding "is not possible because the base property is not virtual", which is true as written, but the base is in this repo:

```csharp
[field: NonSerialized] public IBioPolymer Parent { get; protected set; }   // DigestionProduct.cs:25
```

Making it `virtual` is in scope for mzLib. It is admittedly not a one-word change — the setter is `protected`, so an override has to carry both accessors rather than being a `=> Protein` expression body — but the result is that the base members and the derived property agree, which is what every reader already assumes and what this comment exists to warn them is false.

If that is too much for a documentation PR, entirely fair. In that case I would still rather the comment named the consequence rather than only the mechanism: something to the effect of *"until `SetNonSerializedPeptideInfo` runs, `Parent` is non-null while `BaseSequence` throws"*. As written, a reader has to go and notice the missing guard at `DigestionProduct.cs:36` themselves to work out why any of this matters.

trishorts (commented):
Good comment — the `[NonSerialized]` interaction is exactly the kind of thing that is invisible until it bites, and writing it down beats leaving the `new` bare. One question about whether documenting is the right end state here.

mzLib/Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs
trishorts: Refining my own comment above, because I checked the failure path and it is narrower than I made it sound.

`_baseSequence` is `protected string _baseSequence;` -- **not** `[NonSerialized]`. So it survives deserialization and `_baseSequence ??= Parent.BaseSequence...` short-circuits. For `BaseSequence` to throw, the peptide has to have been serialized *before anything ever read it* -- possible, since the digestion constructor passes `baseSequence: null` through to the base, but not the general case I implied.

And in the one real consumer it closes entirely. `MetaMorpheusTask.cs:1298` loads the index and repopulates in the same method:

```csharp
foreach (var peptide in peptideIndex)
    peptide.SetNonSerializedPeptideInfo(...);
return peptideIndex;
```

Nothing escapes in the divergent state, so no caller observes `Parent` non-null alongside a null base field.

That changes my recommendation rather than withdrawing it. Making `DigestionProduct.Parent` virtual is harder to justify for a state nobody can currently observe, so take the documentation option -- but have the comment say what the divergence *is* and that it is presently unobservable, rather than implying a live crash. Something like: *"until `SetNonSerializedPeptideInfo` runs, `Parent` is non-null while the base field is not; no current caller observes this, because MetaMorpheus repopulates in the same loop that deserialises."*

The unguarded `Parent.BaseSequence` at `DigestionProduct.cs:36` is still the odd one out next to the two guarded members, and still worth a guard on its own merits -- just not as this PR's problem.

trishorts (approved):
Checked every claim the comment makes against `Omics/Digestion/DigestionProduct.cs` on `smith/master`, since on a documentation PR the comment *is* the deliverable:

- `[field: NonSerialized] public IBioPolymer Parent { get; protected set; }` (:25) — not virtual, so `new` is the only option available, and stating it beats leaving a silent CS0108.
- `PreviousResidue` (:31), `NextResidue` (:33) and `BaseSequence` (:35-37) all read the base property, not the shadow. Correct as written.
- The `[field: NonSerialized]` / repopulated-`Protein` divergence is real.

My standing thread is a wording preference, not a defect — and I narrowed it myself last round after checking that `_baseSequence` is not `[NonSerialized]`, so nothing observable is at stake. I would still rather the comment named the consequence and said it is presently unobservable, but that is a suggestion and I am not holding the PR for it. Take it or leave it.

The unguarded `Parent.BaseSequence` at `DigestionProduct.cs:36`, sitting next to two guarded siblings, remains the odd one out. Not this PR's problem and not currently reachable, so I am not filing it — noting it so it is written down somewhere.

Approving.

PR #1222, opened 2026-08-27
Fix the Combinatorics spelling, and record why AreValid is coordinate-only

acesnik:
🤖 A spelling fix, a documentation comment, and a test. **No behaviour change: `AreValid`'s body is now byte-identical to master's.**

That is not what this PR started as, so the short history is worth having. It originally strengthened `AreValid` with two rules beyond the coordinate check — a real-change requirement and a modification-position check. @trishorts reviewed it, established that both were wrong, and both are gone. What is left is the part that turned out to be worth keeping: writing down *why* the predicate is as narrow as it is, so the next person does not make the same change.

## The spelling fix

`maxAllowedVariantsForCombinitorics` appeared 10 times in `VariantApplication.cs`, including the public parameter of `ApplyVariants`. Line 47 read `maxAllowedVariantsForCombinitorics: maxAllowedVariantsForCombinatorics` — both spellings on one line. Renamed throughout. trishorts checked that MetaMorpheus references neither spelling, so nothing downstream breaks.

## Why AreValid is coordinate-only, now recorded on the method

Two things it deliberately does not judge, each of which I tried to make it judge and each of which was a mistake.

**It does not require a real change.** `GetVariantBioPolymers` gates on `All(v => v.AreValid())`, not a per-variation filter, so calling one applicable variation invalid diverts *every* variation on that biopolymer to the genotype-aware path — which returns the base protein alone for XML-loaded variations, since they carry no genotypes. Separately, reversing a start loss produces `original == variant` by construction for every initiator-methionine loss, so a real-change rule drops those decoys while their targets remain: one-decoy-per-target broken for a whole variant class, which biases FDR.

**It does not judge modification positions.** `OneBasedModifications` is keyed in post-edit *variant-protein* coordinates, not parent coordinates. `AdjustModificationIndices` shifts the parent dictionary into the new frame and merges the variant dictionary with its keys untouched, which only composes if those keys are already post-edit — and `TestProteinProperties` has a committed example spanning 4..4 while carrying a modification at position 7, commented as the variant `MPEPPPT[mt]IDE`. A key outside the span is therefore normal. `IBioPolymer.SelectValidOneBaseMods` already removes modifications that do not survive, against the real post-edit sequence and motif, where a predicate on `SequenceVariation` can only guess from lengths.

A related trap in the same area: an empty `VariantSequence` is a **deletion**, not a termination. mzLib tests for the asterisk in four places, and residues after a deletion survive.

## Test

`AreValidJudgesCoordinatesOnly` replaces the previous `AreValid` test and pins the two non-judgements rather than deleting the coverage: a no-op, the reversed-start-loss shape, a modification keyed outside its span, a deletion carrying a following modification, and the coordinate rules that do still apply.

It fails against the version being removed — verified by restoring that version and rerunning, not by inspection. 59 tests across `TestVariantProtein` and `TestDecoyVariantParity` pass in a throwaway net10.0 harness, since `Test.csproj` is `net*.0-windows` via `mzPlot` and cannot execute on macOS.

## On the CI failure in the earlier review

`ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy` no longer exists. This branch was 34 commits behind and master has replaced that test with `TestDecoyVariantParity.cs`; master is merged in now, so CI runs against something current.

## Not included

An `IsEffective()` filter inside `ApplyAllVariantCombinations`, next to the existing `GroupBy(v => v.SimpleString())` dedup. trishorts is right that the duplicate sequences in combinatorial application are a real problem and that this is the place to fix it, but it changes which variant proteins come out, so it wants its own measurement and its own PR rather than riding along in one that no longer changes behaviour at all.


acesnik:
🤖 @trishorts — CI is red here, and having measured it I now think most of this should not land. It is your logic from #969 and I would rather you decide than quietly bury it.

## What is failing

`ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy`, `Expected: 1, But was: 0`. The assert is in the shared `ReverseOne` helper (`TestDecoyProteinGenerator.cs:33`), so it means "the reversal produced nothing".

`MA (1..2) -> A` on `APEPTIDE` reverses to original `"A"`, variant `"A"` — identical. The strengthened `AreValid` calls that a no-op, and `DecoyProteinGenerator.cs:354` filters it out.

## Both new rules share one failure mode

`AreValid` is not only a filter. `VariantApplication.cs:40` uses `All(v => v.AreValid())` to decide whether a protein takes the combinatorial branch. One "invalid" variation changes what happens to the *whole protein*, not to that variation.

**Rule 2, the no-op rule.** I assumed no-ops were a rare degenerate case. The CI failure says otherwise — reversing any start loss produces one, so they arise on the ordinary decoy path. Measured over the four variant fixtures:

| | `VariantModsGPTMD` | other three | `TestDecoyProteinGenerator` |
|---|---|---|---|
| master, position only | 191 | 8 / 776 / 6 | 11/11 |
| + modification-position rule | 191 | 8 / 776 / 6 | 11/11 |
| + no-op rule | **183** | 8 / 776 / 6 | **1 fail** |

**Rule 3, the modification-position rule.** I was going to keep this one. Then I measured what happens when it fires. `PEPTIDEKPEPTIDEK` with a variation `5..7 "IDE" -> "A"` carrying a variant modification at 7:

```
master     GetVariantBioPolymers -> 2   PEPTIDEKPEPTIDEK (0 applied)
                                        PEPTAKPEPTIDEK   (1 applied, modKeys=[7])
this PR    GetVariantBioPolymers -> 1   PEPTIDEKPEPTIDEK (0 applied)
```

The variant proteoform disappears. Not the offending modification — the whole protein form, silently. That is a much larger response than the problem, and it is the same lever as rule 2.

## Something I got wrong along the way

I thought I had found a related defect: `AdjustModificationIndices` bounds-checks the protein's own modifications against the applied sequence (`VariantApplication.cs:440`) but merges variant-borne ones at their literal key with no check (`:455-466`). I was about to file it. Measuring first: a variant modification at 15 or 16 on a protein that shrinks to 14 is dropped, while 6 and 14 survive. Something downstream bounds them. The asymmetry in that method is real; the consequence I inferred from it is not.

## What is actually real

In the example above, master keeps the modification at key 7 — but the residue the variation declared it on was deleted, so it lands on an unrelated residue from further down the sequence. That is worth fixing. The fix is to drop that modification in `AdjustModificationIndices`, not to invalidate the variation that carries it.

## What I would do

Reduce this PR to the `maxAllowedVariantsForCombinitorics` -> `Combinatorics` rename, which is independent, uncontroversial, and stops the misspelling being copied forward. Drop both validity rules. If the modification-on-a-deleted-residue case matters to you, it wants its own change in `AdjustModificationIndices` where the response can be proportionate.

But you wrote this in #969 and may have had a case in mind that I have not reproduced — four fixtures is not a survey. If there is one, I would rather see it than delete the logic. Happy to land whatever you prefer, including keeping it as-is with the test updated.

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -307,13 +308,101 @@ internal bool Includes(int pos)
...
+                return false;
+            }
+
+            bool noSequenceChange = string.Equals(OriginalSequence ?? string.Empty,
+                                                  VariantSequence ?? string.Empty,
+                                                  StringComparison.Ordinal);
+
+            bool hasMods = OneBasedModifications != null && OneBasedModifications.Count > 0;
+
+            if (noSequenceChange && !hasMods)
```
trishorts: Rule 2 is a genuine decision reversal, and two committed tests currently pin the opposite contract in as many words:

- `TestVariantProtein.cs:1364` - *"5) Content fields (Original/Variant) and OneBasedModifications do NOT affect AreValid(), only positions do."*
- `TestVariantProtein.cs:1477` - *"OneBasedModifications is stored but ignored by AreValid()"*, and `:1483-1490` (`svContentIrrelevant`) asserts `AreValid() == true` for `begin=3, end=3, original="M", variant=""`, mods `{1: []}` - a deletion carrying a modification, which is exactly rule 3's target shape.

Reversing a pinned decision is fine, but it is worth saying so in the PR description rather than letting it read as an incidental test edit.

Also worth noting the method is named `AreValid` - plural, on a single object - because it meant "are the *positions* valid". Widening it without renaming makes the two existing call sites mean something they were not written to mean; `VariantApplication.cs:40` in particular reads as a positional sanity check and is now a semantic gate.

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -307,13 +308,101 @@ internal bool Includes(int pos)
...
+                {
+                    if (kvp.Key >= OneBasedBeginPosition)
+                    {
+                        yield return kvp.Key;
+                    }
+                }
+                yield break;
+            }
+
+            int newSpanEnd = OneBasedBeginPosition + VariantSequence.Length - 1;
```
trishorts: This is the blocker. The premise is that these keys are absolute positions in the parent protein; they are consumed as post-edit **variant** coordinates.

`VariantApplication.AdjustModificationIndices` (`VariantApplication.cs:436-468`) shifts the parent dictionary into the new frame:

```csharp
else if (variant.OneBasedEndPosition < kv.Key && kv.Key + sequenceLengthChange <= ...Length)
    mods.Add(kv.Key + sequenceLengthChange, kv.Value);
```

then merges the variant's own dictionary with keys **untouched**. The two can only be merged if the variant keys are already in the post-edit frame.

`TestProteinProperties.cs:88-104` pins it, with the variable named for it - span `[4..4]`, mod key `7`, and the comment *"the variant is MPEPPPT[mt]IDE, with the mt mod at position 7"*. Position 7 in *parent* coordinates is `D`, not `T`.

So `pos >= OneBasedBeginPosition && pos <= OneBasedEndPosition && pos > newSpanEnd` compares variant-frame positions against parent-frame span bounds. For an insertion the variant frame is longer than the span, so legitimate keys land outside `[begin..end]` by design - including the keys the GPTMD write path produces (`ProteinDbWriter.cs:602-620` -> `<subposition>` -> `ProteinXmlEntry.cs:529`).

`IBioPolymer.SelectValidOneBaseMods` (`IBioPolymer.cs:51-79`) already does this check correctly, against the real post-edit sequence and the modification's motif, and runs on every constructed variant protein (`Protein.cs:76, 194, 916`). I would delete rule 3 rather than fix it.

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -307,13 +308,101 @@ internal bool Includes(int pos)
...
+        /// removed. Non-positive keys are invalid regardless.
+        /// </remarks>
+        private IEnumerable<int> GetInvalidModificationPositions()
+        {
+            if (OneBasedModifications == null || OneBasedModifications.Count == 0)
+            {
+                yield break;
+            }
+
+            bool isTermination = VariantSequence == "*" || VariantSequence.Length == 0;
```
trishorts: `VariantSequence.Length == 0` is a **deletion**, not a termination, and mzLib is consistent about this everywhere else - termination is detected only by the asterisk:

- `VariantApplication.cs:276` - `.Split('*')[0]`
- `VariantApplication.cs:395` - `if (variant.VariantSequence.EndsWith("*"))`
- `VariantApplication.cs:409` - `&& !variant.VariantSequence.EndsWith("*")` -> downstream products are shifted and **kept** when there is no stop
- `DecoyProteinGenerator.cs:236` - `bool stopGain = sv.VariantSequence.EndsWith("*");`
- `PeptideWithSetModifications.cs:779` - `if (appliedVariation.VariantSequence == "*")`

and the class doc agrees (`SequenceVariation.cs:10-17`): *"VariantSequence (which may be empty for a deletion)"*, *"Deletion: non-empty OriginalSequence and empty VariantSequence."* The loader says the same (`ProteinXmlEntry.cs:588-591`): an empty `<variation/>` is how UniProt natively encodes a deletion.

Measured on `MPEPTIDEKR`, span 4-6:

```
DELETION  4-6 'PTI'->''  -> MPEDEKR (len 7)  mod at key 5 SURVIVES -> mods at [5]
STOP GAIN 4-6 'PTI'->'*' -> MPE     (len 3)  mod at key 5 dropped (out of range)
```

Because this branch rejects every key `>= OneBasedBeginPosition`, a deletion carrying any surviving downstream modification is declared invalid - and via the `All(...)` gate at `VariantApplication.cs:40` that costs the protein all of its variants.

Note the general branch below already handles deletions correctly: `newSpanEnd = begin + 0 - 1 = begin - 1`, so `pos >= begin && pos <= end && pos > newSpanEnd` reduces to exactly "the mod is inside the deleted span". I ran it both ways:

```
deletion + mod at 15 (outside span 5-7) : general rule -> valid    isTermination branch -> INVALID
deletion + mod at 6  (inside span 5-7)  : general rule -> invalid  isTermination branch -> invalid
```

Dropping `|| VariantSequence.Length == 0` makes this branch mean what its name says and fixes the deletion case in one character-run.

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -307,13 +308,101 @@ internal bool Includes(int pos)
...
+        /// <returns>True if this variation is a well-formed, meaningful edit; otherwise false.</returns>
         public bool AreValid()
         {
-            return OneBasedBeginPosition > 0 && OneBasedEndPosition >= OneBasedBeginPosition;
+            if (OneBasedBeginPosition <= 0 || OneBasedEndPosition < OneBasedBeginPosition)
+            {
+                return false;
+            }
+
+            bool noSequenceChange = string.Equals(OriginalSequence ?? string.Empty,
```
trishorts: This is the line CI is failing on, via `DecoyProteinGenerator.cs:352` (`decoyVariations.Where(v => v.AreValid())`).

`ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy` asserts `result.Count == 1` and now gets `0`. The start-loss reversal branch (`DecoyProteinGenerator.cs:302-305`) writes `original = reverse(OriginalSequence)` minus its last char and `variant = reverse(VariantSequence)`, which are equal precisely when `OriginalSequence.Substring(1) == VariantSequence` - every plain initiator-methionine loss of two or more residues.

Measured on consensus `MPEPTIDEKR` (decoy `MRKEDITPEP`):

```
start loss MP->P alone          : targets=2 decoys=2  -> with rule 2: 1
start loss MP->P + ordinary sub : targets=4 decoys=4  -> with rule 2: 2
```

So the target keeps its variant and the decoy loses it. Plain substitutions, stop gains and deletions are unaffected - it is specific to start-loss. That asymmetry is what `TestDecoyVariantParity.cs` is there to prevent, and it biases variant-peptide FDR rather than just changing a count.

trishorts (changes requested):
The spelling fix is good and is safe to land - I checked, MetaMorpheus references neither spelling, so nothing downstream breaks.

The `AreValid()` rewrite has three problems, and CI is already red on one of them. I verified each by running it rather than reading it.

**1. Rule 3's coordinate premise is inverted.** It assumes `OneBasedModifications` keys are absolute positions in the *parent*. They are consumed as post-edit **variant-protein** coordinates. `VariantApplication.AdjustModificationIndices` (`VariantApplication.cs:436-468`) shifts the *parent* dictionary into the new frame and merges the *variant* dictionary with keys untouched - which only composes if the variant keys are already in the new frame. A committed test names it outright (`TestProteinProperties.cs:88-104`):

```csharp
int mtModLocationInVariant = 7;
new SequenceVariation(4, 4, "P", "PPP", <vcf>,
    new Dictionary<int, List<Modification>> { { mtModLocationInVariant, ... } })
// variant is MPEPPPT[mt]IDE, mt at position 7
```

Span is `[4..4]`, key is `7`. Positions outside the span are normal and expected - that is also exactly what the GPTMD write path emits (`ProteinDbWriter.cs:602-620` stores variant-frame indices as `<subposition>`, read straight back at `ProteinXmlEntry.cs:529`). So rule 3 rejects correct data.

**2. One newly-invalid variation strips *every* variant from the protein.** The gate at `VariantApplication.cs:40` is `All(v => v.AreValid())`, not a per-variation filter. When it fails, control falls to `ApplyVariants`, which requires `Genotypes.Count > 0` (`:129`); XML-loaded variations have none, so `:136` returns the base protein alone. Measured on a protein with one no-op and one real `L10T` substitution, no VCF data:

```
master : noop.AreValid=True   -> 4 variant proteins
#1222  : noop.AreValid=False  -> 1 variant protein   (control, real variation alone: 2)
```

The legitimate `L10T` variant is lost too. Note master's 4 does contain duplicate sequences, so there *is* a real problem here worth fixing - but the fix wants to be a per-variation filter, not a predicate that fails the whole protein.

**3. A deletion is not a termination.** `isTermination = VariantSequence == "*" || VariantSequence.Length == 0` - but mzLib detects termination *only* by the asterisk, in five places (`VariantApplication.cs:276,395,409`, `DecoyProteinGenerator.cs:236`, `PeptideWithSetModifications.cs:779`), and the class doc says empty means deletion: *"VariantSequence (which may be empty for a deletion)"*. Residues after a deletion survive. Measured, same protein, same span 4-6:

```
DELETION  'PTI'->''  -> MPEDEKR (len 7)  mod at key 5 SURVIVES
STOP GAIN 'PTI'->'*' -> MPE     (len 3)  mod at key 5 dropped
```

**4. This is what CI is failing on.** `ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy` - `Expected: 1, But was: 0`. The start-loss reversal branch (`DecoyProteinGenerator.cs:302-305`) produces `original == variant` whenever `OriginalSequence.Substring(1) == VariantSequence`, i.e. for **every initiator-methionine loss**, the canonical start-loss class. Rule 2 drops those decoys while the targets remain:

| target | decoy produced | today | with rule 2 |
|---|---|---|---|
| `1-2 MA->A` | `10-10 A->A` | valid | **dropped** |
| `1-3 MPE->PE` | `9-10 EP->EP` | valid | **dropped** |

That is one-decoy-per-target broken for a whole variant class, which biases FDR - the invariant `TestDecoyVariantParity.cs` exists to protect.

**Suggested shape.** Land the spelling fix and the positional rules as-is. Then, if the duplicate-sequence problem is what you are after, put an `IsEffective()` filter *inside* `ApplyAllVariantCombinations` next to the existing `GroupBy(v => v.SimpleString())` dedup, where dropping one variation costs one variation. And I would drop rule 3 entirely - `IBioPolymer.SelectValidOneBaseMods` (`IBioPolymer.cs:51-79`) already removes non-surviving modifications, against the *actual* post-edit sequence and motif, where rule 3 has to guess.

acesnik:
🤖 You were right on all four, and I checked each against master rather than taking them on trust. Both rules are gone; what is left is the spelling fix and a coordinate-only predicate.

**Rule 3 dropped entirely, as you suggested.** The coordinate frame is the part I had backwards, and `TestProteinProperties` settles it in committed code — `new SequenceVariation(4, 4, "P", "PPP", <vcf>, { { 7, … } })`, span 4..4, modification key 7, with the comment naming the variant `MPEPPPT[mt]IDE`. So a key outside the span is normal, not broken, and `AdjustModificationIndices` merging the variant dictionary untouched only composes because those keys are already post-edit. Your point about `SelectValidOneBaseMods` is the decisive one: it filters against the real post-edit sequence and motif, where my predicate could only guess from lengths. There is no version of rule 3 worth keeping.

It also conflated a deletion with a termination, as you said — mzLib tests for the asterisk in four places and the class doc says an empty variant sequence means deletion, after which the following residues survive.

**Rule 2 dropped too.** The `All(v => v.AreValid())` gate is the thing I should have looked at before changing the predicate: one variation judged invalid diverts *every* variation on that biopolymer to the genotype-aware path, which returns the base protein alone when there are no genotypes. And a reversed initiator-methionine loss produces `original == variant` by construction, so the rule dropped those decoys while their targets stayed — one-decoy-per-target broken for a whole variant class.

**On the failing test.** `ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy` no longer exists: this branch was 34 commits behind, and master has since replaced it with `TestDecoyVariantParity.cs`. I have merged master in, so CI is now running against something current.

**What the test file does now.** The old `AreValid` test is `AreValidJudgesCoordinatesOnly`, and it pins the two things the predicate deliberately does *not* judge rather than just deleting the coverage — a modification keyed outside its span, and a deletion carrying a following modification. Both fail against the version you asked me to remove; I confirmed that by restoring it and rerunning, rather than assuming. 59 tests across `TestVariantProtein` and `TestDecoyVariantParity` pass.

**What I did not do.** I left the `IsEffective()` filter inside `ApplyAllVariantCombinations` out of this PR. You are right that the duplicate sequences in master's 4 are a real problem, and next to the existing `GroupBy(v => v.SimpleString())` is the right place for it — but it changes which variant proteins come out, so it wants its own measurement and its own PR rather than riding along in one that is now purely a revert plus a rename. Happy to take it next if you want it.

mzLib/Omics/BioPolymer/SequenceVariation.cs
trishorts: Addressed in `b25d23db` -- rule 2 is gone, so there is no decision reversal left to declare and the two pinned tests keep their contract. `AreValid()` is back to `OneBasedBeginPosition > 0 && OneBasedEndPosition >= OneBasedBeginPosition`, and the name now matches what it does again. Resolving.

mzLib/Omics/BioPolymer/SequenceVariation.cs
trishorts: Addressed in `b25d23db` -- rule 3 was deleted rather than patched, which was the ask. The new `<remarks>` block records why: the keys are post-edit variant-frame coordinates, `AdjustModificationIndices` merges the variant dictionary untouched, and `SelectValidOneBaseMods` already does this correctly against the real post-edit sequence. Resolving.

mzLib/Omics/BioPolymer/SequenceVariation.cs
trishorts: Addressed in `b25d23db` -- the `VariantSequence.Length == 0` termination branch is gone with the rest of the widening, so a deletion carrying a surviving downstream modification is no longer declared invalid and no longer costs the protein all of its variants through the `All(...)` gate. Resolving.

mzLib/Omics/BioPolymer/SequenceVariation.cs
trishorts: Addressed in `b25d23db`. With rule 2 removed the reversed start loss stays valid, so `ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy` gets its 1 back and target/decoy variant counts stay symmetric. CI is green. Resolving.

trishorts (dismissed):
Approving — this reverses my earlier changes-requested.

`b25d23db` reduces `AreValid()` back to the coordinate check, which resolves all four of my findings at once rather than patching them individually: the pinned no-op contract stands, the modification-position rule that read variant-frame keys against parent-frame bounds is gone, the `VariantSequence.Length == 0` termination branch that failed deletions is gone, and `ReverseSequenceVariations_StartLoss_VariantAtEndOfDecoy` gets its decoy back so target/decoy variant counts stay symmetric.

What's left is the `Combinitorics` → `Combinatorics` spelling fix plus a `<remarks>` block and `AreValidJudgesCoordinatesOnly` recording *why* this method is coordinate-only. That test is the right artifact — it pins the two things the method deliberately does not judge, so the next person to widen it trips a test rather than CI three files away.

Two non-blocking nits, neither worth another round:

- `SequenceVariation.cs:1` — `using System;` came in with the reverted logic and is now unused; the file built without it on master and nothing in the remaining diff needs it.
- `TestVariantProtein.cs:1399-1401` — the rewritten comment reads *"AreValid also requires the variation to be a real change and its modifications to be positionally possible -- neither is AreValid's job"*, which contradicts itself. It wants "does **not** require".

The failing `codecov/patch` is an artifact, not a gap: the 5 flagged lines are all in `VariantApplication.cs` and are the parameter rename touching already-uncovered branches. The coverage delta on both files is `(ø)`.

mzLib/Test/DatabaseTests/TestVariantProtein.cs
```diff
@@ -1351,8 +1396,10 @@ public void IndelDecoyVariants()
         public void SequenceVariationIsValidTest()
         {
             // PURPOSE
-            // Validate the minimal, position-only "validity" rules implemented by SequenceVariation.AreValid():
-            //   AreValid() == (OneBasedBeginPosition > 0) && (OneBasedEndPosition >= OneBasedBeginPosition)
+            // Validates the POSITION rule of SequenceVariation.AreValid():
+            //   begin >= 1 and end >= begin
+            // AreValid also requires the variation to be a real change and its modifications to be
+            // positionally possible -- neither is AreValid's job; see AreValidJudgesCoordinatesOnly above. Every construct here changes the sequence, so only the position rule is in play.
```
trishorts, line 1402: This says `AreValid` requires two rules, then says neither is its job. It reads as the old sentence with the correction appended rather than substituted -- residue from the revision that was withdrawn.

On a PR whose entire payload is recording *why* the predicate is narrow, a comment stating the opposite is the one thing that cannot ship. It also runs onto a single very long line.

```csharp
            // Validates the POSITION rule of SequenceVariation.AreValid():
            //   begin >= 1 and end >= begin
            // AreValid judges nothing else -- not whether the variation is a real change, and not
            // whether its modifications are positionally possible. See AreValidJudgesCoordinatesOnly
            // above for why both are deliberately out of scope.
```

mzLib/Test/DatabaseTests/TestVariantProtein.cs
```diff
@@ -1361,7 +1408,7 @@ public void SequenceVariationIsValidTest()
             // 4) One-position convenience ctor behavior for different originalSequence values (null, "", length > 0).
             //    - This ctor derives end as: end = (original == null) ? begin : begin + original.Length - 1.
             //    - Therefore, empty originalSequence "" makes end = begin - 1 → invalid by design.
-            // 5) Content fields (Original/Variant) and OneBasedModifications do NOT affect AreValid(), only positions do.
+            // 5) Every construct below changes the sequence, so it is the position rule being exercised.
```
trishorts, line 1411: master's line was:

```csharp
            // 5) Content fields (Original/Variant) and OneBasedModifications do NOT affect AreValid(), only positions do.
```

That was true before this PR, is true after it, and is almost verbatim the claim the PR set out to write down. The replacement says strictly less, and points at the constructs instead of at the rule. Please restore the original.

mzLib/Test/DatabaseTests/TestVariantProtein.cs
```diff
@@ -1404,7 +1451,7 @@ public void SequenceVariationIsValidTest()
 
             // -----------------------------
             // 3) Explicit begin/end edge-cases: insertion and deletion modeled by content only
-            //    NOTE: AreValid ignores Original/Variant content; only positions matter.
+            //    NOTE: these all change the sequence, so the position rule is what decides them.
```
trishorts, line 1454: Same shape as line 1411. `NOTE: AreValid ignores Original/Variant content; only positions matter.` was accurate on master and still is -- revert this one too.

mzLib/Test/DatabaseTests/TestVariantProtein.cs
```diff
@@ -1443,7 +1490,7 @@ public void SequenceVariationIsValidTest()
             var svPosCtorLength3 = new SequenceVariation(
                 oneBasedPosition: 20,
                 originalSequence: "PEP",   // len = 3
-                variantSequence: "AAA",    // content irrelevant to AreValid
+                variantSequence: "AAA",    // differs from original, so it is a real change
```
trishorts, line 1493: `// content irrelevant to AreValid` was correct. `// differs from original, so it is a real change` implies `AreValid` cares whether it is a real change, which is the belief this PR is retracting. Revert.

mzLib/Omics/BioPolymer/SequenceVariation.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
```
trishorts, line 1: `using System;` is unused. Everything this PR adds to this file is XML doc comment, and doc comments cannot create a using requirement -- so if master compiled without it, this file still does. Nothing here touches a `System` type either; the `GetHashCode` calls at 237-243 are instance methods on `int`/`string`. Leftover from the withdrawn implementation; drop the line.

mzLib/Omics/BioPolymer/VariantApplication.cs
```diff
@@ -197,15 +197,15 @@ public static List<TBioPolymerType> ApplyVariants<TBioPolymerType>(TBioPolymerTy
 
                         foreach (var ppp in newVariantProteins)
                         {
-                            if (isDeepAlternateAllele && maxAllowedVariantsForCombinitorics > 0 && isDeepReferenceAllele)
+                            if (isDeepAlternateAllele && maxAllowedVariantsForCombinatorics > 0 && isDeepReferenceAllele)
```
trishorts, line 200: The rename stops at the parameter. `combinitoricProteins` (declared line 196, used at 204, 206, 210, 214, 217) keeps the old spelling inside the very method whose parameter you just corrected.

It is a local, so nothing breaks and I am not treating it as blocking. But leaving one `combinitoric` behind here means the next person greps, finds it, and opens the same PR again. Either finish it, or say in the body that locals were left alone on purpose.

(For the record, the two other hits in the repo -- `ProteolyticPeptide.cs:10` and `TestVariantProtein.cs:1246`, both in comments -- are unrelated to this method and clearly out of scope.)

trishorts (commented):
Read the whole diff rather than running the engine on it (3 files, 87 added lines). Confirmed the headline claim: `AreValid`'s body is context in the diff, byte-identical to master. Also confirmed the two pieces of evidence the new doc leans on -- `TestProteinProperties.cs:94-96` really is `SequenceVariation(4, 4, "P", "PPP", ..., {7: mt})`, span 4..4 with a modification at 7; and grepping MetaMorpheus for either spelling of `Combinitorics`/`Combinatorics` returns nothing, so the parameter rename has no downstream caller.

The test is a genuine pinning test, not a restatement of the implementation, and it earns its place.

What I found is all revert residue -- six spots where text from the withdrawn revision survived the revert. One of them (the PURPOSE block in `SequenceVariationIsValidTest`) now asserts the opposite of what this PR exists to record, which on a documentation PR is the one thing that cannot ship. The rest are small.

Two non-blocking notes:
- `codecov/patch` is red. The only production lines in the diff are the ten rename touches in `VariantApplication.cs`, so this is a rename brushing against code that was already uncovered on master -- not a gap you introduced.
- `mergeStateStatus` is `BEHIND`. Update the branch before anyone approves, since the update dismisses approvals.

`ApplyVariants` is public and its parameter name is part of the shipped signature, so any external consumer calling it with a named argument breaks at compile time. Nothing in mzLib or MetaMorpheus does, but it is worth a line in the release notes since mzLib ships as a package.

trishorts (approved):
Approved. Pushed a1bd8bf7 + 5f0c1409 for codecov/patch: 4 tests covering ApplyVariants' capped heterozygous branching, plus removal of two always-true guards (and the unreachable else-if they left) in the uncapped branch — those could never be fully branch-covered. No behavior change; all 172 variant tests pass. Also brought the branch up to date with master.

nbollis (approved):

PR #1227, opened 2026-08-27
Carry precursor metadata through mgf, and trim mgf peaks by MS level

acesnik:
🤖 Two independent mgf fixes, found while making MetaMorpheus able to write calibrated spectra as mgf (MetaMorpheus#1673). Either is useful alone; together they make an mgf round trip lossless for a search.

## 1. The reader trimmed MS1 peaks regardless of `TrimMs1Peaks`

`Mgf.cs` gated peak filtering on `ApplyTrimmingToMsMs` alone and applied it to every block, ignoring MS level. Every other reader dispatches on `msnOrder` — `Mzml.cs:825` and `:1091`, `ThermoRawFileReader.cs:585`, `TofSpectraMerger.cs:141`, and `BrukerFileReader.ShouldTrim`, whose comment already describes this exact failure. Only mgf did not, because mgf historically held nothing but MS2 blocks, so the flag was a fine proxy. Since #1165 the writer emits MS1 blocks too.

**This changes loaded spectra for any mgf containing MS1 scans.** Read with the `FilteringParams` MetaMorpheus builds by default — 200 peaks per window, `TrimMs1Peaks` false, `TrimMsMsPeaks` true — MS1 peaks went **47,169 → 4,723**, a 90% loss. That strips the isotope envelopes, so precursor deconvolution over those scans found **81 envelopes instead of 475**, with 43 of 118 MS2 scans getting none at all. After the fix the mgf and mzML readings of the same data agree exactly at 475.

Nothing changes for an mgf with no MS1 blocks — which is all 17 mgf fixtures in this repo. `ApplyTrimmingToMsN` was also never honoured here and now is.

## 2. Five fields the format had nowhere to put

`TIC` on every scan; `ACTIVATIONMETHOD`, `PRECURSORSCAN`, `ISOLATIONWIDTH` and `ISOLATIONMZ` on MS2 and above. Written only where the source had a value — a reader cannot tell an empty header from a zero — and read back with `TryParse` throughout, since these arrive in files we did not necessarily write.

The specification does not say what a reader must do with an unknown key, so this is de facto rather than de jure. Three that matter skip them silently, which is what makes it safe rather than merely convenient:

| reader | unknown key | block with no `PEPMASS` |
|---|---|---|
| ProteoWizard `SpectrumList_MGF` | `else { continue; // ignored attribute }` | tolerated |
| OpenMS `MascotGenericFile` | silently skipped | tolerated |
| MSToolkit `MSReader` | silently skipped | **`exit(-12)`** |

`TIC` is worth separating out. The recorded value is not the sum of the written peaks — it predates centroiding and thresholding — and MetaMorpheus scores a match partly as a fraction of it, so recomputing it moved scores by up to 0.003. Without the header the reader still falls back to the peak sum, so files written before this read as they did.

## 3. `includeMs1Scans`, because MS1 blocks are out of spec

`WriteMgf` and `ExportAsMgf` gain `includeMs1Scans`, defaulting to `true` so no existing caller changes. MS1 blocks are what let a precursor be re-deconvoluted, but the specification says "At least one PEPMASS parameter is required" and an MS1 has none. That last column of the table above is the consequence: MSToolkit is Comet's reader, so an mgf with MS1 blocks currently kills Comet on load. Passing `false` emits an MS2-only file every reader accepts.

Worth flagging that writing MS1 blocks is **already master's behaviour** from #1165 — this PR adds the way out, it does not introduce the problem.

## This changes search results read from an mgf

With both fixes, a MetaMorpheus Calibrate → Search chain through mgf matches mzML exactly on the bundled yeast file:

| | PSMs | peptides | protein groups | precursors |
|---|---|---|---|---|
| mzML | 80 | 55 | 46 | 482 |
| mgf, before | 61 | — | — | 118 |
| mgf, after | 80 | 55 | 46 | 482 |

`AllPSMs.psmtsv` is identical byte for byte across all 57 columns, as are `AllPeptides.psmtsv`, the Percolator table and `results.txt`.

## Tests

Seven added to `TestMgf.cs`; 30/30 pass. The three reader tests were confirmed to fail against the unfixed reader rather than assumed to — the trimming one fails with `Expected: 100, But was: 5`.

Not verified: the full `Test` project cannot run on macOS (`mzPlot` is WPF), so these ran through a throwaway net10.0 harness compiling `TestMgf.cs` against `Readers`. CI is the check for the rest.


nbollis (dismissed):

trishorts (commented):
I checked the two things that would have been serious here and both are clean:

- **All 21 `MsDataScan` arguments land in the right slots.** Against master the PR substitutes exactly slots 17/18/19/20 (`isolationMZ`, `isolationWidth`, `dissociationType`, `oneBasedPrecursorScanNumber`) plus the slot-10 TIC fallback, each keeping the old value as its `??` default. Your own `ISOLATIONMZ=572` / `ISOLATIONWIDTH=3` -> `570.5/573.5` test independently rules out a 17<->18 transposition.
- **No prefix collisions in the parse chain.** I checked all 55 ordered pairs. The two that look alarming aren't: `"PRECURSORSCAN=1".StartsWith("SCANS")` is `false` (`SCANS` is an infix, not a prefix), and `ISOLATIONMZ`/`ISOLATIONWIDTH` diverge at `M` vs `W`. No real Matrix Science local (`TOL`, `SEQ`, `COMP`, `TAG`, `RAWSCANS`, `USER00`-`12`, ...) begins with any of the five new keys either.

**The MS-level trimming fix is right, and it has direct precedent.** Bruker had this exact bug and fixed it the same way; `TestBruker.cs:262` (`TestMs1NotTrimmedWhenApplyTrimmingToMs1IsFalse`) exists specifically because trimming MS1 under the MsMs flag *"stripped precursor isotope envelopes from the MS1 and precursor deconvolution then found nothing for most MS2 scans."* Same bug, same fix.

Comments below are design/reuse, not correctness - none of them blocks.

mzLib/Readers/Mgf/Mgf.cs
```diff
@@ -265,14 +308,25 @@ public static MsDataFile LoadAllStaticData(string filePath, FilteringParams filt
...
+                totalIonCurrent ?? intensities.Sum(), 0, null, null, precursorMz, charge,
+                precursorIntensity, isolationMz ?? precursorMz, isolationWidth,
+                dissociationType ?? DissociationType.Unknown,
+                precursorScanNumber, precursorMz);
+        }
+
+        // Peak trimming must honor the per-MS-level FilteringParams flags, exactly as the mzML/Thermo/Bruker
+        // readers do. Trimming an MS1 precursor scan strips its isotope envelopes and precursor
+        // deconvolution then finds nothing.
+        private static bool ShouldTrim(IFilteringParams filterParams, int msnOrder)
```
trishorts, line 392: This is a verbatim copy of `BrukerFileReader.ShouldTrim` - `Readers/Bruker/BrukerFileReader.cs:388`, same name, same three clauses, same body. That makes this the third open-coded copy of the predicate; the full set is five:

| site | predicate |
|---|---|
| `BrukerFileReader.ShouldTrim` :388 | `Ms1 && ==1` / `MsMs && ==2` / `MsN && >2` |
| `Mzml.LoadAllStaticData` :825 | same, inline |
| `Mzml.GetMsDataOneBasedScanFromConnection` :1091 | same, inline (verbatim copy of :825) |
| `ThermoRawFileReader.GetSpectrum` :585 | same, inline |
| `TofSpectraMerger.CreateFilteredSpectrum` :141 | **`MsMs && msnLevel > 1`, `ApplyTrimmingToMsN` never read** |

Since you're already touching this, promoting it to one shared seam - an `IFilteringParams` extension in `MassSpectrometry`, or a default interface method - would collapse six call sites onto one. It would also surface the timsTOF outlier, which trims MS3 off the *MsMs* flag and ignores `ApplyTrimmingToMsN` (pre-existing, out of scope here, but it's the one a shared helper would fix for free).

Entirely reasonable to keep the private copy and do the extraction separately - but then it's worth a follow-up issue, because the count only goes up.

While you're here: `FilteringParams.cs:59` says `ApplyTrimmingToMsN` *"is currently only applied in mzml and thermo raw file readers."* Bruker made that three; this PR makes it four.

mzLib/Readers/Mgf/MgfMethods.cs
```diff
@@ -108,6 +121,48 @@ private static void WriteScan(StreamWriter output, MsDataScan scan, string sourc
                 output.WriteLine($"RTINSECONDS={rtInSeconds.ToString(CultureInfo.InvariantCulture)}");
             }
 
+            // Four fields a search needs that the Matrix Science local parameters have no home for.
```
trishorts, line 144: Worth stating the reasoning for five new top-level keys explicitly, because mzLib's existing answer to "the format has nowhere to put this" has been the format's own free-text slot rather than new keys:

- **MSP** - `LibrarySpectrum.ToString()` packs extras into `Comment: Parent=... RT=...` and parses them back out (`SpectralLibrary.cs:382,549,660,792`).
- **MGF, in this very file** - `BuildTitle` already smuggles `File:"..."` and `NativeID:"..."` into `TITLE`, which is the ProteoWizard/msconvert convention and MGF's only sanctioned free-text field. (It's written but never parsed back, so `NativeId` is still `null` on read.)

`MSLEVEL` is the one precedent for a new top-level key, and it's defensible precisely because SIRIUS, pyteomics and MsBackendMgf already emit it. These five are mzLib-only, so an mgf we write is now readable-with-full-fidelity only by us. `TITLE` tokens would round-trip through third-party tools that preserve the title.

Not arguing you're wrong - the `TIC` case in particular is hard to fit in a title, and your remarks about scoring as a fraction of TIC are convincing. Mostly asking that the choice be recorded in the remarks so the next person doesn't have to re-derive it.

One latent snag either way: `PRECURSOR` is a real Mascot search parameter and *is* a proper prefix of `PRECURSORSCAN`. It's inert today (globals are skipped before `BEGIN IONS`), but the day someone adds a `StartsWith("PRECURSOR")` branch above yours, every `PRECURSORSCAN=` is silently swallowed with no test failure. `MsAlign.ParseEntryLines` (`MsAlign.cs:417`) switches on the exact token before `=` for its 16 keys, which is immune to this; at 11 keys this chain is now past the size where MsAlign made that move.

mzLib/Readers/Mgf/Mgf.cs
```diff
@@ -205,6 +210,44 @@ public static MsDataFile LoadAllStaticData(string filePath, FilteringParams filt
...
+                else if (line.StartsWith("TIC") && sArray.Length > 1)
+                {
+                    if (double.TryParse(sArray[1], NumberStyles.Float | NumberStyles.AllowThousands, CultureInfo.InvariantCulture, out double parsedTic))
+                    {
+                        totalIonCurrent = parsedTic;
+                    }
+                }
+                else if (line.StartsWith("ACTIVATIONMETHOD") && sArray.Length > 1)
+                {
+                    if (Enum.TryParse(sArray[1].Trim(), ignoreCase: true, out DissociationType parsedDissociation))
```
trishorts: `Enum.TryParse` accepts numeric strings, so `ACTIVATIONMETHOD=99` returns `true` and yields an undefined `DissociationType` of `99` rather than falling back to `Unknown`. Your negative test uses `"Telekinesis"`, which correctly fails, so this path isn't covered.

```csharp
if (Enum.TryParse(sArray[1].Trim(), ignoreCase: true, out DissociationType parsed)
    && Enum.IsDefined(typeof(DissociationType), parsed))
```

Minor, and `MsAlign.cs:441` has the identical hole with `ACTIVATION`, so it is at least house-consistent. Related: MsAlign defaults an unparseable activation to `Autodetect`, this defaults to `Unknown` - two readers now answer the same question differently, worth picking one.

mzLib/Readers/Mgf/MgfMethods.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
```
trishorts, line 1: Nit: this adds a UTF-8 BOM to the file (`using System;` -> `﻿using System;`), which shows up as a whole-file touch in the blame. Probably an editor artifact rather than intent.

mzLib/Readers/Mgf/MgfMethods.cs
```diff
@@ -108,6 +121,48 @@ private static void WriteScan(StreamWriter output, MsDataScan scan, string sourc
...
+            if (scan.MsnOrder > 1)
+            {
+                if (scan.DissociationType.HasValue && scan.DissociationType.Value != DissociationType.Unknown)
+                {
+                    output.WriteLine($"ACTIVATIONMETHOD={scan.DissociationType.Value}");
+                }
+
+                if (scan.OneBasedPrecursorScanNumber.HasValue)
+                {
+                    output.WriteLine($"PRECURSORSCAN={scan.OneBasedPrecursorScanNumber.Value.ToString(CultureInfo.InvariantCulture)}");
```
trishorts, line 177: This is written unconditionally, but the block it points at may not be in the file.

`WriteMgf` filters `scansToWrite` two ways — empty spectra, and now `includeMs1Scans` — and `WriteScan` never learns which scans survived. So an MS2 keeps `PRECURSORSCAN=1` after MS1 scan 1 was dropped. Reading that back, `OneBasedPrecursorScanNumber` is now populated where master passed `null` for it, and `MsDataFile.GetOneBasedScan` is a raw `Scans[scanNumber - 1]` (`MassSpectrometry/MsDataFile.cs:107`), so the missing slot comes back **null** rather than throwing.

`MzmlMethods.cs:525` then does this, unguarded:

```csharp
string precursorID = myMsDataFile.GetOneBasedScan(scanWithPrecursor.OneBasedPrecursorScanNumber.Value).NativeId;
```

Round-tripping mgf → mzML on this branch:

```
A. includeMs1Scans:false, then ExportAsMzML
   read back: 1 scan(s), MS2 precScan=1
   ExportAsMzML: THREW NullReferenceException

B. control, includeMs1Scans:true
   read back: 2 scan(s)
   ExportAsMzML: OK

C. default includeMs1Scans:true, but the MS1 has no peaks so the writer skips it anyway
   read back: 1 scan(s), MS2 precScan=1
   ExportAsMzML: THREW NullReferenceException
```

Case C is the one I would not want to lose: this is not confined to the new flag. The pre-existing empty-scan skip produces the same dangling reference the moment the reader starts populating the field, so `includeMs1Scans` only widens a door this PR opens.

`WriteMgf` already has `scansToWrite` in hand, so passing a `HashSet<int>` of written scan numbers into `WriteScan` and emitting the header only on a hit would close both cases at once. Writing no `PRECURSORSCAN` is exactly the "absent beats wrong" rule the surrounding comment already argues for — a number that resolves to nothing is worse than an empty header, not better.

Worth a test alongside `MgfWriterCanOmitMs1ScansToStayWithinTheSpecification`, since that test reads the MS2-only file back but never follows the precursor pointer.

mzLib/Readers/Mgf/Mgf.cs
trishorts: `Enum.TryParse` is more permissive than this contract wants. It accepts the *numeric* form of an enum, and it does not require the result to be a defined member:

```
'HCD'         -> parsed=True  value=HCD  isDefined=True
'Telekinesis' -> parsed=False
'3'           -> parsed=True  value=SID  isDefined=True
'999'         -> parsed=True  value=999  isDefined=False
'-1'          -> parsed=True  value=-1   isDefined=False
```

Read back through the reader, `ACTIVATIONMETHOD=999` gives a scan whose `DissociationType` is `(DissociationType)999`. `Enum.IsDefined` is false, so anything that switches on it falls through to whatever its default arm is, carrying a value no `DissociationType` check will match.

`ACTIVATIONMETHOD=3` is the worse one of the two: it silently becomes `SID`. That is not a skipped field, it is a confidently wrong one — and the comment three lines up says these arrive in files we did not necessarily write, which is exactly where a numeric activation code is plausible.

`&& Enum.IsDefined(typeof(DissociationType), parsedDissociation)` restores "unrecognised means Unknown" for both. `MgfReaderIgnoresUnparseableExtensionHeaders` only covers the alphabetic case (`Telekinesis`), which is the one `TryParse` already rejects, so a numeric case there would pin this down.

mzLib/Readers/Mgf/Mgf.cs
trishorts, line 392: This is now the second copy: `BrukerFileReader.cs:387` has the same three-line rule with the same operand order and a comment that differs only in which readers it lists.

Not a problem in this PR — it is the right rule and matching Bruker is the right call — but the PR description's own argument is that every reader should be dispatching on `msnOrder` the same way, and the way to make that true is one shared helper rather than a copy per reader. An extension method on `IFilteringParams` would let Bruker, mgf and the next one call the same code, and would make the next reader that forgets this a compile-time-visible omission instead of a silent one.

Fine to leave for a follow-up if you would rather not touch `BrukerFileReader` here.

trishorts (commented):
Reviewed locally on `bba710e5`: `Readers.csproj` builds clean and all 30 `TestMgf` tests pass. The trimming fix is plainly right and the evidence for it is the good kind — mgf and mzML agreeing at 475 envelopes is the assertion that matters, and `ShouldTrim` matching the other readers rather than inventing a fourth rule is the correct shape. Separating `TIC` out because the recorded value predates centroiding is a nice catch that a lot of round-trip work would have missed.

One thing I think blocks: `PRECURSORSCAN` can name a scan the file no longer contains, and mzLib's own mzML writer dereferences it without a guard. Details inline, with the reproduction.

Every case below was run against this branch rather than read off the diff.

mzLib/Readers/Mgf/MgfMethods.cs
acesnik: 🤖 You are right that this blocks, and case C is the reason — thank you for separating it out. The dangling pointer is not the new flag's doing: `WriteMgf` has always dropped empty scans, so the moment the reader starts populating `OneBasedPrecursorScanNumber` the pre-existing skip produces the same reference into a scan that was never written. `includeMs1Scans` widens a door this PR opens rather than opening it.

Traced the rest of it: `MsDataFile.GetOneBasedScan` is a raw `Scans[scanNumber - 1]`, so a gap comes back null rather than throwing, and `MzmlMethods.cs:525` guards `HasValue` and then dereferences the returned scan unguarded — so the null arrives as an NRE one line later, at export, a long way from the mgf that caused it.

Taking your fix: `WriteMgf` has `scansToWrite` in hand, so threading a `HashSet<int>` of written scan numbers into `WriteScan` and emitting `PRECURSORSCAN` only on a hit closes both cases at the source. That is the same "absent beats wrong" rule the surrounding comment already argues, and you are right that I had it backwards here — a number resolving to nothing is worse than no header, not better.

Test alongside `MgfWriterCanOmitMs1ScansToStayWithinTheSpecification` that follows the precursor pointer rather than just reading the file back, and a second one for case C with a peakless MS1, since that one survives independent of the flag.

mzLib/Readers/Mgf/Mgf.cs
acesnik: 🤖 Confirmed and taking `Enum.IsDefined`. `ACTIVATIONMETHOD=3` silently becoming `SID` is the one I would have missed — it is not a skipped field but a confidently wrong one, and the comment three lines above says these arrive in files we did not write, which is precisely where a numeric activation code shows up. `999` is less dangerous only because nothing matches it.

Adding a numeric case to `MgfReaderIgnoresUnparseableExtensionHeaders`, since `Telekinesis` only covers what `TryParse` already rejects and so cannot fail against this bug.

On the two readers disagreeing: `MsAlign.cs:441` has the same hole and defaults to `Autodetect` where this defaults to `Unknown`. I would rather settle that separately than change `MsAlign`'s default inside an mgf PR — `Unknown` is the honest answer for a value we could not read, and `Autodetect` claims a capability, so my preference is `Unknown` in both. Happy to open that as its own issue unless you would rather I fold it in.

mzLib/Readers/Mgf/Mgf.cs
acesnik: 🤖 Agreed, and your count is right — I checked: the rule is open-coded at `BrukerFileReader.cs:390`, `Mgf.cs:327`, `Mzml.cs:825`, `Mzml.cs:1091` and `ThermoRawFileReader.cs:585`, five copies with the same three clauses, plus `TofSpectraMerger.cs:141-142`, which is the outlier you flagged — `ApplyTrimmingToMsMs && msnLevel > 1`, so it trims MS3 off the MsMs flag and never reads `ApplyTrimmingToMsN` at all.

I would rather do the extraction separately than grow this PR into `BrukerFileReader` and `Mzml`, so I am taking your follow-up option and opening an issue for an `IFilteringParams` extension collapsing all six onto one seam, naming the timsTOF divergence as the thing a shared helper fixes for free. Also noting `FilteringParams.cs:59` still says `ApplyTrimmingToMsN` is "currently only applied in mzml and thermo raw file readers", which Bruker made three and this makes four — I will fix that line here, since it is a comment in a file I am already touching.

mzLib/Readers/Mgf/MgfMethods.cs
acesnik: 🤖 Fair, and the ask is reasonable — the choice should be recorded rather than re-derived. Adding the reasoning to the remarks: `TITLE` tokens are the convention for anything a third-party tool should preserve, and the reason these five are not tokens is that `TIC` and `PRECURSORSCAN` are consumed on read to reconstruct an `MsDataScan`, so they need a parse that does not depend on how another tool rewrote a free-text field. `MSLEVEL` is the precedent for a top-level key and the honest note is that these five are mzLib-only, so full-fidelity round-tripping is ours alone. I will say that in the remarks rather than imply otherwise.

The `PRECURSOR`/`PRECURSORSCAN` prefix snag is the part I want to act on. Inert today, but a latent trap that fails silently is exactly the kind this chain should not be accumulating at 11 keys. `MsAlign.ParseEntryLines` switching on the exact token before `=` is immune, and you are right that this is past the size where that move pays. Converting the chain to an exact-token switch is a bigger diff than this PR should carry, so: issue for the conversion, and in the meantime I am ordering `PRECURSORSCAN` above any bare `PRECURSOR` branch and leaving a comment at the seam saying why the order matters.

mzLib/Readers/Mgf/MgfMethods.cs
acesnik: 🤖 Editor artifact, not intent — you are right, and it is worth removing rather than shrugging at, since it shows up as a whole-file touch in blame. `MgfMethods.cs` starts `EF BB BF` on this branch against no BOM on master. Stripping it so the diff is only the lines I actually changed.

trishorts:
@acesnik — pushed the outstanding items onto this branch directly, same as on #1184. Apologies if you had them in progress; if so take yours and drop mine. Two of them turned out wider than the threads described, so those are worth reading rather than skimming.

**`Enum.IsDefined` does not fix `ACTIVATIONMETHOD`.** That is what the thread proposed and what we both agreed to, and it is not enough. `TryParse` is too permissive in three ways and `IsDefined` closes only the first:

```
ACTIVATIONMETHOD=999      -> DissociationType 999, matched by nothing        (IsDefined catches this)
ACTIVATIONMETHOD=3        -> SID -- a DEFINED member, so IsDefined allows it
ACTIVATIONMETHOD=HCD,CID  -> plain HCD -- also defined, also allowed
```

`3` was the case you singled out as the dangerous one, and it survives `IsDefined` untouched, because there is nothing invalid about `SID`. The third I did not have either: `TryParse` accepts comma-separated lists for *any* enum, not only `[Flags]` ones, and ORs the members — and since `CID` is `0`, `HCD,CID` is just `HCD`. A file naming two activations reads as naming one, and `ETD,HCD` is how some tools spell what mzLib calls `EThcD`. All three are `Unknown` now, by refusing the numeric and comma forms.

`Enum.IsDefined` is then dead weight: with those two guards, `TryParse` is matching a single name against the member list, so anything it accepts is defined by construction. I checked rather than assumed — removing it changes no case in the new test — so it is not in the diff. Mapping known combinations onto their own members (`ETD,HCD` to `EThcD`) is a real question but wants its own evidence.

**`PRECURSORSCAN`** is threaded the way you described: written scan numbers into `WriteScan`, header only on a hit. Fixed at the source rather than in the `includeMs1Scans` branch, because your case C is right that the pre-existing empty-spectrum filter dangles the same pointer at default settings.

One thing worth knowing about the tests for it: they assert the **file** and the **round trip**, not that a downstream mzML export stops throwing. It does not stop throwing. `ExportAsMzML` fails for two separate pre-existing reasons on master — an isolation window that guards `IsolationMz.HasValue` then dereferences `IsolationRange` (null unless the *width* is also set, which mgf has no field for), and a chromatogram loop that treats `NumSpectra` as a scan number. Filed as #1273 with both. Neither is yours; your fix closes the `MzmlMethods:525` dereference specifically, and that part is verified.

Also in: the stale doc clause, the BOM, the `FilteringParams` comment, and the remarks now recording why these are top-level keys rather than `TITLE` tokens.

Still yours to open if you want them: the `IFilteringParams` extension collapsing the six `ShouldTrim` copies (with the `TofSpectraMerger` outlier), and the exact-token switch for the header chain. I have left both alone.

Four mutants, each failing only what it should. Full offline suite 5895 passed, 0 failed. Approving once CI agrees.

trishorts (approved):
Approving at `9295df33`, all five checks green.

The branch update merged #1184's changes — now on master — with this PR's, and both touched `Mgf.cs` and `TestMgf.cs`, including the same `AddPeakIfWellFormed` doc clause. Auto-merging that cleanly is not the same as it being right, so I checked the merged tree rather than the diff: the doc clause resolved to one coherent version, no test names duplicated, and the full offline suite is 5899 passed, 0 failed, 29 skipped.

What this PR ends up doing is more than carrying five headers. The two findings that mattered were both cases of a value being read *confidently and wrongly* rather than skipped, which is the failure mode worth catching in a reader: a `PRECURSORSCAN` naming a block that was never written, and an `ACTIVATIONMETHOD` half-recognised into a real-looking answer. Both now fail closed.

On the activation method, worth recording that the fix we agreed in the thread — `Enum.IsDefined` — was not enough, and I have said so in the commit rather than quietly widening it. `ACTIVATIONMETHOD=3` parses to `SID`, a defined member, so `IsDefined` allows it; and `HCD,CID` ORs to plain `HCD` because `CID` is `0`, so it allows that too. Refusing the numeric and comma forms is what closes them, after which `IsDefined` does no work at all and is not in the diff.

Four threads resolved. Three left open deliberately, because the action agreed on them was a follow-up issue and none has been opened yet:

- the `ShouldTrim` duplication (two threads) — the `FilteringParams` comment is fixed here, but the `IFilteringParams` extension collapsing all six copies, and the `TofSpectraMerger` outlier it would fix for free, still wants an issue;
- the header-key chain — the reasoning is now recorded in the remarks as asked, but the exact-token switch does not have an issue either.

Neither blocks this PR. Both are the kind of thing that evaporates if it stays in a review thread, which is the only reason I am leaving the threads open rather than tidying them away.

Separately filed while verifying this: #1273, `ExportAsMzML` throwing `NullReferenceException` from two independent pre-existing causes. That is why the new writer tests assert the file and the round trip rather than a downstream export — an export assertion would have been asserting something that is false for reasons unrelated to this PR.

nbollis (approved):

PR #1233, opened 2026-08-27
Close the test gaps mutation testing found in Chemistry

trishorts:
# Close the test gaps mutation testing found in Chemistry

Adds seven tests to `Chemistry`'s existing fixtures. **No production code changes.** Every
assertion pins behaviour that master already has — nothing here is a bug fix, and nothing here
should change a single result.

## Why these seven tests, and not seven others

`Chemistry` is one of the best-covered projects in mzLib: **99.6% line coverage, 93.0% branch
coverage**. At that level line coverage has nothing left to say — it can only tell you a line
*executed*, never that a test would *notice* if the line were wrong.

So we asked the second question directly, with mutation testing. [Stryker.NET][stryker] 4.16.0
mechanically introduces small defects into the production code one at a time — `*` becomes `/`,
`>` becomes `>=`, a statement is deleted, a ternary is forced down one branch — rebuilds, and
runs the tests. If a test fails, that defect is *detectable* and the mutant is "killed". If the
whole suite still passes, the mutant **survived**: there is a way to break that line that 6,008
tests would not catch.

```
dotnet stryker -tp ../Test/Test.csproj -c 8 -r html -r json    # run from mzLib/Chemistry
```

Result on master (`3829034d`), 51 minutes:

| | |
|---|---|
| mutants generated | 710 |
| excluded (compile errors, already-covered blocks) | 164 |
| **tested** | **546** |
| killed | 516 |
| timeout (counts as killed) | 4 |
| **survived** | **24** |
| **never reached by any test** | **2** |
| **mutation score** | **95.24%** |

## Triaging the 26 unkilled mutants

Thirteen were **equivalent or worthless** and are deliberately left alone:

- **5** blanked the message string of a `throw new MzLibException(...)`. The tests assert the
  exception type, not its prose. Correct as-is.
- **2** altered arithmetic inside `GetHashCode`. Both still produce a deterministic hash that is
  equal for equal formulas, so the contract is unbroken and no test *should* kill them.
- **4** flipped a floating-point boundary, `Math.Abs(a - b) > 1e-9` → `>=`, in
  `ChemicalFormula.Equals` and `PeriodicTable`'s validators. Killable only by a case whose
  difference is *exactly* 1e-9. Unkillable in practice.
- **2** looked real but are provably equivalent: in `IsotopicDistribution` the `if (size == 1)`
  branch runs only for mono-isotopic elements, where `probability ≡ 1.0` and
  `LogProbability ≡ log(1) = 0`. Multiplying and dividing by 1 agree; adding and subtracting 0
  agree.

The other **13 were real gaps**, and they share one shape: **almost every formula in the test
corpus carries a count of 1.** Where the code multiplies by a count, `* 1` and `/ 1` are the same
number, so no assertion can tell a correct implementation from a broken one.

## What each test pins

| # | test | mutant it kills | what would otherwise go unnoticed |
|---|---|---|---|
| 1 | `TestIsotopicDistributionAddsExplicitIsotopeMass` | `IsotopicDistribution.cs:120` (`+=`→`-=`, `*`→`/`) | The mass correction for explicitly-specified isotopes could have its **sign flipped** and no test would fail |
| 2 | `TestHillNotationPreservesIsotopeCounts` | `ChemicalFormula.cs:515, 527` (ternary forced true) | `H{2}2` would render as `H{2}`, silently **halving the deuterium** on a round-trip |
| 3 | `TestCountSensitivePropertiesOfFormulas` | `ChemicalFormula.cs:79` ×2, `:114`, `:223` | `AverageMass`, `NeutronCount` and `NumberOfUniqueElementsByAtomicNumber` never multiplied their per-entry count through |
| 4 | `TestAddPrincipalIsotopesOfChangesFormula` | `ChemicalFormula.cs:282` (statement deleted) | `AddPrincipalIsotopesOf` could be **gutted to a no-op** and the suite stays green |
| 5 | `TestAddZeroIsotopeCreatesNoEntry` | `ChemicalFormula.cs:308` (`return` deleted) | Adding zero of an isotope could start creating zero-count dictionary entries |
| 6 | `TestIsSupersetOfWithIsotopes` | `ChemicalFormula.cs:426` ×2 — **the 2 no-coverage mutants** | The isotope count comparison in `IsSupersetOf`/`IsSubsetOf` was never reached: existing cases exit on a missing isotope first |
| 7 | `TestGetClosestIndexTieResolvesUpward` | `ClassExtensions.cs:117` (`<`→`<=`) | An exactly-equidistant `ArraySearchOption.Closest` lookup had no defined winner |

Three of these deserve a closer look, because they are the ones where "well covered" was most
misleading:

**`TestIsotopicDistributionAddsExplicitIsotopeMass`.** The only test in the whole suite that put
an explicit-isotope formula through `GetDistribution` was:

```csharp
[Test]
public static void TestIsotopicDistribution2()
{
    IsotopicDistribution.GetDistribution(ChemicalFormula.ParseFormula("AlO{16}"));
}
```

It asserts nothing. It is a smoke test that raises the coverage number and catches nothing. The
new test uses `AlO{16}2` — aluminium is mono-isotopic, so the entire envelope is determined by
the two explicit oxygen-16 atoms — and asserts the most abundant mass equals the monoisotopic
mass. That kills both the sign flip and the count division.

**`TestHillNotationPreservesIsotopeCounts`.** Hill notation appends an isotope's count only when
it is not 1. The diagnostic detail: at `:515` and `:527` the `Conditional (false)` mutant was
**killed** while `Conditional (true)` **survived**, and the identical line for *carbon* at `:506`
was killed by every mutator. So the count==1 branch is tested, and carbon is tested — but no
non-carbon isotope with a count was ever round-tripped through Hill notation. (Formulas such as
`H{1}2` and `N{15}2` do appear in the suite, but those tests compare formulas by equality and
never read `.Formula` back.) The evidence that the gap was real is that `:515` and `:527`
survived. Given how mzLib is used for heavy labelling, that is the gap most likely to have
bitten someone.

**`TestAddZeroIsotopeCreatesNoEntry`.** `AddZeroIsotopeToFormula` already covers this line and
already passes. It asserts `Assert.AreEqual(formulaA, formulaB)` — and `ChemicalFormula.Equals`
compares *masses only*. A zero-count entry changes no mass, so the existing assertion is
structurally incapable of seeing one appear. The new test asserts `NumberOfUniqueIsotopes`
instead. This is the clearest example in the set of a test that looks like coverage and isn't.

## Verification

The same Stryker run was repeated on this branch. The mutant set is enumerated, not sampled, so
it is the identical 710 mutants in the identical places -- the only thing that changed is which of
them the suite can now detect.

| | master | this branch |
|---|---|---|
| mutants tested | 546 | 546 |
| killed | 516 | **527** |
| timeout (counts as killed) | 4 | 6 |
| survived | 24 | **13** |
| never reached by any test | 2 | **0** |
| **mutation score** | **95.24%** | **97.62%** |

**Every one of the 13 targeted mutants is now killed, and Stryker attributes each kill to the
specific test written for it.** Ten of the thirteen are killed by *only* that one test out of 6,015
-- which is the sharpest available statement that the gap was real and that the new test is what
closes it:

| mutant | killed by | total killers |
|---|---|---|
| `IsotopicDistribution.cs:120` `+=` → `-=` | `TestIsotopicDistributionAddsExplicitIsotopeMass` | 1 |
| `IsotopicDistribution.cs:120` `*` → `/` | `TestIsotopicDistributionAddsExplicitIsotopeMass` | 1 |
| `ChemicalFormula.cs:515` forced `true` | `TestHillNotationPreservesIsotopeCounts` | 1 |
| `ChemicalFormula.cs:527` forced `true` | `TestHillNotationPreservesIsotopeCounts` | 1 |
| `ChemicalFormula.cs:79` `AtomicMass / count` | `TestCountSensitivePropertiesOfFormulas` | 1 |
| `ChemicalFormula.cs:79` `AverageMass / count` | `TestCountSensitivePropertiesOfFormulas` | 1 |
| `ChemicalFormula.cs:223` `Neutrons / count` | `TestCountSensitivePropertiesOfFormulas` | 1 |
| `ChemicalFormula.cs:114` statement deleted | `TestCountSensitivePropertiesOfFormulas`, `TestAddZeroIsotopeCreatesNoEntry` | 2 |
| `ChemicalFormula.cs:282` statement deleted | `TestAddPrincipalIsotopesOfChangesFormula` | 1 |
| `ChemicalFormula.cs:308` `return` deleted | `TestAddZeroIsotopeCreatesNoEntry` | 1 |
| `ChemicalFormula.cs:426` `>` → `<` | `TestIsSupersetOfWithIsotopes` | 1 |
| `ChemicalFormula.cs:426` `>` → `>=` | `TestIsSupersetOfWithIsotopes` | 1 |
| `ClassExtensions.cs:117` `<` → `<=` | `TestGetClosestIndexTieResolvesUpward` | 1 |

The 13 mutants still alive are the 13 triaged above as equivalent, with two honest caveats:

- **One mutant I called equivalent was not.** `ChemicalFormula.cs:476`'s `AverageMass / count` in
  `GetHashCode` is now killed -- but by pre-existing tests (`Modification_read_write_into_proteinDb`,
  `MultiMod_ProteinDbWriter`), not by anything in this PR. Those tests round-trip protein databases
  and are sensitive to hash-dependent ordering, so this mutant's status flipped for reasons
  unrelated to these changes. My "equivalent by contract" call on it was wrong.
- **Timeout status is not deterministic.** `ChemicalFormula.cs:464`'s `Sum()` → `Max()` was a
  timeout (scored as killed) on master and survived here. Timeouts are wall-clock classifications,
  so a small amount of run-to-run jitter in the score is expected. It happens to offset the item
  above almost exactly.

Neither caveat touches the 13 kills this PR is actually responsible for.

Test suite: **6,015 tests, 0 failures** (`Category!=ExternalService`), up from 6,008.

- **No production code changes.** Every test pins existing behaviour.
- **No Stryker config, no CI job.** A full-project run is 51 minutes, of which 24 minutes is the
  baseline test run and coverage mapping before a single mutant is tested. If we want this in CI
  the shape is `dotnet stryker --since:master`, which mutates only the lines a PR touches, with
  `--break-at` as a floor — that is a separate discussion and a separate PR.
- **The 13 equivalent mutants are untouched.** Chasing them would mean contorting tests to assert
  exception prose and hash values. If we ever do adopt a mutation gate, the right move is
  `// Stryker disable once` comments at those four float-boundary lines and a config exclusion for
  the string mutator on `throw` arguments — again, separate.
- **`PeriodicTable.cs` was mutated but is mostly a data table** (3,037 of Chemistry's 4,544
  lines). Its 66 mutants are uninformative either way; a future config should exclude it.

## Reproducing this

```bash
dotnet tool install -g dotnet-stryker
cd mzLib/Chemistry
dotnet stryker -tp ../Test/Test.csproj -c 8 -r html
```

Cap concurrency well below the default. mzLib fixtures write to shared paths on disk (the
PSI-MOD database, for one), and parallel test runners racing in one sandbox produce *false
kills* — tests that fail from the race rather than the mutant — which silently inflates the
score.


nbollis (dismissed):

acesnik:
🤖 This is a nice use of mutation testing: rather than chasing a coverage percentage, it goes after specific surviving mutants in `ChemicalFormula` and asks what behaviour nobody had ever pinned. The answers are the interesting part — isotope counts surviving a Hill-notation round trip, the isotope half of `IsSupersetOf`, and the tie-breaking direction in `GetClosestIndex` were all genuinely unobserved, and now they aren't.

I checked all seven tests against `upstream/master` and worked through each named mutant. The kills hold, no production code is touched, and `mzLib.nuspec` ships no test assembly, so nothing here can reach MetaMorpheus or move a search result. Two doc comments are wrong about *why* the gaps existed, and three assertions can't fail — all cosmetic, none of it blocking.

**`TestIsSupersetOfWithIsotopes`** says the isotope half "was reached by no test at all: every existing superset and subset case was built from elements with unspecified isotopes." `IsSuperSetOf()` in the same file parses `CH3NO{17}C` against `CHNO{16}`, so the isotope loop is entered today — it just exits on `!Isotopes.ContainsKey(O{16})`, which short-circuits before `aa.Value > Isotopes[aa.Key]`. That is precisely why Stryker flagged the two `>` mutants and not the loop, so the story is even tidier than the comment suggests: what was unreached is the count comparison.

**`TestHillNotationPreservesIsotopeCounts`** says "every isotope carrying a count in the test corpus was a carbon isotope." `H{1}2` appears as `C2HH{1}2NO` in `AddNegativeIsotopeToFormula` (`ChemicalFormulaConstructionTests.cs:104`). The conclusion survives, because that test asserts formula equality and never reads `.Formula` back — so the gap is real, and "no isotope with a count was ever round-tripped through Hill notation" is both shorter and true.

Both are worth correcting because a comment outlives the PR body, and these two will be read as the reason the gap existed. That is the whole of why I am holding the second checkmark rather than approving and leaving a note - it needs a push either way, and a note on a merged PR tends to stay a note.

Three assertions that can't fail independently of the line above them, worth tightening in the same pass:

- `Assert.AreEqual("H{2}2O{18}3", formula.Formula)` fixes the string, so the `MonoisotopicMass` round trip on the next line reparses the literal the formula was built from.
- `GetClosestValue` is `sortedArray[sortedArray.GetClosestIndex(...)]`, so pinning index 3 already forces value 4.5.
- `masses.All(mass => mass > 0)` — the single mass is pinned exactly one line above, and `All` over an empty array is true regardless. `Assert.That(masses.Length, Is.EqualTo(1))` is the property doing the work. That comment also has the mechanism the other way round: the explicit `O{16}` atoms don't shape the envelope, they bypass the polynomial and get added as a constant (`IsotopicDistribution.cs:120`) — it's aluminium being mono-isotopic that makes it a single peak.

On method, so you know what to trust: I did not run Stryker, so the figures themselves (710/546/24, 95.24% → 97.62%) and the 6,015-test run are yours and CI's. `Test` is `net*.0-windows` through `mzPlot` and won't run here, so I exercised the tests' logic against `Chemistry` directly. What I did check is the falsifiable half — that each named mutant produces a value failing a specific assertion.

acesnik (commented):
🤖 This is a nice use of mutation testing: rather than chasing a coverage percentage, it goes after specific surviving mutants in `ChemicalFormula` and asks what behaviour nobody had ever pinned. The answers are the interesting part — isotope counts surviving a Hill-notation round trip, the isotope half of `IsSupersetOf`, and the tie-breaking direction in `GetClosestIndex` were all genuinely unobserved, and now they aren't.

I checked all seven tests against `upstream/master` and worked through each named mutant. The kills hold, no production code is touched, and `mzLib.nuspec` ships no test assembly, so nothing here can reach MetaMorpheus or move a search result. Two doc comments are wrong about *why* the gaps existed, and three assertions can't fail — all cosmetic, none of it blocking.

**`TestIsSupersetOfWithIsotopes`** says the isotope half "was reached by no test at all: every existing superset and subset case was built from elements with unspecified isotopes." `IsSuperSetOf()` in the same file parses `CH3NO{17}C` against `CHNO{16}`, so the isotope loop is entered today — it just exits on `!Isotopes.ContainsKey(O{16})`, which short-circuits before `aa.Value > Isotopes[aa.Key]`. That is precisely why Stryker flagged the two `>` mutants and not the loop, so the story is even tidier than the comment suggests: what was unreached is the count comparison.

**`TestHillNotationPreservesIsotopeCounts`** says "every isotope carrying a count in the test corpus was a carbon isotope." `H{1}2` appears as `C2HH{1}2NO` in `AddNegativeIsotopeToFormula` (`ChemicalFormulaConstructionTests.cs:104`). The conclusion survives, because that test asserts formula equality and never reads `.Formula` back — so the gap is real, and "no isotope with a count was ever round-tripped through Hill notation" is both shorter and true.

Both are worth correcting because a comment outlives the PR body, and these two will be read as the reason the gap existed. That is the whole of why I am holding the second checkmark rather than approving and leaving a note - it needs a push either way, and a note on a merged PR tends to stay a note.

Three assertions that can't fail independently of the line above them, worth tightening in the same pass:

- `Assert.AreEqual("H{2}2O{18}3", formula.Formula)` fixes the string, so the `MonoisotopicMass` round trip on the next line reparses the literal the formula was built from.
- `GetClosestValue` is `sortedArray[sortedArray.GetClosestIndex(...)]`, so pinning index 3 already forces value 4.5.
- `masses.All(mass => mass > 0)` — the single mass is pinned exactly one line above, and `All` over an empty array is true regardless. `Assert.That(masses.Length, Is.EqualTo(1))` is the property doing the work. That comment also has the mechanism the other way round: the explicit `O{16}` atoms don't shape the envelope, they bypass the polynomial and get added as a constant (`IsotopicDistribution.cs:120`) — it's aluminium being mono-isotopic that makes it a single peak.

On method, so you know what to trust: I did not run Stryker, so the figures themselves (710/546/24, 95.24% → 97.62%) and the 6,015-test run are yours and CI's. `Test` is `net*.0-windows` through `mzPlot` and won't run here, so I exercised the tests' logic against `Chemistry` directly. What I did check is the falsifiable half — that each named mutant produces a value failing a specific assertion.

trishorts:
Thanks — all five points checked out against the code and are addressed in c9cb007a:

- `TestIsSupersetOfWithIsotopes`: comment now says the *count comparison* was unreached, since `IsSuperSetOf` enters the loop and exits on the missing isotope first.
- `TestHillNotationPreservesIsotopeCounts`: comment now reads "no isotope with a count was ever round-tripped through Hill notation." The formula is also built via `Add`, so the round trip compares parse against construction instead of reparsing the pinned literal.
- `GetClosestValue` line dropped (implied by the pinned index).
- `All(mass > 0)` replaced with `Assert.That(masses.Length, Is.EqualTo(1))`, and the comment now describes the mechanism correctly: explicit isotopes are added as a constant, and Al is mono-isotopic.

The four tests pass locally.

acesnik (commented):
🤖 c9cb007a does what it says — I re-read all four files against it. The `IsSupersetOf` comment now names the count comparison rather than the loop, the Hill-notation formula is built through `Add` so the round trip compares the parse path against the construction path instead of reparsing its own literal, and `Assert.That(masses.Length, Is.EqualTo(1))` replaced the vacuous `All`. The corrected mechanism sentence matches the code: `IsotopicDistribution.cs:117-121` sums the explicit isotopes into `additionalMass` and `:125` adds it to every peak, so it is aluminium being mono-isotopic that makes this one peak. CI is green on `74b5c2e9` — `build`, `integration` and `external-service-tests`.

One edit landed half-applied. The Hill-notation comment now reads:

> Hill notation writes the count suffix for an isotope only when that count is not 1. Every isotope with a count was ever round-tripped through Hill notation.

The replacement sentence kept the old opening "Every", so it reads as the inverse of the sentence in your reply ("no isotope with a count was ever round-tripped"). One word, but it is the comment we were correcting, and it is what will be read as the reason the test exists.

The PR body still carries the superseded version of that same claim — "the test corpus contains only `C{13}6` and `C{12}2`. No hydrogen isotope and no other element's isotope ever appeared with a count above 1." On master there is `H{1}2` at `ChemicalFormulaConstructionTests.cs:104`, and `N{15}2`/`N{15}4` at `ModificationEqualityTests.cs:139-140` and `TestFlashLFQ.cs:1178`. It does not touch the conclusion — the evidence that the gap was real is that `:515` and `:527` survived, not the census of literals — and a body edit costs no dismissal.

Holding the checkmark only for the comment, since it needs a push either way. As before, I did not run Stryker, so the mutation figures remain yours and CI's.

trishorts:
@acesnik Thanks, both fixed.

- **Comment (28f65c5d):** the stray "Every" is gone. I also didn't use "no isotope" as-is, because `C2C{13}2H3NO` *is* round-tripped at `ChemicalFormulaPropertiesTests.cs:212`, which is why `:506` was already killed. The comment now reads: "No non-carbon isotope with a count was ever round-tripped through Hill notation."
- **PR body:** I replaced the census of literals with the same claim. It now notes that `H{1}2`/`N{15}2` do appear in the suite but are only compared by equality, and it rests the evidence on `:515`/`:527` surviving. Table row 6 had the old "reached by no test at all" wording too; it now names the count comparison.

PR #1234, opened 2026-08-27
Remove redundant and unreachable code in MzLibUtil, and pin what was untested

trishorts:
# Remove redundant and unreachable code in MzLibUtil, and pin what was untested

Two production changes and five tests, all of them consequences of a single mutation-testing run
over `MzLibUtil`. Two of the four findings turned out **not** to be missing tests at all — they
were redundant and unreachable code that no test could have distinguished from correct code. That
is the part of this PR I'd most like reviewed.

## How we got here

`MzLibUtil` has the **highest branch coverage of any hand-written project in mzLib: 97.7% line,
95.7% branch.** By coverage it is the best-tested code we have. That is exactly why it was worth
asking the question coverage cannot answer — not "did this line run?" but "would any test notice
if this line were wrong?"

[Stryker.NET][stryker] 4.16.0 answers that directly. It introduces one small defect at a time into
the production code — `*` becomes `/`, `>=` becomes `>`, `&&` becomes `||`, a statement is deleted
— then runs the tests. Killed means some test failed, so the defect is detectable. **Survived
means the entire 6,008-test suite passed with broken code in the tree.**

```
cd mzLib/MzLibUtil && dotnet stryker -tp ../Test/Test.csproj -c 8 -r html -r json
```

Twelve minutes, on master (`25f8788f`):

| | |
|---|---|
| tested | 503 |
| killed | 396 |
| timeout (counts as killed) | 1 |
| **survived** | **97** |
| **never reached by any test** | **9** |
| **mutation score** | **78.93%** |

For comparison, the same tool on `Chemistry` — which has **lower** branch coverage, 93.0% —
scored **95.24%**. The two metrics rank the two projects in opposite orders. That inversion is the
whole argument for having run this.

Worst files:

| file | killed | survived | nocov | score |
|---|---|---|---|---|
| `MrsNoiseEstimation/WaveletMathUtils.cs` | 4 | 8 | 6 | **22%** |
| `ObjectPools/DictionaryPool.cs` | 3 | 6 | 0 | **33%** |
| `ObjectPools/ListPool.cs` | 3 | 6 | 0 | **33%** |
| `MrsNoiseEstimation/WaveletFilter.cs` | 4 | 5 | 0 | 44% |
| `ObjectPools/HashSetPool.cs` | 4 | 4 | 0 | 50% |
| `MrsNoiseEstimation/MRSNoiseEstimator.cs` | 19 | 18 | 0 | 51% |
| `DoubleRange.cs` | 24 | 9 | 0 | 73% |
| **`Tolerance` / `PpmTolerance` / `PpmToleranceWithNotch`** | 48 | **0** | 0 | **100%** |

Worth stating plainly, because it was the opposite of what we predicted: **the tolerance family is
perfect.** Every `Within` boundary on the code every search depends on is genuinely pinned. We
went looking for a hole there and there isn't one.

## Change 1 — the pools cleared their collections twice (`ListPool`, `HashSetPool`, `DictionaryPool`)

**How we found it.** All three pools scored 33–50%, and the specific survivor was the statement
deletion of `list.Clear()`. That was surprising, because there is a test named exactly
`ListPool_Return_ClearsListBeforeReturningToPool`, it asserts the list comes back empty, and it
passes. A passing test, covering the line, unable to detect the line's removal, is the signature of
something wrong in the code rather than in the test — so we read it:

```csharp
public void Return(List<T> list)
{
    if (list == null) throw new ArgumentNullException(nameof(list));
    list.Clear();          // ← deleting this changes nothing observable
    _pool.Return(list);    // ← because this reaches the policy below...
}

public override bool Return(List<TItem> obj)
{
    obj.Clear();           // ← ...which clears it a second time
    return true;
}
```

The collection is cleared twice on every return. Neither call is individually necessary, which is
precisely why neither is individually testable — and it means the invariant was being maintained
by accident twice over rather than deliberately once.

**What changed.** Dropped the clear in the public `Return` wrapper and kept the policy's, since
`PooledObjectPolicy<T>.Return` is the API designed for exactly this reset. This is safe because
`DefaultObjectPool.Return` always invokes the policy's `Return` before deciding whether to retain
the instance (the `_isDefaultPolicy` fast path does not apply — these pools supply a custom
policy). The null guard stays where it was.

The change also **converts a permanently-unkillable mutant into a killable one**: with only one
clear left, deleting it now fails `..._Return_ClearsListBeforeReturningToPool`, a test that already
exists and already passes.

## Change 2 — `WaveletMathUtils.QMF` had an unreachable branch, and it was wrong

**How we found it.** Six *no-coverage* mutants — not survived, but never executed at all — clustered
in the `else` branch of `QMF`. Chasing the call graph explains why:

```
WaveletFilter.cs:21:  WaveletMathUtils.QMF(ScalingCoefficients, inverse: true);
```

That is the only call site in the library, `QMF` is `internal`, and it always passes `inverse: true`.
The default path (`inverse = false`) is unreachable from any caller.

**And the two branches are not equivalent.** The `inverse` loop multiplies element `i` by
`(-1)^(n-i)`; the dead loop multiplies it by `(-1)^(i+1)`. Those agree only when the filter length
`n` is odd. Both filters this class defines are even-length — Haar is 2, the commented-out db4 is 8 —
so for every filter mzLib actually has, **the dead branch produces the exact negation of the live
one.** Concretely, for Haar the live path yields `(+y₀, −y₁)` and the dead path `(−y₀, +y₁)`.

So this was not merely unused code. It was unused code that would have silently sign-flipped the
wavelet coefficients of the noise estimator had anyone called the default overload — which is the
easiest possible mistake to make, since it *is* the default.

**What changed.** Removed the `inverse` parameter and the dead branch; updated the single call
site. If the lab would rather keep the non-inverse path for future wavelet families, the
alternative is to keep it and add a two-line test that pins its output — say so and I will swap it.

## The three test gaps

**`DoubleRange` — inclusivity was never asserted.** `IsSuperRange`, `IsSubRange` and
`IsOverlapping` are each a conjunction of two comparisons, and all are documented "(inclusive)".
Both the boundary mutants (`>=` → `>`) and the logical mutants (`&&` → `||`) survived on all three,
which says two things: no existing case put two ranges on a shared endpoint, so the documented
inclusivity was unverified; and no case had exactly one conjunct false, so the two halves were
never independently exercised. `TestRangeComparisonsAreInclusiveAtSharedEndpoints` covers both —
four shared-endpoint cases and six one-conjunct-false cases.

**The pools never asserted that they pool.** `return true` → `false` survived in all three
policies. Returning `false` means the instance is discarded rather than retained, and `Get()` then
allocates a fresh one — observationally identical unless a test looks at instance identity. Three
tests now assert `Is.SameAs` across get → return → get.

**`MRSNoiseEstimator` had no direct test.** Its only caller is `SpectralWeighting`, so every mutant
in the wavelet path could only ever be caught, if at all, through a spectral-averaging assertion
several layers up — and 18 of them weren't. `TestMRSNoiseEstimator` adds two characterization tests
on a fixed, deterministic input (an explicit formula, no RNG, so the pinned value is stable across
platforms and runtimes):

- the estimate on a 512-point signal is pinned to the value master computes today, `6.8650750184874045`;
- tripling the signal triples the estimate.

The second one is worth a note. It was written as a probe — the convergence test inside the
estimator compares against an *absolute* epsilon, which could easily have made the result
scale-dependent. It passed, so the estimator is linear in amplitude, and that is now pinned rather
than assumed.

## Verification

The same Stryker run was repeated on this branch.

| | master | this branch |
|---|---|---|
| mutants generated | 769 | 755 |
| mutants tested | 503 | 492 |
| killed | 396 | **431** |
| timeout (counts as killed) | 1 | 1 |
| survived | 97 | **57** |
| never reached by any test | 9 | **3** |
| **mutation score** | **78.93%** | **87.80%** |

One honest caveat on reading that table: unlike a tests-only change, **the mutant set is not
identical between the two runs.** This PR deletes production code, so there are 14 fewer mutants to
generate -- the unreachable `QMF` branch accounted for six of the nine no-coverage mutants, and the
three removed `Clear()` calls for three more. Deleting code you cannot test does raise the score
without proving anything about the tests. The per-file numbers below separate the two effects.

| file | before | after |
|---|---|---|
| `DoubleRange.cs` | 73% (9 alive) | **97%** (1 alive) |
| `ObjectPools/HashSetPool.cs` | 50% (4 alive) | **100%** (0 alive) |
| `ObjectPools/ListPool.cs` | 33% (6 alive) | **75%** (2 alive) |
| `ObjectPools/DictionaryPool.cs` | 33% (6 alive) | **75%** (2 alive) |
| `MrsNoiseEstimation/MRSNoiseEstimator.cs` | 51% (18 alive) | **76%** (9 alive) |
| `MrsNoiseEstimation/WaveletMath.cs` | 70% (13 alive) | **89%** (5 alive) |
| `MrsNoiseEstimation/WaveletFilter.cs` | 44% (5 alive) | **62%** (3 alive) |
| `MrsNoiseEstimation/WaveletMathUtils.cs` | 22% (14 alive) | **36%** (7 alive) |
| everything else | unchanged | unchanged |

**Attribution.** Stryker records which tests killed each mutant, so the credit can be checked rather
than assumed. **103 mutants are killed with one of the five new tests among their killers, and 30 of
those have a new test as the *only* killer** out of 6,014.

The two `MRSNoiseEstimator` characterization tests are doing most of that work -- 23 mutants in
`MRSNoiseEstimator` and 38 in `WaveletMath` now have a new test among their killers, 9 and 8 of them
exclusively. That is the payoff for pinning a numeric pipeline that previously had no direct test at
all: a single fixed input propagates detection through every arithmetic step behind it.

The remaining two survivors in `ListPool` and `DictionaryPool` are both on
`MaximumRetained = Environment.ProcessorCount * 2` -- pool sizing, which is a performance knob with
no observable behaviour to assert. Those are equivalent for practical purposes and deliberately left.

Test suite: **6,014 tests, 0 failures**. The filtered run over `Test.Util` plus the spectral
averaging fixtures -- the latter deliberately, since `SpectralWeighting` is the only consumer of the
wavelet code touched here -- is **205 passed, 0 failed**.

- **No behaviour changes.** Both production edits are provably behaviour-preserving for every
  existing caller: one removes a duplicated operation, the other removes code nothing calls.
- **No Stryker config or CI job.** Repo-wide mutation testing is not worth its runtime. If we ever
  gate on it, the shape is `dotnet stryker --since:master`, which mutates only the lines a PR
  touches.
- **`MrsNoiseEstimation` is still the weakest area** (`MRSNoiseEstimator` 51%, `WaveletMath` 70%).
  The two characterization tests raise the floor but do not close it. Genuinely closing it means
  understanding the Starck & Murtagh algorithm well enough to write meaningful unit tests for
  `BooleanizeLevels` and `CreateMultiResolutionSupport`, which is a larger piece of work and its
  own PR.
- **`PositionFrequencyAnalysis/QuantifiedProtein.cs` (84%, 10 survivors)** was left alone; the
  survivors there are mostly string and statement mutants in reporting paths.

## Reproducing this

```bash
dotnet tool install -g dotnet-stryker
cd mzLib/MzLibUtil
dotnet stryker -tp ../Test/Test.csproj -c 8 -r html
```

Cap concurrency well below the default. mzLib fixtures write to shared paths on disk, and parallel
test runners racing in one sandbox produce false kills that silently inflate the score.


nbollis (approved):

acesnik:
🤖 Two deletions that have been waiting to happen: a `QMF` overload whose `inverse: false` branch nothing ever called, and a redundant second clear in the object pools. Both are the kind of thing that's easy to leave alone forever because proving them dead takes more work than reading them — so it's good to see it actually done, with characterization tests pinning the noise estimator on the way through.

I verified both independently, and both are safe.

**No API surface change.** I dumped every member of `MzLibUtil.dll` at the merge base and at the PR head. The two dumps differ by exactly one line — `internal static double[] QMF(double[], bool)` becoming `QMF(double[])` — with all 229 public and protected entries identical. `MzLibUtil` declares no `InternalsVisibleTo` anywhere, so `QMF` isn't reachable from `Test` let alone from a package consumer. Nothing for release notes to warn about, which is the question I'd have worried about most on a published library.

**`QMF` is behaviour-preserving bit for bit.** 2,400 randomised differential cases across lengths 1–12 comparing the new single-branch version against master's `inverse: true` path: zero bit-level mismatches. Haar `WaveletCoefficients` is `(0.49999999999999994, -0.49999999999999994)` before and after, and the sign analysis in the body checks out exactly — `live == -dead` for even n, `live == dead` for odd.

**The double-clear removal holds where it could have failed.** The claim rests on `DefaultObjectPool.Return` invoking the policy before deciding whether to retain, so I went at the overflow path specifically. With `MaximumRetained` 20, returning 25 collections to each pool: 0 of 25 lists, 0 of 25 dictionaries, 0 of 25 hash sets left non-empty. No caller anywhere reads a collection after returning it either — all six sites in mzLib and three in MetaMorpheus return in a `finally` or at end of scope.

**One correction, because release notes get written from PR bodies.** The description says the dead branch "would have silently sign-flipped the wavelet coefficients of the noise estimator." I wired that branch back into the live call site to see how bad it was, and `MRSNoiseEstimation` returned **6.8650750184874045 — bit-identical, difference 0.000E+00**. `BooleanizeLevel` compares `Math.Abs(level.WaveletCoeff[i])` and `CreateSmoothedSignal` reads `ScalingCoeff`, so a uniform sign flip of the wavelet filter is unobservable at the estimator's output. Both of the new characterization tests passed with the wrong branch in place. So the deletion is *safer* than the body claims — but as written it reads like a latent numerical bug was averted, and there wasn't one.

That gap is worth two lines of insurance, since `WaveletFilter.WaveletCoefficients` is public and is the only thing the retained branch determines:

```csharp
var filter = new WaveletFilter();
filter.CreateFiltersFromCoeffs(WaveletType.Haar);
Assert.That(filter.WaveletCoefficients, Is.EqualTo(new[] { 0.5, -0.5 }).Within(1e-15));
```

**One thing I can add that CI structurally can't.** All three jobs in `dotnet.yml` are `windows-latest`. On Apple Silicon, macOS 26.6.2, .NET 10.0.302, `MRSNoiseEstimation(BuildSignal(512), 0.01)` returns 6.8650750184874045 exactly — not merely within `1e-6`. It's not perched on a convergence boundary either: identical to all digits at epsilon 0.02, 0.01, 0.005 and 0.001, and a 1-ulp perturbation of every input element moves the result by 8.9e-16. So the pinned constant has roughly nine orders of magnitude of headroom on a second architecture.

Two nits: `ListPool<T>`'s XML docs still say "HashSet" in three places (`Get()`'s `<returns>`, `Return`'s `<param>`, the constructor's `<param>`), and since #1162 those ship to consumers in `MzLibUtil.xml` — this PR is already editing two of those lines. And `TestRangeComparisonsAreInclusiveAtSharedEndpoints` uses `Assert.IsTrue`/`IsFalse` via the file-level `ClassicAssert` alias while its immediate neighbours fully qualify `Assert.That`.

Not verified: any Stryker figure, the coverage percentages, or the full 6,014-test run — `Test.csproj` is `net10.0-windows` through `mzPlot`, so I ran only the `MzLibUtil`-scoped tests plus my own harness. Also `Is.SameAs` is an `ObjectPool<T>` implementation detail rather than a contract; it holds on Microsoft.Extensions.ObjectPool 9.0.0, so those three tests may want a look at the next package bump.

acesnik (approved):

PR #1235, opened 2026-08-27
Cover the unambiguous gaps mutation testing found in Deconvolution

trishorts:
# Cover the unambiguous gaps mutation testing found in Deconvolution

Four tests, no production changes. This is the small, high-confidence slice of a much larger
finding — the rest is recorded in #1236 rather than fixed here, deliberately.

## Where this came from

Mutation testing ([Stryker.NET][stryker] 4.16.0) over `MassSpectrometry/Deconvolution`:

```
cd mzLib/MassSpectrometry
dotnet stryker -tp ../Test/Test.csproj -m "**/Deconvolution/**" -c 8 -r html -r json
```

| | master | this branch |
|---|---|---|
| mutants tested | 909 | 909 |
| killed | 657 | **667** |
| timeout (counts as killed) | 9 | 9 |
| survived | 217 | **211** |
| never reached by any test | 26 | **22** |
| **mutation score** | **73.27%** | **74.37%** |

That is the lowest of the three mzLib projects measured this way — `Chemistry` scored 95.24% and
`MzLibUtil` 78.93% — against a project-level branch coverage of 90.6%.

**Only a handful of those 243 sites are addressed here.** For most of the rest, the code does not
make clear whether `<` or `<=` is the intended behaviour, and writing a test that pins today's value
without knowing the intent would cement whatever is currently there. Those belong in review by
someone who knows the algorithms, which is what the issue is for.

### 1. `ClassicDeconvolutionRejectsChargesOutsideBounds` (renamed from `ClassicDeconvolutionChargeStateBoundsAreInclusive`)

The candidate-charge filter appears **twice**, on consecutive branches:

```csharp
charge = (int)Math.Floor(1 / deltaMass);
if (charge >= MinAssumedChargeState && charge <= MaxAssumedChargeState)   // line 92
    allPossibleChargeStates.Add(charge);

// get the upper bound charge state
charge++;
if (charge >= MinAssumedChargeState && charge <= MaxAssumedChargeState)   // line 99
    allPossibleChargeStates.Add(charge);
```

The test deconvolutes a scan known to contain a charge-2 envelope with `min == max == 2`, and
asserts the envelope is still found and every returned envelope has charge 2.

**What it pins:** the filter cannot be *widened* at either line — swapping `&&` for `||` at line 92
or 99, or loosening either comparison in the wrong direction, now fails. Line 99 goes from 4/6
mutants killed to **6/6**.

**What it does not pin, and this is worth a reviewer's attention:** two mutants at line 92 still
survive — `charge >= Min` → `charge > Min`, and `charge <= Max` → `charge < Max`. **Narrowing line
92 alone is undetectable, because line 99 is a redundant route to the same candidate charge.** When
`floor(1 / deltaMass)` lands one below the true charge, the `charge++` branch admits it anyway, so
line 92's bounds are not load-bearing for that peak.

That is a *code* observation rather than a test gap, and the same shape as the duplicated
`Clear()` found in `MzLibUtil` (#1234): an invariant maintained twice over is an invariant no single
test can hold in place. Killing those two would need a synthetic spectrum whose spacing makes line
92 the only route to the correct charge. I did not write that test, because whether the duplication
is intentional is a question for someone who knows the algorithm.

My earlier draft of this description claimed the test proved inclusivity at both bounds. It does
not, and the verification run is what caught it. Following review, the test is renamed to say what it
does pin.

### 1b. `ClassicDeconvolutionUpperChargeCandidateBoundsAreInclusive` (added in review)

On a clean envelope the isotope spacing is slightly *wider* than 1/z, so `floor(1 / deltaMass)` is
z − 1 and only line 99 proposes z. A single synthetic charge-2 envelope with `min == max == 2`
therefore makes line 99 the only route, and both narrowing mutants there fail this test (checked
by applying each mutant by hand). Line 92's two narrowing mutants still survive: line 92 proposes z
only for a peak spacing at or below 1/z (for example an isotope of a charge z + 1 envelope), and in
any realistic envelope another peak pair lets line 99 propose z as well. Pinning line 92 would need a test seam on the candidate
set, noted on #1236.

### 2–4. Three tests over code no test executed at all

These are `NoCoverage` sites, not merely survivors — nothing in 6,008 tests ever ran these lines.
All are now fully killed, each attributed by Stryker to the specific new test:

- `DeconvolutionScorer.cs:102`, `:103` — null guards on the envelope-only `ComputeFeatures`
  overload. The equivalent guards on the three-argument overload were already covered; these were
  not.
- `DeconvolutionScorer.cs:472` — a zero-intensity envelope peak must be *skipped*, not contribute a
  zero to the per-peak signal-to-noise median. The test scores two envelopes against one identical
  spectrum, differing only by a dead peak in the envelope's own peak list, and asserts the feature
  is unchanged.
- `DeconvolutionScorer.cs:488` — a window whose non-envelope points are all zero has a median noise
  of exactly zero and must be skipped, not divided by. Without the guard the feature becomes
  infinity and propagates into the score. The test asserts the result stays finite.

Both degenerate paths are reachable from real spectra; they simply were not reachable from any test.

- **No production changes.** Every assertion pins behaviour master already has.
- **211 survivors are untouched**, including the 79 unpinned equality boundaries. They are
  enumerated in #1236 with file, line, original expression and mutation, so they can be reviewed by
  hand.
- **No Stryker config or CI job.** A scoped run is 16 minutes, most of it fixed setup cost.


nbollis (approved):

acesnik:
🤖 Going after the specific mutants that survived in `DeconvolutionScorer` is exactly the right way to spend coverage effort, and the two degenerate signal-to-noise paths this pins are ones I'd never have thought to write a test for — a window whose noise samples are all zero, and an envelope peak with zero intensity. Both now have characterization, and I verified the kills at `:472`, `:488` and `:102` all hold. It's genuinely test-only too, so no deconvolution output moves, which given the hot path was the first thing I checked.

I'm holding the second checkmark on one thing, and I think it's worth a conversation rather than a tweak: `ClassicDeconvolutionChargeStateBoundsAreInclusive` is named for a property I don't think it can pin.

The comment says "pinning min == max == 2 means the envelope survives only if both comparisons are inclusive." Look at the pair of guards it's aiming at:

```csharp
charge = (int)Math.Floor(1 / deltaMass);
if (charge >= Min && charge <= Max) { allPossibleChargeStates.Add(charge); }   // :92
charge++;
if (charge >= Min && charge <= Max) { allPossibleChargeStates.Add(charge); }   // :99
```

The two sites add *different* charges — `charge` and `charge + 1`. So charge 2 reaches `allPossibleChargeStates` through `:92` when the floor charge is 2, **or** through `:99` when the floor charge is 1. Narrowing either guard on its own leaves the other as a redundant path, and mutation testing mutates one site at a time — so a narrowed `>` / `<` at either line still yields envelopes and the test still passes.

If that reading is right, it's structural rather than a matter of picking a better scan: on any spectrum where both paths can produce the target charge, inclusivity at these sites isn't observable through `Deconvolute` under a single mutation. What the test *does* pin — that the filter can't be **widened** at either end — is real and worth having. It just isn't what the name promises.

Two ways out and I'm happy with either: rename to something like `ClassicDeconvolutionRejectsChargesOutsideBounds` and trim the comment to the widening claim, or assert on `allPossibleChargeStates` directly if inclusivity is the property you want pinned. My reason for not just approving and leaving a note is that a test name is what the next person reads when they ask "is this branch covered?" — and a name that says yes is the thing that stops them writing the test that would.

Smaller, same file: `LocalSignalToNoise_IgnoresZeroIntensityEnvelopePeaks` depends on the companion peak at `deadMz + 0.5` landing inside the SNR window. Drop `SnrWindowDa` from 2.0 to 0.4 and the test passes *with the `:472` guard removed* — the dead peak's window then holds no noise sample at all, so `:488` skips it and the zero never reaches the median. `SnrWindowDa` is a live tuning constant, so that's reachable, and it fails in the quiet direction. A live twin of the dead peak makes it loud instead:

```csharp
var livePeaks = baseline.Peaks.ToList();
livePeaks.Add((deadMz, 1.0));           // same peak, but alive
var withLivePeak = new IsotopicEnvelope(livePeaks, baseline.MonoisotopicMass,
    baseline.Charge, baseline.TotalIntensity, 0.0);
Assert.That(DeconvolutionScorer.ComputeFeatures(withLivePeak, Model, spectrum).LocalSignalToNoise,
    Is.Not.EqualTo(baselineSnr).Within(1e-9),
    "the extra peak's window must be able to yield an SNR, else the intensity guard is untested");
```

`LocalSignalToNoise_StaysFiniteWhenWindowNoiseIsAllZero` isn't exposed the same way — at `SnrWindowDa = 0.25` with `:488` removed it still fails, because each peak's window picks up the neighbouring peaks' dead-noise points.

One thing worth a second look independent of this PR: `SquashSnr(0.0) == 0.0` against `CoefficientSnr = 2.0`, so a window with no usable noise sample forfeits the full +2.0 logit term and scores like a noise-dominated envelope. Same "is this the intent" category as the 79 boundaries you deferred to #1236 — fine as documentation, less settled as behaviour.

On method, since I'm withholding on it: the redundant-path mechanism above is my reading of the two guards, not a Stryker run, and I haven't reproduced the mutant survivals myself — `Test` is `net*.0-windows` through `mzPlot`, so I couldn't run the suite. If the narrowing mutants are in fact killed at one of those lines then I've got this wrong, and I'd rather you tell me than work around me. Also unverified: the Stryker totals (909 / 667 / 211 / 22, 73.27% → 74.37%), the 6,008-test figure, and whether either degenerate SNR path is reachable from real spectra — I only reached them synthetically, and centroided mzML typically carries no zero-intensity points.

acesnik (commented):
🤖 Going after the specific mutants that survived in `DeconvolutionScorer` is exactly the right way to spend coverage effort, and the two degenerate signal-to-noise paths this pins are ones I'd never have thought to write a test for — a window whose noise samples are all zero, and an envelope peak with zero intensity. Both now have characterization, and I verified the kills at the `envIntensity <= 0.0` guard, the `noiseIntensities.Count == 0` guard and the `medianNoise <= 0.0` guard all hold. It's genuinely test-only too, so no deconvolution output moves, which given the hot path was the first thing I checked.

I'm holding the second checkmark on one thing, and I think it's worth a conversation rather than a tweak: `ClassicDeconvolutionChargeStateBoundsAreInclusive` is named for a property I don't think it can pin.

The comment says "pinning min == max == 2 means the envelope survives only if both comparisons are inclusive." Look at the pair of guards it's aiming at:

```csharp
charge = (int)Math.Floor(1 / deltaMass);
if (charge >= Min && charge <= Max) { allPossibleChargeStates.Add(charge); }   // :92
charge++;
if (charge >= Min && charge <= Max) { allPossibleChargeStates.Add(charge); }   // :99
```

The two sites add *different* charges — `charge` and `charge + 1`. So charge 2 reaches `allPossibleChargeStates` through `ClassicDeconvolutionAlgorithm.cs:94` when the floor charge is 2, **or** through `:101` when the floor charge is 1. Narrowing either guard on its own leaves the other as a redundant path, and mutation testing mutates one site at a time — so a narrowed `>` / `<` at either line still yields envelopes and the test still passes.

If that reading is right, it's structural rather than a matter of picking a better scan: on any spectrum where both paths can produce the target charge, inclusivity at these sites isn't observable through `Deconvolute` under a single mutation. What the test *does* pin — that the filter can't be **widened** at either end — is real and worth having. It just isn't what the name promises.

Two ways out and I'm happy with either: rename to something like `ClassicDeconvolutionRejectsChargesOutsideBounds` and trim the comment to the widening claim, or assert on `allPossibleChargeStates` directly if inclusivity is the property you want pinned. My reason for not just approving and leaving a note is that a test name is what the next person reads when they ask "is this branch covered?" — and a name that says yes is the thing that stops them writing the test that would.

Smaller, same file: `LocalSignalToNoise_IgnoresZeroIntensityEnvelopePeaks` depends on the companion peak at `deadMz + 0.5` landing inside the SNR window. Drop `SnrWindowDa` from 2.0 to 0.4 and the test passes *with the `envIntensity <= 0.0` guard removed* — the dead peak's window then holds no noise sample at all, so the `noiseIntensities.Count == 0` guard skips it and the zero never reaches the median. `SnrWindowDa` is a live tuning constant, so that's reachable, and it fails in the quiet direction. A live twin of the dead peak makes it loud instead:

```csharp
var livePeaks = baseline.Peaks.ToList();
livePeaks.Add((deadMz, 1.0));           // same peak, but alive
var withLivePeak = new IsotopicEnvelope(livePeaks, baseline.MonoisotopicMass,
    baseline.Charge, baseline.TotalIntensity, 0.0);
Assert.That(DeconvolutionScorer.ComputeFeatures(withLivePeak, Model, spectrum).LocalSignalToNoise,
    Is.Not.EqualTo(baselineSnr).Within(1e-9),
    "the extra peak's window must be able to yield an SNR, else the intensity guard is untested");
```

`LocalSignalToNoise_StaysFiniteWhenWindowNoiseIsAllZero` isn't exposed the same way — at `SnrWindowDa = 0.25` with the `noiseIntensities.Count == 0` guard removed it still fails, because each peak's window picks up the neighbouring peaks' dead-noise points.

One thing worth a second look independent of this PR: `SquashSnr(0.0) == 0.0` against `CoefficientSnr = 2.0`, so a window with no usable noise sample forfeits the full +2.0 logit term and scores like a noise-dominated envelope. Same "is this the intent" category as the 79 boundaries you deferred to #1236 — fine as documentation, less settled as behaviour.

On method, since I'm withholding on it: the redundant-path mechanism above is my reading of the two guards, not a Stryker run, and I haven't reproduced the mutant survivals myself — `Test` is `net*.0-windows` through `mzPlot`, so I couldn't run the suite. If the narrowing mutants are in fact killed at one of those lines then I've got this wrong, and I'd rather you tell me than work around me. Also unverified: the Stryker totals (909 / 667 / 211 / 22, 73.27% → 74.37%), the 6,008-test figure, and whether either degenerate SNR path is reachable from real spectra — I only reached them synthetically, and centroided mzML typically carries no zero-intensity points.

nbollis (dismissed):

trishorts:
@acesnik You read it right. I applied each mutant by hand and ran the test. `>`/`<` at `:92` alone and at `:99` alone all pass `ClassicDeconvolutionChargeStateBoundsAreInclusive`; only narrowing both lines at once fails it. Changes in 294dfa7e:

- **Renamed** to `ClassicDeconvolutionRejectsChargesOutsideBounds`, and the comment now claims only widening. Checked: `&&`→`||` and dropping either comparison fails it at both `:92` and `:99`.
- **Added `ClassicDeconvolutionUpperChargeCandidateBoundsAreInclusive`.** A clean envelope's spacing is just over 1/z, so `floor(1/deltaMass)` gives z−1 and only `:99` proposes z. With a single synthetic charge-2 envelope and min = max = 2, both narrowing mutants at `:99` fail it. The two at `:92` still survive, as your reading predicts. That's now written on #1236 (which wrongly claimed `:92` was covered) along with the seam it would take to pin them.
- **`LocalSignalToNoise_IgnoresZeroIntensityEnvelopePeaks`:** reproduced your hole. With `SnrWindowDa = 0.4` and the `envIntensity` guard removed, the original test passed. Added your live-twin assertion; that configuration now fails, as does `SnrWindowDa = 0.4` by itself. Also confirmed your point about `LocalSignalToNoise_StaysFiniteWhenWindowNoiseIsAllZero`: at 0.25 with the count guard removed, it still fails.
- **`SquashSnr(0.0)` forfeiting the +2.0 term:** agreed it's a behaviour question, not a boundary. I added it to #1236 rather than pinning it here.

The PR description is updated to match. All 26 `ClassicDeconvolution*`/`LocalSignalToNoise_*` tests pass locally.

PR #1238, opened 2026-08-27
MGF has default intensity

nbollis:
I came across an mgf file in the wild which did not annotate the precursor intensity, leading to it being unsearchable in MetaMorpheus. This PR fixes that by introducing a default intensity of 1 whenver the mgf does not contain the intensity.

### DIFF
--- a/mzLib/Readers/Mgf/Mgf.cs
+++ b/mzLib/Readers/Mgf/Mgf.cs
@@ -171,10 +171,11 @@ namespace Readers
                 }
                 else if (line.StartsWith("PEPMASS"))
                 {
-                    sArray = sArray[1].Split(' ');
+                    sArray = sArray[1].Split(' ', StringSplitOptions.RemoveEmptyEntries);
                     precursorMz = Convert.ToDouble(sArray[0], CultureInfo.InvariantCulture);
-                    if (sArray.Length > 1)
-                        precursorIntensity = Convert.ToDouble(sArray[1], CultureInfo.InvariantCulture);
+                    precursorIntensity = sArray.Length > 1
+                        ? Convert.ToDouble(sArray[1], CultureInfo.InvariantCulture)
+                        : 1;
                     sawPrecursorMz = true;
                 }
                 else if (line.StartsWith("MSLEVEL"))
--- a/mzLib/Test/FileReadingTests/SpectraFileReading/TestMgf.cs
+++ b/mzLib/Test/FileReadingTests/SpectraFileReading/TestMgf.cs
@@ -171,6 +171,28 @@ namespace Test.FileReadingTests.SpectraFileReading
             NUnit.Framework.Assert.That(scans[0].SelectedIonIntensity, Is.EqualTo(47641904.0).Within(0.00001));
         }
 
+        [Test]
+        public void DefaultsBlankPrecursorIntensityToOne()
+        {
+            string path = Path.Combine(TestContext.CurrentContext.WorkDirectory, "blankPepmassIntensity.mgf");
+            File.WriteAllLines(path, new[]
+            {
+                "BEGIN IONS",
+                "PEPMASS=678.5586751 ",
+                "CHARGE=2+",
+                "RTINSECONDS=215.437",
+                "110.030752706439 39.64286",
+                "END IONS"
+            });
+
+            var reader = MsDataFileReader.GetDataFile(path);
+            NUnit.Framework.Assert.DoesNotThrow(() => reader.LoadAllStaticData());
+
+            var scan = reader.GetOneBasedScan(1);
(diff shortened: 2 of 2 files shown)


mzLib/Readers/Mgf/Mgf.cs
```diff
@@ -171,10 +171,11 @@ public static MsDataFile LoadAllStaticData(string filePath, FilteringParams filt
...
                 else if (line.StartsWith("PEPMASS"))
                 {
-                    sArray = sArray[1].Split(' ');
+                    sArray = sArray[1].Split(' ', StringSplitOptions.RemoveEmptyEntries);
                     precursorMz = Convert.ToDouble(sArray[0], CultureInfo.InvariantCulture);
-                    if (sArray.Length > 1)
-                        precursorIntensity = Convert.ToDouble(sArray[1], CultureInfo.InvariantCulture);
+                    precursorIntensity = sArray.Length > 1
+                        ? Convert.ToDouble(sArray[1], CultureInfo.InvariantCulture)
+                        : 1;
```
trishorts, line 178: Defaulting to `1` discards the distinction the declaration eight lines up is deliberately making:

```csharp
double? precursorIntensity = null; //default when unknown
```

After this change `null` is unreachable, so "the mgf didn't record an intensity" and "the intensity is 1" become the same value. That leaks: `MzmlMethods.cs:559` writes `MS:1000042` only `if (scanWithPrecursor.SelectedIonIntensity.HasValue)`, so an mgf→mzML conversion now stamps a fabricated selected-ion intensity of 1 on every scan, and the mzML no longer records that the value was unknown.

Note this fires for **every** mgf without the optional intensity token — the common case, not just the malformed file — since the token is optional in the Matrix Science spec.

If MetaMorpheus needs a number, it already picks one (`?? 1.0`, `MetaMorpheusTask.cs:308`), and that choice stays local to the consumer that needs it. Suggest keeping the reader honest and dropping to:

```csharp
if (sArray.Length > 1)
    precursorIntensity = Convert.ToDouble(sArray[1], CultureInfo.InvariantCulture);
```

mzLib/Readers/Mgf/Mgf.cs
```diff
@@ -171,10 +171,11 @@ public static MsDataFile LoadAllStaticData(string filePath, FilteringParams filt
                 }
                 else if (line.StartsWith("PEPMASS"))
                 {
-                    sArray = sArray[1].Split(' ');
+                    sArray = sArray[1].Split(' ', StringSplitOptions.RemoveEmptyEntries);
```
trishorts, line 174: This is the line that actually fixes the reported bug, and it's correct.

It does overlap #1184, which hardens this same statement (and `MSLEVEL`, `CHARGE`, `SCANS`, `RTINSECONDS`, and the peak lines) with `TryParse` instead. `RemoveEmptyEntries` handles the trailing space; `TryParse` additionally survives `PEPMASS=` with a non-numeric or empty value. Same hunk in both PRs, so one will need a rebase — worth deciding with @acesnik which one carries the fix.

trishorts (changes requested):
I ran the motivating case on master, on this branch, and on #1184. The crash is real, but the default of `1` is doing something different from what the description says, and the crash is already fixed elsewhere.

| `PEPMASS` form | master | #1184 | this PR |
|---|---|---|---|
| `PEPMASS=678.55` (bare) | `null` | `null` | **`1`** |
| `PEPMASS=678.55 ` (trailing space) | **`FormatException`** | `null` | `1` |

Two things follow.

**1. A missing intensity was never what made the file unsearchable.** A bare `PEPMASS` with no intensity token already read fine on master with `SelectedIonIntensity = null`, and MetaMorpheus already supplies exactly this default itself — `MetaMorpheusTask.cs:308`:

```csharp
double precursorIntensity = ms2scan.SelectedIonMonoisotopicGuessIntensity ?? ms2scan.SelectedIonIntensity ?? 1.0;
```

What actually broke your file is the trailing space: `"678.5586751 ".Split(' ')` yields `["678.5586751", ""]`, so the old `sArray.Length > 1` guard passed and `Convert.ToDouble("")` threw. `RemoveEmptyEntries` fixes that — the default doesn't contribute.

**2. #1184 already fixes it, and is the more thorough fix.** acesnik's open PR replaces `Convert.ToDouble` with `TryParse` across *every* header, so the trailing-space file reads with `SelectedIonIntensity` left `null`. Both PRs edit the same hunk, so they will conflict. Worth coordinating with @acesnik: if #1184 lands, all this PR still adds is the `1`.

Happy to be wrong if you have a file where the *absence* of the token (not the trailing space) is what fails — that would change the analysis.

nbollis:
Closing until anthony's have been merged so I can address whatever issue remain with the .mgfs I have

PR #1240, opened 2026-08-28
Make the quantification table rectangular, so values stay under their headers

trishorts:
A protein group whose samples were all unobserved emitted two columns per sample group where its neighbour emitted four. The header is written once, from `proteinGroups.First()`, while every row writes its own `ToString()` — so every value after the disagreement landed under the wrong column name. No exception, no warning: a file that looks fine and is wrong.

Measured on the delivery fixture: a group observed in one file of two came out **39 fields against a 45-field header**; a group the engine quantified but found nothing for came out **33**.

## Cause

`SampleGroupResult.HasIntensityData` answers a per-group question — did *this* sample group get a value — and both `GetTabSeparatedHeader` and `ToString` were using it to decide whether a column *exists*. Whether the columns exist has to be a property of the entity, or a row cannot line up with a header written from a different row.

The gate is now `HasAssignedSampleIntensities` — a non-empty `SamplesForQuantification` **and** an assigned `IntensitiesBySample`. `QuantificationEngine.OverwriteSampleIntensities` sets both for every row of the matrix, and `GetUniquePeptideToProteinMap` gives every protein group a row, so all groups of one quantified run agree.

Both fields are required. `ConstructSubsetBioPolymerGroup` assigns an empty-but-non-null dictionary beside an empty sample list when no sample matches the requested file — the normal case for a per-file subset, since the engine omits zero cells — and gating on the dictionary alone would advertise intensity columns that `PopulateSampleGroupResults` can never fill. An earlier revision of this PR did exactly that; there is now a test for it.

## What it deliberately does not do

**Store zeros for unobserved samples.** That would also square the table and was the first thing suggested, but it destroys the difference between *not observed* and *observed as zero*, which `Run_OmitsUnobservedSamplesRatherThanStoringZero` pins on the data model. An unobserved sample now renders as an **empty cell** rather than a vanished column.

Worth being precise, because the obvious reading is wrong: absent and a measured zero are **not** distinguishable end-to-end today, because `OverwriteSampleIntensities` drops zero-valued cells, so a genuine zero never reaches `IntensitiesBySample` either. The empty cell is still right — it does not assert a measurement that was not made — but it is not currently carrying that distinction.

## Scope — this narrows the seam, it does not close it

Rectangularity cannot be established from inside `BioPolymerGroup`. The *number* of sample groups comes from `PopulateSampleGroupResults`, whose no-design branch builds one result per file in that groups **own** spectral matches — so two groups covering different files still emit different column counts, and a group that never entered the matrix keeps null intensities while its neighbours do not. The header is written once per file, from one group; the rows come from many. **Only a writer-level decision — one column set chosen for the whole file — makes every table square.**

This fixes the case that matters now: the four columns agree across the groups of one quantified run, which is the shape MetaMorpheus #2755 introduces. The remarks in the code say the rest plainly rather than implying it is solved.

## `virtual`, and its MetaMorpheus counterpart

`ToString` is virtual and `GetTabSeparatedHeader` was not, so a subclass could only *hide* the header with `new` while genuinely overriding the row. `IBioPolymerGroup` declares `GetTabSeparatedHeader`, so an interface-typed call would get this header paired with a derived row.

No such call site exists today — every caller in either repo resolves on a concrete type — but calling `QuantificationEngine` from `PostSearchAnalysisTask` is exactly the change that starts passing groups through `IBioPolymerGroup`-typed plumbing, and it is in flight. Making it `virtual` closes that before it lands. It is a recompile concern rather than a source one.

**Counterpart PR, not yet opened:** MetaMorpheuss `EngineLayer.ProteinGroup` carries its own copy of this rendering block and hides the header with `new`. After a release ships this, that becomes `override`, and its own `Intensity_` gate — still per-group — needs the same treatment. MetaMorpheus is not currently ragged only because it writes FlashLFQs zeros explicitly, which makes `HasIntensityData` uniformly true; the TMT path in #2755 omits them and removes that protection. `ProteinGroupTest.cs:280` is the existing guard rail and should be extended to a group whose intensities were *omitted* rather than zeroed.

## Tests

Five, asserting at the **file** level, because a unit test on `GetTabSeparatedHeader` cannot see a misalignment — render a header and every row, then assert each row has the headers field count.

- a group with unobserved samples
- a group quantified with no values — also asserts an unquantified group describes the same columns as a quantified one, since the writer may take the header from either
- a group no engine touched still has no intensity columns at all
- the absent cell is empty, not a fabricated zero (verified by mutation: with the inner check removed, this fails on three cells with `Expected: <empty> But was: "0"` while the rectangularity tests still pass — which is why it exists)
- a subset group for an unmatched file does not advertise columns it cannot fill (verified by reverting the gate)

**No existing test needed changing.** Full suite: 5827 passed, 29 skipped, 0 failed.

### DIFF
--- a/mzLib/Omics/BioPolymerGroup/BioPolymerGroup.cs
+++ b/mzLib/Omics/BioPolymerGroup/BioPolymerGroup.cs
@@ -285,27 +285,6 @@ namespace Omics.BioPolymerGroup
         /// Header includes columns for biopolymer information, quantification, and statistical metrics.
         /// </summary>
         /// <returns>Tab-separated header string suitable for TSV file output.</returns>
-        /// <summary>
-        /// Whether this group was handed to a quantification engine at all, which is what decides
-        /// whether the intensity columns exist.
-        /// </summary>
-        /// <remarks>
-        /// Deliberately NOT <see cref="SampleGroupResult.HasIntensityData"/>. That answers a per-group
-        /// question -- did this sample group get a value -- and using it to choose columns made the
-        /// table ragged: a group whose samples were all unobserved emitted two columns per sample
-        /// group where its neighbour emitted four. The header is written once, from one group, while
-        /// every row writes its own, so every value after the disagreement landed under the wrong
-        /// column name.
-        ///
-        /// Quantification is a property of the run, not of one sample group: an engine either
-        /// populated this entity or it did not. <see cref="IHasSampleIntensities.IntensitiesBySample"/>
-        /// is assigned for every row of the matrix -- a dictionary, empty if nothing was observed --
-        /// so all groups in one run agree, the table is rectangular by construction, and an unobserved
-        /// sample yields an empty cell rather than a vanished column, which keeps absent
-        /// distinguishable from an observed zero. It stays null when no engine ran, which is what
-        /// omits the columns altogether.
-        /// </remarks>
-        private bool IsQuantified => IntensitiesBySample is not null;
         public string GetTabSeparatedHeader()
         {
             var sb = new StringBuilder();
@@ -327,7 +306,7 @@ namespace Omics.BioPolymerGroup
             #region Quantification Header Building
             if (SampleGroupResults is null) PopulateSampleGroupResults();
 
-            bool quantified = IsQuantified;
+            bool quantified = HasAssignedSampleIntensities;
             foreach (var group in SampleGroupResults!)
             {
                 sb.Append($"SpectralCount_{group.Label}\t");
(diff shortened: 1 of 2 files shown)


nbollis (approved):

acesnik:
🤖 Good catch, and a satisfying one to verify: a row that renders fewer fields than its own header isn't obviously broken until you count the tabs, and then every value after the gap is silently under the wrong column name. Making the shape a property of the run rather than of each group is the right fix.

**First, the red check is not this PR.** `external-service-tests` failed one test of 198 — `TestAltimeterChargeStateBoundaries` — on a 400 from Koina whose body carries a server-side fault: `PyTorch execute failure … RuntimeError: CUDA error: CUBLAS_STATUS_EXECUTION_FAILED` in ensemble `Altimeter_2024_intensities`. That's a GPU fault inside the remote inference server. `TestAltimeterAcceptsValidPeptides` passed 3 s earlier and `TestAltimeterPredictionQuality` 2 s later against the same model, the same job is green on #1242, #1243 and #1244 which ran about three hours later, and the diff touches nothing on the Koina path. `dotnet.yml` marks that job non-blocking in as many words, and master's branch protection requires only `build`. Nothing to fix and no re-run needed.

Worth noting the probe can't catch this class: `KoinaLiveTestFixture` checks `/v2/health/ready` once per fixture to convert outages into skips, and Koina reported ready while individual inferences failed. A 400 whose body is a server-side CUDA error is an outage, not a bad request.

## What changes in the output — measured

I rendered header and every row through both the baseline and PR versions on the delivery fixture and diffed the text. **Two of seventeen rendered lines change, and both were already misaligned:**

| scenario | before | after |
|---|---|---|
| all samples observed | 45 / 45 / 45 | byte-identical |
| a group unobserved in one file | row is **39** fields against a 45-field header | 45, three empty cells |
| a group quantified with no values | row is **33** fields | 45, six empty cells |
| nothing quantified | 21 / 21 | byte-identical |
| subset for an unmatched file | 21 / 21 | byte-identical |

No column added or removed from any header, no observed value altered, no cell that held a number now empty or the reverse. The only rows that move are the ones that were being written under the wrong column names.

**And nothing mzLib writes today changes at all.** `QuantificationWriter` builds all three of its files from `QuantMatrix` directly, and no non-test code in the repo calls `GetTabSeparatedHeader`. This reaches a produced file only once a consumer renders through these methods.

Two corroborations from the code rather than the description. `OverwriteSampleIntensities` assigns `SamplesForQuantification = matrix.ColumnKeys.ToList()` on every row and `GetUniquePeptideToProteinMap` seeds an entry for every group, so all groups of one quantified run do agree — but the other map builder, `GetAllPeptideToProteinMap` (used when `UseSharedPeptidesForProteinQuant` is true), is `throw new NotImplementedException()`. Nothing reachable weakens the invariant today; if that gets implemented it has to seed every group or the raggedness returns. And `OverwriteSampleIntensities` drops zero-valued cells, which is why the absent-versus-zero caveat in the description is the honest version.

## The MetaMorpheus half

`virtual` checks out: on MetaMorpheus master `ProteinGroup : BioPolymerGroup` with `new string GetTabSeparatedHeader()` at line 236 and `override string ToString()` at 284, and the class comment already warns that a base-typed call pairs base column names with MetaMorpheus rows. Since it does derive, `protected` on the new predicate is usable there.

Two notes for the counterpart, both from that file. MetaMorpheus already has the run-level pattern — `HasPeptideLevelQuantification`, documented as "assigned uniformly so every group carries the same schema" — but applies it only to `IntensityOccupancy_`; `Intensity_` is still on per-group `HasIntensityData`. So the counterpart extends an accepted precedent rather than introducing one. And `WasSearched(group)` gates `SpectralCount_`/`CountOccupancy_` off a run-level file set, which is the writer-level information your scope note says `BioPolymerGroup` lacks — that may be the shape of the remaining fix.

No file overlap with #1242, #1243 or #1244, so merge order doesn't matter for this one.

Not verified: mzLib's `Test` project is `net10.0-windows`, so the measurements above went through a purpose-built net10.0 harness rather than the suite — the suite result is CI's. I didn't look at #2755, so the claim about the TMT path omitting zeros is yours rather than something I checked, and I haven't run a real TMT dataset end to end.

acesnik (approved):

PR #1242, opened 2026-08-28
Reject an unset quantification strategy by name, instead of throwing from mid-run

trishorts:
`QuantificationParameters` leaves its strategy properties null and `ValidateEngine` never looked at
them. An external caller who missed one got a `NullReferenceException` from
`QuantificationEngine.cs:348` — a frame several steps removed from the property they forgot, naming
nothing. This is on the released 1.0.588 surface.

### The timing is worse than the message

`Run` starts writing the raw matrix *before* it reads the first strategy:

```csharp
if (Parameters.WriteRawInformation)
    rawWriter = Task.Run(() => QuantificationWriter.WriteRawData(...));   // starts here
RunPeptideQuant(out var peptideMatrix);                                   // throws here
```

So the failing run left `RawQuantification.tsv` on disk and then died. I confirmed that by probe
rather than by reading: against the unfixed engine the file is present after the throw. Validation
now rejects the run before anything is written, and returns a `QuantificationResults` failure rather
than throwing — the contract the rest of `ValidateEngine` already keeps.

### Six required, reported together

The six are reported in one message rather than one per run. A caller who missed one has most likely
missed several, and learning the list a run at a time is the slower way to find out.

### The seventh is conditional, and this is the substantive finding

The plan I was working from called this "seven strategy properties". The repo disagrees, and the repo
is right.

`NoCollapse` already documents that it ignores the aggregator — *"The aggregation strategy is unused,
because no group ever has more than one member"* — and `QuantificationDeliveryTests.SimpleParameters`
builds exactly that configuration: `NoCollapse`, with `CollapseAggregationStrategy` left unset. It
works today. Requiring the property outright would have rejected a working setup and forced callers
to name a behaviour that never happens.

So `ICollapseStrategy` now declares whether it reads one:

```csharp
bool RequiresAggregation => true;
```

A default interface member, so nothing external breaks, and defaulting to `true` is the safe
direction: an implementation that *does* aggregate is covered without opting in, and only a genuine
no-op has to say otherwise. `NoCollapse` overrides it to `false`.

**Those delivery tests pass unmodified.** That is the evidence the conditional is the right shape
rather than a convenience — had I required all seven, the correct move would have been to edit them,
and the bug would have been in the contract instead of the code.

### The cross-check, added in `24b4c01f`

`RequiresAggregation` is itself a second representation of a fact `CollapseSamples`' body already
carries — and `ValidateEngine` **trusts** it, letting a strategy that answers `false` run with a null
aggregator. So a strategy whose declaration and body disagree brings back the exact
`NullReferenceException` this PR removes, from a strategy that declared it could not happen.

`RequiresAggregationMatchesWhetherTheStrategyActuallyReadsIt` checks the declaration against the
behaviour for every concrete `ICollapseStrategy` in the assembly, in both directions — declaring
`false` means a null aggregator must not throw, declaring `true` means it must. The strategies are
found by reflection, so one added later is covered without anyone remembering to add it.

Both mutants verified: `CollapseStrategyBase` claiming `false` fails on `CollapseBiologicalReplicates`;
`NoCollapse` claiming `true` fails on `NoCollapse`.

That arm is still worth validating even though `CollapseStrategyBase` already throws
`ArgumentNullException` for a null aggregator: what changes is not the naming but the timing, since
that throw also comes from inside the run, after the raw matrix is written.

### `Parameters` itself

Checked first. Both the new strategy lookup and the existing output-directory fallback dereference it,
so without that check the validation meant to replace a `NullReferenceException` would throw one of
its own.

### Tests

Ten, in `QuantificationTests`. Each strategy is nulled in turn and asserted to be named — driven off
a *reflected* list of the strategy-typed properties on `QuantificationParameters`, so a strategy added
later is covered without anyone remembering to add it here, and one that nobody registers in
`RequiredStrategies` fails. Both arms of the aggregation condition are pinned, as is the
no-file-written consequence.

All ten fail against the unfixed engine, and the seven per-strategy cases fail with the
`NullReferenceException` each was written for.

122 tests green in `Test.Quantification`; the full suite is 5834 passed / 29 skipped / 0 failed.

Companion to PR-B (`QuantificationResults.Samples`), which is independent of this one.

### DIFF
--- a/mzLib/Quantification/Interfaces/ICollapseStrategy.cs
+++ b/mzLib/Quantification/Interfaces/ICollapseStrategy.cs
@@ -18,6 +18,21 @@ namespace Quantification.Interfaces
         string Name { get; }
 
         /// <summary>
+        /// Whether <see cref="CollapseSamples{T}"/> reads the aggregation strategy it is handed.
+        /// </summary>
+        /// <remarks>
+        /// A strategy that groups nothing never reduces two values to one, so it has no use for an
+        /// aggregator -- and requiring the caller to supply one anyway would mean naming a behaviour
+        /// that does not happen. <see cref="QuantificationEngine"/> reads this to decide whether a null
+        /// <see cref="QuantificationParameters.CollapseAggregationStrategy"/> is a configuration error
+        /// or simply unused.
+        ///
+        /// Defaults to true, which is the safe answer: an implementation that does aggregate gets the
+        /// check without opting in, and only a genuine no-op has to say so.
+        /// </remarks>
+        bool RequiresAggregation => true;
+
+        /// <summary>
         /// Combines the columns this strategy groups together, reducing each group to a single column
         /// with <paramref name="aggregation"/>.
         /// </summary>
--- a/mzLib/Quantification/QuantificationEngine.cs
+++ b/mzLib/Quantification/QuantificationEngine.cs
@@ -198,6 +198,33 @@ public class QuantificationEngine
     }
 
     /// <summary>
+    /// The strategies every run needs, in the order the engine applies them, each paired with the name
+    /// to report when it is missing. <see cref="QuantificationParameters.CollapseAggregationStrategy"/>
+    /// is deliberately absent: it is needed only when the collapse strategy reads it, which is
+    /// <see cref="ValidateEngine"/>'s separate check.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="QuantificationParameters"/> leaves all of these null, and none has a defensible
+    /// default -- summing and taking a median are different answers, not different spellings of one.
+    /// So the only thing validation can do is say which are missing, and it is worth doing early: the
+    /// first dereference is in <see cref="RunPeptideQuant"/>, which <see cref="Run"/> reaches after it
+    /// has already started writing the raw matrix. Without this check the run leaves that file behind
+    /// and then throws a NullReferenceException naming nothing.
+    ///
+    /// Spelled out rather than reflected over, so that a strategy added to the parameters is a
+    /// deliberate addition here too, and the reported name comes from <c>nameof</c> either way.
+    /// </remarks>
+    private static readonly (string Name, Func<QuantificationParameters, object> Get)[] RequiredStrategies =
+    {
+        (nameof(QuantificationParameters.SpectralMatchNormalizationStrategy), p => p.SpectralMatchNormalizationStrategy),
+        (nameof(QuantificationParameters.SpectralMatchToPeptideRollUpStrategy), p => p.SpectralMatchToPeptideRollUpStrategy),
+        (nameof(QuantificationParameters.PeptideNormalizationStrategy), p => p.PeptideNormalizationStrategy),
+        (nameof(QuantificationParameters.CollapseStrategy), p => p.CollapseStrategy),
+        (nameof(QuantificationParameters.PeptideToProteinRollUpStrategy), p => p.PeptideToProteinRollUpStrategy),
+        (nameof(QuantificationParameters.ProteinNormalizationStrategy), p => p.ProteinNormalizationStrategy),
+    };
+
+    /// <summary>
     /// Checks Engine state for validity before running quantification.
     /// </summary>
     /// <param name="badResults"> Return quant results with descriptive Summary if problem was encountered; null otherwise</param>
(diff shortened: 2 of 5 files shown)


nbollis (approved):

acesnik:
🤖 Turning a `NullReferenceException` deep inside a run into a named list of what the caller forgot to set is a real improvement, and the detail I like most is that it reports *all* the missing strategies at once rather than one per round trip. I checked the three things that would have worried me — reachability, ordering, and whether anything that used to work is newly refused — and all three come out clean.

**Reachability.** This can't reach a MetaMorpheus user. `git grep` on MetaMorpheus `upstream/master` finds zero references to `QuantificationEngine`, `QuantificationParameters` or `ICollapseStrategy` (it pins mzLib 1.0.587), and the rejection is a `QuantificationResults` failure return rather than a throw, so it never touches the crash-report path. Programming error against the library surface, reported as data — which is the right shape.

**Ordering.** Strictly earlier than the work it guards: `ValidateEngine` is step 0 of `Run`, ahead of the `Task.Run` raw writer, so nothing side-effecting can precede it. I ran the old behaviour as a probe — against the unfixed engine `RawQuantification.tsv` exists at the moment of the `NullReferenceException`, in 5 of 5 trials.

**Nothing that used to work is newly rejected**, at least for in-repo callers. `CollapseStrategyBase.CollapseSamples` throws `ArgumentNullException` as its first, unconditional statement, so there's no group-size-dependent path where a collapsing strategy previously ran with a null aggregator. And the conditional is load-bearing rather than a convenience — making the aggregation requirement unconditional fails **10 pre-existing `QuantificationDeliveryTests`**, every one built from `SimpleParameters()`, which pairs `NoCollapse` with an unset aggregation strategy.

**One case worth a line in the body.** A default interface member is source- and binary-compatible, but for a third-party `ICollapseStrategy` that genuinely ignores its aggregator it isn't *behaviour*-compatible: it inherits `RequiresAggregation => true` and a run that worked with a null aggregator is now refused. No such implementer exists in mzLib or MetaMorpheus, and `true` is the direction that fails loudly rather than silently, so I wouldn't change anything — just say that rather than "nothing external breaks", which is true of compilation but not of behaviour.

**Optional, on the two-round-trip case the "reported together" rationale is meant to prevent.** `Parameters.CollapseStrategy?.RequiresAggregation == true` short-circuits when `CollapseStrategy` is null, so a caller who set nothing is told six names, sets `CollapseFractions`, reruns, and is only then told about `CollapseAggregationStrategy`. `ParametersWithNoStrategiesSetNameEveryStrategyTheRunNeeds` encodes exactly that exclusion, so it's deliberate — and you genuinely can't know whether it's needed. A clause on the end of the message would close it without guessing.

**On the tests**, which I checked for discrimination rather than reading: reverting only `QuantificationEngine.cs` to master fails 11 of the 12 new cases, with `NoCollapseNeedsNoAggregationStrategy` correctly passing both ways. Both `RequiresAggregation` mutants are caught. Dropping `ProteinNormalizationStrategy` from `RequiredStrategies` fails the reflected per-strategy case, so the unregistered-strategy claim holds.

Four small things:

1. The body says "Ten, in `QuantificationTests`" — it's 6 `[Test]` methods / 12 cases, 11 of which fail against the unfixed engine.
2. `Assert.That(results.WrittenFiles, Is.Empty)` in `AMissingStrategyIsRejectedBeforeAnythingIsWritten` is near-vacuous: `QuantificationResults.Failure` never populates it and the property defaults to an empty list, so it holds on *any* failure whatever landed on disk. The `Directory.GetFiles` assertion is the one doing the work — and against the unfixed engine the test dies on the `NullReferenceException` before reaching either.
3. `EveryCollapseStrategy()` filters on `GetConstructor(Type.EmptyTypes) != null` and drops non-matches silently. Fine for the five that exist; a future strategy with a required constructor argument would leave the cross-check with no signal. Asserting the expected count keeps it honest.
4. The `true` arm asserts `ArgumentNullException` specifically, so it requires a guard clause rather than merely a dereference — a strategy that honestly needs an aggregator and NREs on it would fail. That's the better contract, but `ICollapseStrategy`'s doc for `RequiresAggregation` doesn't state it.

`CollapseStrategyTests.cs:269` already pins `CollapseFractions` plus a null aggregator, and `Names_DescribeTheGroupingAndAreDistinct` hand-lists the same five types the new reflection helper enumerates — both could fold into it.

Nothing here changes a produced value: no arithmetic is touched, and the 51 pre-existing cases in my harness pass unchanged. Not verified: the full suite — `Test` targets `net10.0-windows`, so I built a throwaway net10.0 harness compiling the four relevant files (52 cases). The body's 5834/29 figures are CI's.

acesnik (approved):

PR #1243, opened 2026-08-28
Collapse the peptide matrix where the caller can see it, so one sample list keys both tables

trishorts:
`QuantificationResults.Samples` is documented as the columns of **both** `PeptideIntensities` and
`ProteinIntensities`. It is assigned `proteinMatrix.ColumnKeys` only, and with any collapse strategy
that groups anything the peptide table is keyed differently. This is on the released 1.0.588 surface.

Not "described wrongly" — *disjoint*. In the fixture added here:

```
Samples          : Reference_Bio0_Frac0_Tech0_Collapsed_Channel_0, Control_..._Channel_1
peptide table    : file1_Channel_0, file1_Channel_1, file2_Channel_0, file2_Channel_1
```

A caller who joins the two tables on `Samples` gets nothing back for the peptides.

### The cause is a by-value parameter, which makes this bigger than a doc fix

`Run` calls `RunProteinQuant(peptideMatrix, out proteinMatrix)`, and `RunProteinQuant` collapsed by
reassigning **its own parameter**:

```csharp
internal void RunProteinQuant(QuantMatrix<IBioPolymerWithSetMods> peptideMatrixNorm, ...)
{
    peptideMatrixNorm = Parameters.CollapseStrategy.CollapseSamples(peptideMatrixNorm, ...);
```

The caller never saw that. So the collapse reached the protein matrix and nothing else, and **three**
things followed rather than one:

- `Samples` described the protein table only;
- `PeptideQuantification.tsv` was written from the uncollapsed matrix, so the two written files named
  different samples and neither said so;
- `OverwriteSampleIntensities` wrote uncollapsed per-sample values back onto the peptide entities
  while the protein entities got collapsed ones.

### The fix is where the class already said it was

Moving the collapse to the end of `RunPeptideQuant` fixes all three at once — and it is the order
this class's own summary describes:

> 4) Collapse the peptide matrix, combining fractions and technical replicates
> \* Writes peptide information (if enabled)
> 5) Roll up to proteins

The code had drifted from its own documented order. Now the protein matrix is rolled up **from the
collapsed peptide matrix**, so both tables carry the same columns by construction rather than by
agreement — which is what `Samples` can honestly claim, and the `Samples` doc now says why.

### Blast radius

**None for a run using `NoCollapse`**, which is every production caller today: collapsing nothing is
the identity either way. Callers that do collapse get peptide numbers that match their protein
numbers, which they did not before.

### Tests

Three, all failing against the unfixed engine:

- `SamplesKeysBothIntensityTables` — the direct claim.
- `CollapseReachesThePeptideTableAndNotOnlyTheProteinTable` — the peptide table carries the collapsed
  samples, not the uncollapsed ones it started from.
- `TheWrittenPeptideAndProteinFilesShareOneSetOfSampleColumns` — the same claim at the file level,
  where a user actually meets it. The two files are written from two different matrices at two
  different points in `Run`, so agreement between their headers is the end-to-end statement that one
  collapse governed the whole run.

112 green in `Test.Quantification`, 159 with the FlashLFQ fixtures.

Independent of #1242 — different regions of `QuantificationEngine.cs`, no shared interface. If both
land, expect a trivial conflict in `QuantificationTests.cs` and nothing more.

### DIFF
--- a/mzLib/Quantification/QuantificationEngine.cs
+++ b/mzLib/Quantification/QuantificationEngine.cs
@@ -364,14 +364,27 @@ public class QuantificationEngine
         // 5) Normalize combined peptide matrix
         peptideMatrixNorm = Parameters.PeptideNormalizationStrategy
             .NormalizeIntensities(combinedPeptideMatrix);
+
+        // 6) Collapse samples (fractions, technical replicates).
+        //
+        // This belongs here rather than in RunProteinQuant, where it used to sit. Run hands the peptide
+        // matrix on BY VALUE, so collapsing there reassigned a parameter the caller never saw again: the
+        // protein matrix came out collapsed while the caller's peptide matrix stayed uncollapsed, and the
+        // two were then reported side by side keyed differently. Doing it at the end of the peptide stage
+        // is also the order this class's summary already describes -- collapse, then write peptides, then
+        // roll up to proteins -- so the written peptide file now matches its own documentation.
+        peptideMatrixNorm = Parameters.CollapseStrategy
+            .CollapseSamples(peptideMatrixNorm, Parameters.CollapseAggregationStrategy);
     }
 
+    /// <summary>
+    /// Rolls the peptide matrix up to protein groups and normalizes the result. The matrix arrives
+    /// already collapsed -- <see cref="RunPeptideQuant"/> does that, so the peptide results the caller
+    /// keeps and the protein results derived from them are keyed by the same samples.
+    /// </summary>
     internal void RunProteinQuant(QuantMatrix<IBioPolymerWithSetMods> peptideMatrixNorm,
         out QuantMatrix<IBioPolymerGroup> proteinMatrixNorm)
     {
-        // 6) Collapse samples (technical replicates, fractions)
-        peptideMatrixNorm = Parameters.CollapseStrategy.CollapseSamples(peptideMatrixNorm, Parameters.CollapseAggregationStrategy);
-
         // 7) Roll up to proteins
         var proteinMap = Parameters.UseSharedPeptidesForProteinQuant
             ? GetAllPeptideToProteinMap(peptideMatrixNorm, BioPolymerGroups)
--- a/mzLib/Quantification/QuantificationResults.cs
+++ b/mzLib/Quantification/QuantificationResults.cs
@@ -39,6 +39,12 @@ namespace Quantification
         /// The samples that form the columns of <see cref="PeptideIntensities"/> and
         /// <see cref="ProteinIntensities"/>, in output order. Empty on failure.
         /// </summary>
+        /// <remarks>
+        /// One list covers both tables because the protein matrix is rolled up from the peptide matrix
+        /// after it has been collapsed, so the two carry the same columns by construction rather than by
+        /// agreement. That is worth stating: when the collapse ran on the protein side only, this list
+        /// described the protein table and silently did not match the peptide one.
+        /// </remarks>
         public IReadOnlyList<ISampleInfo> Samples { get; internal set; } = new List<ISampleInfo>();
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


nbollis (dismissed):

acesnik:
🤖 The diagnosis here is the valuable part: collapsing the protein matrix but handing the caller an uncollapsed peptide matrix means the two tables in one result set describe different sample groupings, which is the sort of inconsistency that gets noticed months later by someone comparing columns. Moving the collapse to where the caller can see it is the right call, and the fix is smaller than the explanation.

I measured the blast radius rather than reasoning about it. Dumping `Samples`, `PeptideIntensities`, `ProteinIntensities` and the protein-entity delivery from both master's engine and this one, same fixture, same seed:

| run | peptide table | protein table / `Samples` / entity delivery |
|---|---|---|
| `NoCollapse` | byte-identical | byte-identical |
| `CollapseFractions` + `SumAggregation` | **4 columns → 2, values change** | byte-identical |

The peptide change is different numbers, not only a different shape:

```
master: file1_Channel_0=2300  file1_Channel_1=4600  file2_Channel_0=5400  file2_Channel_1=10800
this:   Reference_..._Channel_0=7700  Control_..._Channel_1=15400
```

So `PeptideQuantification.tsv` and `QuantificationResults.PeptideIntensities` change for any collapsing run, and nothing else does. `NoCollapse` is exact identity, which follows from `NoCollapse.CollapseSamples` returning the same instance — the move can't perturb it.

**Applied once, on every path.** `RunPeptideQuant` and `RunProteinQuant` have exactly one caller each (`Run`) across the repo, so there's no double-apply and no gap — that was my main worry with a moved responsibility. And no MetaMorpheus reference to `QuantificationEngine` exists, so no shipped quantification output moves with this.

**One bullet in the description isn't observable today.** The third listed consequence, `OverwriteSampleIntensities` on peptide entities, is a no-op: `IHasSampleIntensities` is implemented by `IBioPolymerGroup` only, so `OverwriteSampleIntensities(peptideMatrix)` writes nothing onto a `PeptideWithSetModifications` either way. The fix is real for the other two; this one becomes real the day a peptide type implements the interface.

Minor: `RunProteinQuant`'s new doc states the "arrives already collapsed" invariant but nothing enforces it, so a future second caller passing an uncollapsed matrix gets uncollapsed protein output silently rather than an error. Both methods are `internal` and single-caller, so that's a note rather than a request.

**On the tests:** all three fail with master's `QuantificationEngine.cs` restored and the new test file kept, while the five pre-existing tests in the file still pass — so they'd catch a revert. `CollapseReachesThePeptideTableAndNotOnlyTheProteinTable` is the load-bearing one; `SamplesKeysBothIntensityTables` uses `Is.SubsetOf`, which an all-zero peptide row would satisfy vacuously since `BuildIntensityTable` drops zero cells. It happens to be non-vacuous on this fixture.

No existing collapsing test regresses: `TmtSpikeInTests`' two `CollapseFractionsAndTechnicalReplicates` cases assert on `proteinMatrix` only, `QuantificationDeliveryTests` uses `NoCollapse`, and the `Development` collapse tests are `[Explicit]`.

**Merge order.** Independent of #1240 and #1244. Against #1242, `QuantificationEngine.cs` auto-merges — the two touch `ValidateEngine` and `RunPeptideQuant` respectively — and the only conflict is the inserted region in `QuantificationTests.cs`. Either order works, and #1242's remark that the first strategy dereference is in `RunPeptideQuant` gets *more* accurate after this.

Not verified: any real data — the only collapsing fixture is synthetic (3 peptides × 2 files × 2 channels) — and the `[Explicit]` HeLa development runs. `Test.csproj` is `net10.0-windows`, so I ran the 8 tests in `QuantificationTests.cs` plus my own dump via a throwaway harness; `TmtSpikeInTests` and `QuantificationDeliveryTests` were read, not executed.

acesnik (dismissed):

PR #1244, opened 2026-08-28
Audit an SDRF for what it says about quantification, and settle the corpus counts

trishorts:
An SDRF whose `comment[label]` names a **reagent kit** rather than a channel parses cleanly,
validates cleanly, and then yields **one channel for a ten-channel experiment**. So does a file with
two rows per (data file, channel), or one that declares no isobaric modification at all. None of that
is a structural defect, so `SdrfValidator` is right not to report it — it is a defect in what the
annotation *means*, and it needed its own reader.

`SdrfQuantAuditor` reports, read-only, per document:

- channel-level / kit-only / not isobaric;
- the channels found and the plex they imply;
- where the plex came from, with the evidence shown;
- which of the eleven per-(data file × channel) facts are present, absent or unparseable;
- whether an isobaric modification is declared at all;
- the label form **per cell**;
- repeated (data file, label) pairs, ragged rows;
- which data files were found on disk.

Scoped to every SDRF rather than to TMT. "This is not an isobaric experiment" is a useful answer when
a user points it at whatever they just downloaded, and the corpus's 16 iTRAQ files are real targets.

### It settles a discrepancy this repo has been carrying

Three independent surveys of the curated corpus returned **49, 51 and 52** channel-level TMT files,
and 31 versus 32 kit-only. The qualitative findings agreed; the integers did not, and none of them
was produced by code anyone could re-run.

The spread is **scope, not arithmetic**:

| | channel-level | kit-only |
|---|---|---|
| TMT family | **51** | **31** |
| iTRAQ only | 10 | 6 |
| **all isobaric** | **61** | **37** |

51 and 31 are exactly right for the TMT family alone, which is what those surveys were counting.
Both rows are pinned in the expectations table, so the next disagreement is a scope question with an
answer rather than three integers and no way to choose.

Two smaller corrections fell out of doing it:

- The curated corpus is the **`datasets` subtree** — 1,236 files. The clone also carries a `sandbox`
  directory of 295 more that are not part of it, and counting those is one way to get an
  unreproducible number. The test accepts either the clone root or the subtree and scopes itself.
- `PXD040455` does carry exactly the **551** repeated pairs recorded earlier — but it is **not** the
  worst. `PXD017291-mixed-label` carries **781**, which that survey missed.

### Decisions the corpus forced

Each is one a reader written from the specification alone gets wrong:

- **Classification is on label values, never column presence** — all 1,236 files declare
  `comment[label]`.
- **Label form is per cell.** Exactly one document writes both forms, so a per-document sniff
  misreads every cell of the other one.
- **Accessioned cells are parsed, not prefix-matched.** Both key orders occur, and `AC=` first is the
  commoner of the two for label-free.
- **Case is folded for identity and reported as a variant.** `itraq114`, `TMT10Plex` and `tmt126` are
  the same reagents as their canonical spellings — but anyone joining on the raw string needs to know.
- **Reserved words are absence, not answers.** `not available` is spec-correct and tells a consumer
  nothing, which is why it is counted apart from a value like `19b` that is simply unparseable.
- **The plex is reported as unstated rather than guessed.** The source-name partition is right 3 times
  in 9 where ground truth exists, failing by *over*-splitting, and no bulk dataset in the corpus
  carries a usable batch column. Inferring here would manufacture plexes that are not in the file.

### Tests

15 unit tests over a committed fixture pair — including **zero-byte stub raw files**, since the
corpus carries no data files and cannot exercise disk matching, and "0 of 2 found" must not read the
same as "not looked".

Two `[Explicit]` corpus tests gated on `MZLIB_SDRF_CORPUS`, following `TestSdrfValidator`'s existing
calibration pattern: one compares sixteen counts against the checked-in expectations table so drift
in the corpus **or** in the classifier shows up as a diff; the other pins the three named duplicate
documents, including the 21-rows-for-one-pair shape in `PXD007160`.

121 green across the SDRF fixtures.

### Not in this PR

The CMD verb that surfaces this to a user. MetaMorpheus consumes mzLib by NuGet pin, so that lands
after a release carries this. `ToReport()` is what it will print.


acesnik:
🤖 This is a genuinely useful thing to have. "What does this SDRF actually say about quantification" is a question that currently gets answered by squinting at a TSV, and an auditor that reports plex evidence rather than guessing at it, treats reserved words as absence, and distinguishes "not looked" from "looked and found nothing" is the right design. Several of the calls here are ones I'd have got wrong — refusing to infer the plex from a filename, and the three-way `Kind`, especially.

Let me lead with the easy part: it's **purely additive**. One new production file (`Readers/Sdrf/SdrfQuantAudit.cs`), one new test file, four fixtures, four `CopyToOutputDirectory` entries. No existing type is touched and nothing in the repo references `SdrfQuantAuditor` yet, so no produced database, search result or quantification value can change. I built a throwaway net10.0 harness against `Readers` and ran the 15 non-`[Explicit]` tests on macOS/arm64: 15 passed. Backslash handling is right — `D:\runs\f1.raw` in `comment[data file]` matches `f1.raw` on disk off Windows. The one numeric parse in production code uses `NumberStyles.Integer, CultureInfo.InvariantCulture`.

What's holding me up is that three measured behaviours defeat the specific thing the auditor exists to prevent — the audit returns a confident wrong answer rather than "I couldn't tell."

**1. Ten of the eleven column lookups are case-sensitive, so a mis-cased header reads as "not isobaric."** `SdrfHeader.Contains` and `IndexOf` are `StringComparison.Ordinal` (`:48`, `:57`), and the auditor's literals are lowercase. A file with `Comment[label]` and `Comment[data file]` gives, measured:

```
  design            : NotIsobaric
  label form        : 0 bare, 0 accessioned
     1. data file              ABSENT  column absent
     2. channel                ABSENT  not an isobaric experiment
```

A TMT file reported as label-free, with `Kind`, `Channels`, `BareLabelCells`, `DuplicateFileLabelPairs`, `MostRowsForOneDataFile` and `MatchedDataFiles` all silently empty. Three things suggest oversight rather than decision: `FactorValue` in the same method already uses `OrdinalIgnoreCase` (`:512`) and resolves `Factor Value[disease]` fine — one of eleven; `SdrfHeader`'s own doc names `"Material Type"` and `"Factor Value[organism part]"` as real corpus spellings; and `SdrfValidator` next door matches every column with `OrdinalIgnoreCase` and has `CasingHint` (`:140`) built for exactly this, its doc example being `'Comment[data file]' is present and differs only in casing`.

I wouldn't make the auditor *accept* a mis-cased column — the spec makes lowercase a MUST and the validator is right to treat mis-cased as missing. But `NotIsobaric` is a *conclusion*, and it shouldn't be reached because a lookup missed. Reusing `CasingHint`'s loop to report `Unparseable` with the observed spelling keeps the strictness and loses the wrong answer.

This also bears on the sixteen pinned counts. `not-isobaric 1236 → 1138` can't distinguish a genuinely label-free file from one with a capital C, and "all 1,236 files declare `comment[label]`" is only true under this exact lookup. I couldn't check that — no corpus clone here — so it's worth re-deriving before the table is treated as settled.

**2. An unrecognised isobaric label becomes `NotIsobaric`, and nothing counts what fell through.** The `else { continue; }` discards unclassifiable label cells without recording them. Measured, one row per file:

| `comment[label]` | reported `Kind` |
|---|---|
| `TMT-10-plex` | **NotIsobaric** |
| `iTRAQ-8-plex` | **NotIsobaric** |
| `TMT10` | **NotIsobaric** |
| `TMT131N-d` | **NotIsobaric** |
| `TMT 10 plex` | KitOnly |
| `TMTpro16plex` | KitOnly |
| `tmt126c`, `TMT134N` | ChannelLevel |

A hyphen before `plex` silently reclassifies the experiment. Since the point is honesty about what the file says, an `UnclassifiedLabelCells` count with the distinct values (capped) and a line in `ToReport()` would make `NotIsobaric` mean "no isobaric label" rather than "none I recognised." Widening the regexes is optional once the count is visible; the count is the part that matters.

**3. A nonexistent data directory reads exactly like a clean result.** `MatchDataFiles` returns `([], [])` when `!Directory.Exists(dataDirectory)`, while `SearchedForDataFiles` is set from `dataDirectory != null`. Measured, on a document naming two files with a bad path:

```
  data files        : 0 found, 0 missing
```

You make `"0 of 2 found" must not read the same as "not looked"` a design goal and `NoDirectoryMeansNotLookedRatherThanNothingFound` pins one half of it. This is the other half — a typo'd path produces a report with no complaint in it.

**4. One key, two comparers.** `rowsPerFile` is `OrdinalIgnoreCase` and `pairCounts` takes the default ordinal comparer — adjacent lines, `:317` and `:318`. On two rows carrying `plexA_fr1.raw` and `PLEXA_FR1.RAW` with the same channel, one report says `MostRowsForOneDataFile = 2` and `DuplicateFileLabelPairs.Count = 0` simultaneously. `OrdinalIgnoreCase` for both matches the disk matching, which is already case-insensitive.

**Public surface, worth settling before NuGet freezes it.** `SdrfPlexSource.FileName` has no producing code path — `ReadPlex` returns only `Column` or `None` — so it's a member a consumer will write a `case` for that can't occur. `SdrfLabelForm` is public but unreachable through the public API (`ReadLabel` is private; only the two counts are exposed). `NonCanonicalChannelSpellings` also collects *kit* spellings, so the name undersells it. `KitLabel`'s `(?<size>\d+)` group is captured and never read, so a kit-only document reports `ImpliedPlex = 0` while its label says `10` — if that's deliberate, dropping the group name would stop it looking unfinished. And `ReservedWords` re-lists four words `SdrfValidator.ReservedWords` already exposes publicly, with nothing keeping them in step.

**Tests** — the 15 are real and assert exact values, so none would survive the production code being gutted. What's missing is one case per failure mode. `RaggedRows` has no exercising coverage anywhere, and `PXD059974.sdrf.tsv` is *already committed and registered* and measured gives `RaggedRows = 17` — a four-line test. `MostRowsForOneDataFile` and `PlexEvidence` have zero assertions, and the `Audit(SdrfDocument)` overload is never called. Nothing covers the header-case case, an unclassifiable label, a missing data directory, a header with no `comment[label]`, a header-only document, or duplicate `comment[label]` columns (currently one row claims two channels and `ImpliedPlex = 2`, silently). And `ReadExpectations` drops a malformed line silently, guarded only by `Is.Not.Empty` — given this PR exists because counts weren't reproducible, the guard table should guard itself.

Two small ones: `Audit(string)` throws (`MzLibException` on a zero-byte file, `DirectoryNotFoundException` on a bad path) while the type doc says it never does, and the corpus test catches only the first — worth documenting the contract on the path overload or adding a `TryAudit`, since the CMD verb will walk a folder. And `ReadExpectations`' `int.TryParse` uses `CurrentCulture`; not a live bug with `NumberStyles.Integer` and non-negative counts, but the production code got it right.

Not verified: the sixteen corpus counts — no clone of `bigbio/sdrf-annotated-datasets` and `MZLIB_SDRF_CORPUS` unset, so both `[Explicit]` tests skipped. Everything about `files 1236`, the 51/31 TMT split, the 551/781/80 duplicate counts and `unreadable 0` is unconfirmed, including whether fixing item 1 moves any of them. `Test.csproj` is `net10.0-windows`, so my run used a throwaway harness rather than the real `Test` assembly — "fixtures are registered" is read off the csproj diff, not observed in `Test/bin`.

nbollis (approved):

zhuoxinshi (approved):

PR #1245, opened 2026-08-29
Fix two wrong Unimod accessions in the shipped mod files

### LINKED ISSUE #2288: In PTM search assigned unimod id dosen't match real unimod id
Metamorphius assigns the modification "Decarboxylation" the Unimod id 914 and the atom count H-2 C-1 O-1. This doesn't match the entry on Unimod. And "Glu to PyroGlu" is assigned to Unimod id 28, but on Unimod it's the description for Unimod 27.

acesnik:
🤖 Fixes smith-chem-wisc/MetaMorpheus#2288. Two of the hand-written `DR   Unimod; N.` lines in the shipped mod files cite the wrong record.

**`Decarboxylation`** (D/E, `H-2 C-1 O-1`, −30.010565) cited **UNIMOD:914**, which is **Methylmalonylation on Ser** (`H(4) C(4) O(3)`, +100.016044). The record it describes is **UNIMOD:1915, "Decarboxylation"** — identical composition, sites D and E Anywhere.

**`Homoserine lactone on M`** in `protease_mods.txt` (`C-1 H-4 S-1`, −48.003371) cited **UNIMOD:10, `Met->Hse`** (−29.992806). The match is **UNIMOD:11, `Met->Hsl`** — Hsl *is* homoserine lactone, −48.003371, M at any C-term. This is the CNBr cleavage mod, so it is on a live path.

**No searched mass changes. `.mzID` accessions and GPTMD-written protein XML do.** Both compositions were already correct, so nothing about the mass a search looks for moves — verified by loading the fixed resources and printing the masses. What changes is the accession, and the accession is exported: MetaMorpheus's `MzIdentMLWriter.GetUnimodCvParam` emits `UNIMOD:<accession>` verbatim, and mzLib's own `ProteinDbWriter` serialises `Modification.ToString()` — including the `DR   Unimod;` line — into the header of every protein XML it writes. Until now, decarboxylated PSMs went to downstream repositories under an accession denoting a +100 Da modification on Ser while carrying a −30 Da mass on Asp/Glu. Anyone re-exporting old results, or diffing a newly written database against an old one, will see the accession move.

**When each half actually reaches MetaMorpheus, because they do not arrive together.** `protease_mods.txt` is a genuine embedded resource of the mzLib assembly (`Proteomics.csproj:32`, `ProteaseDictionary.cs:45`), so the homoserine-lactone fix propagates with the package. `Mods.txt` does not: MetaMorpheus ships its own duplicate at `EngineLayer/Mods/Mods.txt`, and `GlobalVariables.LoadModifications` reads whatever is in `DataDir\Mods` (`EngineLayer/GlobalVariables.cs:415`). That copy still reads `DR   Unimod; 914.` at line 535 today. So the decarboxylation accession — the one the mzIdentML paragraph above is about — does **not** change in MetaMorpheus output when this merges. The ordering is:

1. this PR merges,
2. an mzLib release carries it,
3. **MetaMorpheus#2749** — which deletes the duplicate `EngineLayer/Mods/Mods.txt` and takes mod data from the package — bumps to that release.

The `.mzID` accession moves at step 3, not step 1. #2749 is currently `CONFLICTING`; resolving it is the follow-up that makes this fix real, and if it stalls, the same two-line edit to MetaMorpheus's own `Mods.txt` is the fallback.

The second half of #2288 is a different failure, in the second commit so it can be dropped independently. `Glu to PyroGlu` targets **Q** at the N terminus with `H-3 N-1`, and **UNIMOD:28 is correct** for that — it is `Gln->pyro-Glu`, "Pyro-glu from Q", −17.026549, Q at any N-term. UNIMOD:27 is the Glu form with a different delta (`H-2 O-1`, −18.010565), so retargeting the accession would break the agreement between accession, mass and motif. Only the name was wrong, and it is renamed to `Gln to PyroGlu`.

**That rename changes a published modification name.** Name-keyed lookups in consumers stop resolving: an existing `.psmtsv`, a saved task `.toml`, or a MetaDraw settings file naming `Glu to PyroGlu on Q` no longer finds a mod. Two tests in this repo indexed it by name and are updated here, which is the same failure mode a downstream caller sees. **Commit 2 is being held out of this PR** on review, and the reason is the same duplication. MetaMorpheus keeps writing `Q[Common Biological:Glu to PyroGlu on Q]` from its own untouched copy, so mzLib's reader of MetaMorpheus output loses the exact-name pin and falls through to `ModificationConverter.GetClosestMod`. That does not fail: `nameContaining` is empty (full-substring `Contains`, and `"Gln to PyroGlu on Q"` does not contain `"Glu to PyroGlu"`), so the `case < 1` arm ranks *every* Q-targeting mod by `GetOverlapScore` and lands on the renamed mod anyway. Right answer, silently, by a ranking artifact — with no test between that and a wrong answer once another similarly named mod exists. The rename is still correct and still wanted; it goes in its own PR once #2749 has landed and there is one `Mods.txt` to rename, together with the three MetaMorpheus fixtures that name the old string. **Commit 1 stands alone and is what this PR is now asking for.**

Added `ShippedUnimodAccessionsMatchTheCitedRecordsComposition`: it walks the embedded mod resources (`Mods.txt`, `RnaMods.txt`, `TMT.txt`, `protease_mods.txt`) and asserts each cited record's composition equals the declared formula. Confirmed it fails against the unfixed files, naming all three offending entries. Two documented skips: entries with no `CF` — the 16 `TMT.txt` references give an MM averaged across label channels on purpose, per that file's own comment — and accessions absent from the shipped `unimod.xml`. One existing test, `TestProteaseDictionaryEmbeddedMods.ParseModificationsFromString_ParsesDatabaseReference_WhenPresent`, asserted the old `"10"`; its expectation is updated in the same commit. Grepping the whole test tree for the accession literals rather than the mod names confirms that assertion is the only hard-coded Unimod accession anywhere in the tests, so the blast radius here is measured rather than assumed.

Three things found by the same audit (103 Unimod references) and left alone:

- `TMT18` cites UNIMOD:2016 (TMTpro), which is **not in the shipped `unimod.xml`**. The accession is likely right and the bundled XML is stale; that deserves its own change, not a guess here.
- `DiLeu-12plex` (MM 145.128005) cites UNIMOD:1327, `SulfurDioxide` (63.9619). A real mismatch, but the shipped Unimod has no DiLeu 12-plex record, so there is no correct accession to substitute.
- The iTRAQ and DiLeu-4plex MM values differ from Unimod by ~2–3 mDa by design (averaged label compositions).

Verified against the shipped `Omics/Resources/unimod.xml` and live unimod.org records 10, 11, 27, 28, 914 and 1915. Note the fix also makes the two copies of the name agree: `unimod.xml` itself carries a record titled "Decarboxylation" at 1915, so `Mods.txt` and the Unimod-sourced entry no longer disagree about which record that name means.

### DIFF
--- a/mzLib/Omics/Resources/Mods.txt
+++ b/mzLib/Omics/Resources/Mods.txt
@@ -177,7 +177,7 @@ NL   ETD:101.0239
 CF   C4 H4 O3
 DR   Unimod; 64.
 //
-ID   Glu to PyroGlu
+ID   Gln to PyroGlu
 TG   Q
 PP   N-terminal.
 MT   Common Biological
--- a/mzLib/Omics/Resources/Mods.txt
+++ b/mzLib/Omics/Resources/Mods.txt
@@ -534,7 +534,7 @@ TG   D or E
 PP   Anywhere.
 MT   Less Common
 CF   H-2 C-1 O-1
-DR   Unimod; 914.
+DR   Unimod; 1915.
 //
 ID   Reduction
 TG   D or S or T
--- a/mzLib/Omics/Resources/Mods.txt
+++ b/mzLib/Omics/Resources/Mods.txt
@@ -177,7 +177,7 @@ NL   ETD:101.0239
 CF   C4 H4 O3
 DR   Unimod; 64.
 //
-ID   Gln to PyroGlu
+ID   Glu to PyroGlu
 TG   Q
 PP   N-terminal.
 MT   Common Biological
--- a/mzLib/Omics/BioPolymerGroup/BioPolymerGroup.cs
+++ b/mzLib/Omics/BioPolymerGroup/BioPolymerGroup.cs
@@ -284,8 +284,15 @@ namespace Omics.BioPolymerGroup
         /// Returns a tab-separated header line for output files, matching the format of <see cref="ToString"/>.
         /// Header includes columns for biopolymer information, quantification, and statistical metrics.
         /// </summary>
+        /// <remarks>
+        /// Virtual because <see cref="ToString"/> is, and the two must be overridden together or a
+        /// writer pairs one type's header with another type's rows. A derived group that hides this
+        /// with <c>new</c> while overriding <see cref="ToString"/> is dispatched inconsistently: a
+        /// call through <see cref="IBioPolymerGroup"/> or a base reference gets this header and the
+        /// derived row. Overriding both keeps the pair together however the call is typed.
+        /// </remarks>
         /// <returns>Tab-separated header string suitable for TSV file output.</returns>
-        public string GetTabSeparatedHeader()
+        public virtual string GetTabSeparatedHeader()
         {
             var sb = new StringBuilder();
             sb.Append("BioPolymer Accession" + '\t');
@@ -306,13 +313,14 @@ namespace Omics.BioPolymerGroup
             #region Quantification Header Building
             if (SampleGroupResults is null) PopulateSampleGroupResults();
 
+            bool quantified = HasAssignedSampleIntensities;
             foreach (var group in SampleGroupResults!)
             {
                 sb.Append($"SpectralCount_{group.Label}\t");
-                if (group.HasIntensityData)
+                if (quantified)
                     sb.Append($"Intensity_{group.Label}\t");
                 sb.Append($"CountOccupancy_{group.Label}\t");
-                if (group.HasIntensityData)
+                if (quantified)
                     sb.Append($"IntensityOccupancy_{group.Label}\t");
             }
             #endregion
@@ -326,6 +334,38 @@ namespace Omics.BioPolymerGroup
             sb.Append("Best Sequence Notch QValue");
             return sb.ToString();
         }
+
+        /// <summary>
+        /// Whether sample intensities have been assigned to this group -- a non-empty sample list
+        /// AND an assigned dictionary -- which is what decides whether the intensity columns exist.
+        /// Named for what it tests rather than for "was quantified": the setters are public and
+        /// non-coalescing, so an assignment is evidence that something populated this group, not
+        /// proof that an engine did.
+        /// Protected rather than private so a derived group can adopt the same predicate instead of
+        /// copying the expression.
+        /// </summary>
+        /// <remarks>
+        /// Deliberately NOT <see cref="SampleGroupResult.HasIntensityData"/>. That answers a per-group
+        /// question -- did this sample group get a value -- and using it to choose columns made rows
+        /// disagree with the header: a group whose samples were all unobserved emitted two columns per
+        /// sample group where its neighbour emitted four, so every value after the disagreement was
+        /// written under the wrong column name.
+        ///
+        /// Both fields are required. <see cref="IHasSampleIntensities.IntensitiesBySample"/> alone is
+        /// not enough: <see cref="ConstructSubsetBioPolymerGroup"/> assigns an empty-but-non-null
+        /// dictionary alongside an empty sample list when no sample matches the requested file, and
+        /// gating on the dictionary alone would then advertise intensity columns that
+        /// <see cref="PopulateSampleGroupResults"/> can never fill.
+        ///
+        /// SCOPE, because this does not make every table rectangular. It makes the four columns per
+        /// sample group agree across the groups of one quantified run. When no engine has run,
+        /// <see cref="PopulateSampleGroupResults"/> builds one result per file in each group's OWN
+        /// spectral matches, so two groups covering different files still emit different column
+        /// counts -- a pre-existing defect on the unquantified path that no per-group flag can fix,
+        /// because a single group does not know the run's full file list.
+        /// </remarks>
+        protected bool HasAssignedSampleIntensities =>
+            SamplesForQuantification is { Count: > 0 } && IntensitiesBySample is not null;
 
         /// <summary>
         /// Returns a tab-separated string representation of this biopolymer group,
@@ -416,23 +456,30 @@ namespace Omics.BioPolymerGroup
                 : AllBioPolymersWithSetMods.Select(p => p.BaseSequence).Distinct().OrderBy(s => s))
                 .ToList();
 
+            bool quantifiedRow = HasAssignedSampleIntensities;
             foreach (var group in SampleGroupResults!)
             {
                 sb.Append(group.SpectralCount);
                 sb.Append("\t");
 
-                if (group.HasIntensityData)
+                if (quantifiedRow)
                 {
-                    sb.Append(group.Intensity);
+                    // Empty rather than 0, so the cell does not assert a measurement that was
+                    // not made. Note this does not make absent distinguishable from a measured zero
+                    // through the engine: QuantificationEngine.OverwriteSampleIntensities drops
+                    // zero-valued cells, so a genuine zero never reaches IntensitiesBySample either.
+                    if (group.HasIntensityData)
+                        sb.Append(group.Intensity);
                     sb.Append("\t");
                 }
 
                 sb.Append(TruncateString(group.FormatOccupancy(orderedKeys, isParentLevel, intensityBased: false)));
                 sb.Append("\t");
 
-                if (group.HasIntensityData)
+                if (quantifiedRow)
                 {
-                    sb.Append(TruncateString(group.FormatOccupancy(orderedKeys, isParentLevel, intensityBased: true)));
+                    if (group.HasIntensityData)
+                        sb.Append(TruncateString(group.FormatOccupancy(orderedKeys, isParentLevel, intensityBased: true)));
                     sb.Append("\t");
                 }
             }
--- a/mzLib/Proteomics/ProteolyticDigestion/protease_mods.txt
+++ b/mzLib/Proteomics/ProteolyticDigestion/protease_mods.txt
@@ -14,7 +14,7 @@ TG   M
 PP   Peptide C-terminal.
 MT   Protease
 CF   C-1 H-4 S-1
-DR   Unimod; 10.
+DR   Unimod; 11.
 //
 ID   Homoserine lactone N-terminal on M
 TG   M
--- a/mzLib/Quantification/QuantificationEngine.cs
+++ b/mzLib/Quantification/QuantificationEngine.cs
@@ -420,14 +420,27 @@ public class QuantificationEngine
         // 5) Normalize combined peptide matrix
         peptideMatrixNorm = Parameters.PeptideNormalizationStrategy
             .NormalizeIntensities(combinedPeptideMatrix);
+
+        // 6) Collapse samples (fractions, technical replicates).
+        //
+        // This belongs here rather than in RunProteinQuant, where it used to sit. Run hands the peptide
+        // matrix on BY VALUE, so collapsing there reassigned a parameter the caller never saw again: the
+        // protein matrix came out collapsed while the caller's peptide matrix stayed uncollapsed, and the
+        // two were then reported side by side keyed differently. Doing it at the end of the peptide stage
+        // is also the order this class's summary already describes -- collapse, then write peptides, then
+        // roll up to proteins -- so the written peptide file now matches its own documentation.
+        peptideMatrixNorm = Parameters.CollapseStrategy
+            .CollapseSamples(peptideMatrixNorm, Parameters.CollapseAggregationStrategy);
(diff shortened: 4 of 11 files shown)


mzLib/Omics/Resources/Mods.txt
```diff
@@ -534,7 +534,7 @@ TG   D or E
 PP   Anywhere.
 MT   Less Common
 CF   H-2 C-1 O-1
-DR   Unimod; 914.
+DR   Unimod; 1915.
```
trishorts, line 537: **The accession this fixes is not the one MetaMorpheus exports — MetaMorpheus ships its own copy of this file.**

The PR body's case for why this matters is the mzIdentML export: `MzIdentMLWriter.GetUnimodCvParam` emits `UNIMOD:<accession>` verbatim. That is true, but it never sees this file. `GlobalVariables.LoadModifications` reads whatever is in `DataDir\Mods`:

```csharp
foreach (var modFile in Directory.GetFiles(Path.Combine(DataDir, @"Mods")))
```

(`EngineLayer/GlobalVariables.cs:415`), and that directory is filled from MetaMorpheus's own `EngineLayer/Mods/Mods.txt`, which on master still reads:

```
ID   Decarboxylation      <- line 530
TG   D or E
PP   Anywhere.
MT   Less Common
CF   H-2 C-1 O-1
DR   Unimod; 914.         <- line 535
```

So after this merges and releases, a MetaMorpheus search still writes `UNIMOD:914` for a decarboxylated PSM. The half of this PR that *does* propagate is `protease_mods.txt`, because that one is a genuine embedded resource of the mzLib assembly (`Proteomics.csproj:32`, `ProteaseDictionary.cs:45`) rather than a duplicated file — so the homoserine-lactone fix lands and the decarboxylation fix does not, which is a confusing pair to ship without saying so.

The route already exists: your own MetaMorpheus#2749 deletes `EngineLayer/Mods/Mods.txt` and takes the mod data from the mzLib package. So the real ordering is #1245 -> mzLib release -> #2749 bumps to it. Worth naming #2749 in the description as the PR that makes this fix reach the tool, because #2749 is currently `CONFLICTING`, and if it stalls this change is invisible to everything the body describes. If it is going to sit a while, the same two-line edit to MetaMorpheus's `Mods.txt` is a cheap companion PR that makes the fix real today.

mzLib/Omics/Resources/Mods.txt
```diff
@@ -177,7 +177,7 @@ NL   ETD:101.0239
 CF   C4 H4 O3
 DR   Unimod; 64.
 //
-ID   Glu to PyroGlu
+ID   Gln to PyroGlu
```
trishorts: **The rename desynchronises the two copies of this file, and mzLib's own reader of MetaMorpheus output is the first thing that notices.**

You flagged that name-keyed lookups in consumers stop resolving. Concretely on this side: `ModificationConverter.AllModsKnown` is `Mods.AllKnownProteinModsDictionary` (`Readers/ExternalResults/SupportClasses/ModificationConverter.cs:16`), which is built from *this* file via `Mods.MetaMorpheusProteinModifications` (`Omics/Modifications/Mods.cs:25-28`). And `GetModificationDictionaryFromFullSequence` is what parses a MetaMorpheus full sequence:

```csharp
if (!allModsKnown.TryGetValue(modId, out mod))
    toThrow = new MzLibException("Could not find modification while reading string: " + fullSequence);
```

MetaMorpheus keeps writing `Q[Common Biological:Glu to PyroGlu on Q]`, because its own `Mods.txt:180` is untouched by this PR (see my other comment). So every such sequence now misses the exact-name lookup and falls through to `GetClosestMod` at line 125 — the fuzzy longest-common-substring matcher. It will almost certainly still land on `Gln to PyroGlu on Q`, 17 characters of overlap being hard to beat, so this degrades *silently and probably correctly*, which is the worst shape a regression can take: no exception, no wrong answer today, and no test standing between you and a wrong answer tomorrow. The two tests this PR updates (`RetentionPredictableTests.cs:146`, `SequenceFormattingTests.cs:272`) were the only exact-name pins on this mod.

Three MetaMorpheus fixtures also name the old string and would need updating in lockstep: `Test/TestData/ModificationAnalysis/GPTMDTaskconfig.toml` (in the `ListOfModsGptmd` list) and both `Test/MetaDraw/112MetaDrawSettingsSavedForTestCoverage_{Success,Failures}.xml`.

I would still take the rename — `Gln` is right and `Glu` is wrong — but it wants to happen in one place, not ahead of the other copy. You already made commit 2 separable, so the cleanest order is: merge commit 1 now, and hold commit 2 until MetaMorpheus#2749 has landed and there is only one `Mods.txt` left to rename.

trishorts (commented):
Two comments inline. The accession work is right and the audit behind it is the good kind — 103 references walked, the three you did not fix each with a stated reason, and a test that fails against the unfixed files. My concern is not the change but its reach: MetaMorpheus does not read this file, so as it stands the mzIdentML impact the body describes does not actually happen until MetaMorpheus#2749 merges.

Verified rather than assumed: `GlobalVariables.LoadModifications` enumerates `DataDir\Mods` (`EngineLayer/GlobalVariables.cs:415`); MetaMorpheus's `EngineLayer/Mods/Mods.txt:535` still says `DR   Unimod; 914.` and its line 180 still says `ID   Glu to PyroGlu`; `protease_mods.txt` by contrast is a real embedded resource (`Proteomics.csproj:32`) so that half does propagate; and MetaMorpheus#2749 is the open PR that removes the duplicate, currently `CONFLICTING`.

Not blocking on my side once the ordering is stated in the description.

nbollis (dismissed):

mzLib/Omics/Resources/Mods.txt
acesnik: 🤖 Confirmed on all four points, checked rather than taken: MetaMorpheus's `EngineLayer/Mods/Mods.txt` still reads `ID   Decarboxylation` at 530 and `DR   Unimod; 914.` at 535, still `ID   Glu to PyroGlu` at 180, and #2749 is `dirty` with no push since 08-27. So you are right that as it stands the mzIdentML impact the body claims does not happen — `protease_mods.txt` propagates because it is a real embedded resource, and `Mods.txt` does not because MetaMorpheus ships a duplicate. Shipping the homoserine-lactone half working and the decarboxylation half not, without saying so, is the confusing pair you describe.

Description updated to state the ordering — #1245 → mzLib release → #2749 bumps to it — and to say plainly that the mzIdentML change lands only at the third step.

On the cheap companion PR: I would rather not, because it would make a third place where these two values live and someone would have to remember to unwind it. #2749's conflict is mine and it is the thing that removes the duplicate permanently, so I would rather spend the same effort there. If it is still stalled when this releases, the two-line edit is the fallback and I will take it then.

mzLib/Omics/Resources/Mods.txt
acesnik: 🤖 Agreed, and I will hold commit 2. Your mechanism is right and the route is slightly different in a way that makes your point stronger rather than weaker, so it is worth writing down.

`AllModsKnown` is exact-name, so `Glu to PyroGlu` misses after the rename — that part is as you describe. But `GetClosestMod` does not reach the overlap scorer the way it looks: `nameContaining` is `IdWithMotif.Contains(trimmedName)`, full-substring containment, and `"Gln to PyroGlu on Q".Contains("Glu to PyroGlu")` is false, so `nameContaining` is empty and `goodMatches` with it. There is exactly one PyroGlu mod left in the file after the rename, so nothing else rescues it either.

What saves it is the `case < 1` arm, which falls back to `nameContaining.Union(motifMatching)` — the union, so effectively every mod that can sit on Q or on X — and ranks that by `GetOverlapScore` against `Glu to PyroGlu`. `Gln to PyroGlu on Q` wins because 13 of 14 characters line up, so you land on the right answer, silently, exactly as you predicted.

The difference matters because it is not a near-exact match beating distant ones; it is the best of *all* Q-targeting mods on an overlap score. Correctness there is a ranking artifact, and any future mod whose name overlaps `Glu to PyroGlu` more can displace it with no test failing. That is a sharper version of your "no test standing between you and a wrong answer tomorrow", and it is the argument for your ordering rather than mine.

So: commit 1 goes now, commit 2 becomes its own PR gated on #2749, and I will fix #2749's conflict. The three MetaMorpheus fixtures you listed go in that PR too, in lockstep.

trishorts:
@acesnik — the body and the diff had come apart, so I have pushed a revert of `9eb2d453` (`0d6b922d`) to bring them back together.

The body says **"Commit 2 is being held out of this PR"** and **"Commit 1 stands alone and is what this PR is now asking for"**, and I think that is exactly the right call — but the commit was still on the branch. So the description told a reviewer the rename was gone while the diff still carried it, and @nbollis's approval predates the rewritten body. That combination is worse than not having updated the body at all: someone trusting the description would have merged the half we agreed to hold.

```
before   6 files, 57 insertions,  8 deletions
after    4 files, 52 insertions,  3 deletions
```

No occurrence of either PyroGlu spelling survives in the diff. What is left is commit 1 alone: the `Mods.txt` accession, `protease_mods.txt`, and the two tests that pin them.

Reverted rather than rebased out, deliberately — dropping a commit from the middle means rewriting published history on a branch that is not mine, and the resulting diff is identical either way. If you would rather the history were clean, force-push a rebase over it and I will not mind.

```
Affected tests     149 passed, 0 failed
Full offline suite 5844 passed, 0 failed, 29 skipped
```

Your analysis in the second thread was better than my original comment, and worth not losing: I had assumed `GetClosestMod` would rescue the rename through near-match ranking, and you showed `nameContaining` is empty because `"Gln to PyroGlu on Q"` does not *contain* `"Glu to PyroGlu"`, so it is the `case < 1` arm ranking **every** Q-targeting mod by overlap score. Right answer by a ranking artifact rather than by a match, which is a sharper argument for holding the rename than the one I made.

Two things for you:

- @nbollis's approval will have been dismissed by this push. That is the correct outcome rather than a nuisance — it was given against a diff that still contained the rename.
- The branch is `BEHIND` master. Worth updating right before merge rather than now, since master has been moving several times a day.

The rename still wants doing. When #2749 lands and there is one `Mods.txt`, it is a clean small PR with the three MetaMorpheus fixtures alongside it.

mzLib/Omics/Resources/Mods.txt
trishorts: Addressed. The description now states the ordering explicitly — merge → mzLib release → #2749 bumps to it — and says plainly that the `.mzID` accession moves at step 3, not step 1. That was all I wanted; the fix itself was never in question. Taking your point on the companion PR too: a third copy of these values is worse than waiting for the one that deletes the duplicate. Resolving.

mzLib/Omics/Resources/Mods.txt
trishorts: Addressed — commit 2 is reverted (`Revert "Rename ..."`, 2026-09-01) and the diff is now the two accession lines plus tests. Thanks for running the `GetClosestMod` path down properly; your correction makes the case stronger than my version of it, and it belongs in the follow-up PR description. Resolving.

trishorts (approved):
Both accessions verified independently against the shipped `Omics/Resources/unimod.xml` rather than taken from the description:

| cited | record | composition | mass |
|---|---|---|---|
| ~~914~~ | Methylmalonylation on Serine | `H(4) C(4) O(3)` | +100.016044 |
| **1915** | Decarboxylation | `H(-2) C(-1) O(-1)` | −30.010565 |
| ~~10~~ | Met->Hse (Homoserine) | `H(-2) C(-1) O S(-1)` | −29.992806 |
| **11** | Met->Hsl (Homoserine lactone) | `H(-4) C(-1) S(-1)` | −48.003371 |

Both new accessions match the `CF` lines exactly (`H-2 C-1 O-1` and `C-1 H-4 S-1`); both old ones do not. The homoserine one is the CNBr mod, so it was live.

`ShippedUnimodAccessionsMatchTheCitedRecordsComposition` is the right shape for this: it pins the invariant (declared formula equals cited record's composition) rather than the two literals, so the next hand-written `DR` line cannot go wrong quietly. The two documented skips are both justified — the TMT entries omit `CF` deliberately, and a missing record is a stale-XML problem rather than a wrong accession.

Both of my threads are addressed and resolved: commit 2 is reverted, and the description now names the merge → release → #2749 ordering and says the `.mzID` accession only moves at step 3.

Approving.

Issue #33, opened 2017-01-06
mzml writing output does not work for TDPortal


stefanks:
- [x] MsConvert
- [x] SeeMs

Need to:
- [x] Write chromatogram
- [x] Write Indexed mzml

stefanks:
Maybe instead of writing index it is easier to write an unindexed mzml file

stefanks:
Rough structure of a method for writing indexed mzmls:

1.  Write non-indexed mzML to memory
2.  Calculated indices of scans by reading the stream as a binary stream
3.  Discard the non-indexed mzML from memory
4.  Write the indexed mzML using the xsd generated class and the computed indices

The specification is here: https://github.com/HUPO-PSI/mzML/raw/master/specification_document/mzML1.1.0_specificationDocument.doc

Note that there is a need to also write the indexListOffset element, which is the file pointer offset (in bytes) of the 'indexList' element, and the fileChecksum element which is the SHA-1 checksum from beginning of file to end of 'fileChecksum' open tag.

acesnik:
🤖 Closing — all four boxes are ticked and the two that were still under discussion are implemented.

**Indexed mzML** (`Readers/MzML/MzmlMethods.cs:1074+`) follows the plan sketched in the second comment almost step for step: build the `indexedmzML`, write it, rescan the file for `<indexList` to get `indexListOffset`, then fill in a per-scan offset array.

**Chromatogram** (`MzmlMethods.cs:346+`) writes a `chromatogramList` carrying a total ion current chromatogram.

Worth noting the fallback floated here — "maybe instead of writing index it is easier to write an unindexed mzml file" — turned out not to be needed; the harder option got done.

Reopen if TDPortal still rejects mzML that mzLib writes, since that was the original symptom and this only confirms the writer's side of it.

PR #1246, opened 2026-08-29
Include the byte-order mark in the indexed-mzML fileChecksum

acesnik:
🤖 The `<fileChecksum>` mzLib writes into an indexed mzML is a SHA-1 over the file **minus its first three bytes**. `MzmlMethods.GetSHA1Hash` read the file through a `StreamReader`, which consumes the UTF-8 byte-order mark `XmlWriter.Create` emits, then re-encoded each line with `Encoding.UTF8.GetBytes`. No checksum-verifying consumer can reproduce the value we write.

**This changes produced output.** Every indexed mzML mzLib writes from now on gets a different `<fileChecksum>`. Nothing else changes: on a 2,094,480-byte rewrite of `Test/DataFiles/SmallCalibratibleYeast.mzml`, before and after are byte-identical except for 39 of the 40 hex characters inside `<fileChecksum>` (one digit happened to coincide). Same length, same spectra, same metadata, same `<indexList>` offsets.

### Which bytes the spec means

`mzML1.1.1_idx.xsd`, on the `fileChecksum` element:

> SHA-1 checksum from beginning of file to end of 'fileChecksum' open tag.

Beginning of file is byte 0, so a BOM counts. "End of the open tag" is the closing `>` of `<fileChecksum>`, not the value inside it. The range is therefore `bytes[0 .. indexOf("<fileChecksum>") + 14)`.

Measured against the 22 mzMLs checked into this repo:

- 13 verify exactly on that range. All are msconvert output with no BOM, so for them byte 0 is the `<` of `<?xml`.
- 3 verify only if the BOM is skipped: `Development/DeconvolutionDevelopment/TestData/Averaged_221110_{Cyto,HGH,Ubiq}Only.mzML`. Those are mzLib's own output, and they carry this bug.
- 6 match nothing, having been edited or truncated after they were written. That is inferred from their names and contents, not confirmed.

Two msconvert files from a local vignette dataset also verify on that range, which is why this reads as our defect rather than a disagreement about the spec.

The index offsets were never wrong. They are absolute from byte 0 and already count the BOM — verified by seeking to `<indexListOffset>` and to the first `<offset>` in a BOM-bearing mzLib-written file and landing exactly on `<indexList ` and `<spectrum `. That is why every reader works and only a verifying consumer notices.

### The fix

`GetSHA1Hash` hashes the raw bytes in 64 KiB blocks and stops at the end of the open tag, holding back 13 bytes between reads in case the tag straddles a block boundary. Hashing bytes rather than re-encoded text also removes the old code's sensitivity to line endings and to the final newline. The public signature is unchanged; the now-unused `NewLine` field goes away.

### Test

`Test/FileReadingTests/SpectraFileReading/TestMzmlFileChecksum.cs` writes an indexed mzML, then computes SHA-1 over `bytes[0 .. end of open tag)` independently and compares against what was written. Two cases: a two-scan `FakeMsDataFile`, and a rewrite of `SmallCalibratibleYeast.mzml`, which is large enough that the tag is found in a later read block than the one the file starts in. A third test pins the index offsets as absolute from byte 0.

Both checksum tests fail against the unfixed code:

```
Failed WrittenFileChecksumCoversTheWholeFileIncludingTheByteOrderMark
  the written checksum skips the byte-order mark
  Expected: "f9132071541a1ac9bc4459c8a3b05e4fe95df933"
  But was:  "25b044f7a62668ec0037988686c7a7108448f78c"

Failed WrittenFileChecksumIsCorrectForAFileLargerThanTheReadBuffer
  the written checksum skips the byte-order mark
  Expected: "ab6da85f1fb8ab178780728721d00ed10350abff"
  But was:  "ce257dde2803e67cd6a0a4c34b157cf5713beaea"
```

The offsets test passes unfixed, as it should.

### Nothing depended on the old value

No test, fixture or golden file asserts an emitted `<fileChecksum>`; the occurrences in checked-in mzMLs are input that gets read, never compared. `GetSHA1Hash` has no callers outside `MzmlMethods` in mzLib, MetaMorpheus or Spritz. The three mzLib-written `Averaged_221110_*.mzML` fixtures keep their now-incorrect checksums — nothing verifies them, and regenerating them would churn test input for no gain.

### Provenance

This is a defect in the indexed-mzML writer added under #33, not what that issue asked for. It surfaced from MetaMorpheus#1955, where a Proxl maintainer reported unindexed calibrated mzML; that issue is closed as fixed, and its closing comment says the checksum is being looked at separately.

### Not verified

`Test` is `net10.0-windows`, so it cannot run on macOS. The new tests were exercised through a throwaway net10.0 NUnit harness compiling only `FakeMsDataFile.cs` and the new test file, run in both the fixed and unfixed states; the harness is deleted. The rest of the suite has not been run locally — CI on windows-latest is the check. Nothing outside this workspace was checked for dependence on the old value.


trishorts (approved):
Approving. I read the rewritten `GetSHA1Hash` as a buffer loop rather than taking the description for it, and the four things that usually break one are all handled:

- **Hold-back length.** `overlap = openTag.Length - 1` = 13, which is exactly right: any occurrence starting at or before `pending - 14` was already scanned this iteration, so retaining 13 bytes is the minimum that cannot lose a straddling tag, and it is not one short.
- **EOF.** `bytesRead == 0` with no tag found hashes the remaining `pending` bytes through `TransformFinalBlock` rather than dropping them, so the "if the tag is absent the whole file is hashed" contract in the new doc comment is what actually happens.
- **Buffer exhaustion.** `buffer.Length` is `65536 + overlap` and after compaction `pending` is always 13, so `buffer.Length - pending` is never 0 and a `Read` of count 0 can never be mistaken for EOF.
- **Small and empty files.** A file shorter than the tag takes `hashable == 0`, skips the compaction and terminates on the next read; an empty file hashes zero bytes. Neither loops.

The two-write scheme is what makes this safe to change at all: the first write emits `fileChecksum = "0"` and the second is byte-identical up to the open tag, so hashing the first and stamping the second is sound. Good that a test pins the index offsets as absolute from byte 0 as well, since that is precisely why this was only ever wrong for a verifier and never for a reader.

One thing I would still do, non-blocking. The three `Development/DeconvolutionDevelopment/TestData/Averaged_221110_*.mzML` fixtures are, by your own audit, mzLib output carrying this bug, so after this merges the repo ships three files that fail the contract its newest test asserts. "Nothing verifies them" is true today and is exactly the sentence that stops being true the moment someone writes a verifying reader. Regenerating them is churn in test input, which I agree is not free, but a line in `TestMzmlFileChecksum.cs` naming them as known-stale would stop the next person reading them as reference output.

Nice catch on the provenance chain back to MetaMorpheus#1955 too, which closes a loop that was left explicitly open.

mzLib/Readers/MzML/MzmlMethods.cs
```diff
@@ -1201,42 +1200,79 @@ instrumentModel is not null
...
-                            foundChecksumField = true;
-                        }
+                    if (bytesRead == 0)
+                    {
+                        sha1hash.TransformFinalBlock(buffer, 0, pending);
+                        return sha1hash;
+                    }
 
-                        int bytesRead = Encoding.UTF8.GetBytes(line, 0, line.Length, buffer, 0);
+                    // Hold back the last (openTag.Length - 1) bytes; the tag may straddle two reads.
```
trishorts, line 1241: This is the subtlest line in the change and I do not think anything pins it.

`WrittenFileChecksumIsCorrectForAFileLargerThanTheReadBuffer` shows the tag being found in a *later* window than the first, which is not the same thing as the tag *straddling* a window boundary -- and straddling is the only reason `overlap` exists. I worked out where the tag actually lands in that file rather than assuming:

- The first read requests `buffer.Length` and leaves `pending == 13`; every read after it contributes 65536 fresh bytes, so window `k` begins at file offset `65536 * k`.
- Your body gives the rewritten file as 2,094,480 bytes. `2094480 = 65536 * 31 + 62864`, so the last window holds 62,864 bytes.
- The tail after the open tag is 40 hex + `</fileChecksum>` + newline + `</indexedmzML>`, about 70 bytes, so the tag ends near offset 62,794 within that window -- roughly 2.7 KB clear of the boundary.

Setting `overlap = 0` should therefore leave both checksum tests green. I derived that rather than building the mutant, so treat the number as arithmetic rather than as a run, but the margin is large enough that I would not expect it to be close.

A direct test is cheap and does not need the writer: hand-build a byte array with `<fileChecksum>` positioned across offset 65536, write it to a temp file, and compare `GetSHA1Hash` against `SHA1.HashData` over the same prefix. That fails with `overlap = 0` and with a wrong `BlockCopy`, which is exactly the pair of mistakes this line is guarding against.

mzLib/Readers/MzML/MzmlMethods.cs
```diff
@@ -1201,42 +1200,79 @@ instrumentModel is not null
             }
         }
 
+        /// <summary>
+        /// SHA-1 of the raw bytes of <paramref name="filePath"/> from the first byte of the file
+        /// through the closing '&gt;' of the &lt;fileChecksum&gt; open tag, per the indexedmzML schema:
+        /// "SHA-1 checksum from beginning of file to end of 'fileChecksum' open tag."
+        /// If the tag is absent the whole file is hashed.
```
trishorts, line 1207: This sentence is a new contract on a `public` method, and nothing exercises it -- the `bytesRead == 0` branch that implements it is only reachable for a file with no `<fileChecksum>` at all, and all three tests write indexed mzML.

The repo has five such files, incidentally: `FAIMS_snip_10665-11578.mzML`, `BinGenerationTest.mzML`, `MS3_TMT11_Mouse_snip.mzML` and the two `XICData/..._noRt.mzML` have no `<fileChecksum>` element. Any of them would do, or a three-line temp file, asserting `GetSHA1Hash` equals `SHA1.HashData` over the whole thing.

Small, and it also happens to be the branch that decides what a caller gets for a truncated or non-mzML input, which is worth having written down in a test rather than only in the summary.

trishorts (commented):
Read the diff directly (two files). The fix is right, and I reproduced the central measurement independently rather than taking it.

**I re-ran your survey and got your numbers exactly: 13 / 3 / 6.** Own script -- SHA-1 over `bytes[0 .. find("<fileChecksum>") + 14)` for every tracked mzML, compared against the embedded value. Thirteen match from byte 0, six match nothing, and the three that match only with the BOM skipped are precisely the ones you named: `Development/DeconvolutionDevelopment/TestData/Averaged_221110_{Cyto,HGH,Ubiq}Only.mzML`, and they are the only three in the repo carrying a BOM. That is about as clean a confirmation of the spec reading as the repo can give.

**One thing worth adding to the body, because it cost me a false negative.** My first run reported *nothing* matching -- all 22. The cause is this repo's own `.gitattributes` (`* text=auto`) plus `core.autocrlf=true`, the Windows default: every checked-in mzML is smudged to CRLF in the working tree, which moves every byte and breaks every checksum. I only reproduced your 13/3/6 after reading the raw blobs with `git cat-file` instead of the working tree.

So the measurement is platform-dependent: on macOS or Linux it reproduces from the working tree, and on any Windows checkout it reports 0 of 22. Anyone re-checking your claim on Windows will conclude the premise is wrong. A line in the body saying which state the files have to be in would save them.

The durable version of that is `*.mzML binary` (and `*.mzml`) in `.gitattributes`, so the fixtures stay byte-stable across platforms. That is a repo-wide change that would touch every mzML in every working tree, so it belongs in its own issue rather than here -- happy to file it if you would like.

**Also verified, since the whole change hangs on it:** the two-pass write is sound. `CreateAndWriteMyMzmlWithCalibratedSpectra` serializes once with `fileChecksum = "0"` (line 1186), hashes, then re-serializes with the real value. Every byte before the open tag is identical between the two passes, so hashing the placeholder file over that range gives the final file's checksum. And `fileChecksum` is the last element of `indexedmzML`, after `indexList` and `indexListOffset`, so the value growing from 1 character to 40 cannot move anything either of those points at. Your "index offsets were never wrong" holds for a structural reason, not just an empirical one.

**The block-reading logic is correct as written.** I traced the buffer bookkeeping: after each `TransformBlock` the retained tail is exactly `openTag.Length - 1` bytes, `pending` returns to 13, and each subsequent window covers 65536 fresh file bytes, so `buffer[0..)` is always unhashed when the tag is found and `TransformFinalBlock(buffer, 0, bytesThroughOpenTag)` is right. `pending` can never reach `buffer.Length` going into a read, so the zero-count read that would be misread as EOF cannot happen.

Two coverage notes inline. Both are about the new code being under-pinned rather than wrong, and neither blocks -- already approved, all five checks pass.

nbollis (approved):

PR #1247, opened 2026-08-29
Skip a Koina test when Koina fails to run the model, instead of failing it

### LINKED ISSUE #1241: Koina GPU failures fail external-service tests instead of skipping them
The `external-service-tests` job goes red when Koina's inference server answers but its GPU
fails mid-inference. That is a third-party availability problem, and the skip machinery in
`ExternalServiceTestHelper` was built precisely so those do not read as code failures — but
the Koina path has two gaps that let this one through.

### What happened

On the run for #1240 ([job log](https://github.com/smith-chem-wisc/mzLib/actions/runs/33190927337/job/98922935402)),
195 of 198 tests passed, 2 skipped, and `TestAltimeterChargeStateBoundaries`
(`mzLib/Test/KoinaTests/FragmentIntensityPrediction/Altimeter2024IntensitiesTests.cs:42`) failed:

```
Expected: No Exception to be thrown
But was:  <System.Net.Http.HttpRequestException: Request failed with status 400 Bad Request:
{"error":"in ensemble 'Altimeter_2024_intensities', PyTorch execute failure: ...
    out = torch.einsum('abc,bd->adc', inp, self.embed)
RuntimeError: CUDA error: CUBLAS_STATUS_EXECUTION_FAILED when calling
  `cublasSgemmStridedBatched(handle, opa, opb, m, n, k, &alpha, a, lda, stridea, b, ldb, strideb, &beta, c, ldc, stridec, num_batches)`"}>
```

That is Koina's own CUDA/cuBLAS call failing inside the model, not a rejection of our request
payload. The PR under test touched only `Omics/BioPolymerGroup/BioPolymerGroup.cs` and
`Test/Quantification/QuantificationDeliveryTests.cs`, and every other recent run of
`Build and Test` was green. Notably, UniProt was also returning HTTP 500 during the same run —
and *that* one was correctly skipped, with the intended message on the log.

### Why the guard missed it

Two independent reasons, and both need addressing:

1. **The readiness probe is fixture-level and coarse.** `KoinaLiveTestFixture` probes
   `/v2/health/ready` once per fixture (`mzLib/Test/KoinaTests/KoinaLiveTestFixture.cs:21-25`).
   Koina reported itself ready — the server was up. What was broken was inference on one
   model, which a health endpoint does not and cannot report. A per-call classification is
   needed, not just a per-fixture probe.

2. **A server-side fault arrives dressed as a 400.**
   `ExternalServiceTestHelper.ThrowIfUnavailable` classifies on status code only, treating
   408/429/5xx as unavailable (`mzLib/Test/ExternalServiceTestHelper.cs:109-123`). Koina
   returns **400** with the real cause in the JSON body, so even if this call were routed
   through the helper the status check alone would still call it a genuine failure.
   `HTTP.InferenceRequest` turns any non-success status into a bare `HttpRequestException`
   carrying that body as text (`mzLib/PredictionClients/Koina/Client/HTTP.cs:28-33`).

### The shape of a fix

The blunt version — route the Koina fixtures through `ExternalServiceTestHelper.RunAsync`,
which already skips on `HttpRequestException` — is **wrong on its own**: it would also swallow
genuine contract breaks, since a real regression in how we build a request would surface as a
400 too. The distinction has to be drawn on *cause*, not on transport.

A better split is to teach the classifier to read the body: a Koina error body naming an
inference-side fault (`PyTorch execute failure`, `CUDA error`, `CUBLAS_`, `failed to load model`)
is an availability problem and should skip; a 400 complaining about our inputs is our bug and
must still fail. `ThrowIfUnavailable` already has precedent for body-sniffing — it does exactly
this for UniProt's error-in-a-200 case. That likely also means having the Koina client preserve
the error body in a typed way rather than only in an exception message string, so the test layer
is not pattern-matching on prose.

### Not urgent, but recurring

Nothing is broken in library code and no release is blocked — the cost is a red required-looking
check on unrelated PRs and the review time spent working out that it is unrelated. This is the
second flavour of "external service misbehaves in a way the guard does not cover", so it is worth
fixing the classifier rather than re-running the job each time.

Filed while reviewing #1240, whose failure this was.

trishorts:
Closes #1241.

Koina answers **400 for both** "your request is malformed" and "my GPU fell over mid-inference",
putting the difference only in the response body. mzLib #1240 carried a red `external-service-tests`
check for the whole of its review because of the second kind:

```json
{"error":"in ensemble 'Altimeter_2024_intensities', PyTorch execute failure: …
 RuntimeError: CUDA error: CUBLAS_STATUS_EXECUTION_FAILED when calling
 `cublasSgemmStridedBatched(...)`"}
```

That's Koina's own cuBLAS call failing. The skip machinery missed it for **two independent reasons**,
and both are fixed here.

### 1. Classify where the body is

`HTTP.InferenceRequest` now throws `KoinaServiceException` when the body names a server-side
execution failure, carrying the server's own text on `ServerError`. Anything else stays a plain
`HttpRequestException`, exactly as before. It **derives from `HttpRequestException`**, so every
existing `catch` keeps working unchanged.

This is a production improvement, not only a test one: a caller that can't tell *retry, the GPU
glitched* from *fix your request* either retries forever or gives up on a request that would have
succeeded.

### 2. The fixture probe is necessary but not sufficient

`KoinaLiveTestFixture` probes `/v2/health/ready` once per fixture — and Koina reported itself
**ready** throughout. The server was up; one model's GPU was not, and no health endpoint reports
that. So the classification has to happen per *call*, and the fixture gains a `[TearDown]` that
rewrites a failure caused by a `KoinaServiceException` into a skip.

**Deliberately narrower than `ExternalServiceTestHelper.RunAsync`,** which skips on any
`HttpRequestException`. Only a service fault is skipped here, so a 400 meaning "your request is
wrong" — which is exactly what a regression in how we build a request looks like — still fails. The
marker list is narrow for the same reason and should stay that way: an unrecognised error is treated
as our problem.

### Two things this got wrong first

Both caught by tests written to catch them, and both worth recording because the failure mode was
silence rather than noise:

- The guard began as an **`IWrapTestMethod` attribute on the abstract base class**. NUnit does not
  apply that to derived fixtures, so it never fired — which would have left all fifteen Koina
  fixtures exactly as exposed as before, with nothing to say so. `[TearDown]` inheritance is part of
  NUnit's documented contract and cannot fail that way.
- The test *observing* the guard was itself a `[TearDown]` in the derived fixture. NUnit runs
  TearDowns derived-first, base-last, so it read the result before the guard had rewritten it and
  reported a failure that was its own. It now observes the finished child results from
  `[OneTimeTearDown]`.

### Tests

15 unit tests on the classifier, using the real response bodies, no network — including four
**request-rejection** cases that must *not* skip, and a non-JSON body (a proxy's HTML error page)
that must be classified rather than thrown on.

Two more prove the guard actually fires: one throws a service fault and must come back Skipped, one
is ordinary and must be left alone.

397 passed, 1 skipped across the Koina fixtures.


mzLib/Test/KoinaTests/TestKoinaServiceFaultClassification.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// </summary>
+    [TestFixture]
+    [ExcludeFromCodeCoverage]
+    public class TestKoinaServiceFaultIsSkippedNotFailed : KoinaLiveTestFixture
+    {
+        /// <summary>
+        /// The base class's [OneTimeSetUp] probes the live server, which this fixture must not do.
+        /// </summary>
+        [OneTimeSetUp]
+        public new void EnsureKoinaReachable() { }
```
acesnik: This fixture says it "needs no network", but NUnit runs *both* the base and the hiding `[OneTimeSetUp]` — reflection returns two distinct methods for a `new`, and only one for an `override`. Compiled your files with `ExternalServiceTestHelper` stubbed to log instead of dial:

```
LIVE-PROBE-WOULD-FIRE service=Koina url=https://koina.wilhelmlab.org:443/v2/health/ready
                      fixture=Test.KoinaTests.TestKoinaServiceFaultIsSkippedNotFailed
```

Two consequences. The fixture carries no `[Category]`, so `cat != ExternalService` in `required.runsettings` selects it into the **required** job — the one #1140 exists to keep off the network. And `EnsureReachable` calls `Assert.Ignore` from `[OneTimeSetUp]`, so when Koina is down the fixture proving the guard works is itself skipped. That is the failure you rejected the `IWrapTestMethod` approach to avoid: the guard goes unproven with nothing to say so.

Making the base method `virtual` and this one `override` fixes it — verified. A non-probing base that `KoinaLiveTestFixture` itself derives from also works, and has the advantage of keeping this proof honest about inheriting through one level.

mzLib/Test/KoinaTests/KoinaLiveTestFixture.cs
```diff
@@ -23,5 +37,45 @@ public void EnsureKoinaReachable()
...
+        public void SkipWhenKoinaFailedToRunTheModel()
+        {
+            var result = TestExecutionContext.CurrentContext.CurrentResult;
+            if (result.ResultState.Status != TestStatus.Failed) return;
+
+            // The recorded message carries both the exception type name and the server's response
+            // body. Either alone identifies the fault; requiring the type name keeps this from firing
+            // on a test that merely mentions CUDA in an assertion message.
+            string message = result.Message ?? string.Empty;
+            if (!message.Contains(nameof(KoinaServiceException), StringComparison.Ordinal)) return;
```
acesnik, line 80: Matching a substring of the recorded message rewrites any failure whose message happens to name the type:

| failure | result |
|---|---|
| `throw new KoinaServiceException(...)` | Skipped (intended) |
| `Assert.DoesNotThrow` wrapping a service fault | Skipped (intended) |
| `Assert.That(() => { }, Throws.InstanceOf<KoinaServiceException>())`, nothing thrown | **Skipped** |
| ordinary assertion failure | Failed (intended) |

Row 3 is latent, not live — every `Throws` in the fourteen live fixtures is `Assert.Throws<ArgumentException>`, so nothing regresses today. But this PR makes `KoinaServiceException` public, and `Assert.Throws<KoinaServiceException>` is the natural next test someone writes against `InferenceRequest`; it would report Skipped instead of Failed. The comment above guards the reverse direction only ("a test that merely mentions CUDA").

mzLib/Test/KoinaTests/KoinaLiveTestFixture.cs
```diff
@@ -23,5 +37,45 @@ public void EnsureKoinaReachable()
...
+
+            string skipped =
+                "Skipping external-service test: Koina failed to run the model "
+                + $"({FirstLine(KoinaServiceException.ExtractServerError(message))}). "
```
acesnik: For the case that motivated the PR, this drops the server's error entirely. `TestAltimeterChargeStateBoundaries` uses `Assert.DoesNotThrow(..., $"Charge {charge} should be valid")`; replicated with the real #1240 body:

```
Skipping external-service test: Koina failed to run the model (  Charge 1 should be valid).
This is a third-party availability problem, not a code failure.
```

`CUBLAS` appears nowhere. Two things compound: `ExtractServerError` is the identity function on an NUnit failure message, since that message is never valid JSON (it is prefixed with the type name, or with `Expected: No Exception to be thrown`), so `JToken.Parse` throws and the raw string comes back; then `FirstLine` cuts at the first newline, which for `Assert.DoesNotThrow` with a user message is the user message. So `ServerError` — the property §1 exists to populate — never reaches CI. Reporting the first line that *contains a matched marker*, or the whole message truncated, would put the cuBLAS text back on the log.

mzLib/PredictionClients/Koina/Client/KoinaServiceException.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// test suite exists to catch.
+        /// </remarks>
+        private static readonly string[] ServiceFaultMarkers =
+        [
+            "pytorch execute failure",
+            "torchscript",
+            "cuda error",
+            "cublas_",
+            "cudnn_",
+            "out of memory",
```
acesnik: Eight of the nine markers are immune to what we send, as the remark claims. This one is not: **a GPU OOM is exactly what an oversized request looks like server-side.** Batch too large, too many peptides in one call, a pathological sequence length — the fault is ours, the body says out of memory, and the guard converts a real failure to Skipped.

It is also the plausible regression in this area, since batch size is the sort of thing that gets tuned: someone raises it, Koina OOMs, and CI goes green with a skip. Dropping it, or requiring co-occurrence with a `cuda`/`torch` marker, would keep the intent without the hole. Reasoned from the marker list and the request shape — I could not provoke a Koina OOM to measure it.

mzLib/PredictionClients/Koina/Client/KoinaServiceException.cs
```diff
@@ -0,0 +1,125 @@
...
+            "failed to load model",
+            "model is not ready",
```
acesnik: This may be dead as written — it needs those four words adjacent, and `{"error":"model 'Altimeter_2024_intensities' is not ready"}` would not match. I could not confirm Triton's exact wording, so it may well be fine as-is; worth a check against a real unready-model body if you have one.

acesnik (commented):
🤖 The thing this fixes is one I watched happen: #1240 sat with a red check for its entire review because Koina's GPU faulted on one model, and there was nothing anyone could do about it from our side. What makes it a genuinely hard problem — and what I think you have read correctly — is that the server answers `400` for both "your request is malformed" and "my GPU fell over", putting the difference only in the response body. A readiness probe cannot see that, so no amount of fixture-level guarding was ever going to catch it. The classification has to happen per call, on the body, which is exactly where you put it.

The classifier is the part I would have gotten wrong. Splitting on the body rather than the status, deliberately under-matching, and keeping the marker list to things only the server's own execution can produce — that is the right instinct, and it means a regression in how we build a request still fails loudly instead of quietly becoming a skip. I checked all six rejection shapes in `RequestRejectionsAreNotServiceFaults` plus four Triton-style bodies of my own, and none of them classify as a fault. `ExtractServerError` falling through to the raw body on a non-JSON page rather than throwing is a nice touch. I ran the two new files against a stubbed probe: 20 passed, 1 skipped, matching your second commit exactly.

I have left four notes on specific lines. Three are in the guard that consumes the classifier rather than the classifier itself, and they share a cause worth stating once here: **the guard never sees the exception, only the recorded failure message.** That is what makes it both over-fire and lose the server's text, so the narrowness you argue for in §2 is real in `ForFailedResponse` and then given away downstream. Passing the exception through, or matching on it rather than on prose about it, would collapse two of the three.

One thing that belongs in the description rather than the code. The title says "a Koina test", but the `[TearDown]` is inherited by fourteen fixtures and 184 `[Test]`-attributed methods — and nothing bounds the skip rate, so if Koina's GPU broke permanently all 184 would report Skipped and the job would read green. It is already non-blocking, so that is red becoming invisible-green rather than a new gate, but it is the kind of thing worth writing down now while it is a choice rather than a surprise.

Verified: the classifier over ten bodies, and that no exact-type `Assert.Throws<HttpRequestException>` exists on the Koina path — all twelve are UniProt/PRIDE, so your "existing catches keep working" claim holds. Not verified: anything against live Koina. `Test.csproj` is `net10.0-windows` via `mzPlot` and cannot run here, so I used a throwaway net10.0 harness with `ExternalServiceTestHelper` stubbed. Your "397 passed, 1 skipped" is CI's to confirm, not mine, and I took Triton's exact error wording on trust.

mzLib/Test/KoinaTests/TestKoinaServiceFaultClassification.cs
trishorts: Fixed in `6eee0601` — base method is `virtual`, this one `override`.

You are right about the mechanism and right about why it matters. `new` leaves two `EnsureKoinaReachable` on the type, NUnit finds both by reflection, and the fixture documented as needing no network dialled out. And the consequence you name is the one that stings: `EnsureReachable` calls `Assert.Ignore`, so with Koina down the fixture *proving the guard works* was itself skipped. That is precisely the failure I rejected the `IWrapTestMethod` approach to avoid — the guard going unproven with nothing to say so — reintroduced one layer down.

Added `TheLiveProbeIsSuppressedRatherThanHidden`, which asserts reflection finds exactly **one** such method and that it is declared here. Same lookup NUnit does, so the next person who reaches for `new` is told.

mzLib/Test/KoinaTests/KoinaLiveTestFixture.cs
trishorts: Fixed in `6eee0601` — it now requires **both** the type name and a server fault in the text.

Your row 3 is the one that matters and I had it backwards: the comment guarded the reverse direction only. A `Throws`-nothing failure names the type and carries no body; a real fault carries both. So the condition is now the conjunction, and `Assert.Throws<KoinaServiceException>` that threw nothing stays Failed.

Pinned by `AFailedThrowsAssertionNamingTheExceptionIsStillAFailure`, which asserts the premise explicitly — the message *does* name the type — so the test is about the guard rather than about the message shape.

Your "latent, not live" framing is also the reason I did not treat it as lower priority: making the exception public in this PR is exactly what puts that test one keystroke away.

mzLib/Test/KoinaTests/KoinaLiveTestFixture.cs
trishorts: Fixed in `6eee0601`. This one was the worst of the five, because it defeated the PR on its own motivating case.

The guard now reports the first line that **names the fault** rather than the first line. Replicated your scenario as a test — `Assert.DoesNotThrow(..., "Charge 1 should be valid")` around the real #1240 body — and it now logs:

```
Skipping external-service test: Koina failed to run the model
(PredictionClients.Koina.Client.KoinaServiceException : Request failed with status 400
Bad Request: {"error":"in ensemble 'Altimeter_2024_intensities', PyTorch execute failure: ...).
```

`ADoesNotThrowFailureStillReportsTheServersError` asserts the message contains `PyTorch execute failure` **and does not contain** `Charge 1 should be valid`, so the regression is pinned in both directions.

Your diagnosis of why was exactly right, including the part I had not seen: `ExtractServerError` is the identity function on an NUnit message because that message is never valid JSON, so the JSON path never ran at all.

mzLib/PredictionClients/Koina/Client/KoinaServiceException.cs
trishorts: Removed in `6eee0601`. You are right and I had the reasoning wrong in the remark directly above the list.

A GPU OOM *is* what an oversized request looks like server-side, so it fails the test I set for the other eight markers — immune to what we send — and I asserted that it passed. Batch size being a tuned parameter makes it the plausible regression rather than a theoretical one.

I considered your co-occurrence suggestion and did not take it: a request-induced OOM reports itself as `CUDA out of memory` too, so pairing it with a `cuda` marker does not separate the two cases. Dropping is the honest fix.

`AnOutOfMemoryIsOurProblemNotTheServersAndMustNotSkip` now asserts both spellings fail rather than skip, and the removal has its reasoning recorded where the marker used to be — otherwise it reads like an oversight and comes back.

mzLib/PredictionClients/Koina/Client/KoinaServiceException.cs
trishorts: Relaxed to `"is not ready"` in `6eee0601`.

I could not confirm Triton's wording either, so I have not claimed to. But the failure mode is asymmetric: if the real body is `model 'X' is not ready` then four adjacent words never match and the marker is dead, whereas two words match both that and the adjacent form. Strictly better under uncertainty.

Added a test using the exact shape you wrote — model name in the middle — so if someone later has a real unready-model body to check against, the expectation is written down rather than inferred from the marker list.

trishorts:
@acesnik — all five threads addressed and resolved in `6eee0601`. Four were defects in what this PR shipped, and one of them defeated it on its own motivating case.

| your finding | fix |
|---|---|
| the proof fixture was dialling Koina — `new` leaves two `[OneTimeSetUp]` and NUnit runs both | base is `virtual`, this one `override`; a reflection test asserts there is exactly **one** |
| the guard could skip a real failure — `Assert.Throws<KoinaServiceException>` that threw nothing | now requires **both** the type name and a server fault in the text |
| the skip dropped the server error for `Assert.DoesNotThrow` | reports the first line that **names** the fault; `PyTorch execute failure` now reaches the log, `Charge 1 should be valid` does not |
| `out of memory` is our problem, not the server’s | marker removed, with the reasoning recorded where it was |
| `model is not ready` needed four adjacent words | relaxed to `is not ready` |

Six new tests, one per finding plus the `DoesNotThrow` shape. 407 passed, 2 skipped by design.

Two things worth flagging back:

**Your first finding cost me a test fixture, correctly.** The proof test constructed `KoinaServiceException` by hand with a body-free message — a shape `ForFailedResponse` never produces. That is what made the new AND condition fail against it honestly, and the fixture now builds through the factory.

**I did not take your co-occurrence suggestion on the OOM marker.** A request-induced OOM reports itself as `CUDA out of memory` too, so pairing it with a `cuda` marker does not separate the two cases. Dropping was the only honest option.

Ready for another look when you have time.

mzLib/PredictionClients/Koina/Client/KoinaServiceException.cs
```diff
@@ -0,0 +1,135 @@
...
+        public string ServerError { get; }
+
+        /// <summary>
+        /// Markers that identify a fault in Koina's own inference stack rather than a rejection of what
+        /// we sent. Matched case-insensitively against <see cref="ServerError"/>.
+        /// </summary>
+        /// <remarks>
+        /// Deliberately narrow, and it must stay that way. Everything here names the SERVER's execution
+        /// failing: a Torch interpreter fault, a CUDA or cuBLAS call failing, a model that would not
+        /// load, the GPU running out of memory. None of them can be provoked by a malformed request.
```
acesnik: This sentence did not follow the removal. It still lists "the GPU running out of memory" among what the markers name, and still asserts "None of them can be provoked by a malformed request" — which is precisely the claim lines 54-60 now refute, in the same remark. A reader who stops at 47 learns the opposite of the decision. Dropping the OOM clause and softening the sentence to cover the eight that remain is the whole fix; it is the same shape as the reasoning error you corrected last round, just left behind one paragraph up.

mzLib/Test/KoinaTests/TestKoinaServiceFaultClassification.cs
```diff
@@ -0,0 +1,316 @@
...
+        /// derived-first, base-last, so a TearDown in this fixture would look at the result BEFORE the
+        /// base class's guard had rewritten it. That is what the first version of this test did, and it
+        /// reported a failure that was its own.
+        /// </summary>
+        [OneTimeTearDown]
+        public void TheGuardMustHaveSkippedTheFaultAndLeftTheOtherAlone()
+        {
+            var children = TestExecutionContext.CurrentContext.CurrentResult.Children.ToList();
+
+            var fault = children.Single(c => c.Name == nameof(AServiceFaultSkipsTheTest));
```
acesnik: The four `children.Single(...)` calls mean this fixture cannot be run one test at a time. `dotnet test --filter Name=AServiceFaultSkipsTheTest` leaves one child, and the `Single` for `AnOrdinaryFailureStillFails` throws `InvalidOperationException` from `[OneTimeTearDown]` — so the fixture errors, and the person debugging one of these tests sees a failure that is the observer's rather than theirs. That is the same trap the doc comment above says the first version of this test fell into. `SingleOrDefault` with a null check, or asserting only over the children actually present, keeps it runnable in isolation.

mzLib/Test/KoinaTests/TestKoinaServiceFaultClassification.cs
```diff
@@ -0,0 +1,316 @@
...
+                "and it carries no server fault, which is what keeps the guard off it");
+        }
+
+        /// <summary>
+        /// Hiding the base [OneTimeSetUp] with `new` leaves TWO of them on the type and NUnit runs both,
+        /// so the live probe fired from a fixture documented as needing no network. Reflection is how
+        /// that is visible, and it is the same lookup NUnit does.
+        /// </summary>
+        [Test]
+        public void TheLiveProbeIsSuppressedRatherThanHidden()
```
acesnik, line 275: Worth knowing this only sees `public` methods: `GetMethods()` with no flags would return zero if someone later made the base probe `protected virtual`, and the count assertion would then fail with a message about `new` versus `override` that no longer describes what happened. `Has.Count.EqualTo(1)` does fire before the indexer on the next line, so it fails cleanly rather than throwing — just the diagnosis would point the wrong way.

acesnik (dismissed):
🤖 All five are fixed, and I checked each against `6eee0601` rather than reading the replies. The probe is `virtual`/`override` with an empty body, so the fixture documented as needing no network genuinely does not dial. The guard is now the conjunction at `KoinaLiveTestFixture.cs:80-81`, so a `Throws`-nothing failure that names the type stays Failed. `LineNamingTheFault` returns the first line for which `IsServiceFault` holds rather than the first line, which is what puts the server's text back. `out of memory` is gone from the marker array with the reasoning recorded where it was. And `is not ready` is two words rather than four.

Two things I want to note about the fixes themselves, because both are better than what I asked for.

Declining the co-occurrence suggestion was right and the reason is stronger than mine: a request-induced OOM reports itself as `CUDA out of memory` as well, so pairing the marker with a `cuda` marker separates nothing. Dropping was the only option that does not leave a hole, and I had not thought that through.

The second is that fixing the probe dissolved my other concern rather than papering over it. I had flagged the missing `[Category]` as putting this fixture into the required job that #1140 exists to keep off the network — but with the probe suppressed the fixture needs no network at all, so belonging to the required job is now correct. Adding a category would have buried the proof in the non-blocking job, which is the opposite of what you want for the test that proves the guard works. Leaving it uncategorised is the right call and the comment at `:200` says so.

I did go looking for the assertion behind `ADoesNotThrowFailureStillReportsTheServersError`, since its own body is only a deliberately-failing `DoesNotThrow`. It is in the `[OneTimeTearDown]` observer at `:308-310` — `Does.Contain("PyTorch execute failure")` and `Does.Not.Contain("Charge 1 should be valid")` — so "pinned in both directions" is exactly right, just asserted from the observer rather than in the test named. Same for the `Passed` expectation on the `Throws`-nothing test, which reads as odd until you see it asserts the premise and then that `IsServiceFault` is false on it.

Approving. Three notes below, none of which I would hold this for. The one in `KoinaServiceException.cs` needs a push, but `mzLib` master is `strict: true` and this branch is `behind`, so a merge-from-master is required before it can land anyway — the doc line can ride along at no extra cost. Approvals are at 0 of 2, so this is the first of two and does not merge on its own.

Still not verified by me: anything against live Koina, and Triton's exact unready-model wording. `Test.csproj` is `net10.0-windows` via `mzPlot`, so all of the above came from reading the head plus a throwaway net10.0 harness. Your 407/2 is CI's to confirm.

mzLib/PredictionClients/Koina/Client/KoinaServiceException.cs
trishorts: Fixed. The sentence no longer mentions the GPU running out of memory, and no longer claims the markers cannot be provoked by a malformed request. It now states the test a marker has to pass to belong in the list, and points at the paragraph below for why "out of memory" fails it — so the two halves of the remark argue the same thing.

mzLib/Test/KoinaTests/TestKoinaServiceFaultClassification.cs
trishorts: Fixed, and confirmed by running it rather than by reading it. All four lookups are now `SingleOrDefault` behind a local `ITestResult Child(string)`, and each block of assertions runs only if that child is present.

Verified: `dotnet test --filter Name=AServiceFaultSkipsTheTest` now reports `Skipped! - Failed: 0, Passed: 0, Skipped: 1, Total: 1` instead of erroring out of `[OneTimeTearDown]`. Whole-fixture run is unchanged: 418 passed, 2 skipped.

mzLib/Test/KoinaTests/TestKoinaServiceFaultClassification.cs
trishorts: Fixed rather than just noted — the diagnosis pointing the wrong way is the whole cost, so making it not arise is cheaper than documenting it. The lookup is now `GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static)`, with a comment recording why the parameterless overload was wrong here. A base probe narrowed to `protected virtual` is now seen, so the count assertion still measures `new` versus `override` — which is what its message claims to measure.

RayMSMS (approved):

zhuoxinshi (approved):

PR #1249, opened 2026-08-30
Add StcE as a protease in its own right, not only paired with trypsin

trishorts:
# Add StcE as a protease in its own right, not only paired with trypsin

mzLib has shipped StcE only as `StcE-trypsin`, the combined double digest. This adds `StcE` alone.
**One data row, one comment block, three tests. No production code changes.**

```
StcE    TX|T,TX|S,SX|T,SX|S    full        StcE
```

## Why StcE alone is worth having

`StcE-trypsin` is the wrong tool for O-glycosite work, because the tryptic cleavages chop up exactly
the long, site-dense peptides StcE exists to produce. Digesting the human reviewed proteome
(`UP000005640`, fully-specific, length 6-50) both ways, counting candidate O-glycosites per peptide
as the Ser/Thr count:

| enzyme | peptides | median len | n=0 | n=1-2 | n=3-5 | n=6-9 | n=10+ | max n |
|---|---|---|---|---|---|---|---|---|
| `StcE-trypsin` | 223,361 | 10 | 20.8% | 56.6% | 20.7% | **1.9%** | 0.1% | 14 |
| **`StcE`** | 44,283 | 21 | 0.1% | 22.4% | **55.6%** | **20.5%** | 1.5% | 16 |

76% of StcE peptides carry three or more candidate sites and 20.5% carry six to nine. Adding trypsin
cuts the six-or-more population from 22% to 2%.

## The motif

Unchanged from the sequence half of `StcE-trypsin`, and it matches the primary source. Malaker et
al., *PNAS* 2019 ([10.1073/pnas.1813020116](https://doi.org/10.1073/pnas.1813020116)): StcE
*"recognizes the consensus sequence S/T\*-X-S/T, and cleaves the peptide backbone before the P1' S/T
only when the P2 S/T is glycosylated."*

## What the motif syntax cannot say, so the file says it instead

The asterisk above is load-bearing: StcE only cleaves when the **first S/T is O-glycosylated**, and
its activity is abolished by deglycosylation. `DigestionMotif` matches sequence alone, so **both**
StcE entries over-digest relative to the real enzyme -- their peptides are a superset of true StcE
peptides. The comment block says so next to the row, where someone choosing a protease will read it.

That same limitation is why **SmE, OgpA/OpeRATOR and IMPa are deliberately not added**. All three
cleave N-terminal to a glycosylated Ser/Thr and differ only in glycan tolerance -- SmE takes complex
and sialylated glycans and cuts between adjacent glycosites, OgpA is blocked by sialic acid, IMPa
cannot cut between adjacent glycosylated residues. As sequence motifs all three collapse to an
identical `|S,|T`, which would put three byte-identical proteases in the file under three names and
invite someone to select "SmE" believing they had modelled SmE. `|S,|T` also cuts at every S and T,
leaving at most **one** candidate site per peptide, which is not what any of them does. Modelling
them honestly needs glycosylation-aware digestion, which is a much larger change.

## Tests

Three, in `Test/Omics/Modifications/TestDigestionMotif.cs`:

- `TestStcEProtease` -- explicit boundaries: `AAATPTGGGSQSCCC` -> `AAATP`, `TGGGSQ`, `SCCC`.
- `TestStcEAgreesWithStcETrypsinAbsentLysineAndArginine` -- pins the new entry against the shipped
  one on a sequence with no K or R, where the two must agree by construction.
- `TestStcEDoesNotCleaveAtLysineOrArginine` -- asserts the property that makes StcE-alone useful.

357 digestion / protease / SDRF tests pass. The two tests that count the dictionary do so relatively
(`initialProteaseCount + 1`), so adding an entry is safe.

Stryker is not useful here -- there is no production C# to mutate. I ran the data-equivalent instead,
mutating the row five ways and confirming each is caught:

| mutant | result |
|---|---|
| drop the `SX\|S` alternative | killed (2 tests) |
| shift cut index left (`T\|XT`) | killed (2) |
| leave the trypsin motifs in | killed (1) |
| specificity `full` -> `semi` | killed (3) |
| `SX\|S` -> `SX\|A` | killed (2) |

Each mutant is caught by a different subset, so none of the three tests is redundant.

---

### Review follow-up

**The over-digestion caveat needs to reach a user, not only whoever next edits the file.** The `.tsv`
comment block is the right place for an editor and the wrong place for the person it protects: a
MetaMorpheus user picks "StcE" from a dropdown and never sees it. These peptides are a **superset**
of true StcE peptides, because the motif matches sequence only and cannot express the glycan
requirement — a result-shaping fact that belongs in a tooltip or the `Description` column. Recorded
here so it is not lost; not this PR's job.


acesnik (approved):
🤖 Approving. StcE paired with trypsin was the only way to reach it, and trypsin's cleavages destroy exactly what StcE is used for — the long, glycosite-dense peptides. Adding it in its own right is the right fix and the entry is additive: a new name in `proteases.tsv`, nothing existing changed.

I traced the motif against your own fixture rather than taking it on faith. `TX|T,TX|S,SX|T,SX|S` on `AAATPTGGGSQSCCC` cuts between P(4) and T(5) for the T-P-T match and between Q(10) and S(11) for S-Q-S, giving `AAATP` / `TGGGSQ` / `SCCC` — which is what `TestStcEProtease` asserts. `TestStcEAgreesWithStcETrypsinAbsentLysineAndArginine` is the test I liked most: pinning a new entry against a shipped one on a sequence carrying neither K nor R is a much stronger statement than asserting a literal, and it will keep holding if either motif set is edited.

The 15 comment lines are safe — `proteases.tsv` already carries 97 of them on master, so the parser demonstrably handles `#`.

The comment block itself is better than most code I read: saying outright that the motif matches sequence only, cannot express the glycan requirement, and therefore **over-digests relative to the real enzyme** is the honest framing, and explaining why SmE, OgpA and IMPa are deliberately absent pre-empts the obvious follow-up PR.

One thing about where that caveat lives. It is in the right place for whoever next edits the file and the wrong place for the person it protects: a MetaMorpheus user picks "StcE" from a protease dropdown and never sees a `.tsv` comment. Those peptides being a superset of true StcE peptides is a result-shaping fact, so it belongs somewhere a user meets — a tooltip on the dropdown, or the `Description` column if anything renders it. Not this PR's job, and not a reason to hold it; worth a line in the description so it does not get lost.

RayMSMS:
### Manual unit-test review

All 3 unit test(s) added by this PR were reviewed individually by the author (not only by automated/AI-generated checks): 3 approved as written, 0 edited, 0 flagged.

This satisfies the author's standing AI-use policy that AI-assisted tests receive human review before relying on them. Recorded 2026-09-03 13:34.

RayMSMS (approved):

PR #1250, opened 2026-08-30
Add deterministic, composition-preserving sequence permutation

trishorts:
# Add deterministic, composition-preserving sequence permutation

Three additive members on `DecoySequenceValidator`. **No existing member changes** -- 375
insertions, 0 deletions against master.

```csharp
public static HashSet<int> CleavageSitePositions(string sequence, List<DigestionMotif> motifs);
public static BigInteger  PermutationSpaceSize (string sequence, List<DigestionMotif> motifs);
public static string      UnrankPermutation    (string sequence, List<DigestionMotif> motifs,
                                                BigInteger index, out int[] swappedPositionArray);
```

## What they do, and why they are not just another shuffle

`ScrambleSequence` rearranges a sequence with a `Random`. These rearrange it by **index**: given a
sequence and an integer, you get one specific rearrangement, and the same one every time.

- **`PermutationSpaceSize`** -- how many distinct rearrangements leave every cleavage-site residue
  in place. Closed-form multinomial over the free positions, built from successive binomials so no
  factorial is ever materialised. Cheap even when the answer is astronomical.
- **`UnrankPermutation`** -- the rearrangement at a given index, in lexicographic order. Returns the
  same old-to-new position map `ScrambleSequence` already returns, so the existing modification
  remapping is reused rather than duplicated.
- **`CleavageSitePositions`** -- every position participating in a motif match.

Two properties fall out, and both are the point:

**The result is isomeric with the input** -- same residues, same mass, different order -- because
only positions are moved. **And it is a pure function of (sequence, index)**: no random state, so it
is reproducible across runs, threads, machines and framework versions, and it parallelises without
coordination. Seeded `System.Random` is not documented to be stable across .NET major versions, so
"seed 42" is not a reproducibility guarantee; an index is.

Callers wanting an unpredictable-but-reproducible choice derive the index from a seed and the
sequence rather than from a random number generator, and can walk `(i + k) mod N` to enumerate the
whole space -- which makes "there is no usable rearrangement" a **proven** statement rather than
"we gave up after N tries".

## `CleavageSitePositions` pins the full motif span, deliberately

`ScrambleSequence` records only the location where a motif matches
(`DecoySequenceValidator.cs:135`). That is complete for a single-residue motif -- `K`, `R`, `E` --
but not for a multi-residue one. `StcE`'s `TX|T` matches three residues, and leaving the trailing
two free lets a rearrangement destroy that cleavage site and invent others elsewhere. Same for
`collagenase`'s `GPX|GPX`.

**This PR does not change `ScrambleSequence`.** Fixing its pinning would alter existing decoy
sequences, which is a separate decision and belongs in its own PR. The new member exists so
rearrangements that must survive a round-trip through digestion can hold the whole span still.

## Tests

Eleven, in `Test/DatabaseTests/TestDecoyProteinValidator.cs`: the multinomial against hand-computed
values (`TTTPAPTTT` -> 252, `SSSSSSR` -> 1, `ACDEFGHK` -> 5040), exhaustive enumeration of a
252-permutation space with composition preserved throughout, cleavage sites held fixed, the
position map, out-of-range indices and their message, empty/null inputs, full-span pinning, the
preventing residue, and no position ever returned past the end of the sequence.

Plus **six golden vectors** pinning the ordering itself. Every other assertion holds for *any*
consistent ordering, so none of them would notice if the ordering changed -- and a caller's pairing
depends on a given index always yielding the same string. The expected values were computed by a
**separate implementation** written to the same lexicographic definition, so the test asserts that
two implementations agree rather than that this one still does what it did.

48/48 in the fixture; 830 passed / 0 failed across the wider database and decoy suites.

## Mutation testing

Stryker.NET 4.16.0 over `DecoySequenceValidator.cs`:

```
dotnet stryker -tp ../Test/Test.csproj --mutate "**/DecoySequenceValidator.cs" -c 8 -r json
```

The new members scored **81.7%** (76 killed of 93 scored) before the changes below, and **93.5%**
(72 killed of 77) after. Reading the seventeen original survivors split them three ways:

**Three were dead code, so they were deleted rather than covered** -- seven mutants killed by
removal, and less code than before:

| survivor | why it could not be killed |
|---|---|
| `i + offset <= sequence.Length` | the inner loop re-checked that the span fits, which `Fits()` has already established |
| `freeCount--`, `false ? One : permutations` | `Multinomial` took a count whose only use was a zero check an empty multiset already satisfies by falling through the loop; the `out` parameter feeding it was dead too |
| `k < 0 && k > n`, `k <= 0` | `Binomial` guarded a range its only caller cannot produce |

**Four were genuine gaps, now covered:** the empty-sequence and null-motif guards in all three
public members, and the *content* of the out-of-range message -- which a caller probing for a free
rearrangement reads to tell exhaustion from a bad argument.

**The remainder are equivalent mutants and were left alone:** widening the outer loop bound is
absorbed by `Fits` returning false at `i == Length`; skipping the `continue` for an exhausted
residue leaves a zero-width branch no rank can fall into; and both mutations of
`Math.Min(k, n - k)` still compute the same binomial by symmetry.

## Notes for review

- Purely additive. `ScrambleSequence`, `ScrambleDecoyBioPolymer` and `IsPalindromic` are untouched.
- `MzLibException` for out-of-range indices, matching `ControlledVocabulary.cs:120` and
  `PrideArchiveClient.cs:224`.
- File is `#nullable enable`; the additions build with zero warnings.
- `BigInteger` is new to mzLib, from `System.Numerics` in the BCL -- no new package reference.

### DIFF
--- a/mzLib/Test/DatabaseTests/TestDecoyProteinValidator.cs
+++ b/mzLib/Test/DatabaseTests/TestDecoyProteinValidator.cs
@@ -4,6 +4,7 @@ using Proteomics.ProteolyticDigestion;
 using Proteomics;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 using MzLibUtil;
 using UsefulProteomicsDatabases;
 using Transcriptomics.Digestion;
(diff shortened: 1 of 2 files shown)


acesnik (approved):
🤖 Approving, and this is the piece the rest of the stack rests on, so I verified it against brute force rather than reading it.

The design decisions are the right ones and each is stated. `CleavageSitePositions` returning the **full span** of a motif match rather than only its start is the one that matters most — and it is exactly what #1249 needs, since StcE's `TX|T` is three residues and holding only the first lets a rearrangement destroy that site and invent others. `UnrankPermutation` consuming no random state is the other: entrapment output has to be identical across runs, threads and machines, and indexing into a lexicographic order gets that for free where a seeded RNG would still depend on call order. And `PermutationSpaceSize` in closed form means the astronomically-large case costs nothing.

## Verified against enumeration

Six sequences, comparing `PermutationSpaceSize` with brute-force enumeration of every distinct arrangement holding cleavage sites fixed, and unranking every index:

| sequence | cleavage sites | size | brute force |
|---|---|---|---|
| `ACDEFGHK` | 7 | 5040 | 5040 |
| `TTTPAPTTT` | — | 252 | 252 |
| `ACDKEFGRHK` | 3, 7, 9 | 5040 | 5040 |
| `MPEPTIDEK` | 8 | 10080 | 10080 |
| `SSSSSSR` | 6 | 1 | 1 |
| `AKAKA` | 1, 3 | 1 | 1 |

At every index, for all six: distinct output (so the unranking is a bijection, not merely a function), composition preserved, and every cleavage-site residue held in place. Where enumeration was feasible the unranked set **equals** the brute-force set, not merely its size. Your golden cases check out by hand too — `TTTPAPTTT` is T×6 P×2 A×1 with no tryptic site, so 9!/(6!·2!·1!) = 252 and the last index is 251.

## One thing worth adding to the remark

`PermutationSpaceSize` returning `BigInteger.One` is documented with a homopolymeric example, `SSSSSSR`, and framed as arithmetic rather than a collision. That framing is right, but the example undersells how often it happens: `AKAKA` also comes back 1, and it is neither homopolymeric nor unusual — two cleavage sites at 1 and 3 pin the sequence down to a single arrangement of three identical free residues. Short tryptic peptides with dense K/R are common, so a caller will meet size-1 regularly rather than only on pathological input, and in that case the only available "rearrangement" is the identity, i.e. an entrapment sequence equal to its target.

That is a caller's problem rather than this method's, and #1259 appears to be where it is counted. Worth naming in the remark anyway, because "homopolymeric tract" reads like a corner case and the dense-cleavage-site case is not one.

nbollis (approved):

PR #1251, opened 2026-08-30
Capture the last pipe-delimited field as the UniProt FASTA accession

acesnik:
🤖 `UniprotAccessionRegex` was `[|](.+)[|]` — unanchored, with a greedy `.+`. `Regex.Matches` therefore started at the *first* pipe and ran to the *last*. On a two-pipe UniProt defline that happens to equal "capture field 2", which is why this survived. On a legacy four-pipe NCBI defline it concatenated three fields:

```
>gi|16128008|ref|NP_414555.1| thr operon leader peptide [E. coli K-12 MG1655]
   old -> 16128008|ref|NP_414555.1
   new -> NP_414555.1
```

`16128008` is the GI number, `ref` names the source database, and `NP_414555.1` is the RefSeq accession.version. The old capture is not an accession under any convention.

**This is output-affecting.** Any database with more than two pipes in a defline gets different accessions — correct ones where they were wrong — and accession is the join key in every file we write, so nothing produced before and after is comparable. Two-pipe UniProt and modern two-pipe RefSeq headers are unchanged.

The fix anchors the pattern and restricts the capture to one field, `^>.*\|([^|]*)\|`, so the greedy prefix pushes the opening pipe as late as possible. That is the same shape as MaxQuant's documented UniProt identifier rule, `>.*\|(.*)\|`; I compared the two forms across every `>` line in the test tree and they agree on all 41.

## Why this is worse than it looks, and also better than it looks

Worse: **`|` is not only inside the accession, it is the separator our own output uses for multi-valued columns.** A pipe inside an accession desynchronises any reader that splits several columns in lockstep, and turns one protein into several tokens in a protein-group name. That is also why `NP_414555.1` is the right answer *for us specifically* rather than universally — MaxQuant can keep `gi|16128008` as an identifier because `proteinGroups.txt` separates grouped identifiers with `;`. Which field is the accession is coupled to the delimiter downstream.

Better: this regex is what mzLib's own auto-detection reaches for on a legacy NCBI file. `DetectFastaHeaderFormat` classifies a header as UniProt when it sees `>`, two lowercase letters and a pipe — which `>gi|` satisfies — so detection labels legacy NCBI as UniProt and then applies this pattern. The fix repairs the detected path, not only the explicitly-passed one.

## What it does not fix

A better default is an improvement, not a solution, and mzLib's own fixtures show why. Seven pre-existing deflines change capture under the new pattern: six Modomics RNA headers and one Gencode header. On the Gencode one:

```
>ENSMUSP00000034487.2|ENSMUST00000034487.3|ENSMUSG00000018620.3|OTTMUSG...|Mmp20-201|Mmp20|...
   old UniProt regex     -> ENSMUST00000034487.3|ENSMUSG00000018620.3|OTTMUSG...|Mmp20-201|Mmp20
   new UniProt regex     -> Mmp20
   GencodeAccessionRegex -> ENSMUSP00000034487.2
```

`Mmp20` is a gene symbol, not an accession. The new pattern is wrong there too, just differently wrong — only the dedicated Gencode rule is right, which is precisely why mzLib carries one. Neither form survives a pipe inside a description either. **None of those seven are regressions:** `DetectFastaHeaderFormat` routes Gencode and Ensembl files to their own regexes, the RNA fixtures go through `RnaDbLoader.ModomicsFieldRegexes`, and no test asserts on any of the old concatenated values — I checked. But they are the honest bound on what this PR claims.

The sibling patterns are untouched and share the same shape. `UniprotNameRegex` is `\|(?:.+)\|(.*?)(\s|$)`, so `Name` on a multi-pipe defline stays wrong. Fixing those is a larger, separately measurable change and I would rather not bundle it with an accession change that already moves output.

## Tests

One test, `LegacyNcbiFastaHeaderYieldsTheAccessionNotConcatenatedFields`, over a new three-entry fixture covering the four-pipe `gi|` form and the two-pipe `ref|` form. It asserts the three expected accessions and, separately, that none contains `|` — the second assertion states the invariant rather than the example.

I could not run mzLib's suite here: `Test.csproj` targets `net*.0-windows` via `mzPlot` and the host needs `Microsoft.WindowsDesktop.App`. The regex comparison and the fixture scan above were run directly; the assertion values are derived, not observed under NUnit. CI is the confirmation.

## Related

MetaMorpheus #1921 is this reported from outside, in 2020. It is not fixed by this PR alone: MetaMorpheus always passes an accession regex, so it never reaches `DetectFastaHeaderFormat` and mangles Ensembl and Gencode databases too. The companion MetaMorpheus PR adds a per-task header-format setting, closes #1936, and includes an `Auto` option that passes no regex specifically so detection runs — which is the half of #1921 this PR cannot reach. Because that PR references mzLib's presets by name rather than copying them, this fix improves its `UniProt` preset with no change needed there.


nbollis (approved):
Reviewed by Sisyphus: no actionable findings.

zhuoxinshi (approved):

PR #1253, opened 2026-08-30
Add the entrapment peptide generator

trishorts:
# Add the entrapment peptide generator

> **Stacks on #1250.** This branch contains #1250's four commits plus this PR's two
> (`08d4db49`, `5ccc7ea5`). **Merge #1250 first** and this diff reduces to the two files below.
> Review just those two if you prefer; the earlier commits are #1250's and were reviewed there.

Adds `UsefulProteomicsDatabases/EntrapmentGeneration/`, a new folder alongside `DecoyGeneration/`.
Two files, 372 insertions, nothing existing touched.

## What it does

Given a target peptide, the cleavage motifs that must stay put, and the set of sequences already
spoken for, it returns that peptide's **entrapment partner** or a typed reason it has none.

```csharp
EntrapmentPeptide result = EntrapmentPeptideGenerator.Create(
    targetSequence, digestionParams.DigestionAgent.DigestionMotifs, everyTargetPeptide,
    fold: k, foldCount: r, seed: 1);
```

The partner is **isomeric** with its target: same residues, different order, so same mass and the
same count of any residue of interest. Cleavage sites stay where they were, so the partner digests
the same way its target does.

## Why isomeric is the point, not a limitation

For localization work the number of candidate sites in a peptide sets the difficulty of placing a
modification on it -- with no evidence the best you can do is 1/n, so the ceiling is 50% at two
sites and 17% at six. A partner carrying a *different* site count would be scored at a difficulty
its target never faced, and the difference would be read as a difference in correctness. Composition
preservation gives matching counts for free.

It also makes target and partner the hardest possible pair to tell apart, which is the
discrimination an entrapment experiment exists to measure. Anything that made partners easier to
distinguish -- a mass offset, a tag -- would throw that away.

## Folds get disjoint stretches, not a shared pool

`stretch = size / foldCount`, and fold *k* searches `[k*stretch, (k+1)*stretch)`. Two properties
follow, and both matter:

- the *r* partners of a target are **distinct by construction** rather than by checking; and
- each fold is **independent of the others** -- fold 5 is the same sequence whether or not folds
  0-4 were ever asked for.

So a database can be extended, regenerated in pieces, or built in parallel without invalidating
anything already issued.

## Three failure reasons, because the remedies differ

| `EntrapmentFailure` | meaning | what the caller can do |
|---|---|---|
| `NoPermutationExists` | one arrangement once cleavage sites are pinned -- a homopolymeric tract like `SSSSSSR` | nothing; arithmetic, not competition |
| `SpaceTooSmallForFoldCount` | real space, too small to give every fold its own partner | ask for fewer folds |
| `AllPermutationsTaken` | every arrangement this fold may draw on is already spoken for | depends on the target set; another may succeed |

Collapsing these into one "failed" would hide the distinction from whoever has to act on it. The
space size is reported even on failure, because that value is what separates an impossible peptide
from a merely crowded one.

## Determinism

Selection is by index, not by chance, so the answer is a pure function of the sequence and the
seed -- unchanged by processing order, by which protein the peptide came from, by threading, or by
framework version. Nothing consumes random state and folds never consult one another.

The starting offset comes from **SHA-256**, not `string.GetHashCode()`, whose value is explicitly
not stable across runs -- let alone across machines. A database regenerated elsewhere has to come
out byte-identical.

Because the search enumerates its stretch rather than sampling it, "no partner exists" is a
**proven** statement rather than a search that gave up.

## Tests

Twelve, in `Test/DatabaseTests/TestEntrapmentPeptideGenerator.cs`: isomerism, candidate-site count
and cleavage sites preserved, purity under unrelated work and under a changed seed, the three
failure reasons distinguished, r folds mutually distinct, fold independence, forbidden sequences
never returned, reported space size and probe count, and each argument guard.

848 passed / 0 failed across the database, decoy and entrapment suites.

## Mutation testing

Stryker.NET 4.16.0 over both this file and `DecoySequenceValidator.cs`:

```
dotnet stryker -tp ../Test/Test.csproj --mutate "**/DecoySequenceValidator.cs" \
                                       --mutate "**/EntrapmentPeptideGenerator.cs" -c 8 -r json
```

The generator first scored **79.1%**. The survivors were all in the argument guards, and for a
reason worth naming: several bad arguments are rejected by **more than one** guard -- `fold: 0`
with `foldCount: 0` trips both, and a fold past the count is caught downstream by the unranker even
with its own guard deleted -- so `Assert.Throws<MzLibException>` kept passing while the guard under
test was removed. The tests now assert the exception **message**, which pins which guard fired, and
cover the empty-sequence guard that nothing reached. The confirmation run scores this file at
**97.7%** (42 killed of 43), the single remaining survivor being the equivalent mutant below.

One survivor is left deliberately: widening the search loop to `step <= stretch` is an **equivalent
mutant**, since `(offset + stretch) % stretch == offset` simply re-probes the first candidate, which
has already been rejected.

## Notes for review

- New folder, new namespace `UsefulProteomicsDatabases.EntrapmentGeneration`. Nothing existing is
  modified by this PR's own commits.
- `MzLibException` for argument errors, matching `ControlledVocabulary.cs:120` and
  `PrideArchiveClient.cs:224`.
- `#nullable enable`; builds with zero warnings.
- This is peptide-level only. Assembling partners back into a protein database -- excision of
  unpairable peptides, accession scheme, modification transfer -- is the next PR.

---

### Review follow-up

**Culture-dependent seed — fixed in #1265.** Interpolation formatted the seed through the current
culture, so a negative seed produced a different database under `sv-SE`, `fi-FI` and `lt-LT`. Now
formatted invariantly, with a test that builds under four cultures and compares.

**Narrowing one claim.** "Distinct by construction" holds *per target, across folds*, which is what
the disjoint-stretch scheme delivers. It is **not** distinctness across targets: two anagram targets
share a permutation space and derive independent offsets into it, so both can be handed the same
partner. #1255's `AmbiguousPeptides` refuses to resolve exactly that population. Anyone computing a
ratio over *distinct sequences* wants the narrower statement.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
@@ -1,0 +1,195 @@
+#nullable enable
+using MzLibUtil;
+using Omics.Digestion;
+using System;
+using System.Collections.Generic;
+using System.Numerics;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>
+/// Why an entrapment peptide could not be produced. These are kept apart because the caller's
+/// remedy differs for each, and merging them into a single "failed" hides that.
+/// </summary>
+public enum EntrapmentFailure
+{
+    /// <summary>A partner was produced.</summary>
+    None,
+
+    /// <summary>
+    /// The sequence has exactly one arrangement once its cleavage sites are held in place -- a
+    /// homopolymeric tract such as "SSSSSSR". Arithmetic, not competition: no seed, no extra
+    /// searching and no smaller fold count can produce a partner for this peptide.
+    /// </summary>
+    NoPermutationExists,
+
+    /// <summary>
+    /// Arrangements exist, but every one this fold may draw on is already spoken for -- it is a
+    /// target peptide, or it has already been issued. Depends on the database, so a different
+    /// target set may succeed where this one did not.
+    /// </summary>
+    AllPermutationsTaken,
+
+    /// <summary>
+    /// The space is real but too small to give every fold its own partner. The remedy is a smaller
+    /// fold count, which distinguishes this from <see cref="AllPermutationsTaken"/>.
+    /// </summary>
+    SpaceTooSmallForFoldCount
+}
+
+/// <summary>
+/// One target peptide's entrapment partner, or the reason it has none.
+/// </summary>
+public sealed class EntrapmentPeptide
+{
+    internal EntrapmentPeptide(string targetSequence, string? entrapmentSequence, int[]? swappedPositions,
+        int fold, BigInteger permutationSpaceSize, int probesUsed, EntrapmentFailure failure)
+    {
+        TargetSequence = targetSequence;
+        EntrapmentSequence = entrapmentSequence;
+        SwappedPositions = swappedPositions;
+        Fold = fold;
+        PermutationSpaceSize = permutationSpaceSize;
+        ProbesUsed = probesUsed;
+        Failure = failure;
+    }
+
+    public string TargetSequence { get; }
+
+    /// <summary>The partner, isomeric with the target. Null when <see cref="Succeeded"/> is false.</summary>
+    public string? EntrapmentSequence { get; }
+
+    /// <summary>
+    /// Maps each position in the target to its position in the partner, the same contract as
+    /// <see cref="DecoySequenceValidator.ScrambleSequence"/>, so modifications ride across unchanged.
+    /// </summary>
+    public int[]? SwappedPositions { get; }
+
+    /// <summary>Zero-based fold this partner belongs to.</summary>
+    public int Fold { get; }
+
+    /// <summary>How many distinct arrangements exist. Reported even on failure, because the value
+    /// is what distinguishes an impossible peptide from a merely crowded one.</summary>
+    public BigInteger PermutationSpaceSize { get; }
+
+    /// <summary>Candidates examined, including the one returned. One means the first choice was free.</summary>
+    public int ProbesUsed { get; }
+
+    public EntrapmentFailure Failure { get; }
+
+    public bool Succeeded => Failure == EntrapmentFailure.None;
+}
+
+/// <summary>
+/// Produces the entrapment partner of a target peptide: the same residues in a different order,
+/// with every cleavage site left where it was.
+/// </summary>
+/// <remarks>
+/// <para>The partner is <b>isomeric</b> with its target -- same composition, so same mass and the
+/// same count of any residue of interest. For localization work that last property is the point:
+/// the number of candidate sites in a peptide sets the difficulty of placing a modification on it,
+/// so a partner carrying a different count would be compared at a difficulty its target never had.
+/// It also makes target and partner the hardest possible pair to tell apart, which is the
+/// discrimination an entrapment experiment is trying to measure.</para>
+/// <para>Selection is by <b>index, not by chance</b>. The starting point is derived from the
+/// sequence and the seed, and search walks forward from there, so the answer is a pure function of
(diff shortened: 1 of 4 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
```diff
@@ -0,0 +1,195 @@
...
+    /// Where in a fold's stretch to start looking, derived from the sequence and the seed.
+    /// </summary>
+    /// <remarks>
+    /// SHA-256 rather than <see cref="string.GetHashCode()"/> or a home-made hash: its output is
+    /// fixed by specification, so a database regenerated on another machine, another runtime or in
+    /// another decade is byte-identical. String hash codes are explicitly not stable across runs.
+    /// </remarks>
+    private static BigInteger DeriveOffset(string sequence, int seed, BigInteger stretch)
+    {
+        byte[] digest = SHA256.HashData(Encoding.UTF8.GetBytes($"{seed}:{sequence}"));
```
acesnik, line 187: `$"{seed}:{sequence}"` formats the seed through the current culture. Measured on this runtime:

```
en-US  "-7:SYKALADQMNLLLSK"   negativeSign U+002D
sv-SE  "−7:SYKALADQMNLLLSK"   negativeSign U+2212
fi-FI  "−7:SYKALADQMNLLLSK"   negativeSign U+2212
lt-LT  "−7:SYKALADQMNLLLSK"   negativeSign U+2212
```

Different bytes into SHA-256, so a negative seed gives a different database per culture. `seed.ToString(CultureInfo.InvariantCulture)` closes it, which is what `EntrapmentAccession.Format` already does for the fold.

acesnik (approved):
🤖 This gives a target peptide its entrapment partner as a pure function of (sequence, seed, fold), holds the cleavage sites so the partner still digests where the target did, and keeps the three reasons a partner can be unavailable apart so a caller knows whether to lower the fold, change the target set, or accept that arithmetic has ruled. The disjoint-stretch scheme is what earns the PR: the r partners come out distinct without a distinctness check, and fold k is independent of folds 0…k−1, which is what actually lets a database be regenerated in pieces.

One hole in the determinism claim, and it is in the claim rather than the design. `DeriveOffset` interpolates `$"{seed}:{sequence}"`, which formats the seed through the current culture. I measured this directly: `en-US` renders `-7` with U+002D, while `sv-SE`, `fi-FI` and `lt-LT` all render it with U+2212 MINUS SIGN. Different string, different SHA-256, different offset, different entrapment sequence. So "byte-identical on another machine, another runtime or in another decade" holds for the default seed of 1 and fails for a negative one — and `seed` is a plain `int` on a public signature all the way up. `EntrapmentAccession.Format` already reaches for `CultureInfo.InvariantCulture`; this is the place it matters more.

Not asking for a rebase over it. #1257 already edits this file, so the fix can ride there.

Worth narrowing one sentence in the body, too. "Distinct by construction" is true per target across folds, which is what the code delivers. It is not distinctness across targets: two anagram targets share a permutation space and derive independent offsets into it, so both can be handed the same partner. #1255's own `AmbiguousPeptides` already refuses to resolve exactly that population, so nothing here needs changing — but anyone computing a ratio over *distinct* sequences wants the narrower statement.

Nothing outside the tests consumes any of this yet, so no existing output moves.

What I did not verify: the proteome-scale figures, and the cross-target collision rate on real data — that mechanism was only forced under a reduced alphabet.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
trishorts: Fixed in #1265. You are right and I could reproduce it — the seed goes through `seed.ToString(CultureInfo.InvariantCulture)` now, matching what `EntrapmentAccession.Format` already did for the fold.

There is a test that builds the same sequence with `seed: -7` under `en-US`, `sv-SE`, `fi-FI` and `lt-LT` and asserts the four sequences are equal, so the claim in the remark is now checked rather than asserted.

trishorts:

PR #1254, opened 2026-08-30
Assemble an entrapment sequence from a target one

trishorts:
# Assemble an entrapment sequence from a target one

> **Stacks on #1253, which stacks on #1250.** This branch carries their commits too; its own are
> the last three (`62692baa`, `7ff...` tightening one assertion). Merge #1250 then #1253 and this
> diff reduces to the two files below.

Adds `EntrapmentAssembler` to `UsefulProteomicsDatabases/EntrapmentGeneration/`. Two files, nothing
existing touched.

## What it does

Splits a target sequence at its digestion agent's own cleavage sites, gives each base piece an
isomeric partner via #1253, and concatenates the partners back into a sequence.

```csharp
EntrapmentAssembly assembly = EntrapmentAssembler.Assemble(
    protein.BaseSequence, digestionParams, everyTargetPeptide, fold: k, foldCount: r);
```

Because the cleavage sites stay where they were, the entrapment sequence **digests into the same
pieces**, each isomeric with its target counterpart -- same residues, same mass, same count of any
residue of interest.

## Missed cleavages come along for free

Rearranging only the zero-missed-cleavage pieces is enough. A missed-cleavage peptide is a run of
adjacent base pieces, and composition adds, so it is isomeric with its target counterpart without
being handled separately.

The exception is a run that **spans an excision**: it has no counterpart at all. Those are counted
(`MissedCleavagePeptidesSpanningAnExcision`) rather than left implicit, because anyone computing
mass statistics over the entrapment set will otherwise assume the invariant holds everywhere.

## Excision is length-aware, and that is not a detail

A piece with no possible partner is **excised only when it is long enough to be identified on its
own**. A shorter one (`K`, `AK`, `SR`) is kept verbatim: it puts no real peptide into the entrapment
database, and identity is a rearrangement, so isomerism survives.

The naive rule -- excise whenever no partner exists -- looks equivalent and is not. Applied to the
human reviewed proteome it excised every short piece and removed **17.7%** of the peptide
population, stratified from -33.5% at zero candidate sites to -0.6% at six to nine. Per-piece
equality could not see it, because equality is pointwise over pieces present in both sequences and
a removed piece satisfies it vacuously. See the distribution test below.

## Reuse

`DigestionAgent.GetDigestionSiteIndices` already returns the complete partition -- 0, every internal
cleavage site, and the length -- so no cleavage logic is reimplemented here. One wrinkle: it is
built from a `HashSet`, so the returned list needs sorting before use as a partition.

## Tests

Ten, in `Test/DatabaseTests/TestEntrapmentAssembly.cs`:

- one partner per target piece, each isomeric, in order;
- the whole sequence isomeric, with the candidate-site count preserved;
- a sub-length piece kept verbatim rather than excised;
- a searchable piece with no partner excised, leaving its neighbours' digestion untouched;
- no target peptide ever left in the entrapment sequence;
- the exact count of missed-cleavage peptides broken by an excision;
- every retained position mapped to a position holding the same residue, excised ones marked;
- determinism, and a different sequence per fold;
- **the candidate-site distribution reproduced, with excision as the only permitted difference.**

That last one is the assertion that matters, and it earned its place immediately: written first as
"the distributions are identical" it failed, correctly -- identical was never the right invariant
once a peptide can be legitimately removed. It now also asserts the fixture actually exercises
excision, so it cannot pass vacuously.

856 passed / 0 failed across the database, decoy and entrapment suites.

## Notes for review

- `MzLibException` for the empty-sequence guard, matching the neighbours.
- `#nullable enable`; builds with zero warnings.
- Sequence-level only. Turning an assembly into a database entry -- accession scheme, pairing by
  composition, and modification transfer along `TargetToEntrapmentPosition` -- is the next PR.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -1,0 +1,221 @@
+#nullable enable
+using MzLibUtil;
+using Omics.Digestion;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>What became of one base piece of a target sequence.</summary>
+public enum PieceOutcome
+{
+    /// <summary>Rearranged into an isomeric partner.</summary>
+    Permuted,
+
+    /// <summary>
+    /// Has no rearrangement, but is shorter than the minimum peptide length, so it is never
+    /// identified on its own. Kept exactly as it was: no real peptide enters the entrapment
+    /// database, and identity is a rearrangement, so isomerism survives.
+    /// </summary>
+    KeptVerbatimTooShort,
+
+    /// <summary>
+    /// Has no usable rearrangement and *is* long enough to be identified, so it is dropped from the
+    /// entrapment sequence. Emitting it unchanged would place a genuine target peptide inside the
+    /// entrapment database, where every hit is supposed to be false.
+    /// </summary>
+    Excised
+}
+
+/// <summary>One base piece of the target and what happened to it.</summary>
+public sealed class EntrapmentPiece
+{
+    internal EntrapmentPiece(int index, string targetPiece, string? entrapmentPiece,
+        PieceOutcome outcome, EntrapmentFailure failure)
+    {
+        Index = index;
+        TargetPiece = targetPiece;
+        EntrapmentPiece_ = entrapmentPiece;
+        Outcome = outcome;
+        Failure = failure;
+    }
+
+    /// <summary>Ordinal of this piece within the target, counted before anything was excised.</summary>
+    public int Index { get; }
+
+    public string TargetPiece { get; }
+
+    /// <summary>The piece as it appears in the entrapment sequence, or null when excised.</summary>
+    public string? EntrapmentPiece_ { get; }
+
+    public PieceOutcome Outcome { get; }
+
+    /// <summary>Why no partner was found, when <see cref="Outcome"/> is not <see cref="PieceOutcome.Permuted"/>.</summary>
+    public EntrapmentFailure Failure { get; }
+}
+
+/// <summary>An entrapment sequence, and the record of how it was built.</summary>
+public sealed class EntrapmentAssembly
+{
+    internal EntrapmentAssembly(string targetSequence, string entrapmentSequence,
+        int[] targetToEntrapmentPosition, IReadOnlyList<EntrapmentPiece> pieces,
+        int missedCleavagePeptidesSpanningAnExcision)
+    {
+        TargetSequence = targetSequence;
+        EntrapmentSequence = entrapmentSequence;
+        TargetToEntrapmentPosition = targetToEntrapmentPosition;
+        Pieces = pieces;
+        MissedCleavagePeptidesSpanningAnExcision = missedCleavagePeptidesSpanningAnExcision;
+    }
+
+    public string TargetSequence { get; }
+
+    public string EntrapmentSequence { get; }
+
+    /// <summary>
+    /// Position in the entrapment sequence holding each target position's residue, or -1 where the
+    /// piece was excised. Modifications ride across on this.
+    /// </summary>
+    public int[] TargetToEntrapmentPosition { get; }
+
+    public IReadOnlyList<EntrapmentPiece> Pieces { get; }
+
+    /// <summary>
+    /// Missed-cleavage peptides of the entrapment sequence whose constituent pieces were not
+    /// adjacent in the target, so they have no isomeric counterpart there.
+    /// </summary>
+    /// <remarks>
+    /// Every other missed-cleavage peptide is isomeric for free, because base-piece compositions
+    /// add. That is true but not obvious, and these exceptions are counted rather than left
+    /// implicit, because anyone computing mass statistics over the entrapment set will assume the
+    /// invariant holds everywhere.
+    /// </remarks>
+    public int MissedCleavagePeptidesSpanningAnExcision { get; }
+
+    public int ExcisedCount => Pieces.Count(p => p.Outcome == PieceOutcome.Excised);
+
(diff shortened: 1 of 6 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,221 @@
...
+        int[] map = Enumerable.Repeat(-1, targetSequence.Length).ToArray();
+        var retainedTargetIndices = new List<int>(sites.Count - 1);
+
+        for (int index = 0; index < sites.Count - 1; index++)
+        {
+            int start = sites[index];
+            int length = sites[index + 1] - start;
+            string piece = targetSequence.Substring(start, length);
+
+            EntrapmentPeptide partner = EntrapmentPeptideGenerator.Create(piece, motifs, forbiddenSequences,
```
acesnik, line 153: The pinned set is computed from motif matches *within* `piece`, which loses two classes of site.

A preventing motif pins nothing at all — `CleavageSitePositions` does `if (!fits || prevents) continue;`, so under `K[P]|` neither the K nor the P is held and moving the P invents a cleavage site. That is `chymotrypsin|P`, `elastase|P`, `Lys-C|P`, `trypsin|P` and `subtilisin|P` in the shipped table.

A multi-residue motif has its span cut by the piece boundary, so neither half matches inside its own piece — `StcE-trypsin`, and `StcE` once #1249 lands.

Same failure `CleavageSitePositions` was written to prevent within a sequence, reappearing at the seams between pieces.

acesnik (changes requested):
🤖 This takes the peptide-level partner up to a whole sequence: split at the agent's cleavage sites, partner each base piece, concatenate. The length-aware excision rule is the idea that earns it — excising only pieces long enough to be identified, with the measurement showing the naive rule removes 17.7% of the population, is a check pointwise per-piece equality cannot make. Reusing `GetDigestionSiteIndices` for the partition and sorting it because it arrives from a `HashSet` is right in both directions.

## Why I am asking for changes

The invariant in the class remarks — the entrapment sequence digests into the same pieces — does not hold for every digestion agent, and the cause is verifiable by reading rather than only by measurement. `DecoySequenceValidator.CleavageSitePositions` (#1250) does:

```csharp
(bool fits, bool prevents) = motif.Fits(sequence, i);
if (!fits || prevents) { continue; }
```

A *preventing* motif therefore pins nothing. Under `trypsin|P`'s `K[P]|`, neither the K nor the P after it is held, so a rearrangement can move that P away — creating a cleavage site the target lacked — or move some other P next to some other K, destroying one.

**This is five shipped proteases, not one.** `proteases.tsv` carries preventing motifs on `chymotrypsin|P` (101), `elastase|P` (103), `Lys-C|P` (106), `trypsin|P` (109) and `subtilisin|P` (121). And the multi-residue case is the same defect from the other end: `TX|T` cuts after the second of three residues, so one piece ends `…TX` and the next begins `T…`, neither fragment matches inside its own piece, and nothing is pinned at either seam — which covers `StcE-trypsin` and the `StcE` #1249 adds.

`trypsin|P` is an ordinary choice in real work, not an exotic one. Measured by a reviewer over 200 random 40-mers per agent: 45/200 sequences digest into a different number of pieces under `trypsin|P`, 61/200 under `StcE-trypsin`, 0/200 under `trypsin`, `Arg-C` and `Asp-N`. I verified the mechanism in the source; I did not reproduce those two counts.

I am not asking for the real fix — pinning across piece boundaries means the assembler stops being a per-piece loop, which is a rewrite. **The ask is a guard**: refuse, or at minimum report, a digestion agent carrying a preventing motif or a motif longer than one residue. #1255 builds the whole no-index pairing on this invariant, so a user picking `trypsin|P` today gets a database that digests differently from its target and peptides that then fail to pair. Landing the guard on a later branch is equally good — #1259 already reworks this file.

Nothing calls this outside the tests, so no output moves today.

One thing that is *not* a change request: the collision check runs per base piece, so a missed-cleavage run is never tested against `forbiddenSequences`. #1259 covers that as `UnrepairableRunCollisions`, so it is handled downstream — noting it only because this PR's "no target peptide ever left in the entrapment sequence" test is narrower than its name.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Fixed in #1265, as the guard rather than the rewrite — `RefuseAgentsWhoseSitesCannotBeHeld` throws for any agent carrying a preventing motif or a motif longer than one residue, with the mechanism spelled out in the message.

Your reading of `CleavageSitePositions` is what settled it: a prevented position is skipped entirely, so under `K[P]|` neither residue is held. Refusing is the honest interim precisely because the failure is silent — a database that digests differently from its target yields peptides that then fail to pair, and nothing says so. Tests cover `trypsin|P` and `StcE-trypsin`.

Pinning across piece boundaries remains the real fix and remains a rewrite.

trishorts:

acesnik (approved):
🤖 Clearing my changes-requested. The ask was a guard rather than the rewrite, and I said at the time that landing it on a later branch was equally good — `RefuseAgentsWhoseSitesCannotBeHeld` does exactly that.

Verified rather than taken from the reply: built `UsefulProteomicsDatabases` at both heads and enumerated the shipped agents. At #1265 every agent carrying a preventing or multi-residue motif throws `MzLibException`; at #1264 none did. The mechanism you confirmed back — a prevented position is skipped entirely by `CleavageSitePositions`, so under `K[P]|` neither residue is held — is the right reading, and refusing is the honest interim precisely because the failure is otherwise silent.

Two things about the guard I raised on #1265 rather than here, so they are tracked but not blocking this: it throws from `Assemble`, so a parameter-level invariant is re-checked once per protein per fold instead of being refused at the call; and the remarks name `StcE`, which is not an agent in `proteases.tsv` — the seventh affected agent is `collagenase` (`GPX|GPX`), which the guard does catch, since it keys on the motif rather than the name.

trishorts:

PR #1255, opened 2026-08-30
Turn an entrapment sequence into a database entry

trishorts:
# Turn an entrapment sequence into a database entry

> **Stacks on #1254 → #1253 → #1250.** This branch carries their commits; its own is `44f9447e`.
> Merge those three and this diff reduces to the files below.

Completes the round trip. With this, an entrapment database can be written, reloaded, searched, and
every discovery paired back to the target peptide it was built from -- **with nothing carried
alongside the file**.

```
target protein -> entrapment protein -> WriteXmlDatabase -> LoadProteinXML
               -> Digest -> EntrapmentPairing.TryResolve -> the target peptide
```

## Three pieces

**`EntrapmentAccession`** -- what an entrapment protein is called, and how to read the target and
fold back out. It follows two conventions already in the codebase rather than inventing a third:

- the **marker goes first**, because `ProteinDbLoader` decides a protein is entrapment by looking
  for the identifier *anywhere* in its accession and prepends it when absent. An accession shaped
  this way reloads as entrapment with no extra argument and is never double-prefixed.
- the **fold goes last**, matching `VariantApplication.GetAccession`, which appends derived detail.

Parsing takes the fold from the **last** `_f<digits>` rather than splitting on underscores --
UniProt entry names always contain them, and a naive split loses the target. Tested with
`SP_HUMAN_P12345`.

**`EntrapmentProteinGenerator`** -- builds the protein. Two steps, both through existing
constructors and neither through reflection: `CloneWithNewSequenceAndMods` for the sequence and
modifications, then `Protein`'s copy constructor for the accession and the entrapment flag.

Modifications move along the position map to the residue they were on, so their motif still fits
without anything being re-checked. A modification whose residue was **excised is dropped, not
relocated** -- putting it elsewhere would invent a site the target never had.

**`EntrapmentPairing`** -- resolves a discovery back to its target peptide.

## Why no index file is needed

A rearrangement preserves the residue multiset and leaves every cleavage-site residue in place, so
those two things together are a key: the partner of an entrapment peptide is the target peptide of
the same protein with the same free-residue composition and the same pinned pattern. That is
**guaranteed by the construction** rather than recorded somewhere, so there is no side file to
write, ship, or keep in step with the database. A search reports a protein accession and a peptide
sequence, and that is enough.

It indexes **missed cleavages as well as base peptides**. A missed-cleavage peptide is a run of
adjacent base pieces and composition adds, so it pairs for the same reason. (This was found by a
failing test, not by design -- the first version indexed base pieces only, and any search reporting
a missed-cleavage peptide would have failed to pair.)

Keys can collide: `P57071` really does contain `LIHTGVK` and `LIHTVGK` -- same residues, same
pinned `K`, different order. Those are reported in `AmbiguousPeptides` and **refuse to resolve**,
because guessing between them would be wrong half the time and silently so. Measured across the
human reviewed proteome, this affects **274 of 1,315,124** base peptides (0.021%).

## One change to existing code, called out deliberately

`ProteinDbLoader` gains `public const string DefaultEntrapmentIdentifier = "Random"`, and the two
existing default parameters that spelled that literal out now reference it. **The value is
identical**, so this is a duplicated literal replaced by a name rather than a behaviour change --
but it does touch existing signatures, so it should be seen rather than found.

## Tests

Thirteen, in `Test/DatabaseTests/TestEntrapmentProtein.cs`, in three groups:

- **accession** -- marker first, target and fold round-trip, non-entrapment accessions refused,
  and a target accession containing underscores;
- **generation** -- isomeric with the target, modifications landing on the residue their motif
  requires, modifications on excised residues dropped, a distinct accession and sequence per fold;
- **pairing and round trip** -- resolution from composition alone, ambiguous keys refusing to
  guess, reload from XML recognised as entrapment by accession alone, and pairing surviving a full
  write/read/digest cycle.

870 passed / 0 failed across the database, decoy, entrapment and protein-DB suites.

## Notes for review

- `MzLibException` throughout, matching the neighbours.
- `#nullable enable`; zero warnings.
- Protein-specific. `IBioPolymer` exposes `CloneWithNewSequenceAndMods` but no way to set an
  accession, so RNA parity needs a separate look rather than a speculative generic signature here.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
@@ -1,0 +1,84 @@
+#nullable enable
+using System;
+using System.Globalization;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>
+/// The accession an entrapment protein wears, and how to read the target and fold back out of it.
+/// </summary>
+/// <remarks>
+/// <para>Two existing conventions are followed rather than a third invented. The entrapment marker
+/// goes <b>first</b>, because <see cref="ProteinDbLoader"/> decides a protein is entrapment by
+/// looking for the identifier anywhere in its accession, and prepends it when absent -- so an
+/// accession shaped this way reloads as entrapment with no extra argument and is never
+/// double-prefixed. The fold goes <b>last</b>, matching
+/// <c>VariantApplication.GetAccession</c>, which appends derived detail as a suffix.</para>
+/// <para>Everything needed to pair a discovery back to its target therefore travels in the
+/// accession, which a search already reports. Nothing has to be carried alongside the database, and
+/// nothing leaks into the sequence -- which matters, because target and partner are deliberately
+/// identical in mass and composition and must stay that way.</para>
+/// </remarks>
+public static class EntrapmentAccession
+{
+    private const string FoldMarker = "_f";
+
+    /// <summary>The accession for one fold of one target's entrapment partner.</summary>
+    /// <exception cref="MzLibUtil.MzLibException">The target accession is missing, or the fold is negative.</exception>
+    public static string Format(string targetAccession, int fold,
+        string entrapmentIdentifier = ProteinDbLoader.DefaultEntrapmentIdentifier)
+    {
+        if (string.IsNullOrEmpty(targetAccession))
+        {
+            throw new MzLibUtil.MzLibException("An entrapment accession needs a target accession.");
+        }
+        if (fold < 0)
+        {
+            throw new MzLibUtil.MzLibException($"Fold must not be negative, but was {fold}.");
+        }
+
+        return $"{entrapmentIdentifier}_{targetAccession}{FoldMarker}{fold.ToString(CultureInfo.InvariantCulture)}";
+    }
+
+    /// <summary>
+    /// Reads the target accession and fold back out, or reports that this is not one of ours.
+    /// </summary>
+    /// <remarks>
+    /// The fold is taken from the <em>last</em> "_f&lt;digits&gt;" rather than by splitting on
+    /// underscores: UniProt entry names contain them, so a naive split loses the target.
+    /// </remarks>
+    public static bool TryParse(string? accession, out string targetAccession, out int fold,
+        string entrapmentIdentifier = ProteinDbLoader.DefaultEntrapmentIdentifier)
+    {
+        targetAccession = string.Empty;
+        fold = 0;
+
+        string prefix = entrapmentIdentifier + "_";
+        if (string.IsNullOrEmpty(accession) || !accession.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+
+        int marker = accession.LastIndexOf(FoldMarker, StringComparison.Ordinal);
+        if (marker <= prefix.Length - 1)
+        {
+            return false;
+        }
+
+        string digits = accession.Substring(marker + FoldMarker.Length);
+        if (digits.Length == 0
+            || !int.TryParse(digits, NumberStyles.None, CultureInfo.InvariantCulture, out int parsedFold))
+        {
+            return false;
+        }
+
+        targetAccession = accession.Substring(prefix.Length, marker - prefix.Length);
+        if (targetAccession.Length == 0)
+        {
+            return false;
+        }
+
+        fold = parsedFold;
+        return true;
+    }
+}
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
@@ -1,0 +1,102 @@
+#nullable enable
+using MzLibUtil;
+using Omics.Digestion;
+using Omics.Modifications;
+using Proteomics;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>
+/// Builds the entrapment partner of a target protein: an isomeric sequence, its modifications
+/// carried onto the residues they were on, and an accession that says what it is and where it came
+/// from.
+/// </summary>
+public static class EntrapmentProteinGenerator
+{
+    /// <summary>
+    /// The entrapment partner of <paramref name="target"/> for one fold.
+    /// </summary>
+    /// <param name="target">The target protein.</param>
+    /// <param name="digestionParams">Supplies the cleavage sites, and the minimum peptide length
+    /// below which a piece is not identifiable on its own.</param>
+    /// <param name="forbiddenSequences">Sequences no partner peptide may equal, normally every
+    /// target peptide in the database.</param>
+    /// <param name="fold">Zero-based fold, in <c>[0, foldCount)</c>.</param>
+    /// <param name="foldCount">Partners per target -- the <c>r</c> of an r-fold database.</param>
+    /// <param name="seed">Changes every choice reproducibly.</param>
+    /// <remarks>
+    /// Built in two steps, both through existing constructors and neither through reflection:
+    /// <see cref="Protein.CloneWithNewSequenceAndMods"/> for the sequence and modifications, then
+    /// Protein's copy constructor for the accession and the entrapment flag.
+    /// </remarks>
+    public static Protein Create(Protein target, IDigestionParams digestionParams,
+        IReadOnlySet<string> forbiddenSequences, int fold = 0, int foldCount = 1, int seed = 1,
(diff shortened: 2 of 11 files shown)


mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
```diff
@@ -82,10 +82,16 @@ public static class ProteinDbLoader
         /// If protein modifications are specified both in the mzLibProteinDb XML file and in allKnownModifications, they are collapsed into a HashSet of Modifications before generating Protein entries.
         /// </summary>
         [SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times")]
+        /// <summary>
```
acesnik, line 85: These two lines landed between the `[SuppressMessage(… CA2202 …)]` above and the `LoadProteinXML` below, so the attribute and this doc comment both bind to the const and the loader's own doc comment is orphaned. Built at this head:

```
ProteinDbLoader.cs(85,9): warning CS1587: XML comment is not placed on a valid language element
```

CS1587 is not in the project's `/nowarn` list and master builds clean. Consequences: CA2202 now suppresses on a `const string`; `LoadProteinXML` loses both the suppression and its documentation; and the shipped `.xml` documents the const with the loader's summary. Moving the const above the loader's doc comment fixes all four.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        EntrapmentAssembly assembly = EntrapmentAssembler.Assemble(target.BaseSequence, digestionParams,
+            forbiddenSequences, fold, foldCount, seed);
+
+        Dictionary<int, List<Modification>> movedMods =
+            MoveModifications(target.OneBasedPossibleLocalizedModifications, assembly.TargetToEntrapmentPosition);
+
+        var withNewSequence = (Protein)target.CloneWithNewSequenceAndMods(assembly.EntrapmentSequence, movedMods);
+
+        return new Protein(withNewSequence,
```
acesnik, line 51: The copy constructor keeps the target's `UniProtSequenceAttributes`, so an excised partner writes its target's length and mass — a 15-residue sequence emitted as `<sequence length="30" mass="3114" …>`. `LoadProteinXML` recomputes length on read, which is why the round trip looks clean, but the file on disk is self-inconsistent for any other reader. Fresh attributes built from the entrapment sequence would settle it. #1256 fixes the sibling annotations; these are still stale at #1260.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
```diff
@@ -0,0 +1,84 @@
...
+        targetAccession = string.Empty;
+        fold = 0;
+
+        string prefix = entrapmentIdentifier + "_";
+        if (string.IsNullOrEmpty(accession) || !accession.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+
+        int marker = accession.LastIndexOf(FoldMarker, StringComparison.Ordinal);
```
acesnik, line 62: Line 57 matches the prefix `OrdinalIgnoreCase`, this matches the `_f` marker `Ordinal`. So `RANDOM_P12345_F0` passes the first test and fails the second. Ordinal throughout would match what `Format` writes.

acesnik (changes requested):
🤖 This closes the loop: an entrapment protein that can be written, reloaded, digested and paired back to the target peptide it came from with nothing carried alongside the file. The argument for why no index is needed is the substance and it holds — a rearrangement preserves the residue multiset and the pinned pattern, so those two together are a key by construction rather than by bookkeeping. Refusing to resolve an ambiguous key rather than guessing, with the rate measured, is the right call and the right way to report it.

## The `ProteinDbLoader` edit needs to go back

The body describes it as "the value is identical, so this is a duplicated literal replaced by a name". Semantically that is true — I checked, and the substring-match classification and the `"Random"` default are both already on `master`. But the two new lines landed **between** `[SuppressMessage(… CA2202 …)]` and the `LoadProteinXML` it applied to, and I built the branch rather than reasoning about it:

```
ProteinDbLoader.cs(85,9): warning CS1587: XML comment is not placed on a valid language element
```

CS1587 is not in this project's `/nowarn` list, and `master` builds it clean. Three further consequences follow from the same placement: the CA2202 suppression now applies to a `const string`, where "do not dispose objects multiple times" means nothing, while `LoadProteinXML` — where the `using` nesting that provoked it lives — no longer has it; and in the generated documentation the const inherits the loader's summary while the loader gets none.

That `.xml` ships (`mzLib.nuspec`), so what reaches every consumer including MetaMorpheus is a `const string` whose tooltip describes XML loading. Moving the two lines above the existing doc comment fixes all of it. `ProteinDbLoader.cs` is untouched by #1256–#1260, so this is still there at the top of the stack.

## The entrapment protein inherits its target's sequence attributes

The copy constructor keeps `UniProtSequenceAttributes`, and excision changes the length, so the written XML disagrees with itself — a 15-residue partner emitted as `<sequence length="30" mass="3114" …>`. `LoadProteinXML` recomputes length on read, which is why the round trip passes, but the file is wrong for anything else that reads it. #1256 fixes the sibling half of this for truncation products and variations with a measurement I would take as the reason to finish the job here; the sequence attributes are still stale at #1260.

## An observation rather than an ask

Naming the identifier makes it easy to notice what the classification actually is: any accession *containing* "Random", case-insensitively (`ProteinDbLoader.cs:374`). That predates this stack, so it is not yours — but before these PRs nothing generated entrapment databases, so the misclassification was latent. GRCh38 names unplaced scaffolds `chr1_KI270706v1_random`, and a proteogenomic database built from such an assembly can carry that in an accession. Worth a follow-up issue rather than anything here.

What I verified: the CS1587 by building; the classification logic and its presence on master by reading. What I did not: the 274/1,315,124 ambiguity rate or any proteome-scale figure.

mzLib/UsefulProteomicsDatabases/ProteinDbLoader.cs
trishorts: Fixed in #1265. The const and its doc comment now sit above the loader's own documentation, so `[SuppressMessage]` binds to `LoadProteinXML` again and the const documents itself.

Confirmed the way you found it — building `UsefulProteomicsDatabases` now reports **zero** CS1587. Thank you for building the branch rather than reasoning about it; I would not have caught the shipped-`.xml` consequence.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Fixed in #1265. `DescribeSequence` derives length and mass from the entrapment sequence instead of inheriting the target's.

Mass turns out to need the same treatment for a different reason than length: a rearrangement is isomeric, so an unexcised partner has exactly its target's mass and only an excised one does not — deriving both from the sequence is correct either way and does not depend on knowing which happened. Applied on the proteoform path too, where nothing is excised, so that it cannot drift if that ever changes.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
trishorts: Fixed in #1265 — `Ordinal` throughout, matching what `Format` writes.

`RANDOM_P12345_F0` passing the prefix test and failing the marker test was the sharp end of it: the accession was neither accepted nor rejected consistently.

trishorts:

acesnik (approved):
🤖 Clearing my changes-requested — all three are addressed, and I checked each at #1265 rather than reading the replies.

The const now sits above the loader's documentation, and a clean build of `UsefulProteomicsDatabases` reports **zero** CS1587 where #1264 reported it at `ProteinDbLoader.cs(85,9)`, with `[SuppressMessage(… CA2202 …)]` back on `LoadProteinXML`. `DescribeSequence` derives length and mass from the entrapment sequence — a 34-residue target excised to 26 now writes `Length=26 Mass=2861` where it wrote the target's `34`/`12345` — and your point that a rearrangement is isomeric, so deriving both is correct whether or not anything was excised, is the better reason than the one I gave. `EntrapmentAccession` is `Ordinal` throughout: `RANDOM_P12345_f0` and `random_P12345_f0` are now both rejected where the first was accepted, and `Random_P12345_f0` still parses.

Three residuals from that same work are on #1265 and not held against this PR: `Checksum` and `SequenceVersion` are still copied from the target while length and mass are derived; `UpdateMassAttribute` yields `mass="0"` for a sequence containing `X` or `B`; and `ProteinDbLoader` still matches the entrapment identifier `OrdinalIgnoreCase`, so the case mismatch moved rather than went away.

trishorts:

PR #1256, opened 2026-08-30
Report what an entrapment database actually contains

trishorts:
# Report what an entrapment database actually contains

> **Stacks on #1255 → #1254 → #1253 → #1250.** Its own commit is `64f061f4`. Merge those four and
> this diff reduces to `EntrapmentReport.cs`, its tests, and a small overload on
> `EntrapmentProteinGenerator`.

An entrapment database is only usable if you can say what came out of it. This adds the report,
and it is a first-class output rather than a debugging aid.

```csharp
var builder = new EntrapmentReportBuilder(digestionParams, foldCount, seed,
                                          EntrapmentReport.CountResidues("ST"));
foreach (Protein target in targets)
    for (int fold = 0; fold < foldCount; fold++)
    {
        Protein entrapment = EntrapmentProteinGenerator.Create(
            target, digestionParams, forbidden, out EntrapmentAssembly assembly, fold, foldCount);
        builder.Add(target, fold, assembly);
    }

File.WriteAllText(path, builder.Build().ToTabSeparated());
```

## The distribution, not a summary of it

There is no universal distribution of peptides to validate a database against -- it moves with the
proteome, the enzyme and the digestion settings. Only this database's own numbers mean anything, so
the report emits **every stratum**, ascending, plus a total. A single headline figure would hide
exactly the stratum-by-stratum differences worth looking at.

## Failures kept apart by cause *and* stratum

| | meaning | remedy |
|---|---|---|
| `UnpairableNoPermutationExists` | one arrangement exists once cleavage sites are pinned | none -- arithmetic |
| `UnpairableSpaceTooSmallForFoldCount` | real space, too small to give every fold a partner | ask for fewer folds |
| `UnpairableAllPermutationsTaken` | every arrangement available was already spoken for | depends on the target set |

Collapsing these into one "failed" hides the distinction from whoever has to act on it.

## Achieved r, and the invariant that does not always hold

`AchievedFoldRatio` is **partners delivered per target peptide**, counting each target once however
many folds run, so it reads directly against the fold count that was requested and a shortfall is
visible per stratum.

`MissedCleavagePeptidesSpanningAnExcision` counts the only peptides in the database that are *not*
isomeric with a target counterpart. Every other missed-cleavage peptide is isomeric for free
because base-piece compositions add; someone will eventually compute mass statistics over the
entrapment set assuming that holds everywhere, so the exceptions are counted rather than left
implicit.

## Stratify by whatever you are measuring

`CountResidues("ST")` gives the candidate O-glycosite count, which is what sets the difficulty of
placing a modification: with no evidence the best possible is 1/n. `CountNGlycoSequons` counts
N-X-S/T with X not proline -- a quantity a rearrangement does **not** preserve, which is precisely
why it is reported. Drift is acceptable; drift nobody can see is not. Any `Func<string,int>` works.

Counts are over peptides a search could actually report -- base pieces at least `MinLength` long.
Counting shorter pieces would flatter every ratio here.

## Provenance travels with the numbers

Method, enzyme, seed, fold count, missed cleavages, min and max length. Without them the numbers can
be neither interpreted nor reproduced, since the peptide population depends on the enzyme and
digestion settings as much as on the method.

## Tests

Twelve, in `Test/DatabaseTests/TestEntrapmentReport.cs`, including one that counts the entrapment
peptides independently and checks the report matches -- a report is only worth reading if it
describes the database that actually came out.

891 passed / 0 failed across the database, decoy, entrapment and protein-DB suites.

## Mutation testing found two real defects in the PRs below this one

Running Stryker over the four files in #1254 and #1255 scored them 86.5%, 74.2%, 76.9% and 54.5%,
and two survivors were dangerous rather than cosmetic. Both are fixed in this PR:

- **`retained[i + span - 1] + retained[i]` survived** in the assembler's broken-run count. My
  fixture's retained ordinals were `[0,1,3]`, where the `+` and `-` forms coincidentally agree. The
  fixture now puts the excision second, giving `[0,2,3]`, where the contiguous run `{2,3}` must not
  count and the two forms diverge.
- **`destination <= 0` survived** in the modification mover. Position **zero is a real destination**,
  not a missing marker -- only a negative means excised. Conflating them would silently drop a
  modification whenever its residue moved to the front of the sequence.

The remaining survivors were untested guard clauses (null targets, empty sequences, malformed
accessions), now covered, and a handful of equivalent mutants left alone deliberately.

---

### Review follow-up

**This PR changes the database it produces, and that should have led the description.**
`EntrapmentProteinGenerator` now clears sequence variations, applied variations, truncation
products, disulfide bonds and splice sites on every entrapment entry. It changes the output **from
one no search could load** — 131 entries carried a feature position beyond their own length and
every search against that database failed. Right, tested, and invisible in a PR titled for a QC
report.

The phantom zero-site row and the `Ambiguous` population mismatch are both fixed in #1265 and
#1260 respectively; the vacuous truncation-product loop is rewritten in #1265.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
@@ -1,0 +1,316 @@
+#nullable enable
+using MzLibUtil;
+using Omics.Digestion;
+using Proteomics;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>How the database was made, recorded alongside what came out of it.</summary>
+/// <remarks>
+/// Without these an entrapment database cannot be regenerated or compared against another, and the
+/// numbers below cannot be interpreted -- the peptide population, and therefore every distribution
+/// in the report, depends on the enzyme and the digestion settings as much as on the method.
+/// </remarks>
+public sealed class EntrapmentProvenance
+{
+    internal EntrapmentProvenance(string method, string enzyme, int seed, int foldCount,
+        int maxMissedCleavages, int minPeptideLength, int maxPeptideLength)
+    {
+        Method = method;
+        Enzyme = enzyme;
+        Seed = seed;
+        FoldCount = foldCount;
+        MaxMissedCleavages = maxMissedCleavages;
+        MinPeptideLength = minPeptideLength;
+        MaxPeptideLength = maxPeptideLength;
+    }
+
+    public string Method { get; }
+    public string Enzyme { get; }
+    public int Seed { get; }
+    public int FoldCount { get; }
+    public int MaxMissedCleavages { get; }
+    public int MinPeptideLength { get; }
+    public int MaxPeptideLength { get; }
+}
+
+/// <summary>Counts for one stratum -- one value of whatever the report is stratified by.</summary>
+public sealed class EntrapmentStratum
+{
+    internal EntrapmentStratum(int siteCount) => SiteCount = siteCount;
+
+    /// <summary>The value being stratified by, e.g. the number of candidate sites in a peptide.</summary>
+    public int SiteCount { get; }
+
+    /// <summary>Distinct target peptides in this stratum, counted once however many folds ran.</summary>
+    public int TargetPeptides { get; internal set; }
+
+    /// <summary>Entrapment peptides produced for them, across every fold.</summary>
+    public int EntrapmentPeptides { get; internal set; }
+
+    /// <summary>Exactly one arrangement exists. Arithmetic: no seed and no fold count can help.</summary>
+    public int UnpairableNoPermutationExists { get; internal set; }
+
+    /// <summary>Arrangements exist but every one available to the fold was already spoken for.</summary>
+    public int UnpairableAllPermutationsTaken { get; internal set; }
+
+    /// <summary>The space is real but too small to give every fold its own partner.</summary>
+    public int UnpairableSpaceTooSmallForFoldCount { get; internal set; }
+
+    /// <summary>
+    /// Target peptides sharing a composition-and-pinning key with another peptide of the same
+    /// protein, so a discovery cannot be traced back to one of them.
+    /// </summary>
+    public int Ambiguous { get; internal set; }
+
+    /// <summary>
+    /// Entrapment missed-cleavage peptides whose pieces were not adjacent in the target, so they
+    /// have no isomeric counterpart. The only peptides in the database that break that invariant.
+    /// </summary>
+    public int MissedCleavagePeptidesSpanningAnExcision { get; internal set; }
+
+    public int Unpairable => UnpairableNoPermutationExists + UnpairableAllPermutationsTaken
+                             + UnpairableSpaceTooSmallForFoldCount;
+
+    /// <summary>
+    /// Partners actually delivered per target peptide. Compare against the requested fold count:
+    /// the shortfall is where the database could not honour what was asked of it.
+    /// </summary>
+    public double AchievedFoldRatio =>
+        TargetPeptides == 0 ? 0d : EntrapmentPeptides / (double)TargetPeptides;
+}
+
+/// <summary>
+/// What an entrapment database actually contains, stratified, next to how it was made.
+/// </summary>
+/// <remarks>
+/// <para>The report emits the <b>distribution</b>, not a summary of it. There is no universal
+/// distribution of peptides to validate against -- it moves with the proteome, the enzyme and the
+/// digestion settings -- so only this database's own numbers mean anything, and a single headline
+/// figure would hide exactly the stratum-by-stratum differences worth looking at.</para>
+/// <para>Failures are kept apart by cause as well as by stratum, because the remedies differ:
+/// nothing helps a peptide with one arrangement, a smaller fold count helps one whose space is too
+/// small, and a different target database may help one whose space is merely crowded.</para>
+/// </remarks>
+public sealed class EntrapmentReport
+{
+    internal EntrapmentReport(EntrapmentProvenance provenance, IReadOnlyList<EntrapmentStratum> strata,
+        EntrapmentStratum total)
+    {
+        Provenance = provenance;
+        Strata = strata;
+        Total = total;
+    }
+
+    public EntrapmentProvenance Provenance { get; }
+
+    /// <summary>One entry per observed stratum, ascending.</summary>
+    public IReadOnlyList<EntrapmentStratum> Strata { get; }
+
+    /// <summary>The same counts across every stratum.</summary>
+    public EntrapmentStratum Total { get; }
+
+    /// <summary>Counts occurrences of any of <paramref name="residues"/> in a peptide.</summary>
+    /// <remarks>O-glycosylation localization stratifies on "ST": the count of candidate sites in a
+    /// peptide is what sets the difficulty of placing a modification on it.</remarks>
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,316 @@
...
+                    stratum.UnpairableSpaceTooSmallForFoldCount++;
+                    break;
+                case PieceOutcome.Excised:
+                    stratum.UnpairableAllPermutationsTaken++;
+                    break;
+            }
+        }
+
+        // Not attributable to any one stratum: a broken run spans several peptides at once.
+        StratumFor(string.Empty).MissedCleavagePeptidesSpanningAnExcision +=
```
acesnik, line 272: `StratumFor(string.Empty)` is the `siteCount == 0` stratum — the implementation short-circuits to 0 and then creates or reuses the real one. So this whole-protein count is emitted in a row beside that stratum's own `targetPeptides`, and the row appears even when every peptide in the database carries at least one site. The comment above is right that the count is not attributable to a stratum; stratum 0 is a stratum. Accumulating on `Total` directly, or labelling the row distinctly, keeps the number without attaching it to a population it does not describe.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,316 @@
...
+    public void Add(Protein target, int fold, EntrapmentAssembly assembly)
+    {
+        if (target is null || assembly is null)
+        {
+            throw new MzLibException("A report entry needs both a target protein and its assembly.");
+        }
+
+        if (!_ambiguousByAccession.TryGetValue(target.Accession, out HashSet<string>? ambiguous))
+        {
+            ambiguous = new EntrapmentPairing(target, _digestionParams).AmbiguousPeptides.ToHashSet();
```
acesnik, line 228: `AmbiguousPeptides` has no length filter and spans missed-cleavage runs, so this counts peptides the `MinLength` guard a few lines below excludes from every other column — measured on cRAP as 6 ambiguous peptides, all 6 below `MinLength`. `Ambiguous / TargetPeptides` is therefore a ratio over two different populations, which is the thing #1260 sets out to fix one layer up.

Separately: the `HashSet` stored here is never read again, only its key is, so a `HashSet<string>` of accessions would do and would stop retaining every ambiguous peptide in the proteome.

acesnik (changes requested):
🤖 This is the piece that makes the rest usable: until now a database came out and there was no way to say what was in it. Emitting every stratum rather than a headline figure is right, keeping the three unpairable causes apart is what makes a shortfall actionable rather than alarming, and `Report_NumbersReconcileWithTheDatabaseActuallyWritten` is the test that gives the other eleven their weight — a report nobody has recounted is decoration. Determinism checks out: `_strata` is a `Dictionary` but `Build()` sorts by `SiteCount` first and every accumulation is commutative, so dictionary order never reaches the output; three runs including one with reordered proteins came out byte-identical.

## The output change belongs in the body

`EntrapmentProteinGenerator` now clears sequence variations, applied variations, truncation products, disulfide bonds and splice sites on every entrapment entry. That changes the produced database, and on the code comment's own account it changes it *from one no search could load* — 131 entries carrying a feature position beyond their own length, every search against that database failing. It is right, it is tested, and it is invisible in a PR titled for a QC report. It should be the first thing the body says.

## Two emitted numbers do not measure what their names say

**A phantom zero-site row.** `StratumFor(string.Empty)` resolves to `siteCount = 0` — I read the implementation, it short-circuits `peptide.Length == 0 ? 0 : _siteCounter(peptide)` and then creates or reuses the *real* zero-site stratum. So a whole-protein figure is published beside that stratum's own per-peptide counts, and the row is materialised even when nothing in the database has zero sites. Someone comparing two reports line for line reads that as a property of zero-site peptides.

**`Ambiguous` counts the peptides every other column excludes.** The builder's remark says counts are over base pieces at least `MinLength` long and line 239 enforces it, but the ambiguity pass takes `EntrapmentPairing.AmbiguousPeptides` whole, and that collection has no length filter and spans runs. On cRAP a reviewer measured 6 ambiguous peptides, all 6 shorter than `MinLength` — so `Ambiguous` and `TargetPeptides` are counts over different populations sharing a denominator.

## And one test made vacuous by this PR's own fix

`Create_NeverLeavesAnAnnotationPointingPastTheSequence` loops `entrapment.TruncationProducts` asserting each end position is within the sequence. This PR passes `proteolysisProducts: new List<TruncationProduct>()`, so that collection is always empty and the loop body never executes — the only live assertion is that the sequence got shorter. The fix is correct; the test named for it cannot observe it. I verified both halves by reading.

Also worth knowing, and not yours to fix: the reconciliation claim holds only for agents with no preventing-cleavage rule, because `CleavageSitePositions` skips a prevented position entirely. That is #1254's territory, and the one test exercising reconciliation uses trypsin only.

What I verified: the stratum resolution, the empty-list pass, the determinism argument. What I did not: the cRAP counts or any proteome-scale figure.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Fixed in #1265. There is now a `_wholeProtein` accumulator that reaches `Total` and never a stratum, so `mcSpanningAnExcision`, `searchSpacePeptides` and `unrepairableRunCollisions` no longer appear in the zero-site row.

The totals test was asserting `Total == sum(Strata)` field by field, which would have hidden this; it now asserts the stratum sum for those three columns is **zero** and that the total still carries them. Stating the contract rather than the arithmetic.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Both fixed, in two places.

The population mismatch is closed in #1260 as you expected — `EntrapmentPairing` now applies the digestion length bounds, so `Ambiguous` counts only peptides a search could report. On the reviewed human proteome that took it from 6,889 to 736, which matches an independent digest of the written database exactly.

The retained `HashSet` is no longer dead: #1261 emits the ambiguous peptides themselves in an exclusion table, because a consumer computing the paired estimator has to exclude the same peptides this generator did or its `r = 1` assumption fails silently. So the collection is now read rather than only its key — the memory is being spent on something.

trishorts:

acesnik (approved):
🤖 Clearing my changes-requested. Both asks are addressed, though they landed in three different places, so worth recording where.

The whole-protein figures no longer reach a stratum row: the `_wholeProtein` accumulator goes to `Total` only, and I confirmed it on a three-protein report — the stratum sum of `mcSpanningAnExcision`, `searchSpacePeptides`, `entrapmentSearchSpacePeptides` and `unrepairableRunCollisions` is 0 across every row while `Total` carries 2 / 18 / 14 / 0. Rewriting the totals test to assert the stratum sum is zero, rather than asserting `Total == sum(Strata)` field by field, is the change that makes it stay fixed; the old shape would have passed either way.

The `Ambiguous` population mismatch landed in #1260, as you said it would, and I verified it there — the pairing index applies the digestion length bounds, so the count and its denominator finally describe the same population. Your 6,889 → 736 on the reviewed human proteome is your measurement, not mine; I have no proteome here to reproduce it.

The retained `HashSet` I called dead is genuinely alive now that #1261 emits the ambiguous peptides in the exclusion table, so the memory buys something. Noting that the exclusion table has its own counting problem, which I raised on #1261.

trishorts:

PR #1257, opened 2026-08-31
Anchor the protein termini so terminal modifications survive

trishorts:
# Anchor the protein termini so terminal modifications survive

> **Stacks on #1256 → #1255 → #1254 → #1253 → #1250.** Its own commits are the last two.

A rearrangement moves an N- or C-terminally restricted modification off the terminus. mzLib then
rejects it as invalid for its location — correctly — and **nothing counts the loss**.

Measured on the reviewed human proteome, entrapment built from a GPTMD-pruned database:

| | modifications out | lost | `N-terminal.` | `C-terminal.` |
|---|---|---|---|---|
| before | 158,783 | **3,922 (2.41%)** | 3,890 | 31 |
| **after** | **162,699** | **6 (0.00%)** | 5 | 0 |

The six that remain sit on pieces that were **excised**, where losing the modification is the
intended behaviour — the residue is genuinely not in the database any more.

## How

The permutation primitive now accepts extra positions to hold still alongside the cleavage sites,
and the assembler supplies them: the **first two** positions of a protein's opening piece and the
last position of its closing one.

Both of the first two matter. A modification annotated **after initiator-methionine cleavage sits on
the second residue**, not the first, so anchoring only position 1 would have left roughly half the
loss in place. That came out of the failure data rather than from reasoning about it.

## Anchoring is unconditional, deliberately

The obvious economy is to anchor only where a terminal modification actually sits. That would make
the entrapment sequence a function of `(sequence, seed, annotations)` instead of `(sequence, seed)`
— so two databases built from the same proteome with different annotation sets would disagree on
their **sequences**, and the determinism the pairing scheme rests on would quietly weaken.

`Create_AnchorsTheTerminiWhetherOrNotAnythingIsModified` enforces this: the entrapment sequence must
be identical whether or not the protein carried a terminal modification.

The cost is two or three held positions per protein, against permutation spaces in the millions.

## Tests

Seven new, in `TestEntrapmentProtein.cs`:

- an N-terminal modification at position 1 survives, and at position 2 (the initiator-Met case);
- a C-terminal modification survives;
- **the modification count is conserved** — the assertion the plan asked for by name, rather than
  "the shortfall is reported";
- the sequence does not depend on whether anything was modified;
- a **single-base-piece protein**, where both anchoring branches fire at once;
- **anchoring actually shrinks the permutation space** — without this, the assertions above could
  pass by luck on a short fixture while anchoring did nothing.

97 entrapment tests pass; 929 across the database, decoy, protein-DB and digestion suites.

## Note for review

`PermutationSpaceSize` and `UnrankPermutation` gain an optional `alsoHeldInPlace` parameter. Both are
members added by #1250 and not yet merged, and the parameter is optional, so no existing call
changes. The two must be passed the same set or the index does not mean the same thing — said in
the XML docs.

---

### Review follow-up

**Every previously built entrapment database is superseded by this PR.** Anchoring changes the
sequence of every protein whose opening or closing piece has more than one arrangement — measured at
110 of 116 on cRAP. That is expected and unavoidable, but the table above is about modification
counts and does not say it.

The anchors were also being computed in *piece* coordinates rather than protein coordinates, so a
single-residue opening piece left the protein's second residue unheld. Fixed in #1265, along with a
position-2 test that could not previously discriminate.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -151,7 +151,7 @@ public static class EntrapmentAssembler
             string piece = targetSequence.Substring(start, length);
 
             EntrapmentPeptide partner = EntrapmentPeptideGenerator.Create(piece, motifs, forbiddenSequences,
-                fold, foldCount, seed);
+                fold, foldCount, seed, TerminalAnchors(index, sites.Count - 1, length));
 
             if (partner.Succeeded)
             {
@@ -181,6 +181,49 @@ public static class EntrapmentAssembler
         int broken = CountRunsSpanningAGap(retainedTargetIndices, digestionParams.MaxMissedCleavages);
 
         return new EntrapmentAssembly(targetSequence, entrapment.ToString(), map, pieces, broken);
+    }
+
+    /// <summary>
+    /// Positions within a piece that must not move because they are the PROTEIN's termini.
+    /// </summary>
+    /// <remarks>
+    /// <para>A modification can be restricted to the N- or C-terminus, and a rearrangement that
+    /// moved that residue away would make it invalid for its location -- mzLib then drops it, and
+    /// nothing counts the loss. Measured before anchoring: 3,946 N-terminal and 31 C-terminal
+    /// modifications lost across the human proteome, 2.4% of all of them.</para>
+    /// <para>Both of the first two positions are held, not just the first: a modification annotated
+    /// after initiator-methionine cleavage sits on the second residue.</para>
+    /// <para>This is unconditional -- the termini are anchored whether or not anything is actually
+    /// modified there. Anchoring reactively would make the entrapment sequence a function of
+    /// (sequence, seed, modifications) rather than (sequence, seed), so two databases built from the
+    /// same proteome with different annotations would disagree on their sequences and the
+    /// determinism the pairing rests on would quietly weaken. The cost is two or three positions per
+    /// protein, against permutation spaces in the millions.</para>
+    /// </remarks>
+    private static int[]? TerminalAnchors(int pieceIndex, int pieceCount, int pieceLength)
+    {
+        bool first = pieceIndex == 0;
+        bool last = pieceIndex == pieceCount - 1;
+        if (!first && !last)
+        {
+            return null;
+        }
+
+        var anchors = new List<int>(3);
+        if (first)
+        {
+            anchors.Add(0);
+            if (pieceLength > 1)
+            {
+                anchors.Add(1);
+            }
+        }
+        if (last && pieceLength > 0)
+        {
+            anchors.Add(pieceLength - 1);
+        }
+
+        return anchors.ToArray();
     }
 
     private static void AppendPiece(StringBuilder entrapment, int[] map, int targetStart,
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,264 @@
...
+        bool last = pieceIndex == pieceCount - 1;
+        if (!first && !last)
+        {
+            return null;
+        }
+
+        var anchors = new List<int>(3);
+        if (first)
+        {
+            anchors.Add(0);
```
acesnik, line 215: Piece coordinates, not protein coordinates. If the opening piece is one residue long the protein's second residue is never held, so an `"N-terminal."` modification at position 2 can still move. Rare — under trypsin it needs the entry to begin with K or R, and cRAP has none — but it is the case the doc above promises to cover, and it may account for part of the residual loss the body reports.

acesnik (approved):
🤖 This closes a loss that had no observer: a permutation moved an N- or C-terminally restricted modification off the terminus, `SelectValidOneBaseMods` correctly dropped it, and no count anywhere went down. Anchoring unconditionally rather than reactively is the right trade, and the reasoning is the convincing part — making the sequence a function of the annotation set would have made two databases over the same proteome disagree on their residues, which is worse than the problem being fixed.

The anchor set is exactly right, and checkable against the thing that does the dropping. `ModificationLocalization.cs:35` rejects `"N-terminal."` only `when bioPolymerOneBasedIndex > 2`, and `:37` rejects `"C-terminal."` only `when bioPolymerOneBasedIndex < sequence.Length`. Holding zero-based 0, 1 and last is the minimal set keeping both satisfiable — the "both of the first two" decision falls straight out of that `> 2`, so the failure data and the code agree.

A reviewer measured the cost on cRAP: across 232 opening and closing pieces, anchoring flipped **zero** pairable pieces to unpairable and excisions stayed at 0, so the space absorbs two or three held positions. The other side of that measurement is that **110 of 116 entrapment sequences change** — every protein whose opening or closing piece has more than one arrangement. Expected and unavoidable, but the body's table is about modification counts, and one line saying every previously built database is superseded would be worth adding.

One place the doc claims more than the code does. `TerminalAnchors` works in piece coordinates, so "the first two positions" are the first two of the *opening piece*, not of the protein. When the opening piece is a single residue the protein's second residue is never anchored — reachable only when the entry starts with K or R under trypsin, so 0 of 116 on cRAP, but it may be part of the residual you still see. If the doc is going to say "both of the first two positions", the anchors want computing in protein coordinates.

And one of the seven tests does not discriminate: `Create_KeepsAModificationOnTheSecondResidue` reportedly passes on #1256's code, because unanchored the unranking's origin queue happens to return the first free `A` to slot 1. Its siblings do discriminate, so the coverage exists — but the position-2 case, which the body says came out of the failure data, has no test that would notice a regression. A fixture whose second residue is not part of a run of identical residues would fix it. I did not re-run that one.

Approving: neither finding is wrong in the tree. What I verified: the `ModFits` line numbers and the anchor-set argument. What I did not: the cRAP measurements, the proteome figures, or the non-discriminating test claim.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Fixed in #1265. `TerminalAnchors` now takes the piece's start and the protein's length and translates protein coordinates into piece coordinates, so position 2 is held wherever it falls.

You are right that it was reachable — an entry beginning with K or R under trypsin — and right that the documentation promised to cover it. It may well account for part of the residual.

trishorts:

PR #1258, opened 2026-08-31
Entrapment: one protein per database entry, not per applied variant

trishorts:
**Stacked on #1257** (which is stacked on #1256 → #1255 → #1254 → #1253 → #1250). Merge after those; the diff here is only the last commit.

## The defect

`ProteinDbLoader.LoadProteinXML` doesn't return database *entries* — it applies each entry's sequence variations and returns one `Protein` per applied combination, all sharing a `ConsensusVariant`. The reviewed human XML gives **52,337 proteins from 20,416 entries**.

Generating an entrapment partner for each of those proteins gives 52,337 entrapment proteins, and because each is constructed fresh, **each is its own consensus**. `ProteinDbWriter.WriteXmlDatabase` writes one entry per distinct consensus, so the targets fold back to 20,416 and the entrapment proteins don't fold at all:

| | target entries | entrapment entries | ratio |
|---|---|---|---|
| from a FASTA | 20,416 | 20,415 | 1.00 |
| from an XML | 20,416 | **52,337** | **2.56** |

The database reaching the search engine held **2.56 entrapment entries per target while the FDP estimators reading it were told `r = 1`**. Combined FDP scales as `(1 + 1/r)`, so that inflated the XML arm's estimate by up to 1.4×. Found by running the reference FDP tool against a real search, not by any test.

The FASTA path was never affected — a FASTA has no variants, so every protein is already its own consensus.

## The fix

The unit of entrapment is the **entry**. A sequence variant is an annotation on an entry, not an entry of its own, and a partner for every point variant is not something anyone asked for — it's only what a loop over the loaded list produces.

- `GenerateEntrapment(targets, digestionParams, forbidden, foldCount, seed, identifier)` — one partner per entry per fold. This is the list-level entry point the API was missing: `Create` was per-protein only, so the decision of *which* proteins get a partner had no owner and fell to each caller.
- `DatabaseEntries(targets)` is public beside it, because a caller building the QC report needs each partner's `EntrapmentAssembly` and only the per-protein overload returns that. It has to walk the same entries without re-deriving what an entry is.

Both key by **reference**, not value: `Protein` overrides `Equals`, and the question is which entry a protein came from, not whether two proteins look alike. `DecoyProteinGenerator` keys its consensus map the same way — and its comment describes this exact hazard, *"a decoy left as its own consensus makes every variant decoy look canonical."* The hazard was already understood for decoys and simply had no counterpart for entrapment.

## Tests

Five. The one that matters counts entries in the **written file** — every existing assertion was true of the in-memory list and still permitted a 2.56 ratio, because nothing counted what survived the write.

Verified by reintroducing the defect: three of the five fail, and the FASTA-shaped test keeps passing, which is the point. Full suite **6,159 passed / 0 failed**.

Measured on the real database afterwards: 52,337 loaded → 20,416 entries → 40,832 written, **ratio 1.000**, modification loss 3 of 55,336 (0.01%, all on excised pieces). The file also drops from 1,582 MB to 840 MB.

---

### Review follow-up

**Correction: no shipped code path produced the 2.56 ratio.** Nothing in mzLib calls either new
method outside the tests, and MetaMorpheus has no entrapment-generation call at all — the ratio was a
property of an external experiment script. As originally written the description reads as though the
library produced it, and someone chasing which release is affected would not find one.

**Which peptides belong in `forbiddenSequences`** is the immediate next question and the doc pointed
neither way. A set built from the expanded protein list carries variant peptides no entry-based
target set produces; one built from entries omits variant peptides a search *will* report, since
consumers apply variations at load. It is the caller's choice, and #1265 adds an observing overload
so a report caller no longer maintains a second loop that could disagree about it.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using MzLibUtil;
 using Omics.Digestion;
+using Omics.BioPolymer;
 using Omics.Modifications;
 using Proteomics;
 using System.Collections.Generic;
@@ -64,7 +65,19 @@ public static class EntrapmentProteinGenerator
         return new Protein(withNewSequence,
             accession: EntrapmentAccession.Format(target.Accession, fold, entrapmentIdentifier),
             isEntrapment: true,
-            oneBasedModifications: movedMods);
+            oneBasedModifications: movedMods,
+            // Every other positional annotation describes the TARGET's sequence and means nothing
+            // once the residues have moved. Carrying them over is not merely untidy: excision
+            // shortens the protein, so a coordinate can point past its end, and a consumer that
+            // indexes with it -- MetaMorpheus applies sequence variations while loading a database
+            // -- throws. Measured on the human proteome: 131 entrapment entries carried a feature
+            // position beyond their own length, and every search against that database failed.
+            // Modifications are the one annotation that survives, because they alone are remapped.
+            sequenceVariations: new List<SequenceVariation>(),
+            appliedSequenceVariations: new List<SequenceVariation>(),
+            proteolysisProducts: new List<TruncationProduct>(),
+            disulfideBonds: new List<DisulfideBond>(),
+            spliceSites: new List<SpliceSite>());
     }
 
     /// <summary>
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
@@ -1,0 +1,102 @@
+#nullable enable
+using MzLibUtil;
+using Omics.Digestion;
+using Omics.Modifications;
+using Proteomics;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>
+/// Builds the entrapment partner of a target protein: an isomeric sequence, its modifications
+/// carried onto the residues they were on, and an accession that says what it is and where it came
+/// from.
+/// </summary>
+public static class EntrapmentProteinGenerator
+{
+    /// <summary>
+    /// The entrapment partner of <paramref name="target"/> for one fold.
+    /// </summary>
+    /// <param name="target">The target protein.</param>
+    /// <param name="digestionParams">Supplies the cleavage sites, and the minimum peptide length
+    /// below which a piece is not identifiable on its own.</param>
+    /// <param name="forbiddenSequences">Sequences no partner peptide may equal, normally every
+    /// target peptide in the database.</param>
+    /// <param name="fold">Zero-based fold, in <c>[0, foldCount)</c>.</param>
+    /// <param name="foldCount">Partners per target -- the <c>r</c> of an r-fold database.</param>
+    /// <param name="seed">Changes every choice reproducibly.</param>
+    /// <remarks>
+    /// Built in two steps, both through existing constructors and neither through reflection:
+    /// <see cref="Protein.CloneWithNewSequenceAndMods"/> for the sequence and modifications, then
+    /// Protein's copy constructor for the accession and the entrapment flag.
+    /// </remarks>
+    public static Protein Create(Protein target, IDigestionParams digestionParams,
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,215 @@
...
+    /// database that reaches a search engine then holds <b>2.56 entrapment entries per target</b>,
+    /// while every FDP estimator reading it is told <c>r = 1</c>. Measured, not hypothetical: it is
+    /// what the human XML arm of this project's own experiment was searched against.</para>
+    /// <para>So the unit of entrapment is the entry. A sequence variant is an annotation on an
+    /// entry, not an entry of its own, and an entrapment partner for every point variant of a
+    /// protein is not something anyone asked for -- it is only what a naive loop produces. The
+    /// FASTA path is unaffected either way, because a FASTA has no variants and every protein is
+    /// already its own consensus.</para>
+    /// </remarks>
+    public static List<Protein> GenerateEntrapment(IEnumerable<Protein> targets,
```
acesnik, line 48: This entry point discards the `EntrapmentAssembly`, so a report caller has to write its own `DatabaseEntries` + per-protein `Create` loop — two loops that must agree about folds and entries. #1260's body reports exactly that divergence ("it still compared the partner count against the *loaded* protein count rather than the *entry* count"). An overload taking an `Action<Protein, int, EntrapmentAssembly>`, or building the report here, would leave one loop.

acesnik (approved):
🤖 The unit of an entrapment database is the database entry, and until now nothing in the API owned that decision — `Create` was per-protein, so a caller looping the list a loader returns gets a partner per applied variant combination, and the ratio becomes whatever the variant expansion happened to be. `GenerateEntrapment` puts that decision in the library, and `DatabaseEntries` sits beside it so the report path can walk the same entries rather than re-deriving what an entry is.

Entry parity was checked against three variant-carrying fixtures already in the tree rather than only the new synthetic one: `DecoyVariants.xml`, `HomozygousHLA.xml` and `MultipleAlternateAlleles.xml` load 2, 1 and 2 proteins, `DatabaseEntries` returns 1 for each, and the written database holds 2 entries of which 1 is entrapment — ratio 1.000 in all three. `EntrapmentEntriesEqualTargetEntriesInTheWrittenDatabase` counting `<entry ` in the file is the only assertion that could have caught the original defect, which is worth saying because five in-memory assertions did not.

The reference-versus-value keying looks like a comparer mismatch and I do not think it is one, which is worth recording because it is the first thing a reader will query. `WriteXmlDatabase` folds targets by value (`ConsensusVariant` … `Distinct()`, accession plus base sequence) while `DatabaseEntries` folds by reference. Two value-equal entries therefore yield two entrapment proteins where the writer keeps one target entry — but those two entrapment proteins are themselves value-equal, because the sequence is a pure function of (sequence, seed, fold) and the accession derives from the target's, so the writer folds them too and the ratio stays 1. No shape was found where the comparers disagree in a way that reaches the ratio.

One correction for the body rather than the code. Nothing in mzLib calls either new method outside the tests, and MetaMorpheus has no entrapment-generation call at all, so the 2.56 ratio was a property of the external experiment script and not of a shipped code path. As written it reads as though the library produced that, and someone chasing which release is affected will not find one.

The question the diff does not answer: should a caller's `forbiddenSequences` now be built from the loaded protein list or from the entries? `GenerateEntrapment` digests entries, so a set built from the expanded list carries variant peptides no entry-based target set produces, while one built from entries omits variant peptides a search against the written database *will* report, since consumers apply variations at load. That is the caller's choice, but it is the immediate next question and the doc points neither way.

What I did not verify: the proteome-scale figures, and the fixture parity numbers above are a reviewer's measurement rather than mine.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Added in #1265: an overload taking `Action<Protein, int, EntrapmentAssembly>`, with the parameterless one delegating to it.

Worth recording that your prediction came true before I read your comment. The two loops did diverge — the report harness compared its partner count against the *loaded* protein count rather than the entry count, and because the assertion sits after the file write it kept producing databases while failing, for a day, unnoticed.

The callback fires **before** the emptiness check, so a report still sees an entry that produced nothing. That is what a reader of the achieved ratio needs, and dropping it silently is how a poly-Q homopolymer with no possible rearrangement went unnoticed in the entry counts.

trishorts:

PR #1259, opened 2026-08-31
Entrapment: collision-check missed-cleavage runs, and count the ones that can't be fixed

trishorts:
**Stacked on #1258.** Merge after it; the diff here is the last commit only.

## The defect

An entrapment peptide is supposed to be verifiably false. A **missed-cleavage** peptide is a run of adjacent base pieces, and each piece was permuted knowing only itself — so `E1+E2` could equal a genuine target peptide even when neither `E1` nor `E2` did. That puts a **real peptide into the entrapment database**, and a search matching it counts a true peptide as an entrapment discovery.

Measured on the reviewed human proteome: **2,094 peptides (0.076% of the target set) appear in both the target and entrapment sets, 2,090 of them with at least one missed cleavage.**

## The fix

Every run ends at some piece, so testing the runs that *end* at the piece being placed covers all of them — and left-to-right assembly has already produced everything such a run needs. The assembler hands the generator a predicate, which it applies alongside the existing forbidden-sequence test while probing. Runs are built once per piece, not once per candidate. Generation over the human proteome: 20.6 s → 21.7 s.

**The cost:** a piece's permutation now depends on its neighbours, so the construction is a pure function of `(protein, seed)` rather than `(piece, seed)`. Determinism, reproducibility, parallel safety across proteins, and the composition-plus-pinning pairing key are all unaffected — the piece is still a permutation of its target piece.

## What it doesn't fix, and why that's structural

**1,794 collisions survive, and 97.9% of them end in a piece with exactly one possible arrangement** — a short low-complexity piece like `AAR`, where holding the cleavage residue leaves nothing to reorder. There is no candidate to move to. Repairing them would mean backtracking into an already-placed piece or excising a good one.

This project's rule is to count collisions rather than silently repair them, so `EntrapmentAssembly` now carries **`UnrepairableRunCollisions`**. Anyone reading an entrapment count needs this beside it. Four fifths of the survivors are exactly seven residues — the minimum searchable length.

| | before | after |
|---|---|---|
| shared peptides | 2,094 | **1,794** |
| …with 1 missed cleavage | 551 | **335** |
| …with 2 | 1,539 | 1,455 |

## Tests

Five. Each derives its fixture by building with nothing forbidden and then forbidding *exactly the run that came out*, so the collision case is guaranteed to be reached rather than hoped for — the first attempt used a three-residue opening piece and never permuted it at all, because #1257's terminal anchoring had already frozen it, and the fixture guard caught that.

Verified by disabling the check: two tests fail, while the determinism and no-missed-cleavage tests correctly keep passing. Full suite **6,164 passed / 0 failed**.

---

### Review follow-up

**The excision delta, which the description owed and did not give: it is zero.** Rebuilt on the
reviewed human proteome with every review fix in, `unpairableAllPermutationsTaken` and
`unpairableRunCollisionsExhausted` are both **0** — the run check causes no additional excisions on
this database. Worth stating precisely because residues *leaving* the entrapment database would have
been the more consequential of the two effects, and a collision delta alone could not rule it out.

The count itself was wrong in both directions and is fixed in #1265: it returned on the first match
(counting pieces while documenting peptides) and applied no length bound (counting concatenations no
search could report). A run collision that exhausts a stretch now reports its own failure cause
rather than borrowing `AllPermutationsTaken`, whose remedy is different.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
 using MzLibUtil;
 using Omics.Digestion;
 using System.Collections.Generic;
@@ -60,13 +61,14 @@ public sealed class EntrapmentAssembly
 {
     internal EntrapmentAssembly(string targetSequence, string entrapmentSequence,
         int[] targetToEntrapmentPosition, IReadOnlyList<EntrapmentPiece> pieces,
-        int missedCleavagePeptidesSpanningAnExcision)
+        int missedCleavagePeptidesSpanningAnExcision, int unrepairableRunCollisions)
     {
         TargetSequence = targetSequence;
         EntrapmentSequence = entrapmentSequence;
         TargetToEntrapmentPosition = targetToEntrapmentPosition;
         Pieces = pieces;
         MissedCleavagePeptidesSpanningAnExcision = missedCleavagePeptidesSpanningAnExcision;
+        UnrepairableRunCollisions = unrepairableRunCollisions;
     }
 
     public string TargetSequence { get; }
@@ -92,6 +94,24 @@ public sealed class EntrapmentAssembly
     /// invariant holds everywhere.
     /// </remarks>
     public int MissedCleavagePeptidesSpanningAnExcision { get; }
+
+    /// <summary>
+    /// Missed-cleavage peptides of this entrapment sequence that equal a real target peptide and
+    /// could not be avoided.
+    /// </summary>
+    /// <remarks>
+    /// <para>Each piece is permuted away from any collision it can be permuted away from, runs
+    /// included. What cannot be repaired is a run whose final piece has exactly one arrangement --
+    /// a short low-complexity piece such as <c>AAR</c>, where holding the cleavage residue leaves
+    /// nothing to reorder. There is no candidate to move to, so the collision stands.</para>
+    /// <para>Repairing it would mean backtracking into an already-placed piece or excising a
+    /// perfectly good one, and this project counts collisions rather than silently repairing them.
+    /// Measured on the reviewed human proteome: 1,794 such peptides, 0.065% of the target set,
+    /// 97.9% of them ending in a piece with no alternative and 80% exactly at the minimum
+    /// searchable length of seven residues. A search that matches one counts a real peptide as an
+    /// entrapment discovery, so anyone reading an entrapment count needs this number beside it.</para>
+    /// </remarks>
+    public int UnrepairableRunCollisions { get; }
 
     public int ExcisedCount => Pieces.Count(p => p.Outcome == PieceOutcome.Excised);
 
@@ -141,6 +161,10 @@ public static class EntrapmentAssembler
 
         var pieces = new List<EntrapmentPiece>(sites.Count - 1);
         var entrapment = new StringBuilder(targetSequence.Length);
+        // The entrapment pieces already placed, in the order they appear in the entrapment
+        // sequence, so a candidate can be tested against the runs it would complete.
+        var placed = new List<string>(sites.Count - 1);
+        int unrepairable = 0;
         int[] map = Enumerable.Repeat(-1, targetSequence.Length).ToArray();
         var retainedTargetIndices = new List<int>(sites.Count - 1);
 
@@ -150,12 +174,19 @@ public static class EntrapmentAssembler
             int length = sites[index + 1] - start;
             string piece = targetSequence.Substring(start, length);
 
+            Func<string, bool>? completesAForbiddenRun =
+                RejectRunCollisions(placed, digestionParams.MaxMissedCleavages, forbiddenSequences);
+
             EntrapmentPeptide partner = EntrapmentPeptideGenerator.Create(piece, motifs, forbiddenSequences,
-                fold, foldCount, seed, TerminalAnchors(index, sites.Count - 1, length));
+                fold, foldCount, seed, TerminalAnchors(index, sites.Count - 1, length),
+                completesAForbiddenRun);
 
             if (partner.Succeeded)
             {
                 AppendPiece(entrapment, map, start, partner.EntrapmentSequence!, partner.SwappedPositions!);
+                placed.Add(partner.EntrapmentSequence!);
+                // A permuted piece was chosen with the run test applied, so this cannot fire -- it is
+                // asserted rather than assumed only because the count below must mean one thing.
                 pieces.Add(new EntrapmentPiece(index, piece, partner.EntrapmentSequence,
                     PieceOutcome.Permuted, EntrapmentFailure.None));
                 retainedTargetIndices.Add(index);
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,355 @@
...
+    /// <para>Every run ends at some piece, so testing the runs that <i>end</i> at the piece being
+    /// placed covers all of them, and left-to-right assembly has already placed everything such a
+    /// run needs. The runs are built once per piece rather than once per candidate, because the
+    /// probe loop may try many candidates and the preceding pieces do not change between them.</para>
+    /// <para>The cost is that a piece's permutation now depends on its neighbours, so the
+    /// construction is a pure function of <c>(protein, seed)</c> rather than of <c>(piece, seed)</c>.
+    /// Determinism, parallel safety across proteins, and the composition-plus-pinning pairing key
+    /// are all unaffected -- the piece is still a permutation of its target piece.</para>
+    /// </remarks>
+    private static Func<string, bool>? RejectRunCollisions(List<string> placed,
```
acesnik, line 289: No length bound on the run, so a concatenation shorter than `MinLength` — or longer than `MaxLength` — still rejects candidates and still increments the count. On `AAKAAR` with a minimum of 7, the two-piece run is six residues and `UnrepairableRunCollisions` comes back 1, while `EntrapmentPairing.SearchablePeptideCount` for the same protein is 0. #1260 makes the length-bounded definition the report's own, so the assembler and the pairing will disagree about what a peptide is the moment it lands.

Testing `run.Length + candidate.Length` against `MinLength`/`MaxLength` before the lookup would make the two agree — and is my best guess at why this PR says 1,794 and #1260 says 1,984 for the same database when the increment between them does not touch this file.

acesnik (changes requested):
🤖 This closes the check the generator structurally could not do for itself. Each base piece is permuted knowing only itself, so a run of adjacent pieces can land on a real target peptide even when no single piece does, and the assembler is the only place that knows the neighbours. Handing the generator a predicate over the pieces already placed is the right shape.

The formulation is right in the two places it could easily have been wrong. Testing the runs that *end* at the piece being placed covers every run, because every run ends somewhere and left-to-right assembly has already produced what such a run needs. And building the predicate from `placed` rather than from target indices is the subtle part done right: `placed` skips excised pieces, so it tracks adjacency in the *entrapment* sequence — which is what actually gets digested. `Math.Min(maxMissedCleavages, placed.Count)` preceding pieces plus the candidate covers runs of up to `MaxMissedCleavages + 1`, which is what a search generates, and returning null at `maxMissedCleavages < 1` is correct rather than merely convenient.

I have no objection to the generation half. The counting half is where I would ask for a pass, and the two problems pull the number in opposite directions so they do not cancel.

## The count and the report disagree about what a peptide is

`UnrepairableRunCollisions` is incremented at most once per piece, because the rejection returns on its first match — while the property documents a count of *peptides*. A reviewer forced two colliding peptides ending at the same piece and got a count of 1.

In the other direction, `RejectRunCollisions` puts no length bound on the run, so a concatenation no search can report still rejects candidates and still counts. On `AAKAAR` under a minimum length of 7, the two-piece run is six residues, `UnrepairableRunCollisions` is 1, and `EntrapmentPairing.SearchablePeptideCount` for the same protein and parameters is 0.

**That second one is also why I would settle the number before either PR merges.** This PR's table and its doc comment both say 1,794. #1260 says 1,984 for the same database. I checked the `#1259 → #1260` increment: it touches only `TestEntrapmentReport.cs`, `EntrapmentPairing.cs` and `EntrapmentReport.cs` — `EntrapmentAssembler.cs` is not in it, so nothing between the two PRs changes how this is computed, and the two figures cannot both be right. #1260 makes the length-bounded definition the report's own, which is a plausible cause, but I could not confirm it at scale.

## A new failure cause absorbed into an existing one

A run collision can exhaust a piece's whole stretch; the piece is then excised and reported as `AllPermutationsTaken`, whose doc says arrangements are "already spoken for -- it is a target peptide, or it has already been issued" and whose stated remedy is a different target database. A run collision's remedy is a different seed or a different neighbour. The enum's own comment is that failures "are kept apart because the caller's remedy differs for each, and merging them into a single 'failed' hides that" — which is the argument for a distinct member here.

That is also the output change I would want in the body. The PR gives a collision delta and no excision delta, and residues leaving the entrapment database is the more consequential of the two: a peptide that vanishes is not merely a peptide that is now safely false.

On determinism, which the run check does put at risk, I found nothing wrong — no dictionary-order dependence, no parallelism, no unseeded randomness; the digestion-site set is sorted before use. Making the construction a function of (protein, seed) rather than (piece, seed) is correctly identified as the whole cost.

What I verified: the increment's file list, and that the assembler is untouched between #1259 and #1260. What I did not: the fixture measurements above, or any proteome-scale figure.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Both halves fixed in #1265, and you were right that they did not cancel.

The finder now returns **every** offending run rather than the first, so several runs ending at one piece are counted as the several peptides they are. And it applies `MinLength`/`MaxLength` to `run.Length + candidate.Length` before the lookup, so a concatenation no search could report neither rejects a candidate nor counts — the assembler and the pairing index now agree about what a peptide is.

Your `AAKAAR` case is exactly the shape: six residues under a minimum of seven, counted as 1 while `SearchablePeptideCount` said 0.

Also added `EntrapmentFailure.RunCollisionsExhaustedTheSpace`, per your point that the remedies differ — a different seed or different neighbours, not a different target database.

On 1,794 against 1,984: your diagnosis was right that they could not both be right, and the length bound is the cause. The numbers were also measuring different things — 1,974/1,984 are the generator's own counts, while 1,794 came from an independent digest of the written file counting distinct *sequences*, which is smaller because one sequence collides in several proteins. Both restated with their denominators.

trishorts:

acesnik (approved):
🤖 Clearing my changes-requested — both halves fixed, and measured at #1265 on the case from my comment.

Every offending run is returned rather than the first: `"AAAAK" + "GGGGK" + "AAAAR"` with both the 10-mer and the 15-mer forbidden counts 2 at #1265 against 1 at #1264. And the length bound applies before the lookup: `AAKAARLLLLLLLLK` with the six-residue `AAKAAR` forbidden under a minimum of 7 counts 0 where it counted 1 — the exact disagreement with `SearchablePeptideCount` I raised. The assembler and the pairing index now agree about what a peptide is.

`RunCollisionsExhaustedTheSpace` as its own cause is the right call for the reason you give: a peptide defeated by the runs it would complete wants a different seed or different neighbours, not a different target database, so collapsing it into `AllPermutationsTaken` would point the remedy the wrong way.

On 1,794 against 1,984 — your explanation holds and is the useful half: they were counting different things, generator-side counts against distinct sequences from an independent digest, so neither was simply wrong. `EntrapmentAssembler.cs:110` still carries 1,794 as a bare measurement in the remarks though, which this PR's own change invalidates; I flagged that on #1265.

trishorts:

PR #1260, opened 2026-08-31
Entrapment: give the QC report units that match, and the counts a consumer must exclude by

trishorts:
**Stacked on #1259.** Merge after it; the diff here is the last commit only.

Raised by the first named consumer of this database, who is computing the Wen *et al.* FDP estimators from search output and needs to read three things out of the report.

## The ambiguity figure was a rate of nothing

`EntrapmentPairing` indexed **every** run of adjacent base pieces however short. Short peptides collide constantly — `AK` shares its key with every other `AK` — so the ambiguous set filled with peptides no search will ever report. Meanwhile `targetPeptides` counts **base pieces**. Dividing one by the other divides two different populations, and turned 0.026% into an apparent **1.19%**.

The index now respects the digestion length bounds. That's safe as well as honest: a key begins with the peptide's length, so peptides of different lengths never collide, and dropping the unsearchable ones **cannot change how a searchable peptide resolves**.

| reviewed human proteome | before | after |
|---|---|---|
| ambiguous peptides | 6,889 | **736** |

736 matches an independent digest of the written database **exactly**.

## The report gains the denominator that number belongs to

`searchSpacePeptides` counts what a search actually reports — runs of up to `MaxMissedCleavages + 1` base pieces, within the length bounds — beside the base-piece counts the generator permutes:

| | |
|---|---|
| `targetPeptides` (base pieces, the permutation unit) | 578,876 |
| `searchSpacePeptides` (what a search sees) | **2,815,370** |

Nearly fivefold apart, so a rate computed against the wrong one is meaningless. It's also the population an FDP estimator's `r` is over, which is the first thing the consumer asked for.

## And the collisions a consumer has to exclude by

`unrepairableRunCollisions` reaches the report (1,984 on this database). Those are missed-cleavage peptides equal to a real target peptide that could not be permuted away, because the run's final piece has only one arrangement (#1259). A consumer needing an exact entrapment count must exclude them, and can't exclude what it can't see.

## Tests

Four new. The column-position test already in the suite caught the inserted column — which is what it's for.

Fixing the build harness turned up a stale assertion of its own: it still compared the partner count against the *loaded* protein count rather than the *entry* count, so it had been failing since #1258 and I hadn't noticed, because I was reading its output lines rather than its result.

Full suite **6,168 passed / 0 failed**.

---

### Review follow-up

**Correction to a figure in this description.** It said the mismatch "turned 0.026% into an apparent
1.19%"; the code comments say 0.24% and the code figure is the coherent one — the denominator alone
accounts for 0.245% against 1.19%, whereas 0.026% folds in the numerator fix as well. Two separate
effects, and the description conflated them.

Also fixed in #1265: the new column no longer lands in the zero-site stratum row, and
`entrapmentSearchSpacePeptides` joins it so a consumer can compute a **peptide-level** `r` rather
than a base-piece one. On the reviewed human proteome that is 2,815,187 against 2,815,370.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPairing.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPairing.cs
@@ -1,0 +1,127 @@
+#nullable enable
+using MzLibUtil;
+using Omics.Digestion;
+using Proteomics;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>
+/// Finds the target peptide an entrapment peptide was built from, using nothing but the target
+/// protein and the peptide itself.
+/// </summary>
+/// <remarks>
+/// <para>A rearrangement preserves the residue multiset and leaves every cleavage-site residue in
+/// place, so those two things together are a key: the partner of an entrapment peptide is the
+/// target peptide of the same protein with the same free-residue composition and the same pinned
+/// pattern. That is guaranteed by the construction rather than recorded anywhere, which is why no
+/// index file has to be written, shipped, or kept in step with the database. A search reports a
+/// protein accession and a peptide sequence, and that is enough.</para>
+/// <para>The key can collide: a protein may contain two different peptides sharing both, and
+/// P57071 really does contain <c>LIHTGVK</c> and <c>LIHTVGK</c>. Those are reported in
+/// <see cref="AmbiguousPeptides"/> and refuse to resolve, because a guess between them would be
+/// wrong half the time and silently so.</para>
+/// </remarks>
+public sealed class EntrapmentPairing
+{
+    private readonly Dictionary<string, string> _byKey;
+    private readonly HashSet<string> _ambiguous;
+    private readonly List<DigestionMotif> _motifs;
+
+    /// <summary>Indexes a target protein's peptides, missed cleavages included, by composition and pinned pattern.</summary>
+    public EntrapmentPairing(Protein target, IDigestionParams digestionParams)
+    {
+        if (target is null)
+        {
+            throw new MzLibException("Cannot build a pairing without a target protein.");
+        }
+
+        _byKey = new Dictionary<string, string>();
+        _ambiguous = new HashSet<string>();
+        var collided = new HashSet<string>();
+
+        _motifs = digestionParams.DigestionAgent.DigestionMotifs;
+        List<int> sites = digestionParams.DigestionAgent.GetDigestionSiteIndices(target.BaseSequence);
+        sites.Sort();
+
+        // Index runs of adjacent base pieces, not just single ones: a search reports missed-cleavage
+        // peptides too, and those are pairable for the same reason -- base-piece compositions add, so
+        // a run is isomeric with the run its partner was built from.
+        int maxPieces = digestionParams.MaxMissedCleavages + 1;
+
+        for (int first = 0; first < sites.Count - 1; first++)
+        {
+            for (int pieces = 1; pieces <= maxPieces && first + pieces < sites.Count; pieces++)
+            {
+                int start = sites[first];
+                string peptide = target.BaseSequence.Substring(start, sites[first + pieces] - start);
+                string key = KeyOf(peptide);
+
+                if (_byKey.TryGetValue(key, out string? existing))
+                {
+                    if (existing != peptide)
+                    {
+                        collided.Add(key);
+                        _ambiguous.Add(existing);
+                        _ambiguous.Add(peptide);
+                    }
+                    continue;
+                }
+
+                _byKey[key] = peptide;
+            }
+        }
+
+        foreach (string key in collided)
+        {
+            _byKey.Remove(key);
+        }
+    }
+
+    /// <summary>
+    /// Target peptides sharing a composition-and-pinning key with another peptide of the same
+    /// protein, and therefore not resolvable. Report these rather than pairing them.
+    /// </summary>
+    public IReadOnlyCollection<string> AmbiguousPeptides => _ambiguous;
+
+    /// <summary>The target peptide <paramref name="entrapmentPeptide"/> was built from.</summary>
+    /// <returns>False when the peptide belongs to no target peptide of this protein, or when its
+    /// key is ambiguous.</returns>
+    public bool TryResolve(string entrapmentPeptide, out string targetPeptide)
+    {
+        targetPeptide = string.Empty;
+        if (string.IsNullOrEmpty(entrapmentPeptide))
+        {
+            return false;
+        }
+
+        return _byKey.TryGetValue(KeyOf(entrapmentPeptide), out targetPeptide!);
+    }
+
+    /// <summary>
+    /// Length, the residues held in place and where, and the sorted free residues -- exactly what a
+    /// rearrangement preserves, and nothing it is free to change.
+    /// </summary>
+    private string KeyOf(string peptide)
+    {
+        HashSet<int> pinned = DecoySequenceValidator.CleavageSitePositions(peptide, _motifs);
+
+        var pinnedPart = new List<string>();
+        var free = new List<char>();
+        for (int i = 0; i < peptide.Length; i++)
+        {
+            if (pinned.Contains(i))
+            {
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,360 @@
...
+            _ambiguousByAccession[target.Accession] = ambiguous;
+
+            foreach (string peptide in ambiguous)
+            {
+                StratumFor(peptide).Ambiguous++;
+            }
+
+            // Not attributable to a stratum: the count is over the protein's whole search space,
+            // and stratifying it would mean re-enumerating every run to classify each one.
+            StratumFor(string.Empty).SearchSpacePeptides += pairing.SearchablePeptideCount;
```
acesnik, line 275: `StratumFor(string.Empty)` is the `siteCount == 0` stratum — the implementation short-circuits to 0 and then creates or reuses the real one, which also holds real zero-site peptides. So this whole-protein figure is emitted beside that stratum's own `targetPeptides`, and every other row reads 0 for a column that is not zero. Lines 313 and 315 do the same.

The comment above is right that the count is not attributable to a stratum; the problem is that stratum 0 *is* a stratum. A `#` provenance line or a distinct row label keeps the number honest — and it matters more here than for `mcSpanningAnExcision`, because this column is sold as a denominator and the documented use case is the stratified one.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPairing.cs
```diff
@@ -0,0 +1,151 @@
...
+        int minLength = digestionParams.MinLength;
+        int maxLength = digestionParams.MaxLength;
+
+        for (int first = 0; first < sites.Count - 1; first++)
+        {
+            for (int pieces = 1; pieces <= maxPieces && first + pieces < sites.Count; pieces++)
+            {
+                int start = sites[first];
+                int length = sites[first + pieces] - start;
+                if (length < minLength || length > maxLength)
```
acesnik, line 68: Target-side only. There is no entrapment-side search-space count in the report, and `EntrapmentPeptides` counts base pieces, so a consumer still cannot get peptide-level `r` — excision makes the two search spaces differ by more than the fold factor. A second column from this same construction over the entrapment protein would close it, and is the difference between a report a consumer can use and one they have to correct first.

acesnik (changes requested):
🤖 The ambiguity figure was a rate over two different populations, and this fixes both halves: the index stops holding runs no search can report, and the report gains the denominator the count actually belongs to. The safety argument for the filter is right and worth having in the code — a key begins with the peptide's length, so dropping short peptides cannot change how a long one resolves.

`SearchablePeptideCount` as `_byKey.Count + _ambiguous.Count` does count distinct sequences, which is not obvious and worth confirming rather than assuming: a repeated sequence hits the `existing != peptide` guard without counting twice, a collided key leaves `_byKey` while its peptides live in `_ambiguous`, and a peptide cannot be in both because `KeyOf` is a function of the sequence alone. The `MaxLength` bound is also safe by default — both digestion-parameter types default `maxPeptideLength` to `int.MaxValue`, so the new `length > maxLength` test cannot silently empty the index the way a zero-means-unlimited convention would have.

**How I read the stack.** All ten PRs are cumulative and based on `master`, so I reviewed `git diff refs/pr/1259...refs/pr/1260` rather than what GitHub renders — 159 lines across three files, not +3437.

## Where the new columns are written

`StratumFor(string.Empty)` resolves to the zero-site stratum. I read the implementation: it short-circuits `peptide.Length == 0 ? 0 : _siteCounter(peptide)` and then creates or reuses the *real* stratum keyed 0. So the whole-protein `searchSpacePeptides` is published in a row beside that stratum's own per-peptide counts, and every other row reads 0 for a column that is not zero.

The convention arrived with the report in #1255, so `mcSpanningAnExcision` has the same shape and this is not the PR that introduced it. But this is the PR that adds a column sold as **a denominator**, and the documented use case is the stratified O-glycosylation one — where dividing within a row is exactly what a reader will do. A `#` provenance line, or a row label distinct from a real stratum, keeps the number without attaching it to a population it does not describe.

## The exclusion is not on the same footing as the population

`unrepairableRunCollisions` is counted over runs with no length bound, while `SearchSpacePeptides` is length-bounded; and it increments at most once per piece even when several runs ending at that piece collide. A consumer told to exclude these from a length-bounded population is over-excluding some and under-excluding others, and cannot tell which from the report.

**This is also the discrepancy I would settle before either PR merges.** #1259's table and doc comment both say 1,794; this body says 1,984 for the same database. I checked the increment between them: it touches only `TestEntrapmentReport.cs`, `EntrapmentPairing.cs` and `EntrapmentReport.cs` — `EntrapmentAssembler.cs` is not in it, so nothing here changes how that count is computed and the two numbers cannot both be right. The unbounded-run definition is the plausible cause but I could not confirm it at scale.

## On sufficiency, which is what the PR is for

`searchSpacePeptides` is the *target*-side search space and there is no entrapment-side counterpart. `entrapmentPeptides` counts base pieces, so `achievedFoldRatio` is a base-piece ratio rather than the peptide-level `r` an FDP estimator is over, and excision means the two search spaces differ by more than a fold factor — `mcSpanningAnExcision` is a correction the consumer would have to derive and apply itself. If the named consumer needs `r` over reported peptides, an entrapment-side search-space count is the missing column, and it is cheap: the same `EntrapmentPairing` construction over the entrapment protein.

Free to fix while the body is open: it says the mistake "turned 0.026% into an apparent 1.19%" while the code comments say 0.24% twice. The code figure is the coherent one — the denominator alone accounts for 0.245% against 1.19%, whereas 0.026% folds in the numerator fix as well.

What I verified: the stratum resolution by reading, and the increment's file list. What I did not: any proteome-scale figure, or which of 1,794 and 1,984 is correct.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Fixed in #1265 — same `_wholeProtein` accumulator as the sibling comment on #1256, so this column never lands in a stratum row.

Your framing is what made it worth doing properly rather than labelling: the convention arrived in #1255 and `mcSpanningAnExcision` shares it, but this is the column sold as a denominator and the documented use is the stratified one, where dividing within a row is exactly what a reader does.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPairing.cs
trishorts: Added in #1265. `EntrapmentPairing.CountSearchablePeptides(sequence, digestionParams)` measures the entrapment side on the same footing, and the report carries `entrapmentSearchSpacePeptides` beside `searchSpacePeptides`.

Your reason for it is the one that decided it: excision makes the two search spaces differ by more than the fold factor, so a consumer could not have derived this from the target side and `mcSpanningAnExcision`. It is the difference between a report a consumer can use and one they have to correct first.

On the body's 0.026% versus the code's 0.24% — you are right that the code figure is the coherent one for the denominator alone. Corrected.

trishorts:

acesnik (approved):
🤖 Clearing my changes-requested — both asks addressed, both verified at #1265.

The denominator no longer lands in a stratum row. The `_wholeProtein` accumulator reaches `Total` and nothing else, and every stratum row now reads 0 for those columns. That mattered more here than for `mcSpanningAnExcision` for the reason you restated: this is the column sold as a denominator, and the documented use is the stratified one, where dividing within a row is exactly what a reader does.

The entrapment-side count is there: `entrapmentSearchSpacePeptides` sits beside `searchSpacePeptides`, built from the same construction over the entrapment protein. Measured across `foldCount` 1 → 2, `searchSpacePeptides` holds at 18 while `entrapmentSearchSpacePeptides` goes 14 → 28, so the pair really is a peptide-level *r* rather than a restatement of the base-piece ratio — which was the point, since excision makes the two search spaces differ by more than the fold factor.

Good catch correcting the body's 0.026% to the code's 0.24%.

trishorts:

PR #1261, opened 2026-08-31
Entrapment: name the peptides a consumer must exclude, rather than counting them

trishorts:
**Stacked on #1260.** Merge after it; the diff here is the last commit only.

A count tells a consumer how much to distrust an entrapment total. Only the **sequences** let it subtract them — "exclude these" isn't actionable in any other form. Requested by the first named consumer of this database, who is computing the Wen *et al.* estimators from search output.

## Both exclusion sets are now lists

`UnrepairableRunCollisions` was an `int`. The run-collision test already knew which run it had rejected, so it now returns that run rather than a bare `true`, and the assembler records the sequence. The count property stays, computed from the list, so nothing that read it has to change.

`EntrapmentReport` gains:

- `AmbiguousPeptidesByAccession`
- `UnrepairableRunCollisionsByAccession`
- `ExclusionsToTabSeparated()` — `accession, peptide, reason`

A sidecar rather than more columns on the stratified table, because these are per-peptide facts and that table is per-stratum. When nothing is excluded it's a header and nothing else — a real answer rather than a missing file.

On the reviewed human proteome: **736 ambiguous, 1,984 unrepairable collisions**, matching the counts already published in #1260.

## Why ambiguity matters more than it looks

A consumer computing the **paired** estimator must exclude the *same* peptides this generator excluded, or its `r = 1` assumption fails **silently** — no error, just a wrong number. That's only possible if it can see which ones they are.

## Tests

Three. The ambiguity fixture uses `LIHTGVK` and `LIHTVGK`, which P57071 really does contain — same composition, same pinned K, so the pairing key cannot tell them apart. It asserts the pair is actually produced before asserting anything about it.

Full suite **6,171 passed / 0 failed**.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
@@ -48,7 +48,24 @@ public sealed class EntrapmentStratum
     public int SiteCount { get; }
 
     /// <summary>Distinct target peptides in this stratum, counted once however many folds ran.</summary>
+    /// <remarks>
+    /// <b>Base pieces</b> -- the unit the generator permutes -- not what a search reports. For that,
+    /// see <see cref="SearchSpacePeptides"/>. The two differ by roughly five-fold at two missed
+    /// cleavages, so a rate computed against the wrong one is meaningless.
+    /// </remarks>
     public int TargetPeptides { get; internal set; }
+
+    /// <summary>
+    /// Distinct peptides in this stratum a search could report: runs of up to
+    /// <c>MaxMissedCleavages + 1</c> base pieces, within the length bounds.
+    /// </summary>
+    /// <remarks>
+    /// This is the population an FDP estimator's <c>r</c> is over, and the denominator
+    /// <see cref="Ambiguous"/> belongs to -- <see cref="Ambiguous"/> is counted over missed-cleavage
+    /// peptides too, so dividing it by <see cref="TargetPeptides"/> divides two different
+    /// populations. On the reviewed human proteome that mistake turned 0.24% into "1.19%".
+    /// </remarks>
+    public int SearchSpacePeptides { get; internal set; }
 
     /// <summary>Entrapment peptides produced for them, across every fold.</summary>
     public int EntrapmentPeptides { get; internal set; }
@@ -73,6 +90,18 @@ public sealed class EntrapmentStratum
     /// have no isomeric counterpart. The only peptides in the database that break that invariant.
     /// </summary>
     public int MissedCleavagePeptidesSpanningAnExcision { get; internal set; }
+
+    /// <summary>
+    /// Missed-cleavage peptides of the entrapment database that equal a real target peptide and
+    /// could not be permuted away from it, because the run's final piece has only one arrangement.
+    /// </summary>
+    /// <remarks>
+    /// A search matching one of these counts a <i>true</i> peptide as an entrapment discovery,
+    /// inflating an FDP estimate directly. They are counted rather than repaired -- repair would
+    /// mean backtracking into a placed piece or excising a good one -- so a consumer that needs an
+    /// exact count must exclude them. See <see cref="EntrapmentAssembly.UnrepairableRunCollisions"/>.
+    /// </remarks>
+    public int UnrepairableRunCollisions { get; internal set; }
 
     public int Unpairable => UnpairableNoPermutationExists + UnpairableAllPermutationsTaken
                              + UnpairableSpaceTooSmallForFoldCount;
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,436 @@
...
+        StratumFor(string.Empty).UnrepairableRunCollisions += assembly.UnrepairableRunCollisions;
+        if (assembly.UnrepairableRunCollisionPeptides.Count > 0)
+        {
+            if (!_unrepairableByAccession.TryGetValue(target.Accession, out List<string>? collisions))
+            {
+                collisions = new List<string>();
+                _unrepairableByAccession[target.Accession] = collisions;
+            }
+
+            collisions.AddRange(assembly.UnrepairableRunCollisionPeptides);
```
acesnik, line 386: Once per fold, so a peptide lands here as many times as it collides. At `foldCount = 3`: list count 4, distinct 2, three identical `AARGGK` rows in the table. The ambiguous side two lines up is a `HashSet` filled once per accession, so the two halves of one table count differently. `HashSet<string>` on the line 298 declaration would settle it.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,436 @@
...
+                text.AppendLine(string.Join("\t", accession, peptide, "ambiguous"));
+            }
+        }
+
+        foreach ((string accession, IReadOnlyCollection<string> peptides) in
+                 UnrepairableRunCollisionsByAccession.OrderBy(kv => kv.Key, StringComparer.Ordinal))
+        {
+            foreach (string peptide in peptides.OrderBy(p => p, StringComparer.Ordinal))
+            {
+                text.AppendLine(string.Join("\t", accession, peptide, "unrepairableRunCollision"));
```
acesnik, line 193: This `accession` is the target's, but the peptide belongs to the *entrapment* protein — a search reports it under `Random_<accession>_f<fold>`. Someone filtering the table on (accession, peptide) matches the `ambiguous` rows and silently misses these, which under-excludes. `Add` already has `fold` in hand and drops it; emitting the entrapment accession, or a fourth column, makes the row usable.

acesnik (commented):
🤖 This turns the run-collision figure from a number you can only distrust into a list you can act on. `RejectRunCollisions` returning the offending run rather than a bare bool is the move that makes it possible — the peptide that could not be permuted away now travels with the assembly instead of being reconstructed by whoever needs it, and `ExclusionsToTabSeparated` puts it beside the ambiguous peptides in one sidecar with a `reason` column. Keeping that out of the stratified table is right for exactly the reason you give: these are per-peptide facts and the other table is per-stratum.

The two halves of that sidecar are not on the same footing, though. `_ambiguousByAccession` is a `Dictionary<string, HashSet<string>>` (`EntrapmentReport.cs:297`) filled once per accession behind a `TryGetValue` guard; `_unrepairableByAccession` on the next line is a `Dictionary<string, List<string>>` that `AddRange`s on every call, and `Add(target, fold, assembly)` is called once per fold. At `foldCount = 3`, on a protein whose kept-verbatim piece collides in each fold, `AARGGK` comes out as three identical rows with `UnrepairableRunCollisionsByAccession["P00009"].Count == 4` against 2 distinct peptides. Set subtraction is idempotent so nothing is over-excluded, but the list is sold as the actionable form of the count, and every tally you could take from it — row count, list length, distinct against total — disagrees with the 2 peptides that actually need excluding. `HashSet<string>` on line 298 makes the two halves agree.

The reason a reader cannot tell over- from under-exclusion is related. The `accession` column means two different things depending on `reason`. An `ambiguous` row names a target peptide of that target protein. An `unrepairableRunCollision` row names a peptide of the *entrapment* protein — the sequence a search reports under `Random_P00001_f0`, not under `P00001` — and the target peptide it collides with may sit in a different protein entirely, since `forbiddenSequences` spans the whole database. A consumer filtering on (accession, peptide) matches nothing for those rows and under-excludes, which is the direction that inflates an entrapment count. `Add` receives `fold` and never uses it, so the report cannot name the entry the collision lives in even though it has been handed the number.

Two things bound all of that. Nothing outside the test project constructs an `EntrapmentReportBuilder` yet, so no database ships this table today and the fix is free to make. And every test that reaches the exclusion table builds at `foldCount: 1` — the two multi-fold builders in the file, `foldCount: 9` at line 254 and the parameterised one at line 34, both pass `NothingForbidden`, so the collision list is empty in each and the duplication has nowhere to show. That is why a structural read finds this and the suite does not.

What I have not measured is how often the duplicate arises in a real database rather than under a constructed forbidden set — it needs a run of consecutive sub-`MinLength` pieces whose concatenation is a genuine target peptide, so it is rarer than a fixture suggests. The footing mismatch itself is structural, not data-dependent.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Fixed in #1267. Already half-fixed higher in the stack -- #1265 replaced the `AddRange` with a `Contains` dedup -- but you are right that the two halves of the table counted differently, and re-keying the list by fold removes the cross-fold question at the root rather than papering over it.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Fixed in #1267, and this was the most consequential of the seventeen. The rows are keyed by the entrapment accession now, `Random_<target>_f<fold>`. You are right about the failure mode: a consumer filtering on (accession, peptide) matched the `ambiguous` rows and silently missed these, which under-excludes and inflates the estimate.

Worth adding that the same commit fixes a second disagreement in that table you did not ask about but which reads together with this one: the stratified column counted PLACEMENTS where the sidecar counts peptides. On the reviewed human database the column read **2,048** against **1,983** sidecar rows, while the `ambiguous` half already agreed at 736 on both sides.

trishorts:

All resolved. Both fixes are in #1267 — the exclusion rows are keyed by the entrapment accession now, and the stratified column counts peptides so it agrees with the sidecar (it read **2,048** against **1,983** sidecar rows on the reviewed human database).

trishorts:

PR #1262, opened 2026-08-31
Entrapment: the foreign-species arm

trishorts:
**Stacked on #1261.** Merge after it; the diff here is the last commit only.

## Why this is required scope, not a nicety

If the entrapment sequences and the decoy sequences are **both shuffles of the target**, they are the same construction, and comparing one against the other demonstrates nothing about either — the circularity Madej & Lam describe. A foreign proteome is the one entrapment source not derived from the target at all, so it is what makes that comparison mean something.

## Nothing is rearranged

A foreign protein already *is* a sequence the sample cannot contain, so it's relabelled rather than permuted. Unlike the permutation path — where every positional annotation had to be dropped because the residues moved — its own annotations describe its own sequence and are kept. Sequence variations are the exception: applying them would expand one entry into several, and the entrapment side is one entry per entry (#1258).

## The accession answers two questions that pull opposite ways

A consumer scanning for the entrapment identifier **must still find it** — miss one and a foreign entry is counted as a *target*, which corrupts the estimate in the worst direction. But `TryParse` **must refuse it**, because there is no target to name.

So a foreign entry reads `Random_foreign_Q9XYZ1`, and `TryParse` refuses on the **marker** rather than on the shape. Otherwise a foreign accession that happened to end in `_f1` would parse as a target called `foreign_ABC`. There's a test for exactly that.

## The hazard is homology, and it's counted rather than repaired

A conserved protein shares peptides with its ortholog, and a shared peptide is a **real target peptide sitting in the entrapment database** — a search matching it counts a true peptide as an entrapment discovery. Nothing can be permuted away here; the foreign sequence is what it is.

So `GenerateForeignEntrapment` reports them by accession, and they join the exclusion table (#1261) under `sharedWithTarget`, beside the ambiguous peptides and the unrepairable collisions. Same rule as everywhere else in this generator: flag and count, never silently repair.

## API

- `EntrapmentAccession.FormatForeign` / `TryParseForeign`
- `EntrapmentProteinGenerator.CreateForeign`
- `EntrapmentProteinGenerator.GenerateForeignEntrapment(foreign, digestionParams, targetPeptides, out sharedWithTarget)`
- `EntrapmentReportBuilder.AddForeign` — separate from `Add`, because a foreign entry has no assembly and no target, so none of the per-stratum permutation figures apply. Folding it in would put entries into the achieved-fold ratio that were never at risk of failing.

## Tests

Seven. The sharing fixture derives its conserved peptide from the target's own digest, so it's guaranteed to actually share rather than hoped to. Full suite **6,178 passed / 0 failed**.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
@@ -1,0 +1,84 @@
+#nullable enable
+using System;
+using System.Globalization;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>
+/// The accession an entrapment protein wears, and how to read the target and fold back out of it.
+/// </summary>
+/// <remarks>
+/// <para>Two existing conventions are followed rather than a third invented. The entrapment marker
+/// goes <b>first</b>, because <see cref="ProteinDbLoader"/> decides a protein is entrapment by
+/// looking for the identifier anywhere in its accession, and prepends it when absent -- so an
+/// accession shaped this way reloads as entrapment with no extra argument and is never
+/// double-prefixed. The fold goes <b>last</b>, matching
+/// <c>VariantApplication.GetAccession</c>, which appends derived detail as a suffix.</para>
+/// <para>Everything needed to pair a discovery back to its target therefore travels in the
+/// accession, which a search already reports. Nothing has to be carried alongside the database, and
+/// nothing leaks into the sequence -- which matters, because target and partner are deliberately
+/// identical in mass and composition and must stay that way.</para>
+/// </remarks>
+public static class EntrapmentAccession
+{
+    private const string FoldMarker = "_f";
+
+    /// <summary>The accession for one fold of one target's entrapment partner.</summary>
+    /// <exception cref="MzLibUtil.MzLibException">The target accession is missing, or the fold is negative.</exception>
+    public static string Format(string targetAccession, int fold,
+        string entrapmentIdentifier = ProteinDbLoader.DefaultEntrapmentIdentifier)
+    {
+        if (string.IsNullOrEmpty(targetAccession))
+        {
+            throw new MzLibUtil.MzLibException("An entrapment accession needs a target accession.");
+        }
+        if (fold < 0)
+        {
+            throw new MzLibUtil.MzLibException($"Fold must not be negative, but was {fold}.");
+        }
+
+        return $"{entrapmentIdentifier}_{targetAccession}{FoldMarker}{fold.ToString(CultureInfo.InvariantCulture)}";
+    }
+
+    /// <summary>
+    /// Reads the target accession and fold back out, or reports that this is not one of ours.
+    /// </summary>
+    /// <remarks>
+    /// The fold is taken from the <em>last</em> "_f&lt;digits&gt;" rather than by splitting on
+    /// underscores: UniProt entry names contain them, so a naive split loses the target.
+    /// </remarks>
+    public static bool TryParse(string? accession, out string targetAccession, out int fold,
+        string entrapmentIdentifier = ProteinDbLoader.DefaultEntrapmentIdentifier)
+    {
+        targetAccession = string.Empty;
+        fold = 0;
+
+        string prefix = entrapmentIdentifier + "_";
+        if (string.IsNullOrEmpty(accession) || !accession.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+
+        int marker = accession.LastIndexOf(FoldMarker, StringComparison.Ordinal);
+        if (marker <= prefix.Length - 1)
+        {
+            return false;
+        }
+
+        string digits = accession.Substring(marker + FoldMarker.Length);
+        if (digits.Length == 0
+            || !int.TryParse(digits, NumberStyles.None, CultureInfo.InvariantCulture, out int parsedFold))
+        {
+            return false;
+        }
+
+        targetAccession = accession.Substring(prefix.Length, marker - prefix.Length);
+        if (targetAccession.Length == 0)
+        {
+            return false;
+        }
+
+        fold = parsedFold;
+        return true;
+    }
+}
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
@@ -23,6 +23,18 @@ public static class EntrapmentAccession
 {
     private const string FoldMarker = "_f";
 
+    /// <summary>
+    /// Marks an entry taken from a foreign proteome rather than rearranged from a target.
+    /// </summary>
+    /// <remarks>
+    /// It sits immediately after the identifier, so a foreign entry reads
+    /// <c>Random_foreign_P12345</c>. Two things had to be true at once: a consumer scanning for the
+    /// entrapment identifier must still find it -- miss one and a foreign entry is counted as a
+    /// <i>target</i>, which corrupts the estimate in the worst direction -- and
+    /// <see cref="TryParse"/> must refuse it, because there is no target to name.
+    /// </remarks>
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
```diff
@@ -0,0 +1,135 @@
...
+        if (string.IsNullOrEmpty(targetAccession))
+        {
+            throw new MzLibUtil.MzLibException("An entrapment accession needs a target accession.");
+        }
+        if (fold < 0)
+        {
+            throw new MzLibUtil.MzLibException($"Fold must not be negative, but was {fold}.");
+        }
+
+        return $"{entrapmentIdentifier}_{targetAccession}{FoldMarker}{fold.ToString(CultureInfo.InvariantCulture)}";
```
acesnik, line 52: Nothing reserves `ForeignMarker` on this side, so a target accession starting `foreign_` mints an accession the new refusal at line 108 then rejects: `Format("foreign_ABC", 0)` gives `Random_foreign_ABC_f0`, `TryParse` false, `TryParseForeign` true with `ABC_f0`. A throw here when `targetAccession` starts with the marker makes the namespace actually disjoint, and makes the comment at 105-107 true.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,492 @@
...
+            total.UnpairableAllPermutationsTaken += stratum.UnpairableAllPermutationsTaken;
+            total.UnpairableSpaceTooSmallForFoldCount += stratum.UnpairableSpaceTooSmallForFoldCount;
+            total.Ambiguous += stratum.Ambiguous;
+            total.MissedCleavagePeptidesSpanningAnExcision += stratum.MissedCleavagePeptidesSpanningAnExcision;
+            total.SearchSpacePeptides += stratum.SearchSpacePeptides;
+            total.UnrepairableRunCollisions += stratum.UnrepairableRunCollisions;
+        }
+
+        var provenance = new EntrapmentProvenance(
+            method: "composition-preserving permutation (deterministic unranking)",
```
acesnik, line 460: Unconditional, so once `AddForeign` has been called the emitted provenance describes only half the database. `ForeignEntries` never reaches `ToTabSeparated` either, and there is no entry-count denominator beside it, so the arm's *r* is not recoverable from the report.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,492 @@
...
+        return new EntrapmentReport(provenance,
+            _strata.Values.OrderBy(s => s.SiteCount).ToList(),
+            total,
+            _ambiguousByAccession.Where(kv => kv.Value.Count > 0)
+                .ToDictionary(kv => kv.Key, kv => (IReadOnlyCollection<string>)kv.Value),
+            _unrepairableByAccession.Where(kv => kv.Value.Count > 0)
+                .ToDictionary(kv => kv.Key, kv => (IReadOnlyCollection<string>)kv.Value))
+        {
+            ForeignEntries = _foreignEntries,
+            ForeignPeptidesSharedWithTarget = _foreignShared,
```
acesnik, line 477: The builder's live dictionary, where the two above it are `ToDictionary` snapshots — a later `AddForeign` mutates an already-built report. And in `AddForeign`, `_foreignShared[accession] = peptides` overwrites while `_foreignEntries += entryCount` accumulates, so two calls on one accession inflate the count and lose peptides.

acesnik (commented):
🤖 This is the arm that makes the comparison mean something. If the entrapment sequences and the decoy sequences are both shuffles of the target, they are one construction under two names, and a foreign proteome is the only entrapment source not derived from the target at all. Relabelling rather than permuting is right, and the consequence you draw from it — that a foreign protein's own positional annotations still describe its own sequence and so survive, where the permutation path had to drop all of them — is the cleanest statement of that asymmetry anywhere in the stack. Counting homologous peptides instead of repairing them is the same rule as everywhere else, and here it is the only rule available, since the sequence cannot be permuted away from itself.

Determinism holds: `DatabaseEntries` preserves input order and both TSV writers sort keys and peptides with `StringComparer.Ordinal`, so storing the collisions as a `HashSet` does not leak iteration order into the output.

Three things I would rather see before a second approval, all small.

`EntrapmentAccession.Format` will mint an accession that `TryParse` then refuses. `Format` guards only null/empty target and negative fold, so `Format("foreign_ABC", 0)` gives `Random_foreign_ABC_f0`; `TryParse` returns false on it via the new block at `EntrapmentAccession.cs:108`, and `TryParseForeign` reads it as a foreign entry with accession `ABC_f0`. Before that block it parsed back to target `foreign_ABC`, fold 0. The comment there says refusing on the marker "keeps that from ever being a guess", but nothing reserves the marker on the minting side, so the ambiguity has moved rather than gone. A guard in `Format` rejecting a `targetAccession` that starts with `ForeignMarker` closes it and makes the comment true. No UniProt accession looks like this; a hand-assembled FASTA can.

`ForeignPeptidesSharedWithTarget = _foreignShared` in `Build()` hands out the builder's live dictionary, where both lists immediately above it are `ToDictionary` snapshots — so a later `AddForeign` mutates an already-built report. In `AddForeign` itself the two figures disagree about what a repeat call means: `_foreignEntries += entryCount` accumulates while `_foreignShared[accession] = peptides` overwrites, so two calls covering one accession inflate the entry count and drop the first call's peptides. Those two are inconsistent inside a single report, which is the one thing this report cannot afford.

The arm is also largely absent from what gets emitted. `ExclusionsToTabSeparated` carries the `sharedWithTarget` rows, but `ToTabSeparated` has no foreign column and no `# foreignEntries` line, and `# method` reads `composition-preserving permutation (deterministic unranking)` unconditionally — so the machine-readable description of a mixed database says it was made entirely by permutation. `ForeignEntries` has no denominator anywhere either: `TargetPeptides` counts base pieces, not entries, so the arm's achieved *r* is not computable from what is written, and `GenerateForeignEntrapment` takes no fold or subsetting parameter, leaving the ratio to whatever the caller pre-selects.

Nothing outside the tests calls `AddForeign`, so no shipped database or report is wrong today — which is why this is a withhold rather than a block, and why the fixes are cheap to make now.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAccession.cs
trishorts: Fixed in #1267. `Format` now throws when the target accession begins with the marker, so the namespaces are actually disjoint and the comment on `TryParse`'s refusal is true. Reproduced first: `Format("foreign_ABC", 0)` gave `Random_foreign_ABC_f0`, `TryParse` false, `TryParseForeign` true with `ABC_f0` -- exactly as you described.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Fixed in #1267. `# foreignEntries` and `# foreignPeptidesSharedWithTarget` are emitted when the arm was used, so the arm's `r` is recoverable from the report and a bottom-up report is byte-identical to before.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Both fixed in #1267. `Build` returns a snapshot like the two beside it, and `AddForeign` merges rather than assigns. Reproduced: two calls naming one accession gave `entries=5` with only the second call's peptides, and a later `AddForeign` mutated a report that had already been built.

trishorts:

All resolved, fixes in #1267 — the marker is reserved in `Format`, `AddForeign` merges instead of assigning, `Build` returns a snapshot, and foreign entries reach the provenance so the arm's `r` is recoverable.

trishorts:

PR #1263, opened 2026-08-31
Entrapment: proteoform mode

trishorts:
**Stacked on #1262.** Merge after it; the diff here is the last commit only. This completes the entrapment stack.

## Top-down doesn't digest, so the protein is the unit

There are no cleavage sites to hold and no pieces to assemble. K and R become ordinary residues, free to move — the whole difference from the bottom-up path, and there's a test asserting they actually *do* move, so the two modes can't quietly become the same thing under two names.

## Nothing is excised, which is what makes annotations carryable

The bottom-up path had to drop **every** positional annotation, because excising an unpairable piece shortens the protein and leaves coordinates pointing past its end — the defect that once made every database this generator produced unloadable. Here the length never changes, so the coordinates stay valid.

A protein with no usable rearrangement is dropped **whole** instead of partly: half a proteoform is not a proteoform.

## Annotations are moved, not copied

| annotation | treatment | why |
|---|---|---|
| disulfide bonds | **both endpoints mapped** | a bond names two *cysteines*; copying coordinates would assert a bond between whatever residues land there, which is false |
| splice sites | both endpoints mapped | same reason |
| modifications | mapped (existing machinery) | a modification names a residue |
| truncation products | **coordinates kept** | a *span* has no contiguous image under a rearrangement, and the span is what defines a proteoform |
| sequence variations | dropped | applying them expands one entry into several (#1258) |

The entrapment proteoform truncated at 25 is the counterpart of the target truncated at 25. Its residues differ — that's the point of an entrapment sequence.

Begin/end are re-ordered after mapping, since a rearrangement can put the later residue first.

## Tests

Seven. One earned its keep immediately by failing: the modification fixture put a threonine-targeted phosphorylation on a valine, and mzLib correctly refused it — so the test was proving that *invalid* modifications get dropped rather than that valid ones survive.

Full suite **6,185 passed / 0 failed**.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
 using MzLibUtil;
 using Omics.Digestion;
 using System.Collections.Generic;
@@ -60,13 +61,14 @@ public sealed class EntrapmentAssembly
 {
     internal EntrapmentAssembly(string targetSequence, string entrapmentSequence,
         int[] targetToEntrapmentPosition, IReadOnlyList<EntrapmentPiece> pieces,
-        int missedCleavagePeptidesSpanningAnExcision)
+        int missedCleavagePeptidesSpanningAnExcision, int unrepairableRunCollisions)
     {
         TargetSequence = targetSequence;
         EntrapmentSequence = entrapmentSequence;
         TargetToEntrapmentPosition = targetToEntrapmentPosition;
         Pieces = pieces;
         MissedCleavagePeptidesSpanningAnExcision = missedCleavagePeptidesSpanningAnExcision;
+        UnrepairableRunCollisions = unrepairableRunCollisions;
     }
 
     public string TargetSequence { get; }
@@ -92,6 +94,24 @@ public sealed class EntrapmentAssembly
     /// invariant holds everywhere.
     /// </remarks>
     public int MissedCleavagePeptidesSpanningAnExcision { get; }
+
+    /// <summary>
+    /// Missed-cleavage peptides of this entrapment sequence that equal a real target peptide and
+    /// could not be avoided.
+    /// </summary>
+    /// <remarks>
+    /// <para>Each piece is permuted away from any collision it can be permuted away from, runs
+    /// included. What cannot be repaired is a run whose final piece has exactly one arrangement --
+    /// a short low-complexity piece such as <c>AAR</c>, where holding the cleavage residue leaves
+    /// nothing to reorder. There is no candidate to move to, so the collision stands.</para>
+    /// <para>Repairing it would mean backtracking into an already-placed piece or excising a
+    /// perfectly good one, and this project counts collisions rather than silently repairing them.
+    /// Measured on the reviewed human proteome: 1,794 such peptides, 0.065% of the target set,
+    /// 97.9% of them ending in a piece with no alternative and 80% exactly at the minimum
+    /// searchable length of seven residues. A search that matches one counts a real peptide as an
+    /// entrapment discovery, so anyone reading an entrapment count needs this number beside it.</para>
+    /// </remarks>
+    public int UnrepairableRunCollisions { get; }
 
     public int ExcisedCount => Pieces.Count(p => p.Outcome == PieceOutcome.Excised);
 
@@ -141,6 +161,10 @@ public static class EntrapmentAssembler
 
         var pieces = new List<EntrapmentPiece>(sites.Count - 1);
         var entrapment = new StringBuilder(targetSequence.Length);
+        // The entrapment pieces already placed, in the order they appear in the entrapment
+        // sequence, so a candidate can be tested against the runs it would complete.
+        var placed = new List<string>(sites.Count - 1);
+        int unrepairable = 0;
         int[] map = Enumerable.Repeat(-1, targetSequence.Length).ToArray();
         var retainedTargetIndices = new List<int>(sites.Count - 1);
 
@@ -150,12 +174,19 @@ public static class EntrapmentAssembler
             int length = sites[index + 1] - start;
             string piece = targetSequence.Substring(start, length);
 
+            Func<string, bool>? completesAForbiddenRun =
+                RejectRunCollisions(placed, digestionParams.MaxMissedCleavages, forbiddenSequences);
+
             EntrapmentPeptide partner = EntrapmentPeptideGenerator.Create(piece, motifs, forbiddenSequences,
-                fold, foldCount, seed, TerminalAnchors(index, sites.Count - 1, length));
+                fold, foldCount, seed, TerminalAnchors(index, sites.Count - 1, length),
+                completesAForbiddenRun);
 
             if (partner.Succeeded)
             {
                 AppendPiece(entrapment, map, start, partner.EntrapmentSequence!, partner.SwappedPositions!);
+                placed.Add(partner.EntrapmentSequence!);
+                // A permuted piece was chosen with the run test applied, so this cannot fire -- it is
+                // asserted rather than assumed only because the count below must mean one thing.
                 pieces.Add(new EntrapmentPiece(index, piece, partner.EntrapmentSequence,
                     PieceOutcome.Permuted, EntrapmentFailure.None));
                 retainedTargetIndices.Add(index);
@@ -167,6 +198,13 @@ public static class EntrapmentAssembler
             if (piece.Length < digestionParams.MinLength)
             {
                 AppendPiece(entrapment, map, start, piece, Identity(piece.Length));
+                placed.Add(piece);
+                // Kept verbatim because it has no alternative arrangement, so if it completes a
+                // forbidden run there is nothing to move to. Count it; do not repair it.
+                if (completesAForbiddenRun is not null && completesAForbiddenRun(piece))
+                {
+                    unrepairable++;
+                }
                 pieces.Add(new EntrapmentPiece(index, piece, piece,
                     PieceOutcome.KeptVerbatimTooShort, partner.Failure));
                 retainedTargetIndices.Add(index);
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        var withNewSequence = (Protein)target.CloneWithNewSequenceAndMods(assembly.EntrapmentSequence, movedMods);
+
+        return new Protein(withNewSequence,
+            accession: EntrapmentAccession.Format(target.Accession, fold, entrapmentIdentifier),
+            isEntrapment: true,
+            oneBasedModifications: movedMods,
+            sequenceVariations: new List<SequenceVariation>(),
+            appliedSequenceVariations: new List<SequenceVariation>(),
+            proteolysisProducts: target.TruncationProducts.ToList(),
```
acesnik, line 75: `Protease.cs:211` yields intact truncation products as searched species, and a span of a whole-protein rearrangement holds a different multiset than the target's — measured on `signal peptide 1-25`: target sorted `AACDEHIIIKKKLNQQRTTVVVVWY`, partner `AAAAAAAACCCCCCCDDDDDDDDIV`, so not the same mass. Either drop these as the bottom-up path does, or partition on the span boundaries and rearrange within each part so every span keeps its own composition.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,428 @@
...
+            AppendPiece(entrapment, map, 0, targetSequence, Identity(targetSequence.Length));
+            pieces.Add(new EntrapmentPiece(0, targetSequence, targetSequence,
+                PieceOutcome.KeptVerbatimTooShort, partner.Failure));
+        }
+        else
+        {
+            pieces.Add(new EntrapmentPiece(0, targetSequence, null, PieceOutcome.Excised, partner.Failure));
+        }
+
+        return new EntrapmentAssembly(targetSequence, entrapment.ToString(), map, pieces, 0,
```
acesnik, line 290: Both collision figures are constants here, so a report over a proteoform database reads zero regardless of what it contains. That compounds with only the whole-sequence candidate being tested against `forbiddenSequences`: a truncation span of the partner can equal a real target proteoform and is neither avoided nor counted. Forbidding the partner's own 1-25 span and regenerating gave back the identical span.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,428 @@
...
+        int[] map = Enumerable.Repeat(-1, targetSequence.Length).ToArray();
+        var entrapment = new StringBuilder(targetSequence.Length);
+
+        if (partner.Succeeded)
+        {
+            AppendPiece(entrapment, map, 0, partner.EntrapmentSequence!, partner.SwappedPositions!);
+            pieces.Add(new EntrapmentPiece(0, targetSequence, partner.EntrapmentSequence,
+                PieceOutcome.Permuted, EntrapmentFailure.None));
+        }
+        else if (targetSequence.Length < minLength)
```
acesnik, line 279: Unreachable at `minLength = 1`, the default `CreateProteoform` passes. Where it does fire it emits a verbatim copy of the target protein as an entrapment entry — defensible for a sub-`MinLength` base piece in bottom-up, but a short proteoform is a searched species in this mode.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,407 @@
...
+            disulfideBonds: MoveResiduePairs(target.DisulfideBonds, map,
+                (b, e, d) => new DisulfideBond(b, e, d)),
+            spliceSites: MoveResiduePairs(target.SpliceSites, map,
```
acesnik, line 78: A splice site names a junction between adjacent residues (`ProteinXmlEntry.cs:650` takes begin/end from the flanking pair), not two residues the way a bond does, so independently-mapped endpoints are no longer adjacent and the annotation asserts nothing. Cosmetic — nothing reads it past `ProteinDbWriter` — but it is not the disulfide case the remarks justify it with.

acesnik (commented):
🤖 This is the top-down arm, and the shape of it is right: no digestion means nothing to pin but the termini, the protein is the unit, and because nothing is excised the length is fixed — which is genuinely why the annotations become carryable here when the bottom-up path had to drop every one of them. Mapping a disulfide bond's endpoints through the position map so it still joins the same two cysteines is the correct reading of what that annotation asserts, and restoring begin/end order afterwards is the kind of detail that only turns up when you run it. Cost is a non-issue — `AssembleWholeProtein` took 415 ms on a 34,350-residue sequence, so pointing a routine documented as "intended for peptide-length sequences" at whole proteins is free.

The truncation products are the problem, and they are why I am withholding rather than approving. `Protease.GetUnmodifiedPeptides` yields intact truncation products as searched species (`Proteomics/ProteolyticDigestion/Protease.cs:211`, under the `//add intact proteolysis products` comment), so in a proteoform workflow the span *is* what gets reported — and the span carried across at `EntrapmentProteinGenerator.cs:75` is not isomeric with the target's. On a 120-residue fixture annotated `signal peptide 1-25` and `chain 26-120`, sorted span 1-25 is `AACDEHIIIKKKLNQQRTTVVVVWY` in the target and `AAAAAAAACCCCCCCDDDDDDDDIV` in the partner. Different composition, so different mass. That is structural rather than fixture-specific: rearranging a whole protein cannot preserve the composition of an arbitrary sub-span. `AProteoformPartnerIsTheWholeProteinRearranged` asserts `Sorted(partner) == Sorted(target)` two tests above `TruncationProductsKeepTheirSpans`, and only the full-length species satisfies it; the coordinates the second test checks are indeed unchanged and in range, but that is not the same as the proteoform being a matched-mass counterpart of anything.

The doc presents this as intended — "Its residues differ, which is the point of an entrapment sequence" — but that conflates residues differing in *order*, which is what makes target and partner the hardest possible pair to tell apart, with differing in *identity*, which makes them an unmatched pair at an arbitrary mass. Two ways out, both cheap: drop truncation products in this mode as the bottom-up path drops every positional annotation, or partition on the truncation-span boundaries and rearrange within each part, which keeps each span isomeric and the whole protein isomeric at the same time.

Same root cause, second consequence: nothing below the whole protein is tested against `forbiddenSequences`. `AssembleWholeProtein` checks only the whole-sequence candidate, so a partner's truncation span can equal a real target proteoform — forbidding the partner's own 1-25 span and regenerating produced the identical span. And the assembly returns `0` and `new List<string>()` for both collision figures unconditionally, so a report over a proteoform database reads zero collisions whatever it contains. That is a count taken from a constant rather than from what was written.

Two smaller ones. `minLength` defaults to 1, so the `KeptVerbatimTooShort` branch is unreachable at the default and the parameter's documented purpose never fires; where it does fire it writes a verbatim copy of the target protein into the entrapment database, which is safe in bottom-up because the piece is below the searchable length, but not here, where a short proteoform is exactly what is searched. And mapping both endpoints of a `SpliceSite` independently is not the disulfide case: a bond names two residues, a splice site names a junction between adjacent ones (`ProteinXmlEntry.cs:650` takes begin/end from the feature's flanking pair), so two independently-mapped images are no longer adjacent. Nothing consumes splice sites past `ProteinDbWriter`, so that one is cosmetic.

What I did not verify is how common truncation products are in a real proteome. `Test/DatabaseTests/cRAP_databaseGPTMD.xml` carries them on 116 of 116 entries (230 features), but that is a contaminant database, and there was no full reviewed proteome XML to hand. Nothing outside the tests calls `CreateProteoform`, so no database ships this today.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Fixed in #1267, taking your second option. The rearrangement is confined to the truncation-span boundaries: every span is a union of whole segments, so permuting within each segment keeps every span isomeric with the one it replaces, and the whole protein stays isomeric too. It is the same shape as cutting at cleavage sites in the bottom-up path. Your measured example is what made the argument -- rearranging the whole protein preserved only the whole protein's composition.

Note #1263 is marked draft for now: the current focus is bottom-up. The findings are fixed regardless.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Fixed in #1267. Segment-aligned spans of the assembled partner are tested against `forbiddenSequences` and named, so the two figures are no longer constants and a report over a proteoform database no longer reads zero however much it holds.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Fixed in #1267. A partner identical to its target is dropped rather than emitted. Your distinction is the right one: keeping a sub-`MinLength` base piece verbatim is defensible in bottom-up because it is not searched on its own, and a short proteoform is.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Fixed in #1267 -- splice sites are dropped rather than mapped. Cosmetic as you say, but the reasoning is right and the remarks were justifying it with the disulfide case, which is a different thing: a bond joins two residues and survives independent mapping, a junction between adjacent residues does not.

trishorts:

All resolved, fixes in #1267 — the rearrangement is confined to truncation-span boundaries so every span stays isomeric, span collisions are counted rather than passed as constants, a partner identical to its target is dropped, and splice sites are dropped rather than mapped. Note this PR is **marked draft** for now: the current focus is bottom-up, so please don't spend review time here — the findings are fixed regardless.

trishorts:
Superseded by **#1272**, which carries proteoform mode with its four review findings folded in, stacked on #1271.

Kept as a draft there for the same reason it was here: the focus is bottom-up, and this arm has been tested but never used to build a real database.

@acesnik — your measurement on this one drove the actual design change. Rearranging the whole protein preserved only the whole protein's composition, and you showed a truncation span came back holding a different multiset. #1272 confines the rearrangement to the span boundaries instead, so every span stays isomeric with the one it replaces.

PR #1264, opened 2026-08-31
Entrapment: don't emit an empty protein, don't repeat an excluded peptide

trishorts:
**Stacked on #1263.** Merge after it; the diff here is the last commit only.

Both found by the first consumer running the pipeline — the fourth and fifth time that has caught something a passing suite did not.

## An empty entrapment protein was being written

`Q156A1` is a methionine followed by **79 glutamines**. Its single base piece has exactly one arrangement once the termini are anchored (#1257), so it's excised and nothing remains — and the generator wrote the resulting empty-sequence protein anyway.

A loader warns and discards it, **and its decoy with it**. That's why a database built from 20,416 entries loads 20,416 targets against **20,415** entrapment, and reports *"2 empty Protein entries ignored"*.

That off-by-one had been sitting in `measure_r.py` output for days and been read as harmless rounding. It was a protein quietly missing from the entrapment arm.

## The exclusion list repeated peptides

The same run can collide at two points in one low-complexity protein, and the list accumulated **placements** rather than peptides:

| | |
|---|---|
| rows | 2,720 |
| distinct accession + peptide | **2,636** |

An exclusion list is read as "these peptides", so a consumer subtracting a duplicate twice would over-correct. Now deduplicated per accession.

## Tests

Two, one guarding the other's premise: if terminal anchoring ever stopped freezing a homopolymer, the empty-protein test would start passing for the wrong reason.

Full suite **6,187 passed / 0 failed**.

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
 using MzLibUtil;
 using Omics.Digestion;
 using System.Collections.Generic;
@@ -60,13 +61,14 @@ public sealed class EntrapmentAssembly
 {
     internal EntrapmentAssembly(string targetSequence, string entrapmentSequence,
         int[] targetToEntrapmentPosition, IReadOnlyList<EntrapmentPiece> pieces,
-        int missedCleavagePeptidesSpanningAnExcision)
+        int missedCleavagePeptidesSpanningAnExcision, int unrepairableRunCollisions)
     {
         TargetSequence = targetSequence;
         EntrapmentSequence = entrapmentSequence;
         TargetToEntrapmentPosition = targetToEntrapmentPosition;
         Pieces = pieces;
         MissedCleavagePeptidesSpanningAnExcision = missedCleavagePeptidesSpanningAnExcision;
+        UnrepairableRunCollisions = unrepairableRunCollisions;
     }
 
     public string TargetSequence { get; }
@@ -92,6 +94,24 @@ public sealed class EntrapmentAssembly
     /// invariant holds everywhere.
     /// </remarks>
     public int MissedCleavagePeptidesSpanningAnExcision { get; }
+
+    /// <summary>
+    /// Missed-cleavage peptides of this entrapment sequence that equal a real target peptide and
+    /// could not be avoided.
+    /// </summary>
+    /// <remarks>
+    /// <para>Each piece is permuted away from any collision it can be permuted away from, runs
+    /// included. What cannot be repaired is a run whose final piece has exactly one arrangement --
+    /// a short low-complexity piece such as <c>AAR</c>, where holding the cleavage residue leaves
+    /// nothing to reorder. There is no candidate to move to, so the collision stands.</para>
+    /// <para>Repairing it would mean backtracking into an already-placed piece or excising a
+    /// perfectly good one, and this project counts collisions rather than silently repairing them.
+    /// Measured on the reviewed human proteome: 1,794 such peptides, 0.065% of the target set,
+    /// 97.9% of them ending in a piece with no alternative and 80% exactly at the minimum
+    /// searchable length of seven residues. A search that matches one counts a real peptide as an
+    /// entrapment discovery, so anyone reading an entrapment count needs this number beside it.</para>
+    /// </remarks>
+    public int UnrepairableRunCollisions { get; }
 
     public int ExcisedCount => Pieces.Count(p => p.Outcome == PieceOutcome.Excised);
 
@@ -141,6 +161,10 @@ public static class EntrapmentAssembler
 
         var pieces = new List<EntrapmentPiece>(sites.Count - 1);
         var entrapment = new StringBuilder(targetSequence.Length);
+        // The entrapment pieces already placed, in the order they appear in the entrapment
+        // sequence, so a candidate can be tested against the runs it would complete.
+        var placed = new List<string>(sites.Count - 1);
+        int unrepairable = 0;
         int[] map = Enumerable.Repeat(-1, targetSequence.Length).ToArray();
         var retainedTargetIndices = new List<int>(sites.Count - 1);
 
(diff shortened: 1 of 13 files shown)


mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,419 @@
...
+            {
+                Protein partner = Create(entry, digestionParams, forbiddenSequences, out EntrapmentAssembly assembly,
+                    fold, foldCount, seed, entrapmentIdentifier);
+
+                // Every piece excised leaves nothing behind. A protein with an empty sequence is not
+                // a database entry: a loader warns and discards it, and its decoy with it, so writing
+                // one produces a database whose entrapment count is quietly short of its target count.
+                // Q156A1 is the real case -- a methionine followed by 79 glutamines, whose only base
+                // piece has exactly one arrangement once the termini are anchored.
+                if (assembly.EntrapmentSequence.Length == 0)
```
acesnik, line 263: The loaded database is unchanged by this — I measured `r = 0.5000` on both sides, with the same target/entrapment/decoy counts. What goes away is the `"N empty entries ignored"` warning that was the only signal the arm was short. Worth saying so in the comment: short by construction rather than short after a warning. An entry-level count in the report would replace what the warning was doing.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,428 @@
...
+    /// a short low-complexity piece such as <c>AAR</c>, where holding the cleavage residue leaves
+    /// nothing to reorder. There is no candidate to move to, so the collision stands.</para>
+    /// <para>Repairing it would mean backtracking into an already-placed piece or excising a
+    /// perfectly good one, and this project counts collisions rather than silently repairing them.
+    /// Measured on the reviewed human proteome: 1,794 such peptides, 0.065% of the target set,
+    /// 97.9% of them ending in a piece with no alternative and 80% exactly at the minimum
+    /// searchable length of seven residues. A search that matches one counts a real peptide as an
+    /// entrapment discovery, so anyone reading an entrapment count needs this number beside it.</para>
+    /// </remarks>
+    public int UnrepairableRunCollisions => UnrepairableRunCollisionPeptides.Count;
```
acesnik, line 115: Not deduplicated, so this counts placements while the report's sidecar now counts peptides. `"AAAAKAAAARAAAAKAAAAR"` forbidding `{"AAAAKAAAAR"}` gives 2 here against one sidecar row. The column and the sidecar are read together, so they should agree.

acesnik (commented):
🤖 This closes the gap between what the generator wrote and what a loader would keep: an entry whose every base piece is excised produced an empty-sequence `Protein`, and the exclusion list accumulated placements where a consumer reads peptides. Both are the kind of thing only a run surfaces. The two tests are also the right shape — `TheHomopolymerReallyHasNoArrangement` guards a premise that, if it ever stopped holding, would make the other test pass for the wrong reason.

The ratio is the part I would state differently. I wrote both variants and loaded them back, filtering empties the way `DatabaseLoadingEngine.cs:182-183` does:

```
targets = { "M" + 79x"Q", 38-mer }, trypsin, minLen 7, mmc 2
before  written=4  loaded=8  emptyDropped=2  target=2 entrapment=1 decoyTarget=2 decoyEntrapment=1  r=0.5000
after   written=3  loaded=6  emptyDropped=0  target=2 entrapment=1 decoyTarget=2 decoyEntrapment=1  r=0.5000
```

The entrapment arm reaching a search is the same size before and after, and so is *r*. What changes is that the shortfall stops announcing itself — MetaMorpheus's `"2 empty entries ignored"` (`DatabaseLoadingEngine.cs:121-124`) was the only thing telling anyone the arm was an entry short. So 20,416 against 20,415 is not reconciled here; the entry is now missing by construction rather than missing after a warning. That is still the right call, since the alternative writes a genuine target peptide into the entrapment arm — but the body reads as though the count came back into line, and someone tuning *r* from it would be surprised.

That matters because the report is the product and it carries no entry-level figure: `ToTabSeparated` is peptide-level throughout and `ForeignEntries` is the only entry count on the type. An `entriesWithoutAPartner` beside it would give a reader what the warning was supplying.

Second, the deduplication fixed the list and left the count. `UnrepairableRunCollisions` at `EntrapmentAssembler.cs:115` is still `UnrepairableRunCollisionPeptides.Count` over an un-deduplicated `List<string>`, so the `unrepairableRunCollisions` column and the exclusions sidecar now disagree — the same over-correction one layer up. On `"AAAAKAAAARAAAAKAAAAR"` with `{"AAAAKAAAAR"}` forbidden the count is 2 and the sidecar has one row; across the real database that is presumably the 84 rows between your 2,720 and 2,636.

I could not check the 2,720 / 2,636 or 20,416 / 20,415 figures themselves — no proteome in the checkout — so those two are your numbers, not mine. The mechanism behind both reproduces on synthetic fixtures.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Both done in #1267. The comment now says the loaded database is unchanged and that what goes away is the loader's warning, and `EntrapmentReport.EntriesYieldingNoPartner` replaces what the warning was doing -- emitted in the provenance when non-zero, so short by construction is said where it is decided.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Fixed in #1267. The column counts peptides, as the sidecar does. Measured on the reviewed human database before the fix: **2,048** in the column against **1,983** sidecar rows, with the `ambiguous` half already agreeing at 736.

trishorts:

All resolved, fixes in #1267 — the comment says the loaded database is unchanged, `EntrapmentReport.EntriesYieldingNoPartner` replaces what the loader's warning was doing, and the collision column now agrees with the sidecar.

trishorts:

PR #1265, opened 2026-08-31
Entrapment: address the review across the stack

trishorts:
**Stacked on #1264**, and carries the fixes for every review comment on #1253–#1260. Landed on one branch because the reviewer noted several could ride a later branch, and #1255's `ProteinDbLoader` edit is still live at the top of the stack.

Full suite **6,192 passed / 0 failed**.

| PR | comment | fix |
|---|---|---|
| #1253 | culture-dependent seed | seed formatted invariantly |
| #1254 | preventing / multi-residue motifs pin nothing | guard refuses those agents |
| #1255 | CS1587 from const placement | const moved above the loader's doc comment |
| #1255 | stale `UniProtSequenceAttributes` | length and mass derived from the entrapment sequence |
| #1255 | `OrdinalIgnoreCase` vs `Ordinal` | `Ordinal` throughout |
| #1256/#1260 | whole-protein figures land in the zero-site stratum | accumulate on the total, no stratum |
| #1256 | vacuous truncation-product loop | asserts the real contract, plus a sibling on the proteoform path |
| #1257 | anchors in piece coordinates | computed in protein coordinates |
| #1257 | non-discriminating position-2 test | fixture whose second residue is unique |
| #1259 | count returns on first match; no length bound | every offending run, bounded |
| #1259 | run collision reported as `AllPermutationsTaken` | new `RunCollisionsExhaustedTheSpace` |
| #1260 | no entrapment-side search space | `entrapmentSearchSpacePeptides` column |

## The three worth reading

**The seed (#1253).** A negative seed rendered its sign as U+002D under `en-US` and U+2212 under `sv-SE`, `fi-FI`, `lt-LT` — different bytes into SHA-256, so the same request produced a different database on a differently-configured machine. There is now a test that builds under four cultures and compares.

**The motif guard (#1254).** `CleavageSitePositions` skips a position where a motif fits but is *prevented*, so under `trypsin|P`'s `K[P]|` neither the K nor the P is held, and a rearrangement can invent or destroy a cleavage site. Five shipped agents carry preventing motifs, and multi-residue motifs have their span cut by the piece boundary. Pinning across boundaries means this stops being a per-piece loop, so refusing is the interim: the alternative is a database that digests differently from its target and whose peptides then fail to pair, **silently**.

**The run-collision count (#1259).** Wrong in both directions, which is why they didn't cancel. It returned on the first match, so several colliding runs ending at one piece counted as one — while the property documented *peptides*. And with no length bound, a concatenation no search can report still rejected candidates and still counted, while the pairing index that defines the searchable population had already excluded it. That should also settle 1,794 against 1,984: the assembler and the pairing now agree about what a peptide is.

## The two tests that couldn't discriminate

`Create_NeverLeavesAnAnnotationPointingPastTheSequence` looped `TruncationProducts` on a path that empties it, so the body never ran. It now asserts the collections *are* cleared — the actual contract — states the range property over everything the entry does carry, and a sibling test covers the range assertion on the proteoform path, where the collection is genuinely non-empty.

`Create_KeepsAModificationOnTheSecondResidue` used `MAAA…`, where an `A` returns to slot 1 whether anchored or not. Its second residue is now the only `S` in the sequence, and the test asserts that uniqueness so it cannot quietly stop discriminating.

### DIFF
--- a/mzLib/Test/DatabaseTests/TestEntrapmentProtein.cs
+++ b/mzLib/Test/DatabaseTests/TestEntrapmentProtein.cs
@@ -1,0 +1,252 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using NUnit.Framework;
+using Omics.Digestion;
+using Omics.Modifications;
+using Proteomics;
+using Proteomics.ProteolyticDigestion;
+using UsefulProteomicsDatabases;
+using UsefulProteomicsDatabases.EntrapmentGeneration;
+
+namespace Test.DatabaseTests;
+
+[TestFixture]
+[ExcludeFromCodeCoverage]
+public class EntrapmentProteinTests
+{
+    private static readonly HashSet<string> NothingForbidden = new();
+    private const string Sequence = "SYKALADQMNLLLSKGGVDTTPFAWENDRQISTLGGYK";
+
+    private static IDigestionParams Tryptic => new DigestionParams("trypsin", minPeptideLength: 7, maxMissedCleavages: 2);
+
+    private static Modification Phospho()
+    {
+        ModificationMotif.TryGetMotif("T", out ModificationMotif motif);
+        return new Modification(_originalId: "Phospho", _modificationType: "Common Biological",
+            _target: motif, _locationRestriction: "Anywhere.", _monoisotopicMass: 79.96633);
+    }
+
+    // ---- accession ---------------------------------------------------------
+
+    [Test]
+    public void Accession_PutsTheMarkerFirstSoTheLoaderFindsIt()
+    {
+        string accession = EntrapmentAccession.Format("P12345", fold: 0);
+
+        // Prefix, because ProteinDbLoader detects entrapment by looking for the identifier anywhere
+        // in the accession and prepends it when absent -- matching the convention means a written
+        // database reloads as entrapment with no extra argument, and is never double-prefixed.
+        Assert.That(accession, Does.StartWith(ProteinDbLoader.DefaultEntrapmentIdentifier));
+        Assert.That(accession, Does.Contain("P12345"));
+    }
+
+    [Test]
+    public void Accession_RoundTripsTheTargetAndFold()
+    {
+        for (int fold = 0; fold < 4; fold++)
+        {
+            string accession = EntrapmentAccession.Format("P12345", fold);
+            Assert.That(EntrapmentAccession.TryParse(accession, out string target, out int parsedFold), Is.True);
+            Assert.That(target, Is.EqualTo("P12345"));
+            Assert.That(parsedFold, Is.EqualTo(fold));
+        }
+    }
+
+    [Test]
+    public void Accession_RefusesSomethingThatIsNotAnEntrapmentAccession()
+    {
+        Assert.That(EntrapmentAccession.TryParse("P12345", out _, out _), Is.False);
+        Assert.That(EntrapmentAccession.TryParse("", out _, out _), Is.False);
+        Assert.That(EntrapmentAccession.TryParse(null, out _, out _), Is.False);
+    }
+
+    [Test]
+    public void Accession_SurvivesATargetAccessionContainingUnderscores()
+    {
+        // UniProt entry names always carry one, so splitting naively would lose the fold.
+        string accession = EntrapmentAccession.Format("SP_HUMAN_P12345", fold: 3);
+        Assert.That(EntrapmentAccession.TryParse(accession, out string target, out int fold), Is.True);
+        Assert.That(target, Is.EqualTo("SP_HUMAN_P12345"));
+        Assert.That(fold, Is.EqualTo(3));
+    }
+
+    // ---- protein generation ------------------------------------------------
+
+    [Test]
+    public void Create_MakesAnEntrapmentProteinIsomericWithItsTarget()
+    {
+        var target = new Protein(Sequence, "P12345");
+        Protein entrapment = EntrapmentProteinGenerator.Create(target, Tryptic, NothingForbidden);
+
+        Assert.That(entrapment.IsEntrapment, Is.True);
+        Assert.That(entrapment.BaseSequence, Is.Not.EqualTo(target.BaseSequence));
+        Assert.That(string.Concat(entrapment.BaseSequence.OrderBy(c => c)),
+            Is.EqualTo(string.Concat(target.BaseSequence.OrderBy(c => c))));
+        Assert.That(EntrapmentAccession.TryParse(entrapment.Accession, out string parsed, out _), Is.True);
+        Assert.That(parsed, Is.EqualTo("P12345"));
+    }
+
+    [Test]
+    public void Create_CarriesModificationsToTheResidueTheyWereOn()
+    {
+        Modification phospho = Phospho();
+        var mods = new Dictionary<int, List<Modification>>();
+        for (int oneBased = 1; oneBased <= Sequence.Length; oneBased++)
+        {
+            if (Sequence[oneBased - 1] == 'T')
+            {
+                mods[oneBased] = new List<Modification> { phospho };
+            }
+        }
+        Assert.That(mods, Is.Not.Empty, "the fixture must actually carry modifications");
+
+        var target = new Protein(Sequence, "P12345", oneBasedModifications: mods);
+        Protein entrapment = EntrapmentProteinGenerator.Create(target, Tryptic, NothingForbidden);
+
+        Assert.That(entrapment.OneBasedPossibleLocalizedModifications.Count, Is.EqualTo(mods.Count),
+            "every modification on a retained residue must survive the move");
+        foreach (int position in entrapment.OneBasedPossibleLocalizedModifications.Keys)
+        {
+            Assert.That(entrapment.BaseSequence[position - 1], Is.EqualTo('T'),
+                "a modification must land on the residue its motif requires");
+        }
+    }
+
+    [Test]
+    public void Create_DropsModificationsOnExcisedResidues()
+    {
+        // "SSSSSSR" has no partner and is long enough to be identified, so it is excised -- and any
+        // modification sitting on it goes with it rather than landing somewhere arbitrary.
+        const string withHomopolymer = "SYKALADQMNLLLSKSSSSSSRGGVDTTPFAWENDR";
+        ModificationMotif.TryGetMotif("S", out ModificationMotif motif);
+        var onExcisedTract = new Modification(_originalId: "Phospho", _modificationType: "Common Biological",
+            _target: motif, _locationRestriction: "Anywhere.", _monoisotopicMass: 79.96633);
+
+        var target = new Protein(withHomopolymer, "P12345", oneBasedModifications:
+            new Dictionary<int, List<Modification>> { { 17, new List<Modification> { onExcisedTract } } });
+
+        Protein entrapment = EntrapmentProteinGenerator.Create(target, Tryptic, NothingForbidden);
+
+        Assert.That(entrapment.BaseSequence, Does.Not.Contain("SSSSSSR"));
+        Assert.That(entrapment.OneBasedPossibleLocalizedModifications, Is.Empty);
+    }
+
+    [Test]
+    public void Create_GivesEachFoldItsOwnAccessionAndSequence()
+    {
+        var target = new Protein(Sequence, "P12345");
+
+        var entrapments = Enumerable.Range(0, 4)
+            .Select(k => EntrapmentProteinGenerator.Create(target, Tryptic, NothingForbidden, fold: k, foldCount: 4))
+            .ToList();
+
+        Assert.That(entrapments.Select(p => p.Accession).Distinct().Count(), Is.EqualTo(4));
+        Assert.That(entrapments.Select(p => p.BaseSequence).Distinct().Count(), Is.EqualTo(4));
+    }
+
+    // ---- pairing -----------------------------------------------------------
+
+    [Test]
+    public void Pairing_RecoversTheTargetPeptideFromCompositionAlone()
+    {
+        var target = new Protein(Sequence, "P12345");
+        Protein entrapment = EntrapmentProteinGenerator.Create(target, Tryptic, NothingForbidden);
+
+        var pairing = new EntrapmentPairing(target, Tryptic);
+
+        foreach (var peptide in entrapment.Digest(Tryptic, new List<Modification>(), new List<Modification>()))
+        {
+            string entrapmentPeptide = peptide.BaseSequence;
+            Assert.That(pairing.TryResolve(entrapmentPeptide, out string targetPeptide), Is.True,
+                "'" + entrapmentPeptide + "' should resolve to the target peptide it was built from");
+            Assert.That(string.Concat(targetPeptide.OrderBy(c => c)),
+                Is.EqualTo(string.Concat(entrapmentPeptide.OrderBy(c => c))));
+        }
+    }
+
+    [Test]
+    public void Pairing_NeedsNoSideFileBeyondTheTargetDatabase()
+    {
+        // The whole point of pairing on composition: a search reports a protein accession and a
+        // peptide sequence, and that is enough. Nothing has to be carried alongside the database.
+        var target = new Protein(Sequence, "P12345");
+        Protein entrapment = EntrapmentProteinGenerator.Create(target, Tryptic, NothingForbidden);
+
+        Assert.That(EntrapmentAccession.TryParse(entrapment.Accession, out string targetAccession, out _), Is.True);
+        Assert.That(targetAccession, Is.EqualTo(target.Accession));
+
+        var pairing = new EntrapmentPairing(target, Tryptic);
+        string first = entrapment.Digest(Tryptic, new List<Modification>(), new List<Modification>())
+            .First().BaseSequence;
+        Assert.That(pairing.TryResolve(first, out _), Is.True);
+    }
+
+    [Test]
+    public void Pairing_FlagsPeptidesOfIdenticalCompositionAsAmbiguous()
+    {
+        // P57071 really does contain LIHTGVK and LIHTVGK: same residues, same pinned K, different
+        // order. Composition cannot tell them apart, so neither is paired -- they are reported.
+        var target = new Protein("LIHTGVKAAADEFGHIKLIHTVGKMMWWYYPPQQR", "P57071");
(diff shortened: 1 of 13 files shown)


trishorts:
@acesnik — this carries the fixes for **every** comment across #1249–#1260. All twelve inline threads are answered and resolved, and the body-level points are appended to each PR's description as a *Review follow-up* section.

**Full suite 6,193 passed / 0 failed.**

Three of your findings changed my mind about something rather than just the code:

**The preventing-motif reading (#1254).** I had taken `CleavageSitePositions` at its name. It skips a prevented position entirely, so `trypsin|P` pins neither the K nor the P — and `trypsin|P` is an ordinary choice, not an exotic one. The guard is in; pinning across boundaries stays a rewrite.

**The run-collision count being wrong in both directions (#1259).** That they did not cancel is the part I would have missed. Returning on the first match counted pieces while documenting peptides; the missing length bound counted concatenations no search could report. Your `AAKAAR` case named it exactly. Both fixed, and that also settles 1,794 against 1,984 — they were measuring different things and neither was stated with its denominator.

**The excision delta you asked for is zero.** Rebuilt on the reviewed human proteome with everything in: `unpairableAllPermutationsTaken` and the new `unpairableRunCollisionsExhausted` are both 0. You were right that a collision delta alone could not rule out residues leaving the database, and it is worth having measured rather than assumed.

Two smaller things where you were right and I was not: the two tests that could not discriminate — the truncation-product loop ran on a collection the fix empties, and the position-2 fixture returned an `A` to slot 1 whether anchored or not. Both rewritten to assert what they were named for, the second asserting its own fixture stays discriminating.

Reviewing a fourteen-PR stack by diffing consecutive heads, and building the branch to confirm CS1587 rather than reasoning about it, is more care than I had any right to expect. Thank you.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,556 @@
...
+        return count;
+    }
+
+    /// <summary>The report as a tab-separated table, provenance in leading comment lines.</summary>
+    public string ToTabSeparated()
+    {
+        var text = new StringBuilder();
+        text.AppendLine($"# method\t{Provenance.Method}");
+        text.AppendLine($"# enzyme\t{Provenance.Enzyme}");
+        text.AppendLine($"# seed\t{Provenance.Seed}");
```
acesnik, line 297: Current culture, where every numeric column further down this method is explicitly `InvariantCulture`. Under sv-SE this writes `# seed` followed by `−7` with `U+2212`, into the line a consumer reads to regenerate the database — the same defect the seed row of this PR fixed one call away.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,474 @@
...
+    /// </remarks>
+    private static UniProtSequenceAttributes? DescribeSequence(Protein target, string entrapmentSequence)
+    {
+        UniProtSequenceAttributes? source = target.UniProtSequenceAttributes;
+        if (source is null)
+        {
+            return null;   // the target carried none, so inventing some would assert more than is known
+        }
+
+        var described = new UniProtSequenceAttributes(source.Length, source.Mass, source.Checksum,
```
acesnik, line 422: `source.Checksum` and `source.SequenceVersion` come across unchanged. The checksum is a CRC64 of the sequence, and it is both parsed (`ProteinXmlEntry.cs:195-212`) and written (`ProteinDbWriter.cs:515`), so every entrapment entry carries its target's CRC64 against a different sequence. Also: the `return null` branch above is unreachable — `Protein.cs:91` synthesizes attributes when null is passed — and it is the branch that would have written an honest `checksum="unknown"`.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
```diff
@@ -0,0 +1,474 @@
...
+            source.EntryModified, source.SequenceVersion, source.IsPrecursor, source.Fragment);
+        described.UpdateLengthAttribute(entrapmentSequence);
+        described.UpdateMassAttribute(entrapmentSequence);
```
acesnik, line 425: `PeptideWithSetModifications.MonoisotopicMass` returns 0 for a sequence containing `X` or `B` (`AAAXAAAK` and `AAABAAAK` both measured at 0; `U` and `O` are fine), so those entries go from a valid inherited mass to `mass="0"`. Separately this is monoisotopic where UniProt's `mass` is average molecular weight, so target and entrapment entries in one file are on different scales — house convention already, so noting rather than asking.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,495 @@
...
+    /// <para>A <b>preventing</b> motif pins nothing at all: <c>CleavageSitePositions</c> skips a
+    /// position where a motif fits but is prevented, so under <c>trypsin|P</c>'s <c>K[P]|</c> neither
+    /// the K nor the P after it is held. Moving that P away invents a cleavage site the target
+    /// lacked; moving another P next to another K destroys one. That is five shipped agents --
+    /// <c>chymotrypsin|P</c>, <c>elastase|P</c>, <c>Lys-C|P</c>, <c>trypsin|P</c> and
+    /// <c>subtilisin|P</c> -- and <c>trypsin|P</c> is an ordinary choice in real work.</para>
+    /// <para>A <b>multi-residue</b> motif has its span cut by the piece boundary: <c>TX|T</c> cuts
+    /// after the second of three residues, so one piece ends <c>…TX</c> and the next begins
+    /// <c>T…</c>, neither fragment matches inside its own piece, and nothing is held at the seam.
+    /// That covers <c>StcE</c> and <c>StcE-trypsin</c>.</para>
```
acesnik, line 314: There is no agent named `StcE` in `proteases.tsv` — only `StcE-trypsin` (line 117). The seventh agent with a preventing or multi-residue motif is `collagenase` (`GPX|GPX`, line 116), which the guard does refuse, because it keys on the motif rather than the name. That is the right way round; the prose just names the wrong two.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,495 @@
...
+    public static EntrapmentAssembly Assemble(string targetSequence, IDigestionParams digestionParams,
+        IReadOnlySet<string> forbiddenSequences, int fold = 0, int foldCount = 1, int seed = 1)
+    {
+        if (string.IsNullOrEmpty(targetSequence))
+        {
+            throw new MzLibException("Cannot assemble an entrapment sequence from an empty sequence.");
+        }
+
+        List<DigestionMotif> motifs = digestionParams.DigestionAgent.DigestionMotifs;
+        RefuseAgentsWhoseSitesCannotBeHeld(digestionParams.DigestionAgent.Name, motifs);
```
acesnik, line 168: This re-checks a parameter-level invariant once per protein per fold, and throws from inside the run rather than at the call. It fires on the first protein so nothing is wasted, but a throw from this depth surfaces to a GUI user as a crash rather than as "pick a different agent" — `GenerateEntrapment` is the natural place to refuse. `EntrapmentPeptideGenerator.Create` and `EntrapmentPairing` are still unguarded for a caller passing motifs directly.

acesnik (commented):
🤖 Twelve review comments across eight branches, answered on the one that keeps the `ProteinDbLoader` edit at the top of the stack. I checked each row against the two heads rather than reading the diff — building `UsefulProteomicsDatabases` at #1264 and at this head and running the same probe against both — and ten are fixed. Three of those I would have wanted measured this way regardless: the culture test genuinely differs (`(-7).ToString()` renders `U+2212` under sv-SE, fi-FI and lt-LT on this machine, and the entrapment sequence is byte-identical across all four here where it was not at #1264), the guard throws for all six shipped agents where #1264 threw for none, and `CS1587` at `ProteinDbLoader.cs(85,9)` is present in a clean build at #1264 and gone here with `CA2202` back on `LoadProteinXML`.

**Three of these change the emitted database, and I would put that at the top of the body rather than leaving it implied.** The anchors move any entry whose first base piece is a single residue: `"K" + "SYALADQMNLLLSK" + "GGVDTTPFAWENDR"` gives `KLLYADMAQNLSSLK…` at #1264 and `KSYLLQLAASLMDNK…` here. The length bound moves any piece whose below-`MinLength` run was forbidden: on `"AK" + "ABCK" + "LLLLLLLLLLLR"` with the 6-mer `AKACBK` forbidden, piece 1 goes `BACK` → `ACBK`. And the new `length`/`mass` attributes change the file directly. On top of that the TSV gains two columns mid-row — `unpairableRunCollisionsExhausted` at position 8, `entrapmentSearchSpacePeptides` at position 10 — which breaks a positional parse.

The seed fix then stops one line short of the thing it is for. `EntrapmentReport.cs:297` is `$"# seed\t{Provenance.Seed}"` — plain interpolation, current culture, while every numeric *column* in that same method is explicitly `InvariantCulture`. So `# seed<TAB>−7` with `U+2212` under sv-SE, in the record a consumer uses to regenerate the database. Same defect as the row above it, one line away.

`DescribeSequence` has the same shape: it fixes length and mass and copies `Checksum` straight across. UniProt's checksum is a CRC64 of the sequence, parsed on read (`ProteinXmlEntry.cs:195-212`) and written back out (`ProteinDbWriter.cs:515`), so on a UniProt-derived database every entrapment entry still carries its target's checksum against a different sequence — the self-inconsistency this row set out to remove, one attribute over. `SequenceVersion` is inherited too. Worth knowing the null branch above it is unreachable: `Protein.cs:91` synthesizes attributes when null is passed, so `target.UniProtSequenceAttributes` never is — and that unreachable path is the one that would have produced an honest `checksum="unknown"`. The recomputed mass has its own edge, since `PeptideWithSetModifications.MonoisotopicMass` returns 0 for a sequence containing `X` or `B`: those entries go from a valid inherited mass to `mass="0"`, where before they inherited the target's. It is also monoisotopic where UniProt's `mass` is average, so a written file now has the two arms on different scales. That last one is the house convention already, so I mention it rather than ask for it.

On 1,794 against 1,984 — the body says the change should settle it and does not say to what, and `EntrapmentAssembler.cs:110` still carries 1,794 as a measurement this PR invalidates. A single new figure will not discriminate either, which is the more useful half: if the caller's `forbiddenSequences` is built from a digestion at the same `MinLength`, no sub-`MinLength` run can be in the set, the length bound is inert, and the whole gap was the first-match half. Nothing in mzLib builds that set, so only the consumer's side can tell you which — two numbers from two callers would, one number from one will not.

Coverage for the new behaviour is thin in a way that matters for a PR about counts being wrong. `RunCollisionsExhaustedTheSpace` appears once in the test tree, as a column-index assertion satisfied by `0 == 0`; nothing drives it non-zero. No assertion on `UnrepairableRunCollisions` is ever 2, and none uses a run shorter than `MinLength`, so neither direction of the count fix has a regression test. Nothing anywhere asserts the partner's sequence attributes. And the `Create_NeverLeavesAnAnnotationPointingPastTheSequence` rewrite is right about the contract — the four `Is.Empty` assertions bite — but the range assertion under them does not: the fixture carries no modifications, so all four collections feeding `everyPosition` are empty and the sequence has 0 elements. The new proteoform sibling asserts `Is.Not.Empty` first, which is the shape the original wanted.

The `Ordinal` change has a loose end of the same kind. `ProteinDbLoader.cs:374` and `:379`, and `ProteinXmlEntry.cs:437` and `:442`, still match the entrapment identifier `OrdinalIgnoreCase` (and with `IndexOf` rather than a prefix test), so `RANDOM_P12345_f0` is now rejected by `EntrapmentAccession.TryParse` while the loader still treats it as entrapment. Those lines are outside this diff, so it is a question of which side is authoritative rather than something to fix here.

Smaller things. The multi-residue prose at `EntrapmentAssembler.cs:314` names `StcE`, which is not an agent in `proteases.tsv` — the seventh affected agent is `collagenase` (`GPX|GPX`, line 116), which the guard does catch, since it keys on the motif rather than the name. That is the right way round; only the prose is wrong. `RunCollisionsExhaustedTheSpace` was inserted ahead of `NoPermutationExists` rather than appended, renumbering the rest of a public enum on an assembly that ships on NuGet. `EntrapmentAssembler.cs:456` adds a `CS8600` and `EntrapmentProteinGenerator.cs:307` a `CS0419`, worth a glance given the row above is about removing a warning. And `CountSearchablePeptides` builds a throwaway `Protein` per protein per fold, paying a monoisotopic-mass computation it never uses — 37 ms per 2,000 400-mers against 0 when attributes are supplied.

I could not check any real-database figure — 1,794 / 1,984, 2,720 / 2,636, 20,416 / 20,415, or how many reviewed-human entries begin with K or R, which is the size of the anchor-driven output change. No proteome in the checkouts. And I could not run the 6,192-test suite: `Test/Test.csproj` is `net*.0-windows` via `mzPlot`, so everything above came from a throwaway net10.0 harness.

mzLib/Test/DatabaseTests/TestEntrapmentProtein.cs
```diff
@@ -0,0 +1,1017 @@
...
+        Assert.That(entrapment.DisulfideBonds, Is.Empty);
+        Assert.That(entrapment.SpliceSites, Is.Empty);
+
+        // And the property that mattered, stated over everything the entry actually carries.
+        IEnumerable<int> everyPosition = entrapment.TruncationProducts
+            .Where(t => t.OneBasedEndPosition.HasValue).Select(t => t.OneBasedEndPosition.Value)
+            .Concat(entrapment.DisulfideBonds.Select(b => b.OneBasedEndPosition))
+            .Concat(entrapment.SpliceSites.Select(x => x.OneBasedEndPosition))
+            .Concat(entrapment.OneBasedPossibleLocalizedModifications.Keys);
+        Assert.That(everyPosition, Is.All.LessThanOrEqualTo(entrapment.BaseSequence.Length));
```
acesnik, line 268: Vacuous as written — the fixture carries no modifications, so all four collections feeding `everyPosition` are empty and this asserts over 0 elements. The four `Is.Empty` assertions above do bite. The new proteoform sibling gets this right by asserting `Is.Not.Empty` first.

acesnik:
Could you convert the others to drafts now that this is the one to review / is the superset of fixes that you'd like to merge?

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: Fixed in #1267 -- every numeric provenance row goes through `InvariantCulture`, not just the seed. Reproduced under `sv-SE`: the row was written with U+2212 and no ASCII hyphen.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Fixed in #1267. Checksum is written `"unknown"` and the sequence version unset, which is what `Protein`'s own constructor synthesises when handed nothing. Your reading of the unreachable branch is right and I confirmed it: a constructed `Protein` never has null attributes, so the branch that would have written the honest value could not fire.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentProteinGenerator.cs
trishorts: Fixed in #1267, with one correction offered in the right spirit: `MonoisotopicMass` returns **NaN**, not 0. `(int)Math.Round(NaN)` is 0, so the `mass="0"` you saw is exactly right -- the mechanism is just one step further back, which matters because the guard has to test for non-finite rather than for zero. Measured: `AAAXAAAK` and `AAABAAAK` both NaN, `U` and `O` fine, and `UpdateMassAttribute("AAAXAAAK")` storing 0.

A rearrangement is isomeric, so the target's own mass is kept when the recomputation is not finite. Noted on the monoisotopic-vs-average point; left as house convention as you suggested.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Fixed in #1267 -- the prose names `collagenase` now. Confirmed against `proteases.tsv` on this branch: no bare `StcE` row, `collagenase` (`GPX|GPX`) at line 116 and `StcE-trypsin` at 117. As you say the guard keys on the motif rather than the name, so behaviour was already right and it will also refuse the bare `StcE` that arrives separately -- the prose is written so it stays true either way.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: Partly done in #1267. `GenerateEntrapment` refuses up front, so a caller hears it at the call they made rather than as a crash from inside the run; the assembler keeps its own check as a backstop for direct callers.

`EntrapmentPeptideGenerator.Create` and `EntrapmentPairing` are still unguarded, and I have left them that way deliberately rather than quietly widening this PR. Called out under **Not addressed** in the description.

mzLib/Test/DatabaseTests/TestEntrapmentProtein.cs
trishorts: Fixed in #1267. The fixture carries two modifications now and the test asserts the collection is non-empty before asserting over it, so `Is.All` can actually fail. You were right that the proteoform sibling had the shape the original should have had.

trishorts:

All resolved, fixes in #1267 — invariant-culture provenance rows, no inherited CRC64, no `mass="0"` from a NaN, `collagenase` named correctly, the guard hoisted to `GenerateEntrapment`, and the vacuous test given something to bite on. One correction: `MonoisotopicMass` returns **NaN** rather than 0 — `(int)Math.Round(NaN)` is 0, so what you saw was right, the mechanism is one step further back.

trishorts:

PR #1267, opened 2026-09-01
Entrapment: answer the second review round

trishorts:
Answers the second review round — the 17 threads opened across #1261, #1262, #1263, #1264 and #1265. Stacked on `feat/entrapment-review-fixes` (#1265); the only new work is here.

**Every finding was reproduced in a throwaway test before it was acted on, and all 17 held.** One correction, offered in the right spirit: `PeptideWithSetModifications.MonoisotopicMass` returns **NaN** for a sequence holding `X` or `B`, not 0 — but `(int)Math.Round(NaN)` is 0, so the `mass="0"` you observed is exactly right.

### #1261 — the exclusion table

- **Rows are keyed by the entrapment accession now**, `Random_<target>_f<fold>`, not by the target. You were right that `accession` meant one thing on those rows and another on the `ambiguous` ones, and that a consumer filtering on (accession, peptide) matched the ambiguous rows and silently missed these. It under-excludes, which inflates an FDP estimate.
- **The stratified column counts peptides, as the sidecar does, not placements.** Measured on the reviewed human database: the column read **2,048** against **1,983** sidecar rows. The `ambiguous` half already agreed at 736 on both sides, which is what made the mismatch legible.
- The `AddRange` you flagged at 386 was already fixed higher in the stack, by the `Contains` dedup in #1265 — but re-keying by fold removes the cross-fold question at the root.

### #1262 — the foreign arm

- `Format` **refuses a target accession beginning with the marker**, so the two namespaces are actually disjoint and the comment on `TryParse`'s own refusal is now true.
- `AddForeign` **merges** rather than assigns. You were right that assigning while the entry count accumulates meant two calls on one accession counted both entries and kept only the second call's peptides.
- `Build` returns a **snapshot**, like the two dictionaries beside it. A later `AddForeign` no longer mutates an already-built report.
- **Foreign entries reach the provenance.** `# foreignEntries` and `# foreignPeptidesSharedWithTarget` are emitted only when the arm was used, so a bottom-up report is byte-identical to before.

### #1263 — proteoform mode *(marked draft; bottom-up is the current focus, but the findings are fixed)*

- **The rearrangement is confined to truncation-span boundaries.** Your measurement was the decisive one: rearranging the whole protein preserved only the whole protein's composition, and a span held a different multiset. Since every span is a union of whole segments, cutting at every span boundary and permuting within each keeps every span isomeric with the one it replaces — the same shape as cutting at cleavage sites in bottom-up.
- **Span collisions are tested and counted**, so the two figures are no longer constants and a proteoform report no longer reads zero however much it holds.
- **A partner identical to its target is dropped** instead of emitted. Defensible for a sub-`MinLength` base piece in bottom-up; not here, where a short proteoform is itself a searched species.
- **Splice sites are dropped rather than mapped.** Your reasoning stands: a bond joins two residues and survives independent mapping; a junction between adjacent residues does not.

### #1264 — the empty-protein change

- The comment now says what you asked it to: **the loaded database is unchanged**, and what actually goes away is the loader's `"N empty entries ignored"` warning, which was the only signal the arm was short.
- **`EntrapmentReport.EntriesYieldingNoPartner`** replaces it, emitted in the provenance when non-zero — short by construction, said where it is decided.
- The collision count and the sidecar now agree (see #1261 above).

### #1265 — the review round itself

- **Provenance rows go through `InvariantCulture`**, like every numeric column below them. Reproduced under `sv-SE`: `# seed` was written with U+2212 and no ASCII hyphen.
- **Partners no longer inherit their target's CRC64 checksum or sequence version.** Both are written as `"unknown"` / unset, which is what `Protein`'s own constructor synthesises when handed nothing — and you were right that the `return null` branch was unreachable for exactly that reason.
- **A NaN mass no longer writes `mass="0"`.** A rearrangement is isomeric, so the target's mass is the right answer.
- **The guard prose names `collagenase`**, not `StcE`. Confirmed against `proteases.tsv` on this branch: there is no bare `StcE` row, and `collagenase` (`GPX|GPX`, line 116) is the seventh. The guard keys on the motif, so it will also refuse the bare `StcE` mzLib gains separately.
- **The guard also refuses at `GenerateEntrapment`**, so a caller hears it at the call they made rather than as a crash from inside the run.
- **The vacuous test now carries modifications** and asserts the collection is non-empty before asserting over it.

### Not addressed

`EntrapmentPeptideGenerator.Create` and `EntrapmentPairing` remain unguarded for a caller passing motifs directly. The guard is hoisted to the whole-database entry point and still fires inside the assembler, which covers every path this project uses; guarding those two as well is a wider change and is left for a follow-up rather than smuggled in here.

---

**375 tests, 370 passing, 5 skipped** across the database, decoy and entrapment suites.

https://claude.ai/code/session_01MLa2EL9kF2er4CASwRnPvz

### DIFF
--- a/mzLib/Test/DatabaseTests/TestDecoyProteinValidator.cs
+++ b/mzLib/Test/DatabaseTests/TestDecoyProteinValidator.cs
@@ -4,6 +4,7 @@ using Proteomics.ProteolyticDigestion;
 using Proteomics;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 using MzLibUtil;
 using UsefulProteomicsDatabases;
 using Transcriptomics.Digestion;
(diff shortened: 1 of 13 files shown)


trishorts:

PR #1271, opened 2026-09-01
Add a managed entrapment database generator

trishorts:
> **Project of origin:** `entrapment`

Gives mzLib a managed entrapment-database generator: from any target protein database and any protease, an **r-fold entrapment database whose peptides are isomeric with their targets**, deterministically paired to them **without a side file**, and **honestly reported** where the construction cannot be achieved. No external executable, no reflection.

### Why one PR

This consolidates work previously split across #1253–#1265 and #1267. That stack was reviewed in two rounds by @acesnik — every finding is folded in here — but a 15-PR chain has a real hazard: it is merged in order and can be stopped at any point, and the intermediate heads carried defects that were only fixed further up. A reviewer merging three PRs and pausing would have landed known-broken code.

**This branch is self-sufficient.** It is cut from current master, it builds, its tests pass, and it does not depend on any later PR to be correct.

### What it contains

| | |
|---|---|
| `EntrapmentPeptideGenerator` | one peptide's partner, by deterministic unranking of the multiset permutation space — a pure function of (peptide, seed), reproducible across runs, threads and machines, and parallel-safe. Emptiness is *proven*, not discovered by a retry budget. |
| `EntrapmentAssembler` | a protein's partner: cleavage sites pinned so it digests into the same pieces, termini anchored so terminally-restricted modifications survive, missed-cleavage runs collision-checked. |
| `EntrapmentProteinGenerator` | a database entry's partner, modifications moved onto the residues they were on, one partner per **entry** rather than per loaded protein. |
| `EntrapmentAccession` | `Random_<target>_f<fold>`, parsed back safely; the foreign form is refused as a partner rather than mis-parsed. |
| `EntrapmentPairing` | the partner recovered from accession and sequence alone, by composition plus pinned pattern. |
| `EntrapmentReport` | the QC report as a first-class output — the distribution rather than a summary, failures kept apart by cause, and the peptides a consumer must exclude **named individually** rather than counted. |

### Measured, not assumed

Reviewed human proteome, trypsin, MC ≤ 2: **2,591,131** distinct peptides in **114 s**; 99.77% first-probe success; achieved r **8.998** at r = 9; pairing unambiguous for **99.979%** of base peptides; end-to-end retention **99.99%**. Searched at scale with MetaMorpheus across twelve fractions of PXD001468, with every discovery paired back to its target with nothing carried alongside the file.

Three findings contradicted the design assumptions and are documented where they bite:

- **Mucins do not fail** — zero unpairable across 4,770 peptides from all 20 MUC genes. The real failure mode is homopolymeric poly-S/poly-T, and site-rich peptides turn out to be the *easiest* to entrap.
- **Achieved r degrades at `n` = 0**, the opposite end from the one predicted.
- **`DigestionMotif` cannot express "glycosylated"**, so a mucinase encoded as a sequence motif over-digests.

### Known limit, counted rather than hidden

**1,794 peptide sequences (0.065%)** appear in both the target and entrapment digests of the written database, because the run's final piece has exactly one arrangement and there is no candidate to move to. They are named in `<db>_entrapment_exclusions.tsv` so a consumer can subtract them. Repairing would mean backtracking into a placed piece or excising a good one.

### Blast radius

The only existing code touched is `ProteinDbLoader.DefaultEntrapmentIdentifier` — a duplicated literal replaced by a name with the same value — and a clarified `<remarks>` on `DecoySequenceValidator.PermutationSpaceSize`. Everything else is additive.

**498 tests passing** across the database, decoy, entrapment and protein-reader suites.

Proteoform (top-down) mode follows in a separate PR, opened as a draft.

https://claude.ai/code/session_01MLa2EL9kF2er4CASwRnPvz

### DIFF
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -33,13 +33,15 @@ public enum PieceOutcome
 public sealed class EntrapmentPiece
 {
     internal EntrapmentPiece(int index, string targetPiece, string? entrapmentPiece,
-        PieceOutcome outcome, EntrapmentFailure failure)
+        PieceOutcome outcome, EntrapmentFailure failure, int targetStart, int entrapmentStart)
     {
         Index = index;
         TargetPiece = targetPiece;
         EntrapmentPiece_ = entrapmentPiece;
         Outcome = outcome;
         Failure = failure;
+        TargetStart = targetStart;
+        EntrapmentStart = entrapmentStart;
     }
 
     /// <summary>Ordinal of this piece within the target, counted before anything was excised.</summary>
@@ -51,6 +53,21 @@ public sealed class EntrapmentPiece
     public string? EntrapmentPiece_ { get; }
 
     public PieceOutcome Outcome { get; }
+
+    /// <summary>Zero-based offset of this piece in the TARGET sequence.</summary>
+    public int TargetStart { get; }
+
+    /// <summary>
+    /// Zero-based offset of this piece in the ENTRAPMENT sequence, or -1 when it was excised.
+    /// </summary>
+    /// <remarks>
+    /// Not the same as <see cref="TargetStart"/> once anything earlier has been excised, and not
+    /// recoverable from <see cref="Index"/> either. Carried because every caller that wants to look
+    /// at a piece in its protein's coordinates -- to ask whether a modification still fits at its
+    /// position within the peptide, say -- otherwise has to re-derive it by re-running the digestion
+    /// and accumulating lengths, which two separate analysis harnesses did before this existed.
+    /// </remarks>
+    public int EntrapmentStart { get; }
 
     /// <summary>Why no partner was found, when <see cref="Outcome"/> is not <see cref="PieceOutcome.Permuted"/>.</summary>
     public EntrapmentFailure Failure { get; }
@@ -210,6 +227,9 @@ public static class EntrapmentAssembler
             int start = sites[index];
             int length = sites[index + 1] - start;
             string piece = targetSequence.Substring(start, length);
+            // Read before anything is appended, so it is where this piece WILL start rather than
+            // where the next one will.
+            int entrapmentStart = entrapment.Length;
 
             Func<string, IReadOnlyList<string>>? completesAForbiddenRun =
                 RejectRunCollisions(placed, digestionParams.MaxMissedCleavages, forbiddenSequences,
@@ -234,7 +254,7 @@ public static class EntrapmentAssembler
                 // A permuted piece was chosen with the run test applied, so this cannot fire -- it is
                 // asserted rather than assumed only because the count below must mean one thing.
                 pieces.Add(new EntrapmentPiece(index, piece, partner.EntrapmentSequence,
-                    PieceOutcome.Permuted, EntrapmentFailure.None));
+                    PieceOutcome.Permuted, EntrapmentFailure.None, start, entrapmentStart));
                 retainedTargetIndices.Add(index);
                 continue;
             }
@@ -265,14 +285,15 @@ public static class EntrapmentAssembler
                     initiatorMethionineCollisions.Add(strippedCollision);
                 }
                 pieces.Add(new EntrapmentPiece(index, piece, piece,
-                    PieceOutcome.KeptVerbatimTooShort, partner.Failure));
+                    PieceOutcome.KeptVerbatimTooShort, partner.Failure, start, entrapmentStart));
                 retainedTargetIndices.Add(index);
                 continue;
             }
 
             // Long enough to be identified, and no partner exists: drop it. Its positions stay
             // unmapped, which is what marks them excised.
-            pieces.Add(new EntrapmentPiece(index, piece, null, PieceOutcome.Excised, partner.Failure));
+            pieces.Add(new EntrapmentPiece(index, piece, null, PieceOutcome.Excised, partner.Failure,
+                start, -1));
         }
 
         int broken = CountRunsSpanningAGap(retainedTargetIndices, digestionParams.MaxMissedCleavages);
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -62,7 +62,8 @@ public sealed class EntrapmentAssembly
     internal EntrapmentAssembly(string targetSequence, string entrapmentSequence,
         int[] targetToEntrapmentPosition, IReadOnlyList<EntrapmentPiece> pieces,
         int missedCleavagePeptidesSpanningAnExcision,
-        IReadOnlyList<string> unrepairableRunCollisionPeptides)
+        IReadOnlyList<string> unrepairableRunCollisionPeptides,
+        IReadOnlyList<string> initiatorMethionineCollisionPeptides)
     {
         TargetSequence = targetSequence;
         EntrapmentSequence = entrapmentSequence;
@@ -70,6 +71,7 @@ public sealed class EntrapmentAssembly
         Pieces = pieces;
         MissedCleavagePeptidesSpanningAnExcision = missedCleavagePeptidesSpanningAnExcision;
         UnrepairableRunCollisionPeptides = unrepairableRunCollisionPeptides;
+        InitiatorMethionineCollisionPeptides = initiatorMethionineCollisionPeptides;
     }
 
     public string TargetSequence { get; }
@@ -125,6 +127,27 @@ public sealed class EntrapmentAssembly
     /// </remarks>
     public IReadOnlyList<string> UnrepairableRunCollisionPeptides { get; }
 
+    /// <summary>
+    /// Peptides that are a real target peptide because a search removes the entrapment sequence's
+    /// INITIATOR METHIONINE, and no arrangement of the opening piece avoided it.
+    /// </summary>
+    /// <remarks>
+    /// <para>Digestion emits the opening piece twice, with and without the initiator methionine, so
+    /// a piece that is distinct from every target peptide can still have an M-stripped form that is
+    /// not. Every other piece is checked as itself and that is enough; this one is checked as itself
+    /// and as itself minus a residue.</para>
+    /// <para>The route survives a build at <c>MaxMissedCleavages = 0</c>, which deletes the run
+    /// route entirely, so it is the only way a real target peptide reaches the entrapment set there.
+    /// Measured on the reviewed human proteome before this check existed: two peptides under Arg-C
+    /// and two under Glu-C, none under Asp-N -- 0.0005% -- and, because a run collision was the only
+    /// reason the exclusion sidecar could name, they appeared in no list at all. A count of zero
+    /// here is the statement that the check ran, which an absent column is not.</para>
+    /// </remarks>
+    public IReadOnlyList<string> InitiatorMethionineCollisionPeptides { get; }
+
+    /// <summary>How many of <see cref="InitiatorMethionineCollisionPeptides"/> there are.</summary>
+    public int InitiatorMethionineCollisions => InitiatorMethionineCollisionPeptides.Count;
+
     public int ExcisedCount => Pieces.Count(p => p.Outcome == PieceOutcome.Excised);
 
     public int KeptVerbatimCount => Pieces.Count(p => p.Outcome == PieceOutcome.KeptVerbatimTooShort);
@@ -178,6 +201,7 @@ public static class EntrapmentAssembler
         // sequence, so a candidate can be tested against the runs it would complete.
         var placed = new List<string>(sites.Count - 1);
         var unrepairable = new List<string>();
+        var initiatorMethionineCollisions = new List<string>();
         int[] map = Enumerable.Repeat(-1, targetSequence.Length).ToArray();
         var retainedTargetIndices = new List<int>(sites.Count - 1);
 
@@ -191,9 +215,17 @@ public static class EntrapmentAssembler
                 RejectRunCollisions(placed, digestionParams.MaxMissedCleavages, forbiddenSequences,
                     digestionParams.MinLength, digestionParams.MaxLength);
 
+            // Keyed on what has actually been placed rather than on the piece's ordinal, because an
+            // excised opening piece promotes the next one -- the entrapment protein then begins
+            // wherever the first surviving piece begins, and that is the residue a search would
+            // treat as an initiator methionine.
+            Func<string, string?>? strippedOfInitiatorMethionine =
+                RejectInitiatorMethionineCollisions(entrapment.Length == 0, forbiddenSequences,
+                    digestionParams.MinLength, digestionParams.MaxLength);
+
             EntrapmentPeptide partner = EntrapmentPeptideGenerator.Create(piece, motifs, forbiddenSequences,
                 fold, foldCount, seed, TerminalAnchors(start, length, targetSequence.Length),
-                completesAForbiddenRun is null ? null : c => completesAForbiddenRun(c).Count > 0);
+                Reject(completesAForbiddenRun, strippedOfInitiatorMethionine));
 
             if (partner.Succeeded)
             {
@@ -222,6 +254,16 @@ public static class EntrapmentAssembler
                 {
                     unrepairable.Add(collision);
                 }
+                // Same rule for the initiator-methionine route, and for the same reason: a piece
+                // kept verbatim has no alternative to move to, so the collision is named rather than
+                // repaired. It cannot fire while MinLength is respected -- the stripped form is one
+                // residue shorter than a piece that was already too short -- and it is written
+                // because "cannot fire" is a claim the count is entitled to check.
+                string? strippedCollision = strippedOfInitiatorMethionine?.Invoke(piece);
+                if (strippedCollision is not null)
+                {
+                    initiatorMethionineCollisions.Add(strippedCollision);
+                }
                 pieces.Add(new EntrapmentPiece(index, piece, piece,
                     PieceOutcome.KeptVerbatimTooShort, partner.Failure));
                 retainedTargetIndices.Add(index);
@@ -236,7 +278,69 @@ public static class EntrapmentAssembler
         int broken = CountRunsSpanningAGap(retainedTargetIndices, digestionParams.MaxMissedCleavages);
 
         return new EntrapmentAssembly(targetSequence, entrapment.ToString(), map, pieces, broken,
-            unrepairable);
+            unrepairable, initiatorMethionineCollisions);
+    }
+
+    /// <summary>
+    /// The two per-candidate tests as one, or null when neither applies.
+    /// </summary>
+    /// <remarks>
+    /// Composed rather than chained inside the generator so that the generator keeps seeing a single
+    /// "is this candidate acceptable in context" predicate. Both tests are about what a candidate
+    /// becomes once something else is taken into account -- its neighbours, or a residue a search
+    /// removes -- and neither is a property of the candidate alone.
+    /// </remarks>
+    private static Func<string, bool>? Reject(Func<string, IReadOnlyList<string>>? completesAForbiddenRun,
+        Func<string, string?>? strippedOfInitiatorMethionine)
+    {
+        if (completesAForbiddenRun is null && strippedOfInitiatorMethionine is null)
+        {
+            return null;
+        }
+
+        return candidate =>
+            (completesAForbiddenRun is not null && completesAForbiddenRun(candidate).Count > 0)
+            || (strippedOfInitiatorMethionine is not null && strippedOfInitiatorMethionine(candidate) is not null);
+    }
+
+    /// <summary>
+    /// A test returning the real target peptide a candidate would become once a search removes its
+    /// initiator methionine, or null if it would become none.
+    /// </summary>
+    /// <remarks>
+    /// <para>Applies only to the piece that will OPEN the entrapment protein, and only when that
+    /// piece begins with methionine, because those are the only conditions under which a search
+    /// emits a shortened form at all (<c>InitiatorMethionineBehavior.Variable</c>, the default).
+    /// Every other piece is fully covered by testing the candidate itself.</para>
+    /// <para>The length bound is the same one <see cref="RejectRunCollisions"/> applies, and for the
+    /// same reason: the pairing index counts a peptide as searchable only within
+    /// <c>[MinLength, MaxLength]</c>, so testing outside that range would reject candidates over
+    /// sequences no search can report and put the exclusion list on different footing from the
+    /// population it is meant to be subtracted from.</para>
+    /// </remarks>
+    private static Func<string, string?>? RejectInitiatorMethionineCollisions(bool willOpenTheProtein,
+        IReadOnlySet<string> forbiddenSequences, int minLength, int maxLength)
+    {
+        if (!willOpenTheProtein)
+        {
+            return null;
+        }
+
+        return candidate =>
+        {
+            if (candidate.Length == 0 || candidate[0] != 'M')
+            {
+                return null;
+            }
+
+            string stripped = candidate.Substring(1);
+            if (stripped.Length < minLength || stripped.Length > maxLength)
+            {
+                return null;
+            }
+
+            return forbiddenSequences.Contains(stripped) ? stripped : null;
+        };
     }
 
     /// <summary>
@@ -291,24 +395,43 @@ public static class EntrapmentAssembler
     }
 
     /// <summary>
-    /// Positions within a piece that must not move because they are the PROTEIN's termini.
+    /// Positions within a piece that must not move: the PROTEIN's termini, and the PIECE's own.
     /// </summary>
     /// <remarks>
-    /// <para>A modification can be restricted to the N- or C-terminus, and a rearrangement that
-    /// moved that residue away would make it invalid for its location -- mzLib then drops it, and
-    /// nothing counts the loss. Measured before anchoring: 3,946 N-terminal and 31 C-terminal
-    /// modifications lost across the human proteome, 2.4% of all of them.</para>
-    /// <para>Both of the first two positions are held, not just the first: a modification annotated
-    /// after initiator-methionine cleavage sits on the second residue.</para>
+    /// <para>A modification can be restricted to a terminus, and a rearrangement that moved that
+    /// residue away makes it invalid for its location. Measured before protein anchoring: 3,946
+    /// "N-terminal." and 31 "C-terminal." modifications lost across the human proteome, 2.4% of all
+    /// of them. Both of the protein's first two positions are held, not just the first: a
+    /// modification annotated after initiator-methionine cleavage sits on the second residue.</para>
+    /// <para><b>The piece's own termini are held for a restriction that is not protein-level:
+    /// "Peptide N-terminal." and "Peptide C-terminal.", which are satisfied per digestion product
+    /// rather than once per entry.</b> That case fails far more quietly than the protein one, and in
+    /// both directions at once. <c>ModificationLocalization.ModFits</c> is called with the
+    /// digestion-product index set to zero while a protein is being built, so a peptide-level
+    /// restriction falls through to "fits" and the annotation is transported and written; it is
+    /// evaluated for real only at digestion, where the modification no longer fits its new position
+    /// and the peptidoform hypothesis simply does not exist. The in-count equals the out-count, no
+    /// counter moves, and the entrapment peptide has silently lost a hypothesis its target still
+    /// has. Measured on the reviewed human proteome at MaxMissedCleavages 0: without this anchor
+    /// 69.00% of Arg-C pieces and 65.42% of Glu-C pieces move their first residue, and 72.68% of
+    /// Asp-N pieces move their last -- complementary, because for a C-terminal cutter the last
+    /// residue IS the pinned cleavage residue and for an N-terminal cutter the first one is.</para>
+    /// <para>The cost was measured before it was paid, because the obvious guess is that holding two
+    /// more positions on every piece is much dearer than holding three per protein. It is not:
+    /// pieces with no rearrangement at all rise from 32 to 52 (Arg-C), 36 to 56 (Glu-C) and 54 to 72
+    /// (Asp-N) out of roughly 400,000 searchable pieces -- 0.0075% to 0.012% -- and the mean natural
+    /// log of the permutation space falls by about two, from 37-51.</para>
     /// <para>This is unconditional -- the termini are anchored whether or not anything is actually
     /// modified there. Anchoring reactively would make the entrapment sequence a function of
     /// (sequence, seed, modifications) rather than (sequence, seed), so two databases built from the
     /// same proteome with different annotations would disagree on their sequences and the
-    /// determinism the pairing rests on would quietly weaken. The cost is two or three positions per
-    /// protein, against permutation spaces in the millions.</para>
+    /// determinism the pairing rests on would quietly weaken.</para>
     /// </remarks>
     private static int[]? TerminalAnchors(int pieceStart, int pieceLength, int proteinLength)
     {
+        // The piece's own termini, in piece coordinates already.
+        var anchors = new HashSet<int> { 0, pieceLength - 1 };
+
         // Protein coordinates, translated into piece coordinates -- not piece coordinates that
         // happen to coincide. When the opening piece is a single residue the protein's second
         // residue lives in the NEXT piece, and computing "the first two" per piece never held it, so
@@ -319,17 +442,20 @@ public static class EntrapmentAssembler
             ? new[] { 0, 1, proteinLength - 1 }
             : new[] { 0 };
 
-        List<int>? anchors = null;
         foreach (int position in protein)
         {
             int within = position - pieceStart;
             if (within >= 0 && within < pieceLength)
             {
-                (anchors ??= new List<int>(3)).Add(within);
+                anchors.Add(within);
             }
         }
 
-        return anchors?.ToArray();
+        // Ordered, because this becomes part of the pinned-position pattern that the pairing key
+        // compares and a set's enumeration order is not a contract.
+        int[] ordered = anchors.ToArray();
+        Array.Sort(ordered);
+        return ordered;
     }
 
     /// <summary>
--- a/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
+++ b/mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
@@ -1,0 +1,441 @@
+﻿#nullable enable
+using System;
+using MzLibUtil;
+using Omics.Digestion;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace UsefulProteomicsDatabases.EntrapmentGeneration;
+
+/// <summary>What became of one base piece of a target sequence.</summary>
+public enum PieceOutcome
+{
+    /// <summary>Rearranged into an isomeric partner.</summary>
+    Permuted,
+
+    /// <summary>
+    /// Has no rearrangement, but is shorter than the minimum peptide length, so it is never
+    /// identified on its own. Kept exactly as it was: no real peptide enters the entrapment
+    /// database, and identity is a rearrangement, so isomerism survives.
+    /// </summary>
+    KeptVerbatimTooShort,
+
+    /// <summary>
+    /// Has no usable rearrangement and *is* long enough to be identified, so it is dropped from the
+    /// entrapment sequence. Emitting it unchanged would place a genuine target peptide inside the
+    /// entrapment database, where every hit is supposed to be false.
+    /// </summary>
+    Excised
+}
+
+/// <summary>One base piece of the target and what happened to it.</summary>
+public sealed class EntrapmentPiece
+{
+    internal EntrapmentPiece(int index, string targetPiece, string? entrapmentPiece,
+        PieceOutcome outcome, EntrapmentFailure failure)
+    {
+        Index = index;
+        TargetPiece = targetPiece;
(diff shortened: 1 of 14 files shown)


trishorts:
**Full-solution test run, green.**

```
Test Run Successful.
Total tests: 6224
     Passed: 6195
    Skipped: 29
     Failed: 0
```

`dotnet test mzLib/Test/Test.csproj` over the whole suite at `381a3a44`, not just the database/decoy/entrapment/protein-reader subsets I had been reporting until now. That was the outstanding pre-merge gate on this PR and it is closed.

Worth stating explicitly since this PR replaced a 15-branch stack: this is the first time the entrapment work has been tested **against current master** rather than against the master it branched from three days ago. That is also how the `#1251` regression surfaced — the old stack would have reverted your UniProt FASTA accession fix, and no test inside the stack could have seen it.

acesnik (commented):
🤖 This gives mzLib an entrapment generator that unranks the permutation space rather than sampling it, which is what makes emptiness provable and lets a discovery pair back from the accession alone. Both headline claims hold under measurement: the unranking is a clean bijection with no overflow and exact division, checked against brute force over ~4,000 randomised cases, and the partner is byte-identical across nine processes including varied tiering, GC and locale. The `ProteinDbLoader` edit is genuinely inert — `od -c` gives `" R a n d o m "` on both sides and every classification site still uses `IndexOf(..., OrdinalIgnoreCase)` untouched.

Three things I'd fix before this lands.

`RefuseAgentsWhoseSitesCannotBeHeld` tests `(motif.InducingCleavage?.Length ?? 0) > 1`, so an *empty* motif passes — and `top-down`, `peptidomics`, `singleN` and `singleC` all have an empty Motif column in `proteases.tsv`, while `non-specific` is `X|`. An empty `InducingCleavage` makes `Fits` return true at every location, so the protein partitions into single residues, each has a permutation space of 1, and at the default `minPeptideLength = 7` each takes the `KeptVerbatimTooShort` branch and is appended unchanged. Nothing compares the finished sequence to its target, so the entry is emitted with `IsEntrapment = true` and an identical `BaseSequence` — and because the report skips every piece below `MinLength`, all four `Unpairable*` columns read 0. Reachable today by anyone reaching for top-down before #1272.

Entrapment partners are also generated for decoys. `DecoyProteinGenerator.cs:222` passes `nonVariantProtein: decoyConsensus`, so a decoy's `ConsensusVariant` is a decoy and `DatabaseEntries` yields it as its own entry. The partner comes out `Random_DECOY_P12345_f0`, which fails the case-sensitive `StartsWith(decoyIdentifier)` at `ProteinDbLoader.cs:399` and reloads as a target-side entrapment entry.

And the foreign arm keys its exclusion rows by `foreign.Accession` (`EntrapmentProteinGenerator.cs:80`) while writing the entry as `Random_foreign_<acc>` (`:103`) — the same table whose `unrepairableRunCollision` rows are deliberately keyed the other way, for the reason at `EntrapmentReport.cs:497-501`.

Minor: the exclusions list is exact-string only, so an I/L reordering is spectrally identical to a real target peptide and reported nowhere (4 pieces on cRAP have *no* distinguishable arrangement); `stretch = size / foldCount` partitions the identity into a fold and then skips it, so at `stretch == 1` that fold reports `AllPermutationsTaken` rather than `SpaceTooSmallForFoldCount`; and `EntrapmentReport.cs:457` has a raw NUL instead of `\0`, which makes `grep` skip the file silently.

I diffed #1272 — it's 534 insertions, zero deletions, no shared line touched, so nothing here is waiting on it.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
```diff
@@ -0,0 +1,244 @@
...
+    /// on top of the cleavage sites. The assembler uses this to anchor a protein's termini, so a
+    /// modification restricted to one of them survives the rearrangement.</param>
+    /// <param name="rejectInContext">An extra test a candidate must pass, on top of not being a
+    /// forbidden sequence itself. The assembler uses this to reject a candidate that would make a
+    /// *missed-cleavage* peptide equal to a real target peptide: this generator sees one base piece
+    /// at a time, and a missed-cleavage peptide is a run of adjacent pieces, so a concatenation can
+    /// collide even when none of its parts does. Measured before this existed: 2,094 peptides
+    /// (0.076%) appeared in both the target and entrapment sets, 2,090 of them with a missed
+    /// cleavage.</param>
+    public static EntrapmentPeptide Create(string targetSequence, List<DigestionMotif> motifs,
```
nbollis, line 146: [review-backend] Create() has no null guard for forbiddenSequences - NullReferenceException on a public API (revised: guard, don't throw)

REVISED per user feedback ('couldnt we just guard instead of throwing?'): Create validates targetSequence, foldCount and fold, but a null forbiddenSequences reaches forbiddenSequences.Contains(candidate) at ~line 191 as an NRE. The same set is passed through unvalidated by EntrapmentProteinGenerator.Create/GenerateEntrapment. Per feedback, prefer a guard over a throw: treat a null set as 'no exclusions' (an empty IReadOnlySet<string>) rather than throwing, consistent with the parameter being optional in spirit.

Suggestion: Guard instead of throw: at the top of Create, use forbiddenSequences ??= empty IReadOnlySet<string> (e.g. a shared static empty set) so a null means 'no forbidden sequences', and null-coalesce the same way in EntrapmentProteinGenerator.Create/GenerateEntrapment. Add a test asserting null yields the same result as an empty set.

nbollis, line 146: [review-oop] Static class with 8-parameter method + context predicates (NON-BLOCKING maintainability note)

REVISED per user feedback ('this is fine but make sure you mention it is non blocking'): EntrapmentPeptideGenerator is static; creation is a single 8-parameter static method extended by two simultaneously-supplied delegates (alsoHeldInPlace, rejectInContext). The assembler observes nothing about what the generator did other than the returned EntrapmentPeptide. Assemble() similarly does digestion, permutation, run rejection, initiator-methionine rejection, excision bookkeeping, mapping and gap counting in one method. NON-BLOCKING: maintainability suggestion only, not a correctness defect.

Suggestion: Non-blocking: consider extracting an IEntrapmentPeptideSource (Create + PermutationSpaceSize) or grouping constraint parameters into a small immutable EntrapmentConstraints value object; extract per-piece strategy units from Assemble. No change required before merge.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,729 @@
...
+                StratumFor(peptide).Ambiguous++;
+            }
+
+            // Straight onto the total, never a stratum. StratumFor("") resolves to the zero-site
+            // stratum -- it short-circuits an empty peptide to 0 and then reuses the real one -- so
+            // a whole-protein figure was being published beside that stratum's own per-peptide
+            // counts, and every other row read 0 for a column that is not zero. This one is sold as
+            // a denominator and the documented use is the stratified one, which is exactly where
+            // dividing within a row would go wrong.
+            _wholeProtein.SearchSpacePeptides += pairing.SearchablePeptideCount;
```
nbollis, line 513: [review-backend] Per-stratum searchSpacePeptides/entrapmentSearchSpacePeptides columns are structurally always 0

EntrapmentStratum.SearchSpacePeptides and EntrapmentSearchSpacePeptides are documented as per-stratum searchable populations and the denominator Ambiguous belongs to, and ToTabSeparated emits them per stratum, but they are only ever incremented on _wholeProtein. Every stratum row reads 0 for these columns; only the 'all' row carries values. A consumer computing the documented per-stratum rate ambiguous/searchSpacePeptides divides by zero.

Suggestion: Populate per-stratum values by binning searchable peptides via the site counter, or drop the always-zero per-stratum columns and fix the docs to say these are whole-database figures reported on the total row only.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,729 @@
...
+                + "its companion, so the companion protein has to be supplied to Add.");
+        }
+
+        _massGroups?.Add(target, companion!, assembly, _digestionParams.MinLength);
+
+        if (!_ambiguousByAccession.TryGetValue(target.Accession, out HashSet<string>? ambiguous))
+        {
+            var pairing = new EntrapmentPairing(target, _digestionParams);
+            ambiguous = pairing.AmbiguousPeptides.ToHashSet();
+            _ambiguousByAccession[target.Accession] = ambiguous;
```
nbollis, line 500: [review-generic] Ambiguous-peptide exclusion rows keyed by target accession while the other two exclusion lists use the entrapment accession

UnrepairableRunCollisionsByAccession and InitiatorMethionineCollisionsByAccession were re-keyed to the entrapment accession because 'a consumer filtering on (accession, peptide) matched the ambiguous rows and silently missed these'. The ambiguous rows kept target-accession keying, so ExclusionsToTabSeparated mixes two accession semantics in one column and a consumer filtering entrapment hits cannot match ambiguous rows. The test ExclusionTableNamesTheAmbiguousPeptides enshrines the target-keyed form.

Suggestion: Make the accession semantics uniform across all four exclusion-table sections (key ambiguous rows by entrapment accession with the target in a separate column, or add the entrapment accession as an extra column), and update the test.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,588 @@
...
+                {
+                    unrepairable.Add(collision);
+                }
+                // Same rule for the initiator-methionine route, and for the same reason: a piece
+                // kept verbatim has no alternative to move to, so the collision is named rather than
+                // repaired. It cannot fire while MinLength is respected -- the stripped form is one
+                // residue shorter than a piece that was already too short -- and it is written
+                // because "cannot fire" is a claim the count is entitled to check.
+                string? strippedCollision = strippedOfInitiatorMethionine?.Invoke(piece);
+                if (strippedCollision is not null)
```
nbollis, line 283: [review-backend] InitiatorMethionineCollisionPeptides can never be populated - dead reporting path

The only site adding to initiatorMethionineCollisions is the KeptVerbatimTooShort branch, but RejectInitiatorMethionineCollisions returns null whenever stripped.Length < minLength, and the stripped form of a kept-verbatim piece is piece.Length-1 < MinLength, so the invocation always returns null. Permuted opening pieces that would strip-collide are rejected, and openings whose every arrangement strip-collides are excised (UnpairableRunCollisionsExhausted), never retained-and-listed. So the list, report column, and exclusion rows are always empty while the docs describe real sequences being reported for consumer exclusion.

Suggestion: Either retain-and-count the collision in the list (matching 'flag and count, never silently repair'), or delete the always-empty list/column/exclusion section and reclassify under UnpairableRunCollisionsExhausted, and fix the documentation.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            if (rejectInContext is not null && rejectInContext(candidate))
+            {
+                // Usable on its own merits, and refused only because of what it sits next to.
+                anyRejectedOnlyByContext = true;
+                continue;
+            }
+
+            return new EntrapmentPeptide(targetSequence, candidate, swapped, fold, size,
+                (int)(step + BigInteger.One), EntrapmentFailure.None);
```
nbollis, line 204: [review-backend] ProbesUsed is a checked BigInteger-to-int cast that throws OverflowException / wraps past 2^31 probes and reports 0 on failure

The fold search walks 'for (BigInteger step = 0; step < stretch; step++)' and reports ProbesUsed as (int)(step + BigInteger.One). stretch is unbounded (lightly-pinned peptides can exceed 2^31 arrangements), and the only way the walk goes far is a forbiddenSequences set covering most of the stretch - adversarial or degenerate. The explicit BigInteger -> int conversion throws OverflowException once step exceeds int.MaxValue, crashing a report build that should have classified the outcome; even short of the throw, the returned count silently wraps. On the failure path Failed() reports ProbesUsed=0 even though the whole stretch was examined.

Suggestion: Widen ProbesUsed to long, or clamp the diagnostic (ProbesUsed = step < int.MaxValue ? (int)(step + 1) : int.MaxValue) so it cannot throw or wrap, and report actual probes examined on failure paths.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
```diff
@@ -0,0 +1,729 @@
...
+            EntrapmentPairing.CountSearchablePeptides(assembly.EntrapmentSequence, _digestionParams);
+        if (assembly.UnrepairableRunCollisionPeptides.Count > 0)
+        {
+            // Keyed by the accession a search will report the peptide under, not by the target it
+            // was rearranged from. These peptides belong to the ENTRAPMENT protein; filing them
+            // under the target made the two halves of one table mean different things by
+            // `accession`, and a consumer filtering on (accession, peptide) matched the ambiguous
+            // rows and silently missed these -- under-excluding, which inflates an FDP estimate.
+            string entrapmentAccession =
+                EntrapmentAccession.Format(target.Accession, fold, _entrapmentIdentifier);
```
nbollis, line 573: [review-oop] Duplicate per-accession exclusion-list collection logic in Add()

The unrepairable-run and initiator-methionine exclusion lists are collected in two nearly-identical blocks: same EntrapmentAccession.Format keying by entrapment accession, same TryGetValue-or-create, same distinct-peptide dedup, same _wholeProtein.<X>++ increment. Any divergence between the two blocks silently changes what the sidecar means for one column (keying history shows this has already happened). Both lists also recompute EntrapmentAccession.Format twice per Add call.

Suggestion: Extract a private helper (e.g. RecordExclusionsByAccession(Dictionary<string,List<string>> target, IEnumerable<string> peptides, Action<int> wholeProteinIncrement)) and have both blocks delegate to it; add a unit test asserting the two sidecars stay in lockstep.

nbollis (commented):
## Review Synthesis

This review posts 10 explicitly approved inline findings against head `d67faec2ed70f3a36bae1a7906101ff7477ca2f2`.

- **Major:** 1
- **Minor:** 7
- **Suggestions:** 2
- **Summary-only:** 0

One finding was denied and one remained undecided; neither is included in this review.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPairing.cs
```diff
@@ -0,0 +1,172 @@
...
+/// <para>The key can collide: a protein may contain two different peptides sharing both, and
+/// P57071 really does contain <c>LIHTGVK</c> and <c>LIHTVGK</c>. Those are reported in
+/// <see cref="AmbiguousPeptides"/> and refuse to resolve, because a guess between them would be
+/// wrong half the time and silently so.</para>
+/// </remarks>
+public sealed class EntrapmentPairing
+{
+    private readonly Dictionary<string, string> _byKey;
+    private readonly HashSet<string> _ambiguous;
+    private readonly List<DigestionMotif> _motifs;
```
nbollis, line 30: [review-oop] KeyOf() re-derives CleavageSitePositions on every peptide; consider caching

KeyOf is called per (peptide = runs of up to MaxMissedCleavages+1 pieces) in the index constructor and per entrapment peptide in TryResolve; each call recomputes CleavageSitePositions (motif matching) plus sorting. For a whole proteome this is O(peptides * motif matches) of repeated recomputation. CountSearchablePeptides also constructs a throwaway EntrapmentPairing and a new Protein(sequence, 'counting-only') per entrapment sequence.

Suggestion: Cache pinned positions/free-multiset per distinct peptide within an index build (memoize KeyOf results), and/or add a lightweight static counter that avoids per-call Protein construction.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentMassGroups.cs
```diff
@@ -0,0 +1,458 @@
...
+    /// <param name="annotations">The protein's modifications, keyed one-based.</param>
+    /// <remarks>
+    /// <b>The fit test is the point of this method.</b> An annotation is counted only where
+    /// <c>ModificationLocalization.ModFits</c> accepts it at its position <i>within the peptide</i>,
+    /// which is where a peptide-level restriction is finally judged. Counting annotations without it
+    /// reproduces the failure this exists to detect: a modification restricted to a peptide terminus
+    /// is transported, written, and counted by every protein-level tally, and then quietly does not
+    /// exist as a hypothesis because it no longer sits at the terminus.
+    /// </remarks>
+    public Dictionary<MassGroup, int> AnnotatedSites(string polymerSequence, int peptideStart,
```
nbollis, line 246: [review-oop] AnnotatedSites() returns a mutable Dictionary keyed by reference MassGroup rather than exposing the data

AnnotatedSites returns a fresh Dictionary keyed by reference MassGroup; MassGroupComparison.Add then does targetSites.GetValueOrDefault(tally.Group), a reference-equality lookup that only works because MassGroup instances are reference-stable within one MassGroupIndex. The dictionary is exposed mutable and keyed by an object whose equality is reference-based, and the intermediate dictionary + per-group lookup per piece is inefficient.

Suggestion: Return a read-only mapping keyed by representative mass or expose CountFor(MassGroup); or accumulate directly into the tallies in one loop over annotations, avoiding intermediate dictionaries and reference-keyed lookups.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
```diff
@@ -0,0 +1,588 @@
...
+        if (completesAForbiddenRun is null && strippedOfInitiatorMethionine is null)
+        {
+            return null;
+        }
+
+        return candidate =>
+            (completesAForbiddenRun is not null && completesAForbiddenRun(candidate).Count > 0)
+            || (strippedOfInitiatorMethionine is not null && strippedOfInitiatorMethionine(candidate) is not null);
+    }
+
```
nbollis, line 326: [review-backend] Empty-motif proteases (top-down/peptidomics/singleN/singleC) pass the guard and emit no-op entrapment entries with identical sequences

RefuseAgentsWhoseSitesCannotBeHeld tests (motif.InducingCleavage?.Length ?? 0) > 1, so an empty motif passes. top-down, peptidomics, singleN and singleC all have an empty Motif column in proteases.tsv (verified: motif=<>), while non-specific is X|. An empty InducingCleavage makes CleavageSitePositions.Fits return true at every location, so the protein partitions into single residues, each with a permutation space of 1, and at the default minPeptideLength=7 each takes the KeptVerbatimTooShort branch and is appended unchanged. Nothing compares the finished sequence to its target, so the entry is emitted with IsEntrapment=true and an identical BaseSequence - a no-op entrapment. Report skips every piece below MinLength, so all Unpairable* columns read 0. Reachable today via top-down before #1272.

Suggestion: Refuse agents whose inducing cleavage is empty or cannot be held (treat empty motif as unpinnable), and add an emission-time check that any IsEntrapment entry's BaseSequence differs from its target.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
trishorts: **Fixed in `094bde48`, guarded rather than thrown** — taking your revised form.

`Create` now does `forbiddenSequences ??= NoForbiddenSequences` (a shared empty static), and `Assemble` does the same before passing it on, so a null set means "nothing is forbidden" at every layer rather than an NRE naming no argument.

Test: `ANullForbiddenSetMeansNothingIsForbidden` asserts null and empty produce the **same sequence**, not merely that neither throws — the weaker assertion would pass against a `Create` that quietly took a different path for null.

trishorts: **Agreed, and taking it as non-blocking.** Not changed here.

The reasoning for leaving it: `EntrapmentPeptideGenerator.Create` being a pure static function of `(sequence, motifs, seed, fold, anchors)` is load-bearing rather than incidental — it is what makes a database regenerable in pieces, in parallel, years apart, and byte-identical. An `IEntrapmentPeptideSource` with state is the shape most likely to lose that quietly. A parameter-object for the constraints would keep it, and I'd rather do that as its own change than fold an API reshape into a correctness pass.

`Assemble` doing six things in one method is the fairer half of this, and I'd take that refactor next.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: **Confirmed, and taken as your second option — docs, not schema — in `094bde48`.**

You're right that the columns are structurally always 0 on stratum rows, and right that the documented per-stratum rate divides by zero. `SearchSpacePeptides` is accumulated once per *protein* (in the `_ambiguousByAccession` cache-miss branch), so it cannot be attributed to a candidate-site stratum at all.

I did **not** drop the columns, because phred is mid-flight against this exact table shape and a positional reader would silently shift. Instead both XML docs now say plainly that these are whole-database figures carried on `Total` alone, that the stratum zero is **structural rather than measured**, and that `ambiguous / searchSpacePeptides` is a whole-database rate.

If you'd rather have them off the stratum rows I'll do it — say so and I'll coordinate the schema bump with the consumer.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: **Real, and fixed in `094bde48` — but by naming the difference rather than removing it.**

Pushing back on "make the semantics uniform", because I think the column holds two things for a *good* reason. `ambiguous` rows describe **target** peptides: two peptides of the same target protein sharing a composition-and-pinning key, so a discovery cannot be traced to one of them. They live in the target database. Re-keying them to an entrapment accession would assert something false about where they are.

So the table gains a fourth column, `side`, with values `target` and `entrapment` — **appended last**, so a consumer reading the first three fields positionally is unaffected. A consumer filtering entrapment hits wants `side == "entrapment"`; one excluding ambiguous targets from a paired estimator wants the other. Both are now expressible, which they were not before.

`ExclusionTableNamesTheAmbiguousPeptides` is updated, and `TheExclusionTableSaysWhichDatabaseEachAccessionNames` asserts the invariant over **every** row rather than the two I happened to think of: exactly four fields, and `side` is `target` iff the reason is `ambiguous`.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: **Your analysis is exactly right, and I'd like to keep the code — with the docs fixed, in `094bde48`.**

Confirmed dead, by the route you describe: the only append is the kept-verbatim branch, whose piece is already below `MinLength`, so `stripped.Length < minLength` returns null before it can match. Permuted candidates that would strip-collide are *rejected*, and an opening piece whose every arrangement strip-collides is excised as `RunCollisionsExhaustedTheSpace`. The list cannot populate.

Where I disagree is the remedy. This is an **assertion channel**, not a reporting path — it is deliberately written to be unreachable, so that if a future change to the excision rule or the length bounds makes it reachable, the peptide is *named* instead of leaking silently. Deleting it removes a tripwire over a route that cost four real peptides across three arms before the check existed. Reclassifying it under `UnpairableRunCollisionsExhausted` would also lose the distinction at MC = 0, where the run route does not exist and the two zeros mean different things.

What was genuinely wrong is that the docs described real sequences being reported for consumer exclusion. They now state that the list is provably always empty under the current rules, say *why*, and warn explicitly: **do not read the zero as a measurement of the route** — the rejections are the measurement and they are counted as excisions.

If you'd still rather it went, I'll take that call — but I'd want the tripwire replaced by an assertion somewhere rather than dropped.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPeptideGenerator.cs
trishorts: **Split verdict, both halves addressed in `094bde48`.**

On the `OverflowException`: the cast is genuinely checked, so you're right in principle — but reaching it needs the walk to pass 2³¹ steps, i.e. 2.1 billion `UnrankPermutation` calls in a single `Create`. That is not a crash anyone will see; it is a run that never finishes. I've clamped it anyway, because your suggestion costs one line and the value is only ever read as a diagnostic: `Probes()` saturates at `int.MaxValue` rather than throwing or wrapping, and "at least this many" is what the column means at that magnitude.

On `ProbesUsed = 0` on the failure path — **that one is the real bug**, and thank you for putting it last where I nearly missed it. The whole point of walking the stretch end to end is that "no partner exists" is a *proof* rather than an abandoned search, and the number that evidences the walk was exhaustive was being thrown away exactly where it matters. `Failed` now takes the probe count and every failure path reports the whole stretch.

`AFailedSearchReportsHowMuchOfTheSpaceItWalked` pins it: a space of 3 with 2 arrangements forbidden reports `ProbesUsed == 3`, not 0.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentReport.cs
trishorts: **Done in `4a36e5d3`** — and your parenthesis is the argument that moved me: *"keying history shows this has already happened."* That is precisely how these rows came to be filed under the target while the ambiguous rows meant something else by `accession`.

Both lists now go through `RecordExclusions(byAccession, entrapmentAccession, peptides, count)`, and `EntrapmentAccession.Format` is computed once per `Add` instead of twice.

I did **not** add the "the two sidecars stay in lockstep" test you suggested, and I want to be explicit about why rather than have it look like an omission: after the extraction there is one implementation, and one implementation cannot disagree with itself — the same reasoning `GenerateEntrapment` already gives for being a single loop rather than two. A test asserting agreement between two things that are now one thing would pass forever without constraining anything. The 169 entrapment tests staying green is the statement that behaviour did not move.

Kept as two *lists* rather than two reasons on one: different constructions, and at MC = 0 the run list is empty by definition while the other is empty only if the check ran.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentPairing.cs
trishorts: **Fair, and not taken here — deliberately, as a measurement question.**

The recomputation is real: `KeyOf` re-derives `CleavageSitePositions` per peptide, and `CountSearchablePeptides` constructs a throwaway `Protein` per entrapment sequence. But the whole-proteome figure this sits inside is 112.8 s for 52,337 proteins (60.3 s generation, 1.15 ms/protein), and I don't currently know what fraction of that is here rather than in digestion itself. Memoising on a guess is how you end up with a cache that costs more than it saves and a determinism guarantee that now depends on cache state.

Happy to do it with a profile behind it. Flagging one constraint for whoever picks it up: any memo has to be scoped to a single index build, because the generator's contract is that output is a pure function of `(sequence, seed)` and nothing may make it depend on what was processed earlier.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentMassGroups.cs
trishorts: **Correct on all three counts, and not changed here.**

Confirmed: `MassGroup` has reference equality, so `targetSites.GetValueOrDefault(tally.Group)` works only because instances are reference-stable within one `MassGroupIndex` — an invariant nothing states and nothing enforces. That it happens to hold is not the same as it being safe.

Leaving it for now because this is public API surface that phred is reading right now (the `massgroups_<protease>_<tolerance>.tsv` sidecars), and `CountFor(MassGroup)` or a mass-keyed read-only map changes the shape they consume. It's the right change and I'd like to make it once, with them, rather than mid-review.

Recording the hazard so it isn't rediscovered: **if anyone ever makes `MassGroupIndex` return fresh `MassGroup` instances per call, the comparison silently returns zeros** rather than failing — the counters would read a clean 100% while measuring nothing. That is the same silent-success failure mode the mass-group invariant exists to catch, which makes it worth a value-equality fix sooner rather than later.

mzLib/UsefulProteomicsDatabases/EntrapmentGeneration/EntrapmentAssembler.cs
trishorts: **Confirmed in full, fixed in `094bde48` — and it is worse than either of us wrote: `non-specific` fails identically.**

Your mechanism is exactly right, and I verified each step: `"".Split(',')` yields one motif with `InducingCleavage == ""`, `DigestionMotif.Fits` runs its comparison loop zero times and falls through to `fits = true`, `GetDigestionSiteIndices` therefore returns every index, and each single residue takes the kept-verbatim branch below `MinLength`. Entry emitted identical, `IsEntrapment = true`, all four `Unpairable*` columns reading 0.

**The addition:** `non-specific` is `X|`, and `MotifMatches` returns `true` for `'X'` against any residue — so it partitions into single residues by the same mechanism and produced the same no-op entry. Your suggested guard ("treat empty motif as unpinnable") would not have caught it, since `"X"` has length 1. So the guard now refuses **empty** motifs *and* **wildcard** motifs, and `B`/`J`/`Z` stay allowed because each matches a definite pair of residues — what they pin is still a function of the sequence, so a rearrangement can neither invent nor destroy a site.

Ordering detail worth knowing: the wildcard test runs *after* the multi-residue test, so `StcE-trypsin`'s `TX|T` keeps the sharper message about the seam rather than being reported as a wildcard. My first cut had it the other way and `AMultiResidueAgentIsRefused` caught it.

On your emission-time check — I took it as a **count, not a throw**, because the two cases are not distinguishable at that point: a genuinely tiny entry whose every piece is below `MinLength` is legitimately kept verbatim throughout and contributes no searchable peptide either way. `EntrapmentAssembly.IsIdenticalToTarget` and a `# entriesIdenticalToTarget` provenance line (emitted only when non-zero, so healthy reports are byte-identical) say so out loud instead.

Six tests, all confirmed to fail with the guard disabled: the four empty-motif agents by name, the wildcard one, and `AnUnpinnableAgentNeverEmitsAnEntrapmentEntryIdenticalToItsTarget` which asserts over the **output** rather than the message — because the message is not the property that matters.

trishorts:
## Both reviews answered — eight fixes in `094bde48` and `4a36e5d3`

@acesnik @nbollis — thank you both, and apologies for the delay: your reviews sat from 2026-09-05 and 2026-09-14 with nothing pushed in between, so everything you found was still live at head.

Twelve distinct findings between you. **Eight were real and are fixed**; two I've pushed back on with evidence; two I've left as measurement questions, each with the reasoning recorded in the thread rather than silently dropped. @nbollis — every one of your ten has a threaded reply. This comment answers @acesnik's review, which was top-level.

Every new test was confirmed to **fail with its fix disabled** rather than assumed to bite, which is this project's standing rule after five silent-wrongness bugs got past planning.

---

### 1. Empty-motif proteases — confirmed, and it is worse than reported

You were right end to end, and I verified each link: `"".Split(',')` yields one motif with `InducingCleavage == ""`; `DigestionMotif.Fits` runs its comparison loop zero times and falls through to `fits = true`; `GetDigestionSiteIndices` therefore returns every index; each single residue is below `MinLength` and takes the kept-verbatim branch. Entry emitted byte-identical, `IsEntrapment = true`, all four `Unpairable*` columns reading 0.

**What neither review caught: `non-specific` fails identically.** Its motif is `X|`, and `MotifMatches` returns `true` for `'X'` against any residue — so it partitions into single residues by the same mechanism. Refusing only empty motifs would have left it. The guard now refuses **empty** and **wildcard** motifs both; `B`, `J` and `Z` stay allowed, because each matches a definite pair of residues, so what they pin is still a function of the sequence and a rearrangement can neither invent nor destroy a site.

The wildcard test deliberately runs *after* the multi-residue test, so `StcE-trypsin`'s `TX|T` keeps the sharper message about the seam. My first cut had the order reversed and `AMultiResidueAgentIsRefused` caught it — which is the existing test doing its job.

On the emission-time "entrapment ≠ target" check you asked for: I took it as a **count rather than a throw**. A genuinely tiny entry whose every piece is below `MinLength` is legitimately kept verbatim throughout and contributes no searchable peptide either way, and that is not distinguishable from a fault at the point of emission. So `EntrapmentAssembly.IsIdenticalToTarget` plus a `# entriesIdenticalToTarget` provenance line, emitted only when non-zero so healthy reports stay byte-identical. The dangerous route is closed at the call; this is the tripwire for any route that opens later.

### 2. Entrapment partners generated for decoys — confirmed

`DecoyProteinGenerator.cs:222` passes `nonVariantProtein: decoyConsensus`, so a decoy's `ConsensusVariant` is a decoy and `DatabaseEntries` — which deduplicates *by consensus* — yields it as an entry of its own. Your accession analysis holds: `Random_DECOY_P12345_f0` fails the case-sensitive `StartsWith(decoyIdentifier)` at `ProteinDbLoader.cs:399` and reloads target-side.

Now refused at `Create` and at `CreateForeign`, rather than skipped. Skipping would generate fewer entries than the caller asked for and say nothing — which is the failure mode this project has now hit three separate times, and a count that quietly disagrees with its input is worse than a throw naming the entry. The message says what to do instead: entrap the targets, then generate decoys from the finished target-plus-entrapment database.

`ADecoyReachesTheGeneratorThroughItsOwnConsensusVariant` pins the reachability rather than just the guard — it asserts `DatabaseEntries` really does yield the decoy, so if that ever changes the test says so instead of passing vacuously.

### 3. Foreign arm keyed by `foreign.Accession` — confirmed

Re-keyed to `EntrapmentAccession.FormatForeign(...)`, for the reason the other two lists were re-keyed. The sidecar has one `accession` column, and a consumer filtering it against what a search reported can only match rows keyed the way the database is; keyed by the foreign protein's own accession those rows matched nothing, so the arm's one real hazard — homology — was named in a form nobody could subtract. The test now also asserts `entrapment.Select(e => e.Accession)` is a superset of `shared.Keys`, so the two can't drift apart again.

### 4. `stretch == 1` reports the wrong failure — confirmed

Exactly as you describe: the identity is partitioned into a fold and then skipped. The root of it is that the guard was `size / foldCount == 0`, but the identity is never a usable partner, so the space to share out is `size - 1`. That's the up-front test now. There's a residual case the up-front test can't see (a stretch that happens to contain the identity at a fold count that otherwise fits), so the end-of-walk classification also distinguishes "refused only as the identity" from "refused as forbidden".

The test states the invariant rather than the instance: **with nothing forbidden, `AllPermutationsTaken` is never an honest answer** — nothing took them. It sweeps fold counts 2–4 asserting that failure is always `SpaceTooSmallForFoldCount`.

### 5. Raw NUL byte — confirmed, and it was doing more damage than grep

Now the two-character escape. Two corrections to the finding, both in your favour:

- It had moved to **line 527** by `d67faec2`.
- Skipping `grep` was the *lesser* harm. A NUL makes git classify the blob as **binary**, which skipped `text=auto` normalization — so `EntrapmentReport.cs` was the only CRLF blob in `EntrapmentGeneration/` while every sibling is LF.

That is why this one file shows as a whole-file diff in `094bde48`: removing the NUL lets normalization finally apply. One time only — the follow-up commit `4a36e5d3` touches the same file and diffs as 49/47.

### 6. I/L reordering is spectrally identical but excluded nowhere

**Not fixed, and I think you're right that it's a real gap** rather than a nicety — an exclusion list that a search can defeat by isomeric substitution is not doing its whole job. I've left it because I don't yet know how to state the fix without changing what the list *means*: exact-string exclusion answers "is this the same peptide", and an I/L-collapsed key answers "could a search confuse these", which is a different and larger set. Widening it silently would over-exclude and deflate an FDP estimate, which is the direction that flatters us.

Your measurement — 4 pieces on cRAP with *no* distinguishable arrangement — is the thing that makes it worth doing properly. I'd like to take it as its own change with a number attached to both readings. Flagging it here so it is not lost.

---

### What I pushed back on, both in-thread

- **`InitiatorMethionineCollisionPeptides` is dead** (@nbollis) — confirmed dead, but it is a deliberate assertion channel over a route that cost four real peptides across three arms before the check existed. The docs were what was wrong: they described sequences being reported. They now state it is provably always empty, why, and that the zero is not a measurement of the route.
- **`ProbesUsed` overflow** (@nbollis) — the `OverflowException` needs 2³¹ unrank calls in one `Create`, so it is a run that never finishes rather than a crash anyone sees. Clamped anyway, since it costs a line. The real bug in that finding is the one listed last: `ProbesUsed = 0` on failure discarded the evidence that the walk was *exhaustive*, which is the whole basis for "no partner exists" being a proof. Fixed.

### Left as measurement questions, reasoning in-thread

`KeyOf` caching and `AnnotatedSites`' reference-keyed dictionary. Both correct findings. The first wants a profile before a cache that could compromise the determinism guarantee; the second is public API that phred is consuming right now, so I'd rather change its shape once, with them. The `AnnotatedSites` hazard is recorded in that thread because it fails *silently* — reference-unstable `MassGroup` instances would make the comparison read a clean 100% while measuring nothing.

---

### State

- **Full suite: 6,302 tests, 6,270 passed, 31 skipped, 1 failed.** The failure is `TestKoinaProsit2020IntensityHCDModelCollisionEnergyBoundaries` — a socket error against the Koina external service, unreachable from this machine and untouched by any of this. CI will say whether it's reachable there.
- 26 new tests. Every behavioural one verified red with its fix disabled; the two schema tests (`side` column, `entriesIdenticalToTarget`) are red at compile time without theirs.
- **One consumer-visible schema change**: the exclusion table gains a fourth column, `side`, appended last so a positional reader of the first three fields is unaffected.

@acesnik — your note that #1272 is 534 insertions with no shared line touched matches what I see; nothing here waits on it either.

PR #1276, opened 2026-09-02
Add Inosine Nucleotide and Psi as alternative Pseudouridine representation

nbollis:
Allow multiple representations in the nucleotide dictionary to map to the same nucleotide


mzLib/Transcriptomics/Nucleotide.cs
```diff
@@ -53,7 +54,9 @@ static Nucleotide()
             CytosineBase = AddResidue("Cytosine", 'C', "Cyt", "C4H4N3O1");
             GuanineBase = AddResidue("Guanine", 'G', "Gua", "C5H4N5O1");
             UracilBase = AddResidue("Uracil", 'U', "Ura", "C4H3N2O2");
+            InosineBase = AddResidue("Inosine", 'I', "Ino", "C5H3N4O1");
```
trishorts: **Critical — this line makes the `Nucleotide` static constructor throw under Turkish/Azeri locales.**

`AddResidueToDictionary` does `ResiduesByLetter[Char.ToLower(residue.Letter)]` (Nucleotide.cs:79). `Char.ToLower` is *culture-sensitive*: under `tr-TR`/`az-AZ`, `'I'` maps to `'ı'` (U+0131 = 305), while `ResiduesByLetter` is `new Nucleotide['z' + 1]` — length 123. `'I'` is the first letter added to this table whose case mapping is locale-dependent, so master is unaffected; this is a regression introduced by adding Inosine.

Reproduced in a clean process (console app referencing `Transcriptomics.csproj` directly, culture set before first touch of the type):

```
[tr-TR] this branch : TypeInitializationException -> IndexOutOfRangeException
[tr-TR] master      : OK -> Y Psu (PseudoUracil)
[en-US] this branch : OK -> I Ino (Inosine)
```

A `TypeInitializationException` is cached for the life of the AppDomain, so the whole `Transcriptomics` namespace is unusable for any user on a Turkish system locale. CI runs en-US, so it cannot catch this — the existing suite passes.

Fix: `Char.ToLowerInvariant` at line 79 (and at line 248 in the new `TryAddAlternativeRepresentation`).

mzLib/Transcriptomics/Nucleotide.cs
```diff
@@ -209,6 +218,38 @@ public static bool TryGetResidue(string symbol, out Nucleotide residue)
...
+        /// returns false if it maps to a different residue (leaving it unchanged).
+        /// </summary>
+        public static bool TryAddAlternativeRepresentation(Nucleotide residue, char alternative)
+        {
+            if (alternative <= 'z')
+            {
+                if (ResiduesByLetter[alternative] is { } existing && !existing.Equals(residue))
+                    return false;
+                ResiduesByLetter[alternative] = residue;
+                ResiduesByLetter[Char.ToLower(alternative)] = residue;
```
trishorts: **Major — returns `false` ("leaving it unchanged") after it has already mutated `ResiduesByLetter`.**

When `alternative <= 'z'` and the slot is free, this writes both `alternative` and its lowercase form into `ResiduesByLetter` and *then* delegates to the string overload, which can still fail. The caller is told nothing happened, but the two lookup tables are now permanently out of sync.

Reproduced on this branch:

```
TryAddAlternativeRepresentation(A, "K")  -> True
TryAddAlternativeRepresentation(B, 'K')  -> False    // doc comment: "leaving it unchanged"
  GetResidue('K')     -> ResB    // mutated anyway
  GetResidue('k')     -> ResB    // lowercase slot, never clash-checked
  TryGetResidue("K")  -> ResA    // string table still says A
```

Two separate issues here:
1. Do the string-table clash check *before* writing `ResiduesByLetter`, so the method is genuinely all-or-nothing.
2. The clash guard inspects only `ResiduesByLetter[alternative]`, never `ResiduesByLetter[Char.ToLower(alternative)]` — a previously registered lowercase alias is silently clobbered (see line 248 output above).

mzLib/Transcriptomics/Nucleotide.cs
```diff
@@ -209,6 +218,38 @@ public static bool TryGetResidue(string symbol, out Nucleotide residue)
...
+        /// <summary>
+        /// Registers an additional string representation that resolves to the given residue.
+        /// Returns true if the representation was added or already mapped to an equivalent residue;
+        /// returns false if it maps to a different residue (leaving it unchanged).
+        /// </summary>
+        public static bool TryAddAlternativeRepresentation(Nucleotide residue, string alternative)
+        {
+            if (AllKnownResidues.TryGetValue(alternative, out Nucleotide existing))
+                return existing.Equals(residue);
+            AllKnownResidues[alternative] = residue;
```
trishorts: **Major — a single-character *string* alias registers successfully but is unreachable through `GetResidue`.**

This overload writes only `AllKnownResidues`. `GetResidue(string)` forwards length-1 input to `GetResidue(char)`, which reads `ResiduesByLetter` — so an ASCII single-character alias added here is invisible to every lookup that matters, including `NucleicAcid.ParseSequenceString` (which goes through `TryGetResidue(char)`). It returns `true`, but nothing can resolve it.

Reproduced on this branch:

```
TryAddAlternativeRepresentation(A, "K") -> True
  TryGetResidue("K")  -> ResA
  GetResidue("K")     -> KeyNotFoundException   // Try/Get pair disagree on the same key
  GetResidue('K')     -> KeyNotFoundException
```

The new tests don't cover this because `TestCustomResidueWithAlternateCodes` uses a multi-character alias (`"FkpAlt"`) and a non-ASCII char (`Ω`), both of which take different paths.

Suggest either routing single-character strings through the char overload, or rejecting them here.

trishorts (commented):
Reviewed the nucleotide-alias payload (`Nucleotide.cs`, `RnaMods.txt`, the new tests). The chemistry checks out — `C5H3N4O1` is hypoxanthine-less-one-H, and it reconciles exactly with the `A->I` mod's `N-1H-1O1` against `AdenineBase` `C5H4N5`; `BL Modified` is the right base-loss behavior for a base conversion. The `mzLib.sln` churn is VS 18 reformatting and is harmless: no config rows were removed, and the 38 added rows are all `TestAndRelease|Any CPU`. In particular the `Development` project still has no `Release|Any CPU.Build.0`, so the release workflow's solution-wide build/test still excludes it.

Three findings below, all reproduced by running them rather than by reading. One is a hard crash gated on system locale that CI cannot see.

mzLib/Transcriptomics/Nucleotide.cs
trishorts: **Correction to my own comment above — I overstated the impact, and the repro I described is easy to get wrong. Downgrading this from Critical to Major.**

The defect itself stands. `Char.ToLower` is culture-sensitive, `'I'` maps to `'ı'` (U+0131 = 305) under `tr-TR`/`az-AZ`, `ResiduesByLetter` has length 123, and `'I'` is the only ASCII letter with a locale-dependent lowercase mapping — so master, which has no `'I'` residue, is genuinely unaffected. Reproduced clean-process on this branch against a passing master control.

**Two things I got wrong:**

1. **"the entire Transcriptomics namespace is dead for any user with a Turkish system locale" is too broad.** `Loaders` pins invariant culture in its *static constructor* (`Loaders.cs:40-45`), and MetaMorpheus trips that at `GlobalVariables.cs:94` (`LoadModifications` -> `Loaders.LoadUnimod`), one line before `LoadRnaModifications()` at line 95 — and nothing between lines 78 and 94 touches `Nucleotide`. **MetaMorpheus users are protected by that ordering.** What is actually exposed is a consumer that uses `Transcriptomics` without ever going through `Loaders`, i.e. a bare mzLib package consumer. Real, but far narrower than I implied, and it is why this is Major rather than Critical.

2. **A test inside mzLib's own suite will not reproduce it** — which would make this look like a false alarm if that is the first thing you try. It was the first thing *I* tried, and it passed. `FixtureSetUp` calls `Loaders.SetCultureAsInvariantCulture()` in `[OneTimeSetUp]`, and a static constructor runs once per process, so by the time any test body switches culture `Nucleotide` is already initialized. It only reproduces in a fresh process with the culture set before the type is first touched:

```csharp
// console app referencing Transcriptomics.csproj; culture set as the first statement
Thread.CurrentThread.CurrentCulture = new CultureInfo("tr-TR");
var n = Nucleotide.InosineBase;   // TypeInitializationException -> IndexOutOfRangeException
```

The fix is unchanged and still worth making: `Char.ToLowerInvariant` at line 79, and at line 248 in the new `TryAddAlternativeRepresentation`. Sorry for the overstatement on the severity.

nbollis: That is pretty intersting. Is this something we should worry about with this PR? 

I am leaning towards calling that out of scope because we have sooooo much string parsing in our codebases that this cannot be the only case of turkish failures.

nbollis: Good catch. We ended up removing lowercase lookup entirely for built-in residues and alternate aliases, which matches the direction we want long term and also removes the culture-sensitive `Char.ToLower(...)` failure mode. I added regression coverage around the exact-case behavior and the RNA DB lowercase normalization path so lowercase inputs are normalized during DB reading instead of being accepted at lookup time.

mzLib/Transcriptomics/Nucleotide.cs
nbollis: Good catch. One-character string aliases now route through the same exact-case char registration path, so registration, `GetResidue`, `TryGetResidue`, and sequence parsing all agree. I added regression coverage for the single-character string alias path to verify the lookup is reachable and consistent.

mzLib/Transcriptomics/Nucleotide.cs
nbollis: Good catch. I updated the char/string alias registration path so it checks for conflicts before mutating either lookup structure, which keeps the operation all-or-nothing on failure. The new tests cover the conflicting ASCII alias case and verify the original mapping stays unchanged.

Alexander-Sol (dismissed):

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -238,6 +241,13 @@ public static List<RNA> LoadRnaFasta(string rnaDbLocation, bool generateTargets,
                         Dictionary<string, string> additonalDatabaseFields =
                             regexResults.ToDictionary(x => x.Key, x => x.Value);
 
+                        if (!readFirstSequence)
+                        {
+                            if (sb.ToString().IsAllLower())
```
trishorts: **Major — regression: a mixed-case RNA FASTA that loads on `master` now throws and aborts the entire DB read.**

Dropping `ResiduesByLetter[Char.ToLower(residue.Letter)]` from `AddResidueToDictionary` is the right call for the Turkish-locale problem, but the compensating normalization here is all-or-nothing in two ways:

1. `IsAllLower()` requires *every* character to be lowercase, so a soft-masked sequence — the Ensembl/UCSC convention where lowercase marks repeat-masked regions and the rest is uppercase — gets no transform.
2. The check is sampled from the **first record only** and then applied to the whole file, so a file whose first record is uppercase and whose later records are lowercase gets no transform either.

In both cases the first lowercase character reaches `NucleicAcid.ParseSequenceString` and hits the `default:` case, which throws `ArgumentException`. That is not a per-entry message appended to `errors` — the whole `LoadRnaFasta` call fails.

I ran both cases against this branch and against the merge-base (`a8250440`):

| case | `master` | this branch |
|---|---|---|
| `GUACUGCCUcuagugaagca` (soft-masked, one record) | loads → `GUACUGCCUCUAGUGAAGCA` | `ArgumentException: Nucleic Acid Letter c does not exist...` |
| record 1 uppercase, record 2 lowercase | loads both | `ArgumentException: Nucleic Acid Letter g does not exist...` |
| record 1 lowercase, record 2 uppercase | loads both | loads both :white_check_mark: |

The added test `TestDbReadingDifferentExtensionsWithLowercaseSequence` covers only the uniform-lowercase single-record case — the one case that works.

Since the only transform is case normalization and `ToUpper()` is a no-op on already-uppercase input, the simplest fix is to drop the detection and always normalize. If you'd rather keep the flag, key it per-sequence on `sequence.Any(char.IsLower)` instead of once per file on `IsAllLower()`.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -364,7 +376,14 @@ public static List<RNA> LoadRnaXML(string rnaDbLocation, bool generateTargets, D
                         }
                         if (xml.NodeType == XmlNodeType.EndElement || xml.IsEmptyElement)
                         {
-                            RNA newProtein = block.ParseRnaEndElement(xml, modTypesToExclude, unknownModifications, isContaminant, rnaDbLocation, decoyIdentifier, entrapmentIdentifier, isEntrapment);
+                            if (block.Sequence != null && !readFirstSequence)
+                            {
+                                if (block.Sequence.IsAllLower())
```
trishorts: **Minor — same first-entry-only sampling on the XML path.**

Same shape as my comment on the FASTA path: `readFirstSequence` latches on the first entry that has a sequence, so an XML db whose first entry is uppercase and whose later entries are lowercase gets no `ToUpper` and throws part-way through the read.

Worth saying explicitly that this one is *not* a regression — on `master` the lowercase XML path already threw, because `ProteinDbLoader.SanitizeAminoAcidSequence(Sequence, 'X')` coerced lowercase to `'X'` and `X` is not a nucleotide. Replacing that call with `SanitizeAndTransform` is a genuine improvement, and it's also what lets `Ψ` survive the XML reader at all. I'm only flagging that the new normalization handles the uniform-lowercase file and not the mixed one, so whatever you settle on for FASTA should apply here too.

mzLib/Omics/Resources/RnaMods.txt
```diff
@@ -126,6 +126,13 @@ MT   Simple Biological
 CF   C3 H6 O1
 //
 ################################## Base Conversion
+ID   A->I
```
trishorts, line 129: **Question — the new `I` residue and this mod are exactly mass-degenerate.**

`InosineBase` is `C5H3N4O1` and `AdenineBase` is `C5H4N5`, so the residue difference is `N-1H-1O1` — precisely the `CF` on this entry. Measured on this branch: adenine 329.05252012475, inosine 330.03653570766, delta 0.98402 Da, the same delta this mod applies.

So an inosine at a given position can now be represented two indistinguishable ways: as the residue `I` in the sequence, or as an `A` carrying `A->I`. A search with both enabled scores them identically and has no signal to prefer either.

Is that intentional — residue for databases that already encode `I`, mod for discovering A-to-I editing against an unedited database? If so it's worth a line in this file saying so, otherwise the next person to read `RnaMods.txt` will see it as a duplicate.

trishorts (commented):
Reviewed the nucleotide-alias and lowercase-normalization payload. My three earlier threads are addressed and resolved.

One **Major** left: removing the lowercase char lookup is right, but the `IsAllLower()`-on-the-first-record heuristic that replaces it is narrower than the behaviour it removes, so mixed-case (soft-masked) and case-varying multi-record FASTA files that load on `master` now throw. Verified by running both cases on this branch and on the merge-base. Details inline, plus a minor note on the XML path and a question about the `I` residue / `A->I` mass degeneracy.

mzLib/Omics/Resources/RnaMods.txt
nbollis: yes this was intentional. It allows a base conversion to be discovered by GPTMD or the non-canonical base directly encoded in the sequence database

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: Good catch on the first-record sampling. I changed the reader to check each sequence independently: only a fully-lowercase sequence gets an uppercase transform applied to it, and the transform is scoped to that sequence rather than latched per file. So a file whose first record is uppercase and whose second record is lowercase now loads both correctly, and mixed-case sequences keep their case as-is (case is preserved for future case-distinct one-letter modified-nucleotide codes). Added a regression test for the record-1-upper/record-2-lower FASTA case.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: Agreed, and I applied the same per-entry approach to the XML path: each entry's sequence is checked independently, so an all-lowercase entry gets normalized to uppercase while mixed-case and uppercase entries keep their case. Added a regression test with a first-uppercase/lowercase-second XML entry. The `SanitizeAndTransform` replacement of the coercing `SanitizeAminoAcidSequence` remains, so non-canonical letters like Ψ still survive the XML reader.

mzLib/Omics/Resources/RnaMods.txt
nbollis: Confirmed — this was intentional, as noted in the thread: the `I` residue accommodates databases that already encode inosine, while `A->I` is for discovering A-to-I editing against an unedited database. The two mass-degenerate representations are kept deliberately; no additional documentation was added to RnaMods.txt.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Still open — I can't find these changes on the branch.

`RnaDbLoader.cs` last changed in `128f55db` (2026-09-03), and the newest commit on the branch is `2b5bed52`, a master merge from **2026-09-06T17:29Z** — 72 minutes *before* I posted the comment above. The `readFirstSequence` latch is still there verbatim at `RnaDbLoader.cs:244-249`, and `sequenceTransformations` is still a file-level list. I suspect the work is sitting on an unpushed local branch.

Re-ran the same probe against head `2b5bed52`, with `smith/master` (`8e555e96`) as the control:

| FASTA case | `master` | `2b5bed52` |
|---|---|---|
| all-lower, 1 record | loads | loads |
| **soft-masked `GUACUGCCUcuagugaagca`, 1 record** | loads | **`ArgumentException: Nucleic Acid Letter c does not exist...`** |
| **rec1 UPPER, rec2 lower** | loads both | **`ArgumentException: Nucleic Acid Letter g does not exist...`** |
| rec1 lower, rec2 UPPER | loads both | loads both |

Unchanged from what I measured on 09-06, and still a regression against master on two of four cases. The throw escapes `LoadRnaFasta` rather than landing in `errors`, so it aborts the whole read.

One thing in your reply I want to flag separately, because it isn't just a coding detail: *"mixed-case sequences keep their case as-is (case is preserved for future case-distinct one-letter modified-nucleotide codes)."* If that's the intended design, then soft-masked FASTA — the Ensembl/UCSC repeat-masking convention, and the first row in the table above — is a permanent hard failure rather than something the per-sequence fix would have caught. Reserving case for future modified-nucleotide codes and accepting soft-masked input are in direct conflict, and that's a design call worth making explicitly rather than as a side effect. If you do want to reserve case, the read should reject soft-masked input with a clear per-entry message in `errors`, not an `ArgumentException` from the residue table.

Also no regression test for the record-1-upper/record-2-lower case landed — `TestDbReadingDifferentExtensionsWithLowercaseSequence` is still the only lowercase test, and still covers only the uniform case that already works.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Same as the FASTA thread — the per-entry change isn't on the branch. `readFirstSequence` still latches on the first entry with a sequence at `RnaDbLoader.cs:379-384`, and `transformsToApply` is still file-scoped.

Measured on head `2b5bed52`:

| XML case | `master` | `2b5bed52` |
|---|---|---|
| all-lower, 1 entry | throws (`X` coercion) | loads |
| soft-masked, 1 entry | throws (`X` coercion) | throws (`Letter c`) |
| **entry1 UPPER, entry2 lower** | throws (`X` coercion) | **throws (`Letter g`)** |
| entry1 lower, entry2 UPPER | throws (`X` coercion) | loads both |

Restating what I said last time so the severity stays honest: this is **not** a regression. `master` fails all four, and the `SanitizeAndTransform` replacement of `SanitizeAminoAcidSequence` is a real improvement that fixes two of them — it's also what lets `Ψ` survive the XML reader. I'm only noting that whatever you land for FASTA should apply here too. Not blocking on its own.

mzLib/Omics/Resources/RnaMods.txt
trishorts: Thanks — that answers it. Residue `I` for databases that already encode inosine, `A->I` for discovering A-to-I editing against an unedited database; the mass degeneracy is deliberate and the two are reached by different routes. Happy to leave `RnaMods.txt` undocumented now that the rationale is on the record here. Resolving.

trishorts (changes requested):
Not approving yet — one blocking item, and it's a mechanical problem rather than a disagreement.

**The three fix replies posted 2026-09-08 describe changes that aren't on the branch.** `RnaDbLoader.cs` last changed in `128f55db` (2026-09-03); the newest commit is `2b5bed52`, a master merge from 2026-09-06T17:29Z — 72 minutes before I posted the review those replies answer. The `readFirstSequence` latch is still present verbatim on both the FASTA (`RnaDbLoader.cs:244-249`) and XML (`:379-384`) paths. Best guess is an unpushed local branch; a `git push` may be all this needs.

Re-measured on head `2b5bed52` against `smith/master` (`8e555e96`) as control:

| FASTA case | `master` | `2b5bed52` |
|---|---|---|
| soft-masked `GUACUGCCUcuagugaagca` | loads | **throws** |
| rec1 UPPER, rec2 lower | loads both | **throws** |
| rec1 lower, rec2 UPPER | loads both | loads both |
| all-lower, 1 record | loads | loads |

Two of four cases regress against master, and the `ArgumentException` escapes `LoadRnaFasta` instead of landing in `errors`, so it aborts the entire DB read rather than skipping one entry. The XML path shows the same shape but is **not** a regression — master fails all four there — so it's a net improvement and non-blocking on its own.

**Threads:**
- `RnaMods.txt` inosine/`A->I` mass degeneracy — answered and **resolved**. Deliberate: residue for databases that encode inosine, mod for discovering A-to-I editing. Rationale is on the record in the thread.
- FASTA normalization — **still open**, verified unfixed.
- XML normalization — still open, non-blocking, follows whatever FASTA lands.

One design question inside the FASTA thread that I'd rather settle before merge than after: the reply says mixed-case sequences deliberately keep their case, to reserve case for future modified-nucleotide codes. That's defensible, but it makes soft-masked FASTA (the Ensembl/UCSC repeat-masking convention) a permanent hard failure rather than an oversight. Reserving case and accepting soft-masked input are in conflict — if you want to reserve case, the reader should reject soft-masked input with a per-entry message in `errors`, not an `ArgumentException` out of the residue table.

For the record, the three earlier findings (Turkish-locale `Char.ToLower`, the non-atomic alias registration, the unreachable single-character string alias) are genuinely fixed and remain resolved. `dotnet test` across Transcriptomics / Nucleotide / DbLoader: 426 passed, 0 failed — the suite has no coverage for either failing case, which is why it's green.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: This is intentional. Modomics one letter codes are upper and lower case characters to represent modified residues. The intention is to support reading those in after this

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: agreed. I had a branch issue so that change did not get pushed last time

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Confirmed fixed on head `32b5c904` — thanks for pushing it through.

Re-ran the same probe I used before. The per-sequence decision is now at `RnaDbLoader.cs:259`, and the latch is gone:

| FASTA case | `master` | `2b5bed52` (before) | `32b5c904` (now) |
|---|---|---|---|
| all-lower, 1 record | loads | loads | loads |
| **record 1 UPPER, record 2 lower** | loads | **throws (`Letter g`)** | **loads both** |
| soft-masked (mixed) | loads | throws (`Letter c`) | throws (`Letter c`) |

The mixed-case row is the one I was asking about and it is resolved. And understood on soft-masking — keeping case significant so Modomics case-distinct one-letter codes can be read later is a deliberate trade, and it is your call; I am not asking for `IsAllLower` to be loosened.

Resolving. Two small things I noticed in the new normalization code are on separate comments so they do not get lost in this thread.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Confirmed fixed on head `32b5c904`.

Per-entry now at `RnaDbLoader.cs:383`, and `SanitizeAndTransform` replaces the `'X'`-coercing `SanitizeAminoAcidSequence` at `ProteinXmlEntry.cs:490` so non-canonical letters survive the XML reader.

| XML case | `master` | `2b5bed52` (before) | `32b5c904` (now) |
|---|---|---|---|
| all-lower, 1 entry | throws (`X` coercion) | loads | loads |
| **entry 1 UPPER, entry 2 lower** | throws (`X` coercion) | **throws (`Letter g`)** | **loads both** |
| soft-masked, 1 entry | throws (`X` coercion) | throws (`Letter c`) | throws (`Letter c`) |

No worries on the branch mix-up. Resolving.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -463,19 +483,20 @@ public static IEnumerable<RNA> Merge(IEnumerable<RNA> mergeThese)
...
-                case SequenceTransformationOnRead.None:
-                    return cleanedSequence;
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(sequenceTransformation), sequenceTransformation, null);
-            }
+            if (sequenceTransformations.Count == 0)
+                return cleanedSequence;
+
+            if (sequenceTransformations.Contains(SequenceTransformationOnRead.ToUpper))
+                cleanedSequence = cleanedSequence.ToUpper();
```
trishorts: **Major — `ToUpper()` is culture-sensitive, and this PR just added `I` as a residue.**

`cleanedSequence.ToUpper()` uses the *current* culture. Under `tr-TR`/`az-AZ`, `'i'` maps to `'İ'` (U+0130), not `'I'` — so the lowercase normalization this method exists to perform emits a letter no residue table knows, and the whole DB read aborts.

This is the same class of defect as the `Char.ToLower` call we removed from `AddResidueToDictionary`, just in the opposite direction. It was unreachable before because there was no `I` residue; inosine makes it reachable.

Reproduced clean-process on head `32b5c904`, all-lowercase sequence `guacigccucuagugaagca`:

| culture | `'i'.ToUpper()` | FASTA | XML |
|---|---|---|---|
| `en-US` | U+0049 `I` | loads `GUACIGCCUCUAGUGAAGCA` | loads `GUACIGCCUCUAGUGAAGCA` |
| `tr-TR` | U+0130 `İ` | **throws** `Letter İ does not exist` | **throws** `Letter İ does not exist` |
| `az-AZ` | U+0130 `İ` | **throws** | **throws** |

Fix is one word: `ToUpperInvariant()`. Residue letters are an ASCII alphabet, never user-facing text, so the invariant mapping is the correct one. (mzLib is split roughly evenly between the two spellings today, so this is not a house-style point — it is that *this* call site is the one where the distinction has teeth.)

Flagging rather than blocking: two locales and a one-word change. I would like it in before the next release.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -249,7 +252,15 @@ public static List<RNA> LoadRnaFasta(string rnaDbLocation, bool generateTargets,
                             }
                         }
 
-                        var sequence = SanitizeAndTransform(sb.ToString(), sequenceTransformation);
+                        string rawSequence = sb.ToString();
+                        var transformsForThisSequence = transformPool.Get();
+                        transformsForThisSequence.AddRange(sequenceTransformations);
+
+                        if (rawSequence.IsAllLower())
```
trishorts, line 259: **Minor — `IsAllLower()` is evaluated on the raw sequence, before `SanitizeAndTransform` strips whitespace.**

`sb?.Append(line.Trim())` trims only line ends, so whitespace *inside* a sequence line survives into `rawSequence`. `char.IsLower(' ')` is false, so a single space anywhere makes `IsAllLower()` false, `ToUpper` never gets added, and the lowercase sequence reaches the residue lookup unchanged.

Measured on `32b5c904`:

| FASTA sequence | result |
|---|---|
| `guacugccucuagugaagca` | loads `GUACUGCCUCUAGUGAAGCA` |
| `guacug ccucuagugaagca` | **throws** `Letter g does not exist` |
| same, split across two lines | loads (`line.Trim()` handles it) |

So the multi-line case is fine — it is intra-line whitespace that bites. `SubstituteWhitespace` (`\s+`) exists precisely because sequences carry whitespace, so the case decision wants to be made on the *cleaned* string: either clean first and test that, or move the `IsAllLower` check inside `SanitizeAndTransform` after the `Replace`.

Not a live bug on the XML path — `ProteinXmlEntry.cs:208` already strips whitespace before `Sequence` is set, so `block.Sequence.IsAllLower()` at line 383 sees a clean string — but worth keeping the two paths consistent.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -46,6 +48,7 @@ public static class RnaDbLoader
         private static readonly Regex _ncbiAssemblyHeaderRegex = new Regex(@"^>NM_\d+\.\d+ ", RegexOptions.Compiled);
         private static readonly Regex _ncbiRefSeqGeneHeaderRegex = new Regex(@"^>NC_\d+\.\d+:", RegexOptions.Compiled);
 
+        private static readonly ListPool<SequenceTransformationOnRead> transformPool = new(4);
```
trishorts, line 51: **Minor — pooled list is not returned on the exception path; and a `[Flags]` enum would retire the pool entirely.**

Four small things converge on this field, so they are in one comment.

1. `ListPool`'s remarks say the list "should be pulled from outside a try finally loop and finally should return the List to the pool." Neither site does. If `SanitizeAndTransform` throws between `Get()` and `Return()` (lines 256/263), or `ParseRnaEndElement` throws (lines 380/387), the list is never returned. This is only a pool leak, not corruption — `ListPooledObjectPolicy.Return` does `Clear()`, so no stale contents can leak into the next `Get()`. The in-repo precedent is explicit about it: `DigestionAgent.GetDigestionSiteIndices` wraps its rented `HashSet` in `try`/`finally` with the comment *"Try block is to ensure that, even if an error gets thrown, the hashset is returned to the pool."*

2. On the XML path the `Get()`/`AddRange`/`Return()` triple runs on **every** end element, not just `</entry>` — once per `<feature>`, `<organism>`, `<name>`, and so on, rather than once per entry.

3. `sequenceTransformations.Add(...)` at lines 202/207/212 mutates a caller-supplied list. `headerType is null` guards it to fire once per file, so it cannot grow unboundedly, and the `Contains` checks make a duplicate harmless — but a caller reusing one list across two `LoadRnaFasta` calls gets `ConvertAllTtoU` appended twice, which is a surprising side effect on a parameter.

4. Worth asking whether the pool earns its keep at all: the list holds at most two enum values for a few microseconds. `SequenceTransformationOnRead` appears only in `RnaDbLoader.cs` and `ProteinXmlEntry.cs`, so making it `[Flags]` and threading a single value through `SanitizeAndTransform` would drop the list, the pool, points 1-3, and the `IList<>` parameter — net less code than the current shape.

None of this blocks; 1 is the one I would actually fix.

trishorts (dismissed):
Round 3 review. **Approving.**

Both blockers from my previous round are fixed, and I verified each against head `32b5c904` rather than taking the diff's word for it — replies with the before/after tables are on those two threads, now resolved.

| previously open | status on `32b5c904` |
|---|---|
| FASTA: `readFirstSequence` latch meant record 1 UPPER + record 2 lower threw | **fixed** — per-sequence decision at line 259, both records load |
| XML: same latch per file | **fixed** — per-entry decision at line 383, both entries load |
| `A->I` mass-degenerate with residue `I` | **answered** — deliberate, two different discovery routes; resolved last round |

Also checked and found clean: `TryAddAlternativeRepresentationCore` now validates both lookup tables *before* mutating either, so the half-written-state bug is gone and the all-or-nothing contract holds; single-character string aliases route through the char path, so registration, `GetResidue`, `TryGetResidue` and sequence parsing finally agree; dropping the lowercase table entry removed the `Char.ToLower` Turkish hazard; `Ψ` resolves via the non-ASCII dictionary fallback. The added tests pin real behaviour, including the negative cases (exact-case-only lookups, the ASCII alias clash leaving the original mapping intact) — those are the ones I would have asked for.

Three non-blocking findings inline, all reproduced on `32b5c904`:

- **Major** — `SanitizeAndTransform`'s `ToUpper()` is culture-sensitive, and the new `I` residue makes it bite: under `tr-TR`/`az-AZ` a lowercase sequence containing inosine aborts the DB read. One word, `ToUpperInvariant()`. Same class of bug as the `Char.ToLower` one this PR removed.
- **Minor** — `IsAllLower()` is tested against the raw sequence, before whitespace is stripped, so an intra-line space defeats the lowercase normalization on the FASTA path.
- **Minor** — the pooled transform list is not returned on the exception path, against the repo's own `try`/`finally` precedent in `DigestionAgent`; a `[Flags]` enum would remove the pool and two related wrinkles outright.

I am not holding the merge for any of them, though the `ToUpperInvariant()` change is worth making before this reaches a release. Soft-masked/mixed-case sequences still throw by design, which you have explained as keeping case significant for Modomics case-distinct codes — your call, recorded, not contested.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Fixed and pushed directly to this branch as `d0f7596a` — hope that is alright, it was a one-word change and easier than another round trip. Revert it if you would rather take it yourself.

`ToUpper()` -> `ToUpperInvariant()`, plus a regression test in `TestDbLoader` over `tr-TR`/`az-AZ`/`en-US`. I checked the test actually catches the defect rather than just passing: reverting the one word makes the `tr-TR` and `az-AZ` cases fail and leaves `en-US` green. Full `Test.Transcriptomics` suite is 432/432.

One correction to my comment above, in your favour. After posting it I found `Loaders`' static constructor calls `SetCultureAsInvariantCulture()`, which sets `CultureInfo.DefaultThreadCurrentCulture` process-wide — so in any application that touches `UsefulProteomicsDatabases.Loaders` before reading a database (MetaMorpheus does, when it loads mods) the culture is already invariant and this was never reachable. My repro was a bare process that set its culture and called `RnaDbLoader` directly, which is a real but narrower scenario than "the whole DB read aborts" implied. So: still worth fixing, and now fixed, but I overstated the blast radius and should have found that before posting rather than after.

Resolving.

trishorts (approved):
Re-approving. My own push (`d0f7596a`) dismissed the approval I posted a few minutes ago, so this replaces it — nothing else has changed in my assessment.

That commit is the `ToUpperInvariant()` fix for the Major I raised, made directly on this branch at trishorts' request, with a regression test across `tr-TR`/`az-AZ`/`en-US`. Verified the test fails on the two Turkic locales with the old `ToUpper` and passes on all three with the fix; `Test.Transcriptomics` is 432/432.

The two Minor findings from my previous review are still open and still non-blocking — the `IsAllLower()`-before-whitespace-strip ordering on the FASTA path, and the pooled list not being returned on the exception path. Both are yours to take or leave.

I also posted a correction on the `ToUpper` thread: `Loaders`' static constructor already forces invariant culture process-wide, so that defect was narrower in practice than my original comment implied. The fix is still right; my severity framing was not.

RayMSMS (approved):

PR #1277, opened 2026-09-02
Add Reader for Pytheas RNA Search Match Output

nbollis:
## Motivation

Pytheas is an RNA database search tool (Smith lab, https://github.com/smith-chem-wisc/Pytheas). Its match-output files describe spectrum-to-oligomer identifications, but mzLib had no reader for them, so Pytheas results could not be loaded into the Readers framework or compared against MetaMorpheus results.

## Summary of changes

- Add `Software.Pytheas` and `SupportedFileType.PytheasResult`.
- Add `PytheasResult`, a record for one match line of the space-delimited Pytheas match output, and `PytheasMatchedIon` for the normalized fragment-ion tokens. Raw strings are retained on each record so writing round-trips the file unchanged.
- Add `PytheasResultFile : ResultFile<PytheasResult>` that parses the custom line format (16 fixed columns, variable-length MS2 match section, final SCORE token), exposes the `#` metadata header lines as typed properties (enzyme, ppm tolerances, m/z windows, alpha/beta, etc.), and preserves the `PRECURSOR_ION=` grouping on write.
- Register the file type in `SupportedFileTypes`: `.txt` extension plus header-based detection (`#theoretical_digest`) for both `.txt` and renamed `.tsv` files, mirroring the DIA-NN detection approach.
- Exclude the unrelated Visual Studio 17->18 solution-file change that was already on the branch.

## Testing

- New `TestPytheasResultFile` suite against a real Pytheas match output (4744 match lines / 822 precursors): load count, first/last record fields, typed header properties, rename-by-header detection, factory read through `FileReader.ReadResultFile`, write/read round trips, and normalized `PytheasMatchedIon`/interpreted-field checks.
- Extended `TestSupportedFileExtensions` with the Pytheas fixture case.
- Verified: `dotnet build` 0 errors; FileReadingTests pass (1062/1062).


Alexander-Sol (dismissed):

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
```diff
@@ -0,0 +1,203 @@
...
+    public override SupportedFileType FileType => SupportedFileType.PytheasResult;
+    public override Software Software { get; set; }
+
+    /// <summary>The '#' metadata lines at the top of the file, kept so writing round-trips them.</summary>
+    public List<string> HeaderLines { get; } = new();
+
+    #region Header Properties
+
+    /// <summary>Path of the theoretical digest file, from "#theoretical_digest".</summary>
+    public string TheoreticalDigestPath { get; private set; }
```
trishorts: **Major — every typed header property silently returns `null`/`0` until something touches `Results`.**

`ParseHeaderLines()` runs inside `LoadResults()`, and `LoadResults()` is only invoked lazily by the `Results` getter. So immediately after construction, `Enzyme` is `null`, `Ms1Ppm` is `0`, `Alpha`/`Beta` are `0`, and `HeaderLines` is empty — with no exception to tell the caller the values aren't real yet.

I measured it on the fixture: `HeaderLines.Count` read before touching `Results` is **0**; after the load it's 21.

The tell is that the PR already works around this in four places — `TestPytheasHeaderPropertiesParsed` and `TestPytheasResultHeaderLinesRoundTrip` both open with `_ = file.Count();` before asserting anything, and `WriteResults` has an explicit `// Reading Results forces the lazy load, which populates HeaderLines.` A consumer who hasn't read this file won't know to do that, and `Ms1Ppm == 0` reads as a legitimate tolerance rather than "not loaded".

Simplest fix is to make the header properties force the load the same way `Results` does — back them with fields and have the getters do `_ = Results;` first — or parse the `#` block eagerly in the constructor, since it's only the first ~21 lines.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
```diff
@@ -0,0 +1,203 @@
...
+        string precursorIon = null;
+
+        foreach (var line in File.ReadLines(FilePath))
+        {
+            if (line.Length == 0)
+                continue;
+
+            if (line.StartsWith("#", StringComparison.Ordinal))
+            {
+                HeaderLines.Add(line);
```
trishorts: **Major — `HeaderLines` is never cleared, so a second `LoadResults()` duplicates every header line.**

`Results` is reassigned on each load, but `HeaderLines` is a `{ get; } = new()` that only ever accumulates. Calling `LoadResults()` again — directly, or via any future reload path — appends the whole `#` block a second time, and `WriteResults` then emits every header twice, producing a file that no longer round-trips.

Measured on the fixture: 21 header lines after the first load, **42** after calling `LoadResults()` a second time.

One line fixes it: `HeaderLines.Clear();` at the top of `LoadResults()`.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        foreach (var header in HeaderLines)
+            writer.WriteLine(header);
+
+        string lastPrecursorIon = null;
+        foreach (var result in results)
+        {
+            if (result.PrecursorIon != lastPrecursorIon)
+            {
+                writer.WriteLine();
```
trishorts: **Minor — the write does not round-trip the file unchanged, contrary to the PR description and the `RawLine`/`RawMatch` doc comments.**

This emits a blank line before *every* `PRECURSOR_ION` group including the first, but the fixture has **two** blank lines between the `#` header block and the first `PRECURSOR_ION`. Everything from there on is shifted by one line.

Read → write on the fixture: 5,554,117 bytes in, 5,554,115 out; 6,410 lines in, 6,409 out; first divergence at line 22.

The data itself survives — re-reading the written file gives back all 4,744 records with identical `RawLine`s — so this is formatting fidelity, not data loss. But `PytheasResult.RawLine` says "kept verbatim so that writing round-trips the file", `PytheasMatchedIon.RawMatch` says the same, and the PR body says "writing round-trips the file unchanged", and none of that is quite true today. `TestPytheasResultReadWriteRoundTrip` asserts record equality rather than bytes, so it passes either way.

Either preserve the original blank-line layout (a byte-comparison assertion in the round-trip test would pin it), or soften the three claims to "round-trips every record unchanged".

mzLib/Readers/Util/SupportedFileTypes.cs
```diff
@@ -182,6 +190,15 @@ public static SupportedFileType ParseFileType(this string filePath)
                 }
 
                 case ".txt":
+                    using (var reader = new StreamReader(filePath))
```
trishorts: **Major — `ParseFileType` on a `.txt` path now requires the file to exist and be openable; it used to be a pure string check.**

Verified against the merge-base (`a8250440`): on `master`, `ParseFileType` for *any* `.txt` path returned `CruxResult` without touching the disk, because `CruxResult.GetFileExtension()` is `".txt"` so the `EndsWith` was always true. On this branch the `StreamReader` runs first, so:

- a `.txt` path that doesn't exist yet now throws `FileNotFoundException`
- a `.txt` file another process holds with `FileShare.None` now throws `IOException`

Both were confirmed by probe. Note these are the only paths out of `ParseFileType` that throw something other than `MzLibException`, so a caller catching `MzLibException` — which is what every other branch of this method promises — won't catch them. I didn't find an in-repo caller that classifies a not-yet-existing `.txt`, so this may well be harmless in practice; flagging it because `ParseFileType` is public and the contract changed.

Guarding the block with `File.Exists(filePath)` and falling through to the existing name-based Crux check restores the old behavior for missing paths, and `new StreamReader(new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))` covers the locked-file case.

Separately: this loop scans five lines, but the `.tsv` branch above checks only `firstLine` and its comment says "The first line is always `#theoretical_digest`". Worth making the two consistent — if the first line is guaranteed, one `ReadLine()` here does it and Crux `.txt` files stop being partially read on every classification.

mzLib/Readers/ExternalResults/IndividualResultRecords/PytheasResult.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// "892.167114(0.8ppm)[680]:892.166417[M-P(-1)] 362.051178(2.8ppm)[100]:362.050174[y1(-1)]".
+    /// Kept as text rather than parsed so the line round-trips unchanged.
+    /// </summary>
+    public string Ms2Matches { get; set; }
+
+    /// <summary>
+    /// The MS2 <see cref="Ms2Matches"/> tokens parsed into typed <see cref="PytheasMatchedIon"/> records.
+    /// Populated only when the tokens are requested; reading stays a single pass over the raw line.
+    /// </summary>
+    public List<PytheasMatchedIon> MatchedIons => _matchedIons ??= ParseMatchedIons();
```
trishorts: **Minor — a property getter that throws.**

`MatchedIons` defers parsing, and `PytheasMatchedIon.Parse` throws `MzLibException` on a token the regex doesn't match. So a malformed file loads cleanly and then throws from a property access at some arbitrary later call site — confirmed: reading `MatchedIons` on a record with an unparseable token throws `MzLibException: Could not parse Pytheas MS2 match token: ...`.

For what it's worth the parser handles the real data fine — all **95,489** ion tokens across the 4,744 fixture records parse without a single failure — so this is robustness, not a live bug. But an exception out of a getter is awkward to defend against. Either make it a method (`GetMatchedIons()`), or keep the property and have it skip/collect unparseable tokens instead of throwing.

trishorts (commented):
Reviewed the Pytheas reader. The line-format parser is solid — I ran every MS2 token in the fixture through it and all 95,489 parse, and a read/write/re-read gives back all 4,744 records with identical `RawLine`s. Nice job on the header-based detection comment too.

Two things I'd want fixed before merge, both in `PytheasResultFile`: the typed header properties are silently empty until something touches `Results`, and `HeaderLines` is never cleared so a second load duplicates it. One more on the `.txt` branch of `ParseFileType`, which quietly became a disk operation. Plus two minors. Everything below was checked by running it, against this branch and against the merge-base `a8250440`.

trishorts (commented):
Re-reviewed at your request. Nothing has landed since my 2026-09-06 pass — the newest commit is a master merge from 2026-09-02, and `RnaDbLoader`-style follow-ups aside, none of the five threads has been touched. Rather than repeat myself inline, I re-verified the two that would cost the most to ship and left the threads where they are.

**Still true — `ParseFileType` on a `.txt` path now requires the file to exist and be openable.** `SupportedFileTypes.cs:193` opens `new StreamReader(filePath)` unconditionally at the top of the `.txt` case. On master that case was a pure string comparison, so `"whatever.txt".ParseFileType()` was answerable without touching disk. It now throws `FileNotFoundException` for a path that does not exist and `IOException` for one another process holds — neither of which is `MzLibException`, so nothing catching the documented exception type will catch these. The `.tsv` case reads the file too, so there is precedent, but it is precedent for `.tsv`, and callers that pass speculative `.txt` paths have had no reason to expect I/O. The probe also runs *before* the Crux check, so every Crux file now pays an open and five `ReadLine`s to reach a comparison that never needed them — moving the block below the `EndsWith` would cost nothing and shrink the blast radius to files that are not already recognized.

**Still true — `HeaderLines` is never cleared.** It is `{ get; } = new()` (:23) and `LoadResults()` only ever `Add`s to it (:109) while reassigning `Results` (:123). A second `LoadResults()` therefore leaves `Results` correct and `HeaderLines` doubled — and `WriteResults` iterates `HeaderLines` (:194), so the re-read-then-write path emits every header line twice. One `HeaderLines.Clear()` at the top of `LoadResults()` closes it.

The other three (lazy `ParseHeaderLines` leaving typed header properties null/0, the blank line before every `PRECURSOR_ION` group breaking round-trip, the throwing `MatchedIons` getter) I have not re-measured, but nothing in the diff has moved under them.

Not approving yet — the two above are the ones I would want closed first. Happy to look again as soon as there is a push.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
nbollis: Good catch — a header accessor shouldn't depend on someone having touched `Results` first. I've made the typed header properties (and `HeaderLines`) force the lazy load before returning, so `Enzyme`, `Ms1Ppm`, `Alpha`, etc. are populated immediately after construction, keeping the lazy file read as the single load path.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
nbollis: Agreed. Added `HeaderLines.Clear()` at the top of `LoadResults()` so a second load resets the header block instead of doubling it (21 → 42). Added a regression test that reloads and asserts the count stays stable.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
nbollis: You're right — `WriteResults` emitted one blank line before every group while the fixture has two before the first. Write now reproduces the original per-group layout (tracked during load), and a byte-comparison test pins it — the written file is byte-identical to the fixture, so the round-trip doc claims are now exact.

mzLib/Readers/Util/SupportedFileTypes.cs
nbollis: Agreed the contract changed. I've guarded the probe with `File.Exists(filePath)` so a missing `.txt` falls through to the name-based Crux result as before, and opened via `FileStream(FileMode.Open, FileAccess.Read, FileShare.ReadWrite)` to tolerate files another process holds for read. One correction to the rationale: moving the probe below the `EndsWith` check would break Pytheas-in-`.txt` detection, because `CruxResult`'s extension is `.txt` so `EndsWith(".txt")` is always true — the header probe has to stay above it. I kept the five-line scan rather than matching the `.tsv` first-line-only check so a renamed Pytheas file with a leading non-`#` line still classifies; happy to align if you prefer strict first-line matching.

mzLib/Readers/ExternalResults/IndividualResultRecords/PytheasResult.cs
nbollis: Agreed — a throwing getter is hard to defend against. I've renamed it to `GetMatchedIons()` and had the parse skip malformed tokens, collecting them in `UnparseableMs2Matches` so no exception escapes and no data is silently dropped without a trace. A malformed token now leaves the parsed ions intact alongside the raw token for inspection.

nbollis:
Thanks for the thorough pass. The five findings were tracked in their inline threads; all are fixed now (header properties load eagerly on access, `HeaderLines` resets per load, the `.txt` probe is guarded by `File.Exists`, the write is byte-faithful, and `GetMatchedIons()` no longer throws).

nbollis:
Thanks for re-checking. Both blockers are closed: `HeaderLines` is cleared at the top of `LoadResults()`, and the `.txt` probe in `ParseFileType` is guarded by `File.Exists` and opened with `FileShare.ReadWrite` (kept above the Crux name check so renamed Pytheas `.txt` files still classify). The other three items are addressed too. This was just pushed — happy to look again.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
trishorts: Verified fixed at `867081ba`. All 22 typed header getters plus `HeaderLines` now call `EnsureLoaded()` before returning, and `LoadResults()` writes the backing fields directly so there's no getter recursion. `TestPytheasHeaderPropertiesAvailableWithoutTouchingResults` pins it with no `Results`/`Count()` touch first. The two tests that used to open with `_ = file.Count();` no longer need to. Resolving.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
trishorts: Verified fixed at `867081ba`. `LoadResults()` now opens with `_headerLines = new List<string>()` (and `_blankLinesBeforeGroups.Clear()`), so a second load rebuilds the header block instead of appending to it. `TestPytheasResultHeaderLinesDoNotDuplicateOnReload` pins 21 -> 21 across an explicit reload. Resolving.

mzLib/Readers/ExternalResults/ResultFiles/PytheasResultFile.cs
trishorts: Verified fixed at `867081ba`. `_blankLinesBeforeGroups` records the run of blank lines preceding each `PRECURSOR_ION` during load and `WriteResults` replays it per group, so the two-blank-line gap before the first group survives. `TestPytheasResultReadWriteByteRoundTrip` compares `File.ReadAllBytes` on both sides, which is the assertion that was missing — the doc claims on `RawLine`/`RawMatch` and in the PR body are now exact. Resolving.

mzLib/Readers/Util/SupportedFileTypes.cs
trishorts: Verified fixed at `867081ba`. The `File.Exists(filePath)` guard restores the old pure-string behavior for a `.txt` path that doesn't exist yet (falls through to the name-based Crux check), and the explicit `FileStream(..., FileShare.ReadWrite)` covers a file another process holds open for writing.

And you're right about the ordering — the probe has to stay above the `EndsWith` check precisely because `CruxResult`'s extension is `.txt`. I wasn't asking you to move it, just to guard it, so nothing to correct there.

Keeping the five-line scan over strict first-line matching is your call and a defensible one; the cost is that Crux `.txt` files get partially read on every classification, which is not worth another round. Resolving.

mzLib/Readers/ExternalResults/IndividualResultRecords/PytheasResult.cs
trishorts: Verified at `867081ba` — `GetMatchedIons()` is a method now, and malformed tokens land in `UnparseableMs2Matches` instead of throwing, which is exactly what I asked for. `TestPytheasResultGetMatchedIonsCollectsUnparseableTokens` covers it.

Resolving this one, but there's a residual leak I've raised separately on `PytheasMatchedIon.cs` — `catch (MzLibException)` doesn't cover the `double.Parse` calls inside `Parse`, so one shape of malformed token still escapes. Non-blocking.

mzLib/Readers/ExternalResults/IndividualResultRecords/PytheasMatchedIon.cs
```diff
@@ -0,0 +1,80 @@
...
+            offsetPpm: ParseDouble(match.Groups[2].Value),
+            normalizedIntensity: ParseDouble(match.Groups[3].Value),
+            theoreticalMz: ParseDouble(match.Groups[4].Value),
+            ionAnnotation: annotation,
+            charge: charge,
+            rawMatch: token);
+    }
+
+    private static double ParseDouble(string value) =>
+        double.Parse(value, NumberStyles.Float, CultureInfo.InvariantCulture);
```
trishorts, line 79: **Minor (non-blocking) — `catch (MzLibException)` doesn't cover this, so one shape of malformed token still escapes `GetMatchedIons()`.**

`ParseMatchedIons` catches `MzLibException`, which is what `Parse` throws when `TokenRegex` fails to match. But when the regex *does* match, these `double.Parse` calls run unguarded — and the numeric groups are `[\d.]+`, which happily matches strings `double.Parse` rejects.

Probed the regex against the real pattern:

| token | `regex.Success` | result |
|---|---|---|
| `892.167114(0.8ppm)[680]:892.166417[M-P(-1)]` | true | parses, 892.167114 |
| `892.16.7114(0.8ppm)[680]:892.166417[M-P(-1)]` | true | **`FormatException`** on group 1 |
| `...(0.8ppm)[680]:892.166417[M-P(-1)]` | true | **`FormatException`** on group 1 |
| `garbage` | false | `MzLibException` — caught, collected |

`FormatException` derives from `SystemException`, not `MzLibException`, so it propagates straight out of `GetMatchedIons()` — past the `try` that exists to prevent exactly that. `int.Parse` on line 65 is the same shape (`OverflowException` on an absurd charge), though that one is hard to reach in practice.

That makes the new doc comment on `GetMatchedIons` — "Tokens that do not match the expected layout are skipped and collected in `UnparseableMs2Matches` rather than throwing, so malformed input cannot raise from a type accessor" — not quite true for a numerically malformed token.

Switching `ParseDouble` to `double.TryParse` and returning `false`/throwing `MzLibException` on failure closes it, and keeps everything funnelling through the one exception type the caller already handles. As before this is robustness, not a live bug: all **95,489** ion tokens in the fixture parse clean.

trishorts (approved):
Re-reviewed at `867081ba`. **All five of my earlier findings are fixed** — I verified each against the current tree rather than taking the replies at face value, and replied inline on each thread with what I checked before resolving it:

| # | Finding | Status |
|---|---|---|
| 1 | Header properties silently `null`/`0` until `Results` is touched | Fixed — `EnsureLoaded()` on all 22 getters + `HeaderLines`, no getter recursion |
| 2 | `HeaderLines` never cleared, duplicates on reload | Fixed — `_headerLines` rebuilt in `LoadResults()`; 21 -> 21 pinned |
| 3 | Write didn't round-trip the leading blank lines | Fixed — `_blankLinesBeforeGroups` replayed on write, byte-comparison test added |
| 4 | `ParseFileType` on `.txt` started requiring the file to exist | Fixed — `File.Exists` guard + `FileShare.ReadWrite` |
| 5 | Throwing property getter | Fixed — `GetMatchedIons()` + `UnparseableMs2Matches` |

Each fix also came with a regression test that pins the behavior rather than restating the implementation, which is the part I care about most — `TestPytheasResultReadWriteByteRoundTrip` in particular is the byte-level assertion the original round-trip test was missing.

**Approving.** The one comment below is Minor and explicitly non-blocking — a narrow gap where a numerically malformed MS2 token still throws past the new `try`. Fix it here or in a follow-up, your call.

Two smaller observations, not worth threads:

- `PytheasResultFile.cs` — because the base `Results` getter reloads whenever `_results` is empty *and* the file exists, a headers-only Pytheas file (a search with zero matches) re-reads and re-parses the whole file on every header property access. Correct, just O(file) per read, and it's inherited base-class behavior rather than anything this PR introduced. Mentioning it only because this PR is what routes 22 getters through that path.
- `UnparseableMs2Matches` reads empty before `GetMatchedIons()` has ever been called, which is indistinguishable from "parsed, all clean" — the same shape as finding #1, but far milder since `GetMatchedIons()` is the obvious entry point and its name says so.

For the record on CI: `external-service-tests` is red, but the three failures are `TestProsit2023XLCMS2_ChargeStateBoundaries`, `TestProsit2023XLCMS3_ChargeStateBoundaries` and `TestProsit2023XLCMS3_PredictionQuality` — live Koina/Prosit calls, unrelated to this PR and non-blocking by design. `build`, `integration` and both codecov gates are green.

zhuoxinshi (approved):

PR #783, opened 2024-05-24
Update nearly all nuget packages and .Net to 8.0

trishorts:
.net 6 and 7 no longer have support, which prompted this move to .net 8.  To keep up w/ project dependencies and compatibilities, all nuget packages were attempted to be updated. 

I generated a local nuget package with this PR and put it into MetaMorpheus. I analyzed 6 Mann files with calibration, averaging, gptmd and search. Results loaded into MetaDraw successfully. I also performed a successful glyco search with test data. I don't see any difficulties integrating this into MetaMorpheus.

OxyPlot.Wpf COULD NOT BE UPDATED
SharpLearning.Optimization COULD NOT BE UPDATED

A close look at the impact of updating these nuget packages on MetaMorpheus and other downstream apps should be performed.

Updating NuUnit was imperfect. The Assert function usage has changed substantially. To avoid massive refactoring, Assert was recast as ClassicAssert. Going forward, we should attempt to use the new Nunit usage.

TargetFramework was updated to 8.0 for all projects in the properties menu of each project in the solution explorer.
The .nuspec was updated to .net 8.0

The github/workflows/dotnet.yml and release.yml were updated to .net 8

coverlet.collector in dotnet.yml updated to -v 6.0.2

PR #1278, opened 2026-09-04
Align mzLib.nuspec dependencies with actual PackageReferences

Alexander-Sol:
## Summary

Brings the hand-authored `mzLib.nuspec` `<dependencies>` back in line with what the shipped projects actually reference. The nuspec's declared versions had drifted from the csproj `PackageReference`s, so a NuGet consumer restored dependency versions that no longer matched the packaged DLLs. Applied to **both** dependency groups (`net10.0` and `net10.0-windows7.0`).

This didn't affect MetaMorpheus, as MetaMorpheus independently required the correct version for affected packages. It did not inherit the old, transitive dependencies. This will affect non-metamorpheus consumers that don't explicitly pin versions for the three affected packages.

### Corrected drifted versions

| Package | Declared in nuspec | Actual `PackageReference` |
|---|---|---|
| `Microsoft.ML` | 2.0.0 | **3.0.1** |
| `Microsoft.ML.FastTree` | 2.0.0 | **3.0.1** |
| `NetSerializer` | 4.1.1 | **4.1.2** |

`FlashLFQ.csproj` references these newer versions; the nuspec lagged behind. A consumer that doesn't independently pin them restored the stale versions and then failed at runtime — e.g. `Could not load file or assembly 'NetSerializer, Version=4.1.2.0'`. Because the nuspec is never exercised by the unit tests or by a `ProjectReference` build, the mismatch shipped silently.

### Pinned dependency
FlashLFQ depends on a specific version of SharpLearning.Optimization, version 0.28.0. Previously, the nuspec accepted any version >= 0.28.0. Now, the .nuspec demands that version exactly.

### Removed unused dependencies

- **`Google.Protobuf`** (3.28.3) — leftover from the pre-refactor Koina gRPC client; `PredictionClients` no longer references protobuf.
- **`SharpZipLib`** (1.4.2) — no longer used.
- **`SkiaSharp`** (2.88.9) — leftover from earlier plotting code.

None of the three are referenced by any csproj (there is no central package management), none resolve into any shipped project's `bin/Release` (they appear only transitively in the `Test` project), and none are referenced by MetaMorpheus.

### Package version

Nuget version number changes, but this doesn't affect anything. The version number of the published nuget is set by a github action and doesn't care about what's listed in the nusped

## Validation

- No CI workflows check the mzLib project dependencies against the .nuspec dependencies, which is why extra dependencies were able to slip in, and some package numbers diverged between nuspec and mzLib dependencies. We would need an additional github action to check this. This is stated in issue #1281 


acesnik (commented):
🤖 Good catch, and a genuinely hard one to find: the nuspec is the only file here that no build validates, so the only way to see the drift is to read it against nineteen csprojs by hand. `Microsoft.ML` and `FastTree` moved to 3.0.1 in `FlashLFQ.csproj:16-17` back in #802, and `NetSerializer` to 4.1.2 in #783, so these had been stale well over a year. I re-derived the comparison from scratch and after this change every declared dependency maps one-to-one onto a shipped project's `PackageReference`, nothing missing in either direction. The three removals check out too.

Two things I'd rather you work out than take from me, because both are more useful than the diff itself.

On the validation paragraph: what would the `integration` job do *differently* if the nuspec still said `Microsoft.ML 2.0.0`? Have a look at what MetaMorpheus's own `EngineLayer.csproj` and `TaskLayer.csproj` declare, then at how NuGet chooses between a direct and a transitive reference. If the answer turns out to be "nothing," that's worth saying in the body — and it points at what sort of check *would* have caught this, which is the more valuable thing to come out of the PR.

Second, you audited version *numbers*. Run the same audit over version *ranges*: one entry still matches on the number but not in kind, and that package has published well past what we build against. Worth working out what a consumer could resolve to, and I think the fix belongs in this PR.

Small one while you're pushing: the file loses its trailing newline.

Two facts so you don't have to dig: `1.0.589` is already tagged and published with release notes on master, so the bump names the last release rather than the next — though it's inert, since `release.yml:73` packs with `-Version $(git describe --tags)`. And the branch is 42 commits behind, so `gh pr diff` credits master's `.xml` and `modificationData` entries to you; the real delta is 22 lines and it merges clean.

Alexander-Sol:
> 🤖 Good catch, and a genuinely hard one to find: the nuspec is the only file here that no build validates, so the only way to see the drift is to read it against nineteen csprojs by hand. `Microsoft.ML` and `FastTree` moved to 3.0.1 in `FlashLFQ.csproj:16-17` back in #802, and `NetSerializer` to 4.1.2 in #783, so these had been stale well over a year. I re-derived the comparison from scratch and after this change every declared dependency maps one-to-one onto a shipped project's `PackageReference`, nothing missing in either direction. The three removals check out too.
> 
> Two things I'd rather you work out than take from me, because both are more useful than the diff itself.
> 
> On the validation paragraph: what would the `integration` job do _differently_ if the nuspec still said `Microsoft.ML 2.0.0`? Have a look at what MetaMorpheus's own `EngineLayer.csproj` and `TaskLayer.csproj` declare, then at how NuGet chooses between a direct and a transitive reference. If the answer turns out to be "nothing," that's worth saying in the body — and it points at what sort of check _would_ have caught this, which is the more valuable thing to come out of the PR.
> 
> Second, you audited version _numbers_. Run the same audit over version _ranges_: one entry still matches on the number but not in kind, and that package has published well past what we build against. Worth working out what a consumer could resolve to, and I think the fix belongs in this PR.
> 
> Small one while you're pushing: the file loses its trailing newline.
> 
> Two facts so you don't have to dig: `1.0.589` is already tagged and published with release notes on master, so the bump names the last release rather than the next — though it's inert, since `release.yml:73` packs with `-Version $(git describe --tags)`. And the branch is 42 commits behind, so `gh pr diff` credits master's `.xml` and `modificationData` entries to you; the real delta is 22 lines and it merges clean.

Good notes, the previous summary comment was wrong. I've updated the comment, and opened #1281, suggesting a new CI workflow.

mzLib/mzLib.nuspec
```diff
@@ -15,16 +15,13 @@
...
-    <dependency id="NetSerializer" version="4.1.1" />
+    <dependency id="Microsoft.ML" version="3.0.1" />
+    <dependency id="Microsoft.ML.FastTree" version="3.0.1" />
+    <dependency id="NetSerializer" version="4.1.2" />
     <dependency id="OpenMcdf" version="2.3.1" />
     <dependency id="OpenMcdf.Extensions" version="2.3.1" />
-    <dependency id="SharpLearning.Optimization" version="0.28.0" />
-    <dependency id="SharpZipLib" version="1.4.2" />
-    <dependency id="SkiaSharp" version="2.88.9" />
+    <dependency id="SharpLearning.Optimization" version="[0.28.0]" />
```
trishorts, line 24: **Worth recording — the three removals are safe, but not for the reason the body gives, and they silently downgrade three DLLs.**

The body says the removed packages "don't resolve into any shipped project's `bin/Release`" and are "not referenced by MetaMorpheus." The first half is right about *direct* references; the second doesn't hold in effect. I restored MetaMorpheus master's `GUI.csproj` and read `project.assets.json`:

```
Google.Protobuf  <- ['mzLib', 'TorchSharp']
SharpZipLib      <- ['mzLib', 'TorchSharp']
SkiaSharp        <- ['mzLib', 'TorchSharp']
```

All three reach MM's GUI output today, and MM's installer ships each one by name (`Product.wxs:213-220`, plus `libSkiaSharp.dll` at `:531`) with no `PackageReference` of its own for any of them.

**The good news, and the actual reason this is safe:** `TorchSharp 0.105.1` is a second parent and is retained by this PR, so all three stay present and the WiX components keep resolving. No WIX0103, no companion MetaMorpheus PR needed.

**The part worth a deliberate call:** TorchSharp's floors are lower than the ones being removed, so the resolved versions drop. Restoring a project with TorchSharp 0.105.1 as the only parent:

| package | resolved today | resolved after this PR |
|---|---|---|
| `Google.Protobuf` | 3.28.3 | **3.21.9** |
| `SharpZipLib` | 1.4.2 | **1.4.0** |
| `SkiaSharp` | 2.88.9 | **2.88.6** |

Since no mzLib assembly binds to any of them, this is very likely fine — arguably more correct, since TorchSharp then gets the versions it was built against instead of floors mzLib was raising for no reason. But it's a change in what consumers ship, and MetaMorpheus is one of them, so it should be a decision rather than a side effect.

Two things I'd suggest, both cheap:

1. Correct the body's reasoning. It becomes the historical record for why these were dropped, and "TorchSharp already brings them, at slightly lower versions" is the accurate version.
2. Note the latent coupling for #1281: MetaMorpheus's installer now depends on **TorchSharp** to supply `Google.Protobuf.dll`, `ICSharpCode.SharpZipLib.dll` and `SkiaSharp.dll`. Whoever removes or replaces TorchSharp later will break `Product.wxs` with no signal from either repo's CI. A check that diffs the packaged closure against `Product.wxs` would catch both that and the class of drift this PR fixes.

trishorts (approved):
Approving. I re-derived the audit independently rather than checking the diff, and it holds.

**The 1:1 claim is true in both directions.** Parsed the nuspec against the `PackageReference`s of all 17 projects the `<files>` section actually ships:

- declared in nuspec with no backing `PackageReference`: **none**
- referenced by a shipped project but missing from the nuspec: **none**
- version mismatches: **none**

The three corrected versions match `FlashLFQ.csproj` (`Microsoft.ML` / `FastTree` 3.0.1, `NetSerializer` 4.1.2), and `SharpLearning.Optimization [0.28.0]` now matches `FlashLFQ.csproj` exactly in kind as well as number — `mzPlot.csproj` still declares the floating `0.28.0`, but the nuspec correctly follows the stricter of the two. That closes the range point from the earlier review. Trailing newline is back too.

**One thing worth recording, non-blocking**, inline on the removals: they're safe, but not for the reason the body states, and they quietly lower three shipped DLL versions. `TorchSharp` — retained here — is a second parent for all three, so nothing disappears from MetaMorpheus's installer; but its floors are lower, so `Google.Protobuf` 3.28.3→3.21.9, `SharpZipLib` 1.4.2→1.4.0, `SkiaSharp` 2.88.9→2.88.6 for any consumer that doesn't pin them, MetaMorpheus included. Almost certainly fine since no mzLib assembly binds to them, but worth being a decision rather than a side effect, and worth correcting in the body since that text becomes the record for why they were dropped.

Two smaller notes:

- `<version>1.0.589</version>` names a release that already shipped — `1.0.589` is tagged (2026-09-02) and an ancestor of master, so the bump points at the last release rather than the next. Inert, because `release.yml:73` packs with `-Version $(git describe --tags)`, but in a PR specifically about the nuspec telling the truth it's an odd thing to leave pointing backwards.
- For #1281: the useful check here isn't only nuspec-vs-csproj. MetaMorpheus's `Product.wxs` now depends on TorchSharp to supply `Google.Protobuf.dll`, `ICSharpCode.SharpZipLib.dll` and `SkiaSharp.dll`, and neither repo's CI would notice if TorchSharp were swapped out. Diffing the packaged closure against `Product.wxs` would cover that and the drift this PR fixes.

Also confirmed `Newtonsoft.Json 13.0.4` matches MetaMorpheus's own pin exactly, so no NU1605 risk there. CI green on all five checks.

nbollis (approved):

PR #1284, opened 2026-09-08
Stop keying the peptide-occupancy form lookup on PSM objects

pcruzparri:
First of four related PRs. Independent of the other three — this can merge on its own. **#1286 carries the same unresolved-PSM guard this now takes; it needs a rebase to drop that hunk.**

`CalculateDigestionProductLevelOccupancy` built a `Dictionary<ISpectralMatch, …>` and then read it once per element in a loop over the same collection — a 1:1 materialization that never indexed out of order. Keying it on `ISpectralMatch` is the problem: the interface fixes no equality contract, and the two implementations in play disagree. `BaseSpectralMatch` compares on `(FullFilePath, OneBasedScanNumber, FullSequence)` with a matching value hash; MetaMorpheus's `SpectralMatch` deliberately uses reference equality, with a comment saying value equality "would start deduplicating `AllPsmsBelowOnePercentFDR` and change the reported PSM counts." So two `BaseSpectralMatch`-derived matches agreeing on those three throw `ArgumentException: An item with the same key has already been added`, and nothing in the signature warns you.

The map is gone and the lookup is inline. The comment above it claimed the match was on "FullSequence + Accession"; only `FullSequence` was ever compared, so that goes too.

In-repo callers happen to be safe today — `AllPsmsBelowOnePercentFDR` is a `HashSet<ISpectralMatch>`, whose default comparer is exactly that equality — but the method is public and takes a bare `IEnumerable<ISpectralMatch>`.

**Second change, moved forward from #1286 at trishorts's request:** the same method crashed a second way, and it made no sense to close half the class. `AllSame` opens with `enumerable.First()` (`ClassExtensions.cs:128`), so an empty list throws `InvalidOperationException`. Getting there is easy — "unresolved" has two spellings. MetaMorpheus leaves `BaseSequence` null for a PSM ambiguous across base sequences (`PsmTsvWriter.Resolve` returns null when the best matches disagree); `BaseSpectralMatch` coerces that null to empty at construction. The old filter tested only null, so it caught one spelling and let the other through to fail the all-same check. Now it drops both and returns empty when nothing resolves.

This one is reachable from the in-repo caller, not just the public surface: `BioPolymerGroup.PopulateOccupancy` groups by `BaseSequence` and passes each group straight in, so a null-key group filters to empty.

**Out of scope, taken anyway:** a dead local in the same method — a `result` dictionary that was never read or written — since the method was already being trimmed.

Ran: three tests in `ModificationOccupancyCalculatorTests`, all new. Each was checked against the pre-change file and fails there for its own reason — `ArgumentException` for the duplicate-PSM one, `InvalidOperationException` from `AllSame` for the empty-list one, `ArgumentException` from the all-same check for the mixed resolved/unresolved one. Testing the empty-list guard needed an `ISpectralMatch` that implements the interface directly, since every existing mock derives from `BaseSpectralMatch` and cannot produce a genuinely null base sequence; `UnresolvedSpectralMatch` is that, reference equality included, so it stands in for the MetaMorpheus shape. Full suite green on re-run, 6146 passed. An earlier run of the same suite reported one failure that did not reproduce and that I can't name — I piped that run through `tail` and lost the detail — so treat it as unexplained rather than cleared.

Not run: anything in MetaMorpheus. The claims here about its `SpectralMatch` and `PsmTsvWriter` are read off the source at `92dfdd6c`, not exercised.


trishorts (commented):
Read the whole diff. The change is right — the map was built over `psmsWithBaseSeq` and read once per element in a loop over the same list, so with no duplicates an equality lookup returns exactly what computing inline returns. Same number of `GetIdentifiedBioPolymersWithSetMods()` calls, one fewer dictionary. And the `Is.EqualTo` precondition is what keeps the new test from passing vacuously, which is the part that makes it worth having.

One thing I'd fold in before it merges, because it is the same method and the same argument you make in the body about the public surface.

A few lines above your change:

```csharp
var psmsWithBaseSeq = psmList.Where(p => p.BaseSequence != null).ToList();

if (!psmsWithBaseSeq.Select(p => p.BaseSequence).AllSame())
```

`MzLibUtil.ClassExtensions.AllSame` opens with `T comparand = enumerable.First()` (`ClassExtensions.cs:128`), so an empty sequence throws `InvalidOperationException`. Null base sequences are real: `SpectralMatch.cs:332` and `:492` assign `PsmTsvWriter.Resolve(...).ResolvedValue`, and `Resolve` returns null whenever the best matches disagree (`PsmTsvWriter.cs:137-150`). A group of PSMs all ambiguous across base sequences filters to empty and throws.

You have already written the fix — it is in #1286:

```csharp
if (psmsWithBaseSeq.Count == 0)
    return new Dictionary<int, List<SiteSpecificModificationOccupancy>>();
```

Two lines, same method. It means this PR closes that method's crash class rather than half of it, and it gets the fix out from behind #1286, which has no consumer yet and may sit for a while.

One caveat on lifting it: I would take the early return but **not** the `!= null` to `!string.IsNullOrEmpty` widening that #1286 pairs it with. The comment there says a PSM whose sequence could not be resolved "carries an empty base sequence, not null" — but `Resolve` returns null, and null is already filtered by the existing predicate. What the widening actually changes is a PSM with `BaseSequence == ""`: today it survives the filter, fails `AllSame()` against the real sequences, and throws `ArgumentException`; with the widening it is dropped silently and the rest proceed. That turns a loud failure into a quiet one, and I could not find anything that produces `""` on this path. Happy to be shown one — if there is a producer, the widening is right and the comment just needs to name it.

Existing coverage is unaffected either way: `PeptideLevelWithMultipleBaseSequences` passes two genuinely distinct base sequences, so it still throws `ArgumentException`.

Approving as soon as that is in — holding off only because the push would dismiss the approval anyway.

pcruzparri:
Took the early return. You're right that it makes no sense to close half the crash class and leave the rest behind a PR with no consumer.

I also took the `IsNullOrEmpty` widening, which you asked me not to — here's the case, because I think two things settle it.

You asked to be shown a producer of `""`. There's one, and it's ours: `BaseSpectralMatch.cs:44` does `BaseSequence = baseSequence ?? string.Empty`, deliberately, and the constructor's XML doc says so ("Null values are converted to empty string"), same as it does for `FullFilePath`, `FullSequence` and `Accession`. So the two implementations spell "unresolved" differently — the MetaMorpheus path you found gives null, our own base class gives `""` — and a filter testing only null catches one of them.

The other thing is that #1286's own test already depends on the widening. `UnresolvedPsmDoesNotBreakOccupancy` builds its unresolved PSM as `new MockSpectralMatch(File1, null, null, …)`, and the coercion above turns that into `""`. Without the widening it survives the filter, fails `AllSame()` against the real sequence, and the test throws instead of asserting `1/1`. So the two lines aren't actually separable — rejecting the widening means #1286 keeps it and the hunks diverge.

There's a third, smaller reason: without the widening the early return is unreachable from every mock we have, because none of them can produce a genuinely null `BaseSequence`. I added `UnresolvedSpectralMatch` for that — implements `ISpectralMatch` directly, reference equality and all, so it stands in for the MetaMorpheus shape rather than pretending a `BaseSpectralMatch` can be null.

Your objection isn't hand-waved, though. A `""` PSM mixed with real ones did go from `ArgumentException` to a silent drop, and that's a real change. It's now pinned by `DigestionProductLevelSkipsUnresolvedPsmWithoutInflatingDenominator`, so it's a decision with a test on it rather than a side effect. The argument for making it: null was already dropped silently, and treating two spellings of the same condition differently is harder to defend than dropping both.

You were right that #1286's comment is wrong — but for a different reason than "it should name a producer." It says the unresolved PSM "carries an empty base sequence, not null," which describes the test mock, not the MetaMorpheus path you correctly identified. Both happen. The comment here names both.

Two notes on the citations, for anyone reading this thread inside mzLib: `SpectralMatch.cs:332`/`:492` and `PsmTsvWriter.cs:137-150` are MetaMorpheus, not this repo. I checked them against `92dfdd6c` and the line numbers are exact.

Last thing, and it cuts against my own PR body rather than your review. Checking your claims sent me into MetaMorpheus's `SpectralMatch`, where `Equals(ISpectralMatch)` is `ReferenceEquals` (`EngineLayer/SpectralMatch.cs:161`) — deliberately, with a comment saying value equality "would start deduplicating `AllPsmsBelowOnePercentFDR` and change the reported PSM counts." Distinct objects are therefore never equal there, so the `ArgumentException` this PR fixes isn't reachable from MetaMorpheus at all, and "a chimeric search produces exactly that" was wrong. The fix still stands, on better ground: we ship two `ISpectralMatch` implementations with incompatible equality, so keying a dictionary on the interface depends on a contract the interface doesn't fix. Body rewritten around that, and the test comment making the same claim is fixed too.

#1286 will need a rebase to drop the duplicated hunk.

trishorts (approved):
The early return is in, and you were right about the widening — I said I couldn't find a producer of `""` and there is one, in this repo: `BaseSpectralMatch` coerces a null base sequence to empty at construction, documented, alongside `FullFilePath`, `FullSequence` and `Accession`. So the two spellings of "unresolved" are real and a filter testing only null catches one of them. I was reading the interface and the MetaMorpheus implementation and never opened the base class between them.

`DigestionProductLevelSkipsUnresolvedPsmWithoutInflatingDenominator` is what makes it a decision rather than a side effect — the `ArgumentException`-to-silent-drop change I objected to is now pinned, with `TotalCount` asserted at 1 rather than 2. And `UnresolvedSpectralMatch` is the right shape for the early return; deriving it from `BaseSpectralMatch` would have tested the coercion instead of the guard.

The body correction is the part I'd have missed entirely. `SpectralMatch.Equals` being `ReferenceEquals` does mean the `ArgumentException` isn't reachable from MetaMorpheus, and the PR is better argued for saying so — two shipped implementations with incompatible equality, keyed on an interface that fixes neither.

One follow-up, not blocking this: #1286 hasn't moved since 09-08 and still carries the duplicated hunk.

RayMSMS (approved):

PR #1285, opened 2026-09-08
Move biopolymer group TSV rendering into an external schema and writer

pcruzparri:
Second of four related PRs. **This diff also contains #1284**, because GitHub can only base a cross-fork PR on a branch in this repository, so all four target `master` and each carries its predecessors' commits. The change to review here is the final commit, the one matching this PR's title. Merge #1284 first.

`BioPolymerGroup` both held data and rendered itself. This moves rendering into `BioPolymerGroupTsvSchema` + `TsvColumn<T>` / `TsvWriter`, with `SampleGroupBuilder` and `SampleGroupLabels` extracted alongside. A schema is one column list applied to every record, so header and row widths agree by construction rather than by review.

**#1240 already fixed the quantified path.** This does not re-fix it — it adopts the same run-level gate, computed from `SamplesForQuantification` and `IntensitiesBySample` since both are public. #1240's five delivery tests are rewritten to render through the schema and all 26 tests in `QuantificationDeliveryTests` pass, which is the evidence this subsumes that change rather than reverting it.

What it does fix, both reproduced against master:

- **The unquantified path is still ragged.** Two groups observed in different files with no engine run gives header 25 fields and rows of 25 and 21. `BioPolymerGroup.cs:360` calls this unfixable per-group because "a single group does not know the run's full file list" — a dataset-level schema does. This covers glyco pre-quantification, TMT, quantification off, malformed design, and quantification throwing partway.
- **Same-named files in one sample group crash.** `BioPolymerGroup.cs:610` keys by `FilenameWithoutExtension`; two fractions stored as `FracA/sample.raw` and `FracB/sample.raw` throw `ArgumentException … Key: sample`. `SampleGroupBuilder` keys by full path.

Also: colliding sample labels are widened with parent directories rather than silently merged, and a site with no measured intensity renders blank instead of `fraction=0.0000(0/0)` — 339 of 5,244 sites on a real 18-file run.

**Out of scope, deliberately:** the declared-section API. `SampleGroupColumnBuilder` still infers whether intensity columns exist rather than being told, so MetaMorpheus cannot express its `WasSearched` / `HasPeptideLevelQuantification` gating through it yet. That is the next piece and it needs a design decision first.

Ran: full mzLib suite, green. Not run against MetaMorpheus — it consumes mzLib from NuGet.


trishorts (commented):
🤖 The diagnosis behind this is right and the mechanism is the one the problem needs. A dataset-level schema makes header and row widths agree by construction, and four separate defects in this codebase — #1240's ragged table, the 24-to-22 loss on MetaMorpheus #2755, the individual-file column loss fixed in MetaMorpheus #2802, and the per-channel duplication still on master — are one bug that this makes structurally impossible.

Two things in it are notably right, and I want to name them before the problem.

**The run-level gate is correct, and the comment on it is doing real work.** `BioPolymerGroupTsvSchema` computes `quantified` from `SamplesForQuantification` and `IntensitiesBySample` across the dataset, and says explicitly: *"Deliberately not `SampleGroupResult.HasIntensityData`, which answers 'did this sample group get a value' — using that to choose columns made a group whose samples were all unobserved describe a narrower table than its neighbour."* That is precisely the predicate #1240 rejected, rejected again for the same reason, in writing. Please keep it.

**The per-channel duplication is now visible in one place.** The loop emitting `SpectralCount_`, `Intensity_`, `CountOccupancy_` and `IntensityOccupancy_` per sample-group identity is where TMT's eleven identical spectral counts come from. The schema makes the table rectangular without making it correct, which is the right split — that fix is a separate change on top of this one, not part of it.

## The blocking problem: this removes four members MetaMorpheus master binds to

Verified against `smith-chem-wisc/MetaMorpheus` `master` as of today, after #2789, #2802, #2801, #2790 and #2804:

| Removed here | MetaMorpheus master | Result |
|---|---|---|
| `BioPolymerGroup.GetTabSeparatedHeader()` (`virtual`) | `ProteinGroup.cs:243`, `public override` | **CS0115** |
| `IBioPolymerGroup.GetTabSeparatedHeader()` | — | interface break |
| `BioPolymerGroup.HasAssignedSampleIntensities` (`protected`) | `ProteinGroup.cs:271` and `:403` | **CS0103** |
| `BioPolymerGroup.ToString()` (`override`) | `ProteinGroup`'s own `override` | falls back to `object`, legal |

**And `integration` is green in a way that inverts the truth.** That job clones `smith-chem-wisc/MetaMorpheus` at `ref: master` *at run time*. This run started `2026-09-08T21:03:30Z`. **MetaMorpheus #2789 merged at `21:09:01Z`** — five and a half minutes later — and #2789 is precisely the PR that deleted MetaMorpheus's own copy of the intensity gate so that it would inherit mzLib's, which is what created two of the three real bindings above. #2802 followed at `22:02Z`, then #2801, #2790 and #2804. So the green was earned against a MetaMorpheus that did not yet contain the code this breaks. A re-run today should fail, and that failure is the useful signal rather than a setback.

The fix is small and it is not a rework: keep `GetTabSeparatedHeader()` and `ToString()` on `BioPolymerGroup`, `GetTabSeparatedHeader()` on `IBioPolymerGroup`, and `HasAssignedSampleIntensities`, each delegating to the schema. One implementation, no duplicated rendering, and the release stays adoptable. They come out in a later PR, after MetaMorpheus has switched — mzLib **adds**, MetaMorpheus **switches**, mzLib **removes**. A single PR that both adds the schema and deletes the group's rendering is unadoptable by construction, and it would gate every future mzLib bump behind the renderer replacement.

I have that change written and will open it against this branch so it is a diff you can read rather than a paragraph you have to implement.

## Two rectangularity assertions cannot fail

`TsvWriter.RowLine` is `string.Join('\t', schema.Select(c => c.GetValue(item) ?? string.Empty))`, so it always emits exactly `schema.Count` fields. Both of these take their header and their row from `BioPolymerGroupTsvSchema.For` on the same input:

- `RenderedTable_IsRectangular_WhenAGroupHasUnobservedSamples` (`QuantificationDeliveryTests.cs:621-622`) — `GroupTsv.Header(dataset)` against `GroupTsv.RowInDataset(g, dataset)`.
- `SampleGroupColumnTests.cs:287-291` — `HeaderLine(schema)` against `RowLine(schema, group)`, literally the same schema object.

Before the refactor the header came from `GetTabSeparatedHeader()` and the row from `ToString()` — two paths that could genuinely disagree, which is how the #1240 bug was caught in the first place. That property is what wants pinning, not width.

Something that can fail: assert the column *names* from a partially-quantified group's own single-group schema (`For([group])`) equal the dataset schema's (`For(dataset)`) — two independent derivations — or assert the unfilled `Intensity_` cells are **present and empty** rather than absent, which is #1240's property stated as behaviour. The acceptance test is that reverting the run-level gate to per-group `HasIntensityData` turns a named test red.

The comment above line 620 also still says the row comes "from its own `ToString()`", which is no longer true.

## Smaller things

`codecov/patch` is red at 86.17%, 43 lines. Two files carry it: `SampleGroupLabels.cs` at 57.57% (23 missing, 5 partials) and `TsvColumn.cs` at 37.50%. `SampleGroupLabels.Disambiguate` is the collision rule the labelling work will build on, so its coverage matters more than the raw number suggests. `TsvColumn.cs` is low for a specific reason: `TsvWriter.Write(TextWriter, schema, items)` has **no consumer anywhere on this branch**, production or test — only `HeaderLine` and `RowLine` are called, and only from test helpers. It gets its consumer when MetaMorpheus adopts the writer, so saying so may be better than testing it now.

`mergeStateStatus` is `DIRTY` as of today: #1284 merged squashed as `9c160b714`, and this branch still carries its pre-squash commit `dc81a61b`. A rebase onto master drops it.

**Not verified by me:** I did not run the mzLib suite or re-trigger `integration`. The line numbers, the timestamps and the `RowLine` implementation are read from the source at `83c120f8` and from `master` today.

trishorts:
🤖 The adoptability change from the review is now a diff rather than a paragraph: **pcruzparri/mzLib#9**, opened against `TsvColumnSchema` so you can take it, amend it, or reject it in one place.

It restores `GetTabSeparatedHeader()` and `ToString()` on `BioPolymerGroup`, `GetTabSeparatedHeader()` on `IBioPolymerGroup`, and `HasAssignedSampleIntensities`, each delegating to your schema rather than reimplementing anything. Nothing about the schema, the run-level gate, `SampleGroupBuilder`, `SampleGroupLabels` or the golden tests is touched.

No rendered output changes: the single-group schema evaluates the same predicate `HasAssignedSampleIntensities` gated on before, so those members produce exactly what they produced, through your code.

It also adds `DerivedGroupRenderingTests` — a subclass standing in for MetaMorpheus's `ProteinGroup`. If any of the three members is removed, that file stops **compiling**, so the break lands in mzLib's own build instead of in a downstream bump or a stale `integration` run. Delete it in the PR that removes them.

mzLib suite on that branch: 6148 passed / 0 failed / 31 skipped. I did not re-trigger `integration`; that is still the authority here and is worth running once this is in.

The other review points — the two rectangularity assertions that cannot fail, `codecov/patch`, and the rebase now that #1284 merged squashed as `9c160b714` — are left to you, deliberately.

trishorts (approved):
Approving. The mechanism is right and the adoptability blocker is resolved.

**What this gets right, and it is the important part.** A schema built from the dataset and applied to every record makes column count rectangular *by construction* rather than by each output path getting it right. That is the root of a defect this codebase has now hit four separate ways, and `TsvWriter.RowLine` being `string.Join('\t', schema.Select(...))` is what retires the class rather than patching another instance of it.

The run-level `quantified` gate in `QuantificationColumns` is #1240's predicate stated correctly — computed from `SamplesForQuantification` and `IntensitiesBySample` rather than from per-group `HasIntensityData`. The comment above it explains why, which is worth keeping verbatim. Same for `SampleGroupIndex`: decoupling a column from its position is what fixes the ragged rows, and keying the cache on the sample-group list instance rather than the group is the detail that keeps a mid-write repopulation from serving a stale index.

`SampleGroupBuilder` keying on `FullFilePathWithExtension` also fixes a live crash — two fractions named the same thing in different directories throw on master today.

**Non-blocking, but worth having on the record.**

1. The two rectangularity assertions cannot fail, and I think the criterion behind them needs re-aiming rather than just satisfying. `Test/Quantification/QuantificationDeliveryTests.cs:621-622` takes header and rows from the same `For(...)`; `SampleGroupColumnTests.cs:288-292` takes header width and row from literally the same schema object. Since `RowLine` always emits exactly `schema.Count` fields, no width assertion is capable of being wrong any more — which is a compliment to the design. The falsifiable version is aimed at the **gate**: revert `quantified` to the per-sample-group predicate and a named test should go red.

2. Three things in `SampleGroupBuilder` that are lifted verbatim from master, so not introduced here — flagging them because this PR is where they become easy to fix:
   - **`:96` orders channels by `ChannelLabel`, not `ReporterIonMz`.** An ordinal sort puts `127C` before `127N` in every N/C pair. Values are safe (intensities are looked up by sample object, not position), so this is column order only.
   - **`:94` gives every channel the whole file's PSMs**, so a plex reports one spectral count and one occupancy string eleven times over. This is the artifact that makes counts-per-file and intensities-per-channel two different key spaces.
   - **`:39` decides the label scheme with `File.Exists`.** That gets sharper once designs can come from SDRF, where a design may legitimately name a file that is not on the machine doing the writing.

3. `BioPolymerGroupTsvSchema.IsParentLevel` (`:191-192`) and `OccupancyKeys` (`:198-202`) read `BioPolymerGroupType`, so #1287 will need to touch this file too, not only `ProteinGroup`.

The restored `GetTabSeparatedHeader()`, `ToString()` and `HasAssignedSampleIntensities` all delegate to the schema, and `For([group])` evaluates exactly the predicate the old gate did — so there is one implementation and no rendered output changes. `DerivedGroupRenderingTests` should be deleted in the same PR that finally removes them.

nbollis (approved):

PR #1289, opened 2026-09-09
Support tRNA databases

nbollis:
Stacked on #1276 

 >tdb entries have literally nothing besides an accession. We can now read that in. 

Weird things in tdb format:
lower case letters: They represent a non-conserved residue. This means it is likely the canonical, but variants have been observed in the past. Future work could make it so we load those as sequence variants. 
_ characters: this means that there is a new residue in that slot that does not align with the canonical residue (how this is useful, I am not sure) but does not state what it is. Therefor we do not support these entries.


trishorts (commented):
Reviewed only this PR's own payload, not #1276's.

Worth stating how, because the numbers are misleading: GitHub reports **12 files / 612 additions** against `master`, but this branch carries #1276 inside it. `git merge-base` puts the real fork point at `32b5c904`, and the isolated diff is **4 files / 82 lines** — `RnaDbLoader.cs` (+20), `TestDbLoader.cs` (+36), a 22-line fixture, and a `Test.csproj` entry. Everything below is about those 20 lines of loader code. I reviewed by hand rather than running the engine; on a payload this size that is both faster and more accurate.

The feature itself is clean and does what it says: `>tdb` detection, a two-field regex set for headers that carry nothing but an accession, accession-as-name, and an unconditional uppercase for the format's non-conserved lowercase. `TestTRNAdbUnmodifiedFasta` passes and is a real test — it pins the count, `Accession == Name`, two specific mixed-case sequences, and the empty organism, rather than just asserting it loaded.

Four findings inline, all reproduced on `ba573e3c`:

- **Major** — one `_` character anywhere in a tdb file aborts the *entire* read. A good entry followed by a `_` entry loses both, and `out errors` stays empty. The description calls `_` unsupported, but unsupported currently means "lose the whole database", not "skip the record".
- **Major (stacking, not this file)** — adding `ToUpper` unconditionally for tdb makes the culture-sensitive `ToUpper()` reachable for every tdb record, and lowercase `i` is expected input here. The fix is `d0f7596a` on #1276, and **this branch forks from `32b5c904`, which predates it**. Reproduced: an unconserved inosine loads on `en-US` and throws on `tr-TR`. Nothing to change here — just make sure the rebase picks it up.
- **Minor** — the `(?!.)` organism pattern yields `null` because it declares a capture group that does not exist, not because it matches an empty string. Works, but by an accident a later maintainer would not recognise.
- **Minor** — `>tdb` detection and the `^>(\S+)` accession capture are much looser than the anchored patterns used for Ensembl/NCBI, so `>tdbanything` is accepted as a tdb accession. Plus two typos in the new comment.

Not approving or requesting changes — leaving the call to you. The `_` one is the only finding I would actually want addressed before merge; the stacking note needs nothing but attention at rebase time.

One unrelated thing in the diff: this also strips the UTF-8 BOM from `Test.csproj`. Harmless, just unrelated to tRNA support.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -196,22 +211,30 @@ public static List<RNA> LoadRnaFasta(string rnaDbLocation, bool generateTargets,
...
                                     regexes = MzLibRegexes;
                                     identifierHeader = "Accession";
                                     break;
+                                case RnaFastaHeaderType.TRNAdb:
+                                    regexes = TRNAdbFieldRegexes;
+                                    identifierHeader = "Accession";
+
+                                    // TDB sequences somethimes throw in a lowercase for unconserved modificaitons.
+                                    // A future implementation could encode sequence variants, for now we just convert. 
+                                    sequenceTransformations.Add(SequenceTransformationOnRead.ToUpper);
```
trishorts, line 240: **Major — a single `_` anywhere in the file aborts the entire database read, taking every good entry with it.**

The description says `_` entries "are not supported", which I read as *skipped*. What actually happens is an unhandled `ArgumentException` out of the residue lookup that kills the whole load. `LoadRnaFasta`'s `out List<string> errors` — the channel built for exactly this — stays empty.

Reproduced on `ba573e3c`, minimal tdb files:

| file | result |
|---|---|
| one entry, contains `_` | `ArgumentException: Nucleic Acid Letter _ does not exist…` |
| good entry, then `_` entry | **throws — the good entry is lost too** |
| `_` entry, then good entry | **throws — never reaches the good entry** |

The third row is the one that matters in practice: a real tRNAdb download with one `_` record in the middle costs the user all ~11k of the others, with an exception message that names a character rather than an accession or a line number.

Given the description already treats these as out of scope, the cheap version is to skip the record and append to `errors` — same shape the FASTA path already uses for other bad input — so the load degrades instead of failing. If you would rather not special-case `_` here, even catching and re-throwing with the accession and line number attached would turn this from unactionable into actionable.

Worth a test either way — the description documents this behaviour, but nothing pins it.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -463,19 +503,20 @@ public static IEnumerable<RNA> Merge(IEnumerable<RNA> mergeThese)
...
-                case SequenceTransformationOnRead.None:
-                    return cleanedSequence;
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(sequenceTransformation), sequenceTransformation, null);
-            }
+            if (sequenceTransformations.Count == 0)
+                return cleanedSequence;
+
+            if (sequenceTransformations.Contains(SequenceTransformationOnRead.ToUpper))
+                cleanedSequence = cleanedSequence.ToUpper();
```
trishorts: **Major — this makes the culture-sensitive `ToUpper` reachable for *every* tdb file, and the fix for it is not on this branch.**

This is a stacking issue rather than anything wrong with the line itself, so it is easy to miss.

On #1276, `ToUpper` was only ever added for a sequence that was *entirely* lowercase — rare, so the culture-sensitivity I flagged there was a narrow edge. Here it is added unconditionally for the whole file, which is right for the format, but it means every tdb record now goes through `cleanedSequence.ToUpper()`. And lowercase `i` is expected input by your own description: an unconserved inosine is exactly the case the `I` residue was added for in #1276.

`ToUpper()` is culture-sensitive — under `tr-TR`/`az-AZ`, `'i'` maps to `'İ'` (U+0130), which no residue table knows. Reproduced on `ba573e3c`:

| sequence | `en-US` | `tr-TR` |
|---|---|---|
| `…GUAGAGCiCUCGC…` (unconserved inosine) | loads, `i` -> `I` | **throws** `Letter İ does not exist` |
| `…GUAGAGCICUCGC…` (already uppercase) | loads | loads |
| `GGCCCGGAUAGCUCagUCGG…` (fixture shape) | loads | loads |

The fix is already written — `d0f7596a` on #1276 changes that one call to `ToUpperInvariant()` and adds a regression test across the three cultures. **It is not on this branch:** `git merge-base` of this PR and #1276 is `32b5c904`, which is #1276's head *before* that commit. So if this merges on its current base, or rebases onto anything older than `d0f7596a`, the widened version of the bug ships.

Nothing to change in this file — just make sure the rebase picks up `d0f7596a`, and that `SanitizeAndTransform` still reads `ToUpperInvariant()` afterwards. Flagging it because the merge order makes it silently losable.

Your fixture only contains lowercase `a` and `g`, so it would not catch this; a lowercase `i` somewhere in `TRNAdbUnmodifiedTrimmed.fasta` would.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -126,27 +132,36 @@ public static RnaFastaHeaderType DetectRnaFastaHeaderType(string line)
                 { "Gene", new FastaHeaderFieldRegex("Gene", @"GN=([^\s]*)", 0, 1) },
             };
 
+        public static readonly Dictionary<string, FastaHeaderFieldRegex> TRNAdbFieldRegexes =
+            new()
+            {
+                // >tdbR00000016 — tRNAdb accessions; the bare header's single identifier serves as both accession and name
+                { "Accession", new FastaHeaderFieldRegex("Accession", @"^>(\S+)", 0, 1) },
+                { "Name", new FastaHeaderFieldRegex("Name", @"^>(\S+)", 0, 1) },
+                { "Organism", new FastaHeaderFieldRegex("Organism", @"(?!.)", 0, 1) },
```
trishorts: **Minor — `(?!.)` returns `null` rather than an empty string, via a path that looks accidental.**

`ApplyRegex` only reads a group when `matches[Match].Groups.Count > Group`, i.e. `> 1` here. `(?!.)` is a zero-width negative lookahead with **no capture group**, so `Groups.Count` is 1, the condition is false, and the method falls through to `return result!` with `result` still `null`.

Measured on `ba573e3c`:

```
TRNAdbFieldRegexes["Organism"].ApplyRegex(">tdbR00000016")  -> null
TRNAdbFieldRegexes["Accession"].ApplyRegex(">tdbR00000016") -> "tdbR00000016"
```

So the field is `null` out of the regex and only becomes `""` further downstream, which is what your test observes. It works, but it reads as "match nothing at end-of-input" when the actual mechanism is "declare a group that does not exist so the accessor bails." A maintainer changing `ApplyRegex` to return `""` instead of `null`, or adding a null-guard, would not connect it to this.

If the loader needs an `Organism` key present (`regexResults["Organism"]` would throw if absent), something like `@"(?<Organism>^$)"` — or any pattern with a real group that cannot match — says the same thing legibly. A one-line comment on *why* the organism is deliberately empty for this format would also do it.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
```diff
@@ -55,6 +59,8 @@ public static RnaFastaHeaderType DetectRnaFastaHeaderType(string line)
                 return RnaFastaHeaderType.MzLib;
             if (line.StartsWith(">ENST"))
                 return RnaFastaHeaderType.Ensembl;
+            if (line.StartsWith(">tdb"))
```
trishorts: **Minor — `>tdb` is matched far more loosely than the other header types.**

`StartsWith(">tdb")` accepts any header beginning with those four characters, and `Accession`/`Name` are `^>(\S+)`, which accepts any non-whitespace token. So:

```
>tdbR00000016  -> TRNAdb   (intended)
>tdb           -> TRNAdb   accession ">tdb" -> "tdb"
>tdbanything   -> TRNAdb   accession "tdbanything"
>TDBR00000016  -> Unknown  (throws "Unknown fasta header format")
```

Compare the neighbours, which anchor on the real accession shape — `^>NM_\d+\.\d+ `, `^>NC_\d+\.\d+:`, `^>(ENST\d+\.\d+)`. tRNAdb accessions are `tdb` followed by a letter and digits (`tdbR00000016`, `tdbPDB000010`), so the same treatment is available: detect on something like `^>tdb[A-Z]+\d+` and capture that rather than `\S+`.

Low severity — `>tdb` is a distinctive prefix and a false positive is unlikely in practice. Mostly consistency, plus it would turn a misdetected file into a clear "unknown header format" instead of a bag of junk accessions.

Two typos in the comment just above while you are here: *somethimes* -> sometimes, *modificaitons* -> modifications. That comment also says "modifications" where the description says lowercase marks a **non-conserved residue**, which is a different claim — the description's wording is the one I would keep.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Following up with the actual fix shape, since trishorts asked how big this is: **small — about six lines, and the pattern it needs is already three lines away.**

The throw is a plain `ArgumentException` from `NucleicAcid.ParseSequenceString` (`NucleicAcid.cs:369`), raised inside the `new RNA(...)` constructor. Directly after that constructor this method already does exactly the skip-and-record thing I was asking for:

```csharp
RNA rna = new RNA(sequence, identifier, /* ... */);
if (rna.Length == 0)
    errors.Add("Line" + line + ", Rna length of 0: " + rna.Name + "was skipped from database: " + rnaDbLocation);
else if (rna.IsDecoy) decoys.Add(rna);
else targets.Add(rna);
```

So it slots into that existing chain — make `rna` nullable, wrap the construction, add one branch:

```csharp
RNA rna = null;
try
{
    rna = new RNA(sequence, identifier, /* ... unchanged ... */);
}
catch (ArgumentException e)
{
    errors.Add($"Line {line}, unsupported sequence character for {identifier}, skipped from database {rnaDbLocation}: {e.Message}");
}

if (rna is null) { }                     // already recorded above
else if (rna.Length == 0) errors.Add(/* ... */);
else if (rna.IsDecoy) decoys.Add(rna);
else targets.Add(rna);
```

No control-flow restructuring and no new helper. The `if (!targets.Any()) errors.Add("No targets were loaded from database: ...")` at the end of the method already covers the every-record-was-bad case, so that path stays sensible too.

The one judgement call is how wide to catch. Three options, my order of preference:

1. **Catch `ArgumentException` around the constructor**, as above. Simplest, and it covers `_` *plus* any other unexpected character, which is the more useful behaviour. Mild over-catch risk if the constructor ever throws `ArgumentException` for an unrelated reason.
2. **Pre-check for `_` only** — `if (sequence.Contains('_'))` before constructing. Three lines, no over-catch, exactly matches what the description claims. But every other malformed character still takes the whole file down.
3. **A dedicated exception type** in `ParseSequenceString` so the catch is precise. Cleanest long term, but it touches `NucleicAcid` in the Transcriptomics assembly, which is a wider blast radius than this PR deserves — that one wants its own PR.

**One correction to how I framed this originally.** I filed it against your tdb change, but it is not tdb-specific and you did not introduce it. I checked `master`: the `new RNA(...)` call there is equally unguarded and `LoadRnaFasta` contains no `try`/`catch` at all, so an unexpected character in *any* RNA FASTA has always aborted the whole read. What your change does is make it routine rather than exotic, because `_` is a documented part of the tdb format and tdb files are expected to contain it.

So this is a pre-existing bug that your PR surfaces, not one it causes — which makes it entirely fair for you to say "out of scope, file it separately," and I would not argue. My only ask is that it not stay invisible: either the six lines here, or an issue. Happy to open the issue myself if you would rather keep this PR tight.

nbollis: Agreed — one unsupported letter was aborting the whole load with `errors` left empty. Both loaders now degrade: `LoadRnaFasta` and `LoadRnaXML` catch the failing record, report it with its accession through the `errors` channel, and continue, so a single `_` entry no longer costs the rest of the database. `LoadRnaXML` gained an `out List<string> errors` overload (the previous signature still works unchanged). Added tests with a good → `_` → good layout for both readers, pinning that only the offending record is dropped.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: Fixed — `>tdb` detection now anchors on the accession shape (`^>tdb[A-Z]+\d+`), so `>tdb` / `>tdbanything` fall through to the clear "Unknown fasta header format" path, and the accession/name capture uses the same shape instead of `\S+`. Also fixed the two comment typos and aligned the wording with the description's non-conserved-residue claim. Added `TestTRNAdbHeaderDetectionIsAnchored` pinning the new detection boundaries.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: Right — `(?!.)` only worked because `ApplyRegex` bails on a group index that doesn't exist, which reads as intentional when it isn't. Since bare tRNAdb headers carry no organism at all, the field should be an explicit empty string: I replaced the regex with a real group that matches the zero-width start of the header (`(?<Organism>^)`), so it now always yields `""`, and added a comment stating why. `TestTRNAdbUnmodifiedFasta` also now asserts every entry's organism is empty.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
nbollis: Confirmed — this branch now carries #1276. `SanitizeAndTransform` reads `cleanedSequence.ToUpperInvariant()` (d0f7596a came in via the #1276 merge), and `TestLowercaseSequenceNormalizationIsCultureInvariant` pins `i → I` under tr-TR/az-AZ/en-US. Nothing to change here.

nbollis:
Thanks for the careful review. Disposition:
- `_` records: agreed — they now skip-and-report instead of aborting the load (`errors` carries the accession), with a regression test pinning the good → bad → good layout.
- Culture-sensitive `ToUpper`: resolved by the #1276 merge now on this branch — `SanitizeAndTransform` uses `ToUpperInvariant()` and #1276's cross-culture regression test is present. Nothing further needed here.
- `(?!.)` organism pattern: replaced with a legible non-matching-group pattern plus an explanatory comment.
- Loose `>tdb` detection + comment typos: detection and capture now anchor on the tRNAdb accession shape; typos fixed and wording aligned with the description's non-conserved-residue claim.
- `Test.csproj` BOM: restored to keep the diff scoped to tRNA support.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Verified fixed at `698d5628`, and the fix is better than what I proposed — you did both loaders, not just the FASTA one.

What I checked:

- `LoadRnaFasta` wraps construction in `try` / `catch (ArgumentException)` and records the accession through `errors`, so a `_` record is dropped and the rest load.
- `LoadRnaXML` does the same, plus `block.Clear()`. That `Clear()` is genuinely load-bearing and I want to note why, since it isn't obvious: `ParseRnaEntryEndElement` only calls `Clear()` at its line 516, *after* the `new RNA(...)` on line 512. When the constructor throws, that clear is skipped and the block still holds the failed entry's accession, sequence and parsed mods — so without your explicit `Clear()` the next entry would inherit them. Making `Clear()` `internal` is the minimal way to get that, and the comment explains it.
- The catch can't swallow the contaminant/entrapment guard: that throws `MzLibException`, which derives directly from `Exception` (`MzLibException.cs`), not from `ArgumentException`. I checked because over-catching that one would have been a real regression.
- The new `out List<string> errors` overload is genuinely non-breaking. `errors` has no default, so the old signature stays the only match when it's omitted, and an `int` positional can't bind to `out List<string>`. The `integration` job (which clones MetaMorpheus master and runs its suite) is green, which is the evidence that matters for downstream call sites.
- Moving `transformPool.Return(...)` into `finally` in both loaders is a real bonus fix — on the old code an exception leaked the pooled list.

Tests pin the good -> `_` -> good layout for both readers and assert only the offending record is dropped. Resolving.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Verified fixed at `698d5628`. `_trnaDbHeaderRegex` is `^>tdb[A-Z]+\d+` and the accession/name captures are `^>(tdb[A-Z]+\d+)` instead of `\S+`.

Probed the new pattern against the boundaries, including every accession shape in your fixture:

| header | detected |
|---|---|
| `>tdbR00000016` | TRNAdb |
| `>tdbPDB000010` | TRNAdb |
| `>tdb` | Unknown |
| `>tdbanything` | Unknown |
| `>TDBR00000016` | Unknown |

Both accession shapes actually present in `TRNAdbUnmodifiedTrimmed.fasta` (`tdbR*` and `tdbPDB*`) match, which was my main worry about tightening it — I suggested the pattern, so it was on me to confirm it doesn't reject real headers. `TestTRNAdbHeaderDetectionIsAnchored` pins the boundaries.

Typos and the comment wording are fixed too. Resolving.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Verified fixed at `698d5628`. `(?<Organism>^)` is a real named group, so the value now comes from the group rather than from `ApplyRegex` bailing out.

Confirmed the mechanism changed rather than just the text:

```
(?<Organism>^)  on ">tdbR00000016"  ->  Success=True  Groups.Count=2  Groups[1].Value=""   (returned)
(?!.)           on ">tdbR00000016"  ->  Success=True  Groups.Count=1                        (falls through to null)
```

`ApplyRegex`'s `Groups.Count > 1` guard is now satisfied, so `""` is returned deliberately instead of arriving as `null` and becoming `""` downstream. The comment explaining why tRNAdb headers have no organism is the part that keeps this from being re-broken. Resolving.

mzLib/UsefulProteomicsDatabases/Transcriptomics/RnaDbLoader.cs
trishorts: Verified at `698d5628` — settled, and #1276 merging to master on 2026-09-10 removed the ordering risk entirely.

Checked on this branch rather than assuming the merge carried it: `SanitizeAndTransform` reads `cleanedSequence.ToUpperInvariant()` (`RnaDbLoader.cs:568`), with the comment above it naming the tr-TR `i -> İ` case, and `TestLowercaseSequenceNormalizationIsCultureInvariant` is present with `tr-TR` / `az-AZ` / `en-US` cases.

One note for the record, not a request: your fixture's sequences contain lowercase `a` and `g` only, so the fixture itself still wouldn't catch a culture regression here. The dedicated culture test is what covers it, which is the right place for it. Resolving.

trishorts (approved):
Re-reviewed at `698d5628`. **All four of my earlier findings are fixed**, and CI is fully green this round. Verified each against the current tree rather than taking the replies at face value; details are in the individual threads, which I've resolved.

| # | Finding | Status |
|---|---|---|
| 1 | A single `_` aborted the entire database read | Fixed in **both** loaders, with `errors` reporting and per-record skip |
| 2 | `>tdb` matched far more loosely than its neighbours | Fixed — anchored on the accession shape, captures tightened |
| 3 | `(?!.)` returned `null` via an accidental path | Fixed — real named group, `""` returned deliberately |
| 4 | Culture-sensitive `ToUpper` widened to every tdb file | Settled — `ToUpperInvariant`, and #1276 is now on master |

Three things I want to call out as better than what I asked for:

- **You fixed the XML loader too.** I only filed against `LoadRnaFasta`. `LoadRnaXML` had the same failure mode and now degrades the same way.
- **The `block.Clear()` in the XML catch is genuinely load-bearing**, and it would be easy for a future reader to delete as redundant. `ParseRnaEntryEndElement` only clears at its line 516, after the `new RNA(...)` on 512 — so when that constructor throws, the block keeps the failed entry's accession, sequence and parsed mods and the *next* entry inherits them. Promoting `Clear()` to `internal` is the minimal enabler, and the comment says why.
- **Moving `transformPool.Return(...)` into `finally`** in both loaders fixes a pooled-list leak on the exception path that neither of us had flagged.

I also checked the two things that could have made this fix worse than the bug, and both are clean: `catch (ArgumentException)` cannot swallow the contaminant/entrapment guard (that's an `MzLibException`, which derives straight from `Exception`), and the new `out List<string> errors` overload is genuinely non-breaking — `errors` has no default, so omitting it still binds unambiguously to the old signature, and the `integration` job that builds MetaMorpheus master against this is green.

**Approving. Nothing blocking and no new findings.**

Four observations that are deliberately *not* threads — none worth another round, recorded only so they're not invisible:

- The FASTA skip message carries the accession but not the line number, while the `rna.Length == 0` branch three lines below it does (`"Line" + line`). Minor inconsistency in the same `errors` channel.
- In the XML path, the entrapment prefix is applied to `Accession` (lines 503-509) *before* the constructor throws on 512, so a skipped entrapment record reports its prefixed accession rather than the one in the file.
- `catch (ArgumentException)` also spans `ParseAnnotatedMods` and `PruneOutOfRangeSequenceVariants`. That's the mild over-catch I named when I proposed this shape, and I still think it's the right trade — noting it only so the choice is on the record.
- `Clear()` doesn't reset `OneBasedBeginPosition` / `OneBasedEndPosition`. Pre-existing, hits the success path identically, and squarely outside this PR — mentioning it in case it's worth an issue someday.

For the record on the review process: I reviewed only this PR's own payload. Now that #1276 has merged to master, the diff is honest at 5 files — the earlier 12-file / 612-addition count was #1276 showing through the stack.

zhuoxinshi (approved):

PR #1291, opened 2026-09-10
Add support for mzLib Peptide -> MaSS-Simulator input file

nbollis:
Mass simulator is a lc-ms/ms data simulator. This PR adds support for:
mzlib peptides/proteins -> MaSS-Simulator input file
MaSS-Simulator output spectra -> mgf


trishorts (commented):
This is a nice addition, but it doesn't compile right now: `build`, `integration` and `external-service-tests` all fail with the same CS1003 in `SupportedFileTypes.cs` (details inline). The other comments are non-blocking.

mzLib/Readers/Util/SupportedFileTypes.cs
```diff
@@ -36,6 +37,8 @@ public enum SupportedFileType
         CasanovoMzTab,
         DiaNnReport,
         Sdrf,
+        MaSSSimulatorPeptides,
+        MaSSSimulatorSpectra
```
trishorts, line 41: **Blocking:** this line is missing a trailing comma, which breaks the build:

```
SupportedFileTypes.cs(41,29): error CS1003: Syntax error, ',' expected
```

It looks like a merge with master, where `PytheasResult` was added. While you're fixing it, could you put the two new members *after* `PytheasResult`? Inserting them before it shifts `PytheasResult`'s numeric value.

mzLib/Tools/MaSSSimulatorConverter/MaSSSimulatorConverter.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
trishorts, line 1: Non-blocking: this project isn't added to `mzLib.sln` (and there's no other `Tools/` folder in mzLib), so CI never builds it and it can break without anyone noticing. Could you either add it to the solution or leave the converter out of this PR? The three lines in `Program.cs` are easy to call from a script or from MetaMorpheus.

mzLib/Readers/MaSSSimulator/MaSSSimulatorSpectrumFile.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// <summary>Writes spectra as MS2-only Mascot Generic Format.</summary>
+    public void WriteMgf(string outputPath)
+    {
+        using var writer = new StreamWriter(outputPath);
+        foreach (var spectrum in Results.Where(s => s.Peaks.Count > 0))
+        {
+            writer.WriteLine("BEGIN IONS");
+            writer.WriteLine($"TITLE=MaSS-Simulator scan {spectrum.ScanNumber}");
```
trishorts, line 84: Non-blocking: `ApplyTruth` fills in `PeptideSequence`, but `WriteMgf` never writes it, so the converter's truth step has no effect on the MGF it produces. Should this write a `SEQ=` line (or put the peptide in `TITLE`) when `PeptideSequence` is set? Otherwise the truth argument to the converter does nothing.

mzLib/Readers/MaSSSimulator/MaSSSimulatorSpectrumFile.cs
```diff
@@ -0,0 +1,119 @@
...
+    public void ApplyTruth(string rstPath)
+    {
+        foreach (var line in File.ReadLines(rstPath))
+        {
+            var separator = line.IndexOf(" peptide:", StringComparison.Ordinal);
+            if (!line.StartsWith("scan:", StringComparison.Ordinal) || separator < 0)
+                continue;
+            var scan = int.Parse(line[5..separator], CultureInfo.InvariantCulture);
+            var peptide = line[(separator + 9)..];
+            var spectrum = Results.FirstOrDefault(s => s.ScanNumber == scan);
```
trishorts, line 104: Nit: `FirstOrDefault` inside the per-line loop is O(spectra x truth lines), which will get slow on a full simulated run. Building a `Dictionary<int, MaSSSimulatorSpectrum>` by scan number once would avoid that.

PR #1292, opened 2026-09-10
Make the decoy isotope model charge-invariant

trishorts:
> **Project of origin:** `consensus-deconvolution-paper`

## The problem

`DecoyAveragine` builds its decoy by **moving peaks**: the n-th isotope peak is displaced by `n * (decoySpacing - C13MinusC12)` — a fixed offset in **neutral mass**, 58.955 mDa per isotope step at the default 0.9444 Da spacing.

Deconvolution matches peaks in **m/z**, not neutral mass. The matcher therefore sees that offset divided by the charge, and the decoy's distance from the real lattice shrinks as `1/z`. At m/z ≈ 900, where both bottom-up and top-down ions sit:

| charge | first decoy tooth | vs 20 ppm tol | vs 4 ppm tol |
|---:|---|---|---|
| z = 2 | 29.5 mDa · 33 ppm | distinct | distinct |
| z = 5 | 11.8 mDa · 13 ppm | **collapsed** | distinct |
| z = 15 | 3.9 mDa · 4.4 ppm | **collapsed** | distinct |
| z = 30 | 2.0 mDa · 2.2 ppm | **collapsed** | **collapsed** |

Past roughly **z = 4** at a 20 ppm tolerance — or **z = 17** at 4 ppm — the "physically impossible" comb lands within tolerance of the real peaks. The decoy is then scored as though it were a target, the decoy score distribution creeps toward the target distribution, and any FDR estimated from it is biased low.

The failure is silent. Decoys are still produced and still scored. They simply stop being wrong.

**This is not a defect in the regime the constant came from.** 0.9444 is OpenMS FLASHDeconv's `noise_iso_delta_`, and bottom-up precursors are mostly z = 2–4, where the model behaves exactly as intended. It matters for top-down proteoforms, which routinely carry z = 10–40.

## The change

Adds **`ShuffledAveragine`**, which falsifies envelope **shape** instead of peak **position**: theoretical intensities are permuted (deterministically, seeded) while masses stay on the ¹³C lattice. With nothing displaced there is no offset to be divided by charge, so the model is charge-invariant by construction. What a scorer must then fail to fit is the intensity pattern, which does not depend on charge state.

Two deliberate choices worth reviewing:

- **The apex is held at index 0** rather than permuted. Callers and the deconvolution algorithms treat index 0 of the intensity-descending arrays as the apex; permuting it would change envelope-selection behaviour rather than just the decoy's shape, and would confound "this decoy is hard to match" with "this decoy is indexed oddly."
- **Envelopes with fewer than three peaks** have nothing to permute below the apex, so they are not decoys. `IsDegenerate(index)` reports them rather than passing them off silently, so a caller computing FDR can exclude them.

Also documents the charge dependence on `DecoyAveragine` itself, at the point of use.

**No default changes.** `ToDecoyParameters()` still produces `DecoyAveragine`. Which decoy is stronger is an empirical question that depends on the regime, and this exists so the comparison can be made rather than assumed.

## Tests

16 new tests: positions unchanged, intensities a true permutation (same multiset, same sum), apex fixed and still tallest, determinism under a fixed seed, divergence under different seeds, equality/hashing, degenerate-envelope reporting, and the `1/z` collapse itself asserted directly. The 21 existing `DecoyAveragine` tests are untouched and still pass; the full deconvolution suite is green at **718 passed / 0 failed / 5 skipped**.

## Why draft

The charge-collapse arithmetic is straightforward and I would defend it as written. What I have **not** done is show empirically that the shuffled model is the better decoy on real high-charge data — that needs a target-decoy run on top-down files with both models and an AUC comparison, and I would rather have that number in hand, and agreement on the approach, before this is treated as ready.

@Alexander-Sol — you have benchmarked decoy envelope models on top-down data directly, and your results are the obvious arbiter here. Two questions: does the `1/z` analysis match what you have seen, and is a shape-permuting decoy the right shape of fix, or would an exotic-composition envelope be stronger? Happy to change the approach on the strength of your data.

### DIFF
--- a/mzLib/MassSpectrometry/Deconvolution/AverageResidue/DecoyAveragine.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/AverageResidue/DecoyAveragine.cs
@@ -55,6 +55,28 @@ namespace MassSpectrometry;
 /// var decoys = Deconvoluter.Deconvolute(spectrum, decoyParams);
 /// </code>
 ///
+/// <para><b>Charge dependence — read before using this on high-charge data</b></para>
+/// <para>
+/// The shift above is a fixed offset in <b>neutral mass</b>: 58.955 mDa per isotope step at the
+/// default spacing. Deconvolution matches peaks in <b>m/z</b>, so the matcher sees that offset
+/// divided by the charge, and the decoy's distance from the real lattice shrinks as
+/// <c>1/z</c>. At m/z ≈ 900 the first decoy tooth sits 29.5 mDa (≈ 33 ppm) from the real one at
+/// z = 2, but only 3.9 mDa (≈ 4.4 ppm) at z = 15 and 2.0 mDa (≈ 2.2 ppm) at z = 30.
+/// </para>
+/// <para>
+/// Past roughly z = 4 at a 20 ppm matching tolerance — or z = 17 at 4 ppm — the "physically
+/// impossible" comb lands within tolerance of the real peaks. The decoy is then scored as though
+/// it were a target, the decoy score distribution creeps toward the target distribution, and any
+/// FDR estimated from it is biased low. The failure is silent: decoys are still produced and
+/// still scored, they simply stop being wrong.
+/// </para>
+/// <para>
+/// This is not a defect for the regime the constant came from. Bottom-up precursors are mostly
+/// z = 2–4, where the model behaves exactly as intended. It matters for top-down proteoforms,
+/// which routinely carry z = 10–40. For that regime prefer <see cref="ShuffledAveragine"/>,
+/// which falsifies envelope shape instead of peak position and so is charge-invariant.
+/// </para>
+///
 /// <para>Reference: Käll et al. (2008) target-decoy approach; OpenMS FLASHDeconv
 /// <c>SpectralDeconvolution.cpp</c> <c>noise_iso_delta_ = 0.9444</c>.</para>
 /// </summary>
--- a/mzLib/MassSpectrometry/Deconvolution/AverageResidue/ShuffledAveragine.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/AverageResidue/ShuffledAveragine.cs
@@ -59,11 +59,37 @@ namespace MassSpectrometry;
 /// var decoys = Deconvoluter.Deconvolute(spectrum, decoyParams);
 /// </code>
 ///
+/// <para><b>Measured behaviour — read this before choosing a decoy</b></para>
 /// <para>
-/// Which decoy is stronger is an empirical question and is expected to depend on the regime;
-/// this class exists so the comparison can be made rather than assumed. Nothing here changes
-/// the default: <see cref="DecoyAveragine"/> remains what
-/// <see cref="DeconvolutionParameters.ToDecoyParameters"/> produces.
+/// Benchmarked over 14.4M envelopes from 20 Jurkat top-down files and 6 Velos bottom-up files
+/// (target vs decoy ROC-AUC on <c>GenericScore</c>, deconvolution at 4 ppm, z 1–60):
+/// </para>
+/// <list type="bullet">
+///   <item>This model is <b>charge-invariant, as designed</b>: mean AUC 0.592 at z ≤ 4 and 0.583
+///   at z ≥ 10, a degradation of 0.009. <see cref="DecoyAveragine"/> over the same range falls
+///   from 1.000 to 0.762, a degradation of 0.238, and its AUC correlates with its own predicted
+///   1/z tooth displacement at r = +0.747.</item>
+///   <item>But its <b>absolute separation is poor</b>: AUC 0.593 (top-down) and 0.700 (bottom-up),
+///   against 0.999 and 1.000 for the shifted model. Shuffled decoys score a median 0.916 where
+///   targets score 0.948 — the current scorer largely cannot tell a shape-falsified envelope from
+///   a real one, because it rewards finding peaks at predicted <i>positions</i> and this model
+///   does not move any.</item>
+///   <item>The two cross over at roughly <b>z ≈ 18–20</b>. Above it this model is the better decoy
+///   (z = 25: 0.607 vs 0.472; z = 30: 0.708 vs 0.555 — note the shifted model drops below 0.5,
+///   meaning its decoys outscore real envelopes). Below it the shifted model is far better.</item>
+/// </list>
+/// <para>
+/// So this is <b>not</b> a drop-in replacement. Shape falsification is only as strong as the
+/// scorer's sensitivity to shape, and the present envelope score saturates: it returns above 0.9
+/// for 55% of shuffled decoys. Pairing this model with a scorer that measures fit to the observed
+/// intensity pattern is what would make it work; on its own it trades a decoy that fails at high
+/// charge for one that is mediocre everywhere.
+/// </para>
+/// <para>
+/// Nothing here changes the default: <see cref="DecoyAveragine"/> remains what
+/// <see cref="DeconvolutionParameters.ToDecoyParameters"/> produces. This class is provided so the
+/// comparison can be made, and so a caller working at very high charge has an alternative whose
+/// failure mode is known.
 /// </para>
 /// </summary>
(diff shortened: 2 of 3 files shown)


trishorts:
## Benchmark results — mechanism confirmed, this fix is not sufficient

Ran the experiment this draft was waiting on: **14,366,686 envelopes**, 20 Jurkat top-down files and 6 Velos bottom-up files, 250 scans sampled per file, deconvolution at 4 ppm / z 1–60, target vs decoy ROC-AUC on `GenericScore`.

It confirms the defect cleanly and **partly refutes my proposed fix**.

### The `1/z` collapse is real

Detection rate — how many envelopes each decoy finds relative to the target. This needs no scores at all:

| regime | target | shifted decoy | shuffled decoy |
|---|---:|---:|---:|
| bottom-up (z 2–4) | 107,705 | 6,544 — **6.1%** | 23,575 — 21.9% |
| top-down (z 10–40) | 5,915,406 | 5,759,683 — **97.4%** | 2,553,773 — 43.2% |

The shifted decoy goes from finding 6% of what the target finds to finding 97%. It is no longer modelling false positives; it is finding the real envelopes.

Score separation by charge, with the predicted tooth displacement alongside:

| z | predicted offset | shifted AUC | shuffled AUC |
|---:|---:|---:|---:|
| 2 | 32.8 ppm | 1.000 | 0.602 |
| 5 | 13.1 ppm | 1.000 | 0.596 |
| 9 | 7.3 ppm | 0.992 | 0.586 |
| 12 | 5.5 ppm | 0.933 | 0.532 |
| 15 | 4.4 ppm | 0.775 | 0.621 |
| 18 | 3.6 ppm | 0.640 | 0.625 |
| 20 | 3.3 ppm | 0.610 | 0.652 |
| 25 | 2.6 ppm | **0.472** | 0.607 |
| 30 | 2.2 ppm | 0.555 | 0.708 |

Mean AUC z ≤ 4 → z ≥ 10: shifted **1.000 → 0.762** (degradation 0.238), shuffled **0.592 → 0.583** (degradation 0.009). The shifted model's AUC correlates with its own predicted `1/z` displacement at **r = +0.747**; the shuffled model's does not (r = −0.215). At z = 25 the AUC is below 0.5, meaning decoys outscore real envelopes.

**One caveat that matters for how this gets measured:** 84.9% of top-down target envelopes are z ≤ 4. The aggregate top-down AUC is **0.999** — computed almost entirely where the shifted decoy still works. The aggregate does not merely miss the collapse, it conceals it. Only the stratified view shows anything.

### My fix does not work as shipped

| | top-down AUC | bottom-up AUC |
|---|---:|---:|
| shifted | **0.999** | **1.000** |
| shuffled | 0.593 | 0.700 |

`ShuffledAveragine` is charge-invariant exactly as designed — but its absolute separation is poor. Shuffled decoys score a median **0.916** against targets at **0.948**, with 55% of them above 0.9.

The reason is worth stating plainly: **the current envelope score largely cannot tell a shape-falsified envelope from a real one.** It rewards finding peaks at predicted *positions*, and this model deliberately moves none. Shape falsification is only ever as strong as the scorer's sensitivity to shape.

So on its own this trades a decoy that fails at high charge for one that is mediocre everywhere. The two do cross over around z ≈ 18–20, and above that the shuffled model is genuinely the better of the two — but both are weak there.

### Where that leaves the PR

I have updated the class documentation to say all of the above rather than leave the earlier optimistic framing standing. What I think the data supports:

1. **The documentation of the charge collapse should land** regardless — it is a real, quantified, silent failure mode in the default decoy, and nothing currently warns anyone about it.
2. **`ShuffledAveragine` should not become the default**, and I am not proposing it. It is worth keeping as a known-characterised alternative above z ≈ 20, and as the right building block *if* the scorer gains shape sensitivity.
3. **The decoy and the scorer have to be fixed together.** A charge-invariant decoy paired with a position-only scorer is not an improvement. That is the more interesting finding here and it is not one I set out to make.

Happy to split (1) into its own small PR and park (2) if that is cleaner.

@Alexander-Sol — does the `1/z` degradation match what you have seen benchmarking decoy models on top-down data? And given the scorer limitation above, would an exotic-composition envelope (which changes *both* positions and weights) sidestep the problem better than pure shape permutation?

Benchmark harness and analysis: `DecoyBench` + `decoy_model_benchmark.py`, happy to share.

trishorts:
## Independent confirmation from a second measurement stack

The benchmark above was measured with mzLib's own `GenericScore` on individual envelopes. A fair objection is that the saturated logistic could be doing the work rather than the decoy geometry, and that the population is 85% z ≤ 4.

So I re-tested the same question with **nothing in common except the decoy model**: a different level (consolidated features, not envelopes), a different score (a union-grid apex cosine that is not mzLib's), and a different comparison set. `DecoyAveragine`'s model comes down to feature level cleanly, since the comb spacing is a scorer parameter — 0.9444 / 1.0033548 = **0.941236**.

**Controlled comparison.** One pool of 2,666,108 consolidated features from 20 top-down files, one score, one target set. The only thing that varies is how the null is built:

- **spacing-shifted** — comb laid at 0.9444 Da. This is `DecoyAveragine`.
- **mass-relocated** — feature moved 5–50 Da off the ¹³C lattice, charge and RT kept.
- **RT-randomised** — mass and charge kept, moved to a random retention time.

### At 5% FDR — threshold demanded, and share of that charge stratum surviving

| z | features | spacing-shifted | mass-relocated | RT-randomised |
|---|---:|---|---|---|
| 1–4 | 1,486,822 | 0.616 · 1% | 0.575 · 2% | 0.997 · 0% |
| 5–9 | 1,086,235 | 0.589 · 1% | 0.636 · 1% | 0.694 · 1% |
| 10–14 | 65,747 | 0.573 · 19% | 0.585 · 17% | 0.571 · 19% |
| 15–19 | 18,606 | 0.738 · 28% | **0.507 · 78%** | 0.574 · 64% |
| 20–24 | 6,969 | **0.890 · 3%** | **0.520 · 84%** | 0.629 · 60% |
| 25+ | 1,729 | **0.942 · 0%** | **0.600 · 79%** | 0.586 · 82% |

The spacing null's threshold climbs **0.573 → 0.738 → 0.890 → 0.942** as charge rises. That is the collapse stated in units that matter operationally: because its decoys score ever higher, holding a fixed error rate costs progressively more real signal, until above z = 25 it retains **none of the stratum**. The other two nulls sit between 0.51 and 0.64 throughout.

### Identification coverage — the column nothing is fitted to

Median features per file, 17 files with confident proteoform IDs (5,585 total), 17-file totals in brackets:

| null | 1% FDR | IDs | 5% FDR | IDs |
|---|---:|---:|---:|---:|
| mass-relocated | 1,319 (23,998) | **85.6%** | 3,568 (58,455) | **94.6%** |
| **spacing-shifted** | 900 (16,807) | **67.1%** | 3,056 (46,280) | **90.9%** |
| RT-randomised | 39 (1,507) | 3.8% | 1,606 (26,492) | 87.2% |

At 1% FDR the spacing null costs **18.5 percentage points** of real identifications relative to a null that does not collapse.

### What this settles, and what it doesn't

**Settles:** the charge collapse is a property of the decoy's geometry, not an artifact of mzLib's scorer. Two measurement stacks with no shared components beyond the decoy model itself reach the same conclusion, and in both the effect tracks the predicted `1/z` displacement.

**Also settles something against my earlier comment:** I reported an aggregate AUC of 0.999 for the spacing model on top-down data and flagged that 85% of that population is z ≤ 4. This confirms the flag was the important part — scored properly and stratified, the same model is the *weakest of three* above z = 15. Anyone validating a decoy on pooled numbers would conclude it is fine.

**Doesn't settle:** which replacement is right. `ShuffledAveragine` as proposed here is charge-invariant but under-powered against the current scorer, for the reason in the previous comment. The mass-relocated null above works well, but it is a *feature-level* construct — relocating a whole feature to a different mass — and there is no obvious envelope-level equivalent, since `AverageResidue` is a mass-indexed lookup with no charge context. That gap is the open design question, and I would rather have your read on it than guess.

So my suggestion stands: land the documented failure mode, hold the replacement.

Analysis: `decoy_side_by_side.py`, and `decoy_model_benchmark.py` + `DecoyBench` for the envelope-level run.

PR #1294, opened 2026-09-11
Assign feature-level q-values by competing features against decoy features

trishorts:
> **Project of origin:** `consensus-deconvolution-paper`

Second of two. **Contains #1293 as its first commit**, because GitHub will not accept a base branch that lives only in the fork. Review #1293 first; the FDR work is the second commit, `Assign feature-level q-values…`. Once #1293 merges, this diff reduces to that commit alone.

## Why

#1293 gives every feature a score that varies. A score is only useful if a consumer knows where to cut, and a feature list with no error rate can only be used whole or thresholded by guesswork.

mzLib already estimates FDR for individual isotopic envelopes during deconvolution (`DeconvolutionQValueCalculator` against `DecoyAveragine`). That operates one scan at a time and says nothing about an assembled feature, which is what a consumer of an `_ms1.feature` file actually receives.

## `IsotopicEnvelopeProbe`

A public primitive: given a mass, a charge and a spectrum, gather the observed peaks a species there would occupy and return them as an `IsotopicEnvelope`.

Deconvolution answers *"what is in this spectrum"*. This answers *"how well does this spectrum support THIS hypothesis"*, which deconvolution cannot be asked because it only ever reports the hypotheses it chose. That distinction is the whole reason the class exists: **a decoy feature has no peaks of its own**, so it cannot be scored by any path that starts from an observed envelope. `FindIsotopicEnvelope` in the classic algorithm is private and needs a seed peak, so it cannot serve.

Returns `null`, and a `NaN` score, when nothing matches. Not an empty envelope and not zero, so a hypothesis that could not be evaluated stays distinguishable from one that evaluated badly.

## `ConsensusFeatureFdr`

One decoy per feature, keeping its **charge and retention time** and displacing its mass by a **half-integer multiple of the ¹³C spacing, 5 to 50 Da** away.

- Keeping charge and RT means the decoy probes the same spectrum at the same density as its target, so the comparison isolates the one thing that makes a feature real: whether a complete isotope envelope actually sits at that mass.
- The half-integer offset places the decoy comb *between* the real teeth rather than on them.

**Why displace the feature rather than perturb the isotope spacing.** A spacing perturbation, which is how the envelope-level decoy works, is a fixed offset in **neutral mass**. Matching happens in **m/z**, so the matcher sees that offset divided by the charge. Past roughly z = 4 at a 20 ppm tolerance the perturbed comb lands back on the real peaks and the decoy quietly stops being wrong. A 5–50 Da whole-feature displacement does not shrink with charge, and `B2_DecoySeparationDoesNotShrinkWithCharge` asserts that directly across z = 2 to 50.

(That charge dependence in the existing envelope-level decoy is a separate finding, written up in #1292. Nothing here changes it or depends on it.)

**Two choices worth a reviewer's attention:**

1. **Target and decoy are both scored by probing the apex scan**, rather than reusing the trace-aggregated `QualityScore` from #1293. A decoy has no trace to aggregate over, so comparing a trace-averaged target against a single-scan decoy would favour the target for a reason unrelated to being real. The trace aggregate stays the descriptive per-feature number; this is the matched one.

2. **A decoy the spectrum cannot support at all enters the null at the lowest score rather than being dropped.** Dropping it would bias the null upward and make the estimated FDR look better than it is.

## What the q-value means

It estimates the fraction of surviving features that are **not real isotope envelopes**. It does **not** estimate the fraction that fail to correspond to an identifiable molecule: most true features are never identified by any search, and a feature can be a perfectly real envelope of a species no database contains. That is stated in the XML docs, because it is the way this number will otherwise be misread.

## Output

`MassFeature` gains `QValue`; `Ms1Feature` gains a `Q_value` column alongside the score columns from #1293. Written only on request, so the default output stays byte-compatible with the TopFD schema.

## Tests

12 new: probe hit and miss; `NaN` rather than zero for an unsupported hypothesis; a real hypothesis outscoring a displaced one; argument validation; empty spectra; half-integer off-lattice decoy construction over 500 draws; the charge-invariance property; q-value assignment and bounds; real features ranking ahead of fabricated ones; monotonicity in score; seed reproducibility; empty input.

Surrounding suites pass: **749 passed, 0 failed, 5 skipped**.

## Still to come

These two PRs make the scoring and error-rate model available in mzLib. A third piece of work, not in either PR, is regenerating the manuscript's feature-level numbers with this implementation rather than the external one they were computed with, and confirming the two agree.

### DIFF
--- a/mzLib/MassSpectrometry/Deconvolution/Consensus/ConsensusFeatureFdr.cs
+++ b/mzLib/MassSpectrometry/Deconvolution/Consensus/ConsensusFeatureFdr.cs
@@ -1,0 +1,166 @@
+using Chemistry;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MassSpectrometry.Deconvolution.Consensus
+{
+    /// <summary>
+    /// Assigns a false-discovery-rate q-value to each <see cref="MassFeature"/> by competing the
+    /// real features against decoy features built from them.
+    ///
+    /// <para><b>Why features need their own null</b></para>
+    /// <para>
+    /// mzLib already estimates FDR for individual isotopic envelopes during deconvolution
+    /// (<see cref="DeconvolutionQValueCalculator"/> against <see cref="DecoyAveragine"/>). That
+    /// operates one scan at a time and says nothing about an assembled feature, which is what a
+    /// consumer of an <c>_ms1.feature</c> file actually receives. A feature list with no error
+    /// rate can only be used whole or thresholded by guesswork.
+    /// </para>
+    ///
+    /// <para><b>How the decoys are built</b></para>
+    /// <para>
+    /// Each decoy keeps a real feature's charge and retention time and displaces its mass by a
+    /// half-integer multiple of the <sup>13</sup>C spacing, 5 to 50 Da away. Keeping charge and
+    /// retention time means the decoy probes the same spectrum at the same density as its target,
+    /// so the comparison isolates the one thing that makes a feature real: whether a complete
+    /// isotope envelope actually sits at that mass. The half-integer offset places the decoy comb
+    /// between the real teeth rather than on them.
+    /// </para>
+    /// <para>
+    /// Displacing the mass, rather than perturbing the isotope spacing as the envelope-level decoy
+    /// does, matters at top-down charge states. A spacing perturbation is a fixed offset in neutral
+    /// mass, so the matcher, which works in m/z, sees it divided by charge; past roughly z = 4 at a
+    /// 20 ppm tolerance the perturbed comb lands back on the real peaks and the decoy stops being
+    /// wrong. A whole-feature displacement of 5 to 50 Da does not shrink with charge.
+    /// </para>
+    ///
+    /// <para><b>What the q-value means, and does not</b></para>
+    /// <para>
+    /// It estimates the fraction of surviving features that are not real isotope envelopes. It does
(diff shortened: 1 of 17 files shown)


trishorts:
Pushed a third commit: **the consuming half of the q-value**.

`FromFileDeconvolutionParameters` gains `MaxFeatureQValue`, so a search fed an external feature list can be given only the features that pass. A q-value nothing can act on is decoration.

The filter has to be applied at the `Ms1Feature` row level, because `IMs1FeatureFile.GetMs1Features` already expands rows into per-charge features and drops the q-value at that boundary. The interface gains an overload taking the threshold, with a **default implementation that returns everything** — a format carrying no q-value cannot be thresholded on one, so Dinosaur and friends are unaffected.

**One decision worth checking:** rows with no q-value are **kept**, not dropped. An external TopFD or FLASHDeconv file has no q-values at all, so treating absent as failing would silently empty the precursor list the moment a caller set a threshold, and the only symptom would be a search that quietly finds less. `D2_RowsWithoutAQValueSurviveAnyThreshold` pins that.

`MaxFeatureQValue` is also carried through `Clone`, `EqualProperties` and `AddHashCodes`. `Clone` dropping it would silently lose the threshold; equality ignoring it would let two parameter sets over the same file that admit *different* feature subsets compare equal, so a cached instance could serve the wrong list.

Suite at **757 passed, 0 failed**.

PR #1299, opened 2026-09-14
Keep the sign of PSI-MOD formal charges read from the XML ontology

trishorts:
## Problem

`ModificationLoader.GetFormalChargesDictionary(obo)` parsed PSI-MOD's `FormalCharge` value with `[^\d]`, keeping only the digits. Every negative charge (`1-`, `2-`, `3-`) came back positive.

That dictionary drives the proton correction in the ptmlist loaders (`AdjustMonoIsotopicMassForFormalCharge`), which subtracts charge × proton from the MM line and charge × H from the formula. With the sign flipped, an anionic modification **loses** a proton it should **gain**, an error of two protons.

The OBO-text overload in `Loaders.GetFormalChargesDictionary(IEnumerable<OboTerm>)` already handles the sign through `GetChargeString`. The XML overload, which MetaMorpheus uses (`GlobalVariables.LoadModifications` → `Loaders.GetFormalChargesDictionary(PsiModDeserialized)`), never got the same fix.

## Change

- `ParseFormalCharge` reads the magnitude and the sign separately.
- Two tests in `TestDatabaseLoaders`:
  - `PsiModXml_FormalChargesKeepTheirSign`: the XML reader returns +1 for MOD:00083, −1 for MOD:01701, −2 for MOD:00145 and −3 for MOD:00147, and agrees with the OBO-text reader on every term.
  - `NegativeFormalCharge_AdjustmentAddsAProton`: a ptmlist entry cross-referencing a 1− term gains exactly one proton on the MM line and one H in the formula.

  Both failed before the fix: the XML reader returned `1` where `-1` was expected, and the mass moved by −1.00728 instead of +1.00728.

## Impact

No current output changes. Every ptmlist entry that cross-references a charged PSI-MOD term is 1+ (trimethyllysine, diphthamide and so on), which is why the bug has been silent. It would have become a real mass error as soon as UniProt cross-referenced a negatively charged term.

## Testing

Full suite excluding `ExternalService`: 6,025 passed, 0 failed, 31 skipped.

https://claude.ai/code/session_01JCt4yaoSMk84YrzXj8hKDt


nbollis (dismissed):

acesnik (commented):
🤖 This closes the half of the formal-charge sign bug that #659 missed. That PR fixed the OBO-text reader back in 2022 and left the XML overload's parse as it was, so the digits-only regex survived in exactly the overload `GlobalVariables.LoadModifications` calls (`GlobalVariables.cs:407-408` → `Loaders.GetFormalChargesDictionary(PsiModDeserialized)`). Pinning the two readers against each other is the right way to stop them drifting apart a second time.

The two new tests are real regression tests. Mutating `ParseFormalCharge` back to `return magnitude;` fails both: `PsiModXml_FormalChargesKeepTheirSign` reports `Expected: -1 / But was: 1` on MOD:01701, and `NegativeFormalCharge_AdjustmentAddsAProton` lands 2.0145529 Da low — the two-proton error from the body, measured. The XML-vs-OBO cross-check is doing work rather than passing by construction: all 85 of the XML's charged keys are present in the OBO dictionary, so 85 assertions execute and none disagree, with `MOD:01982` the single OBO-only accession that the `TryGetValue` guard skips.

"No current output changes" holds, downstream too. mzLib's shipped `ptmlist.txt` cross-references 487 distinct PSI-MOD accessions, of which 11 carry a formal charge and all 11 are `1+`; the `EngineLayer/Data` pair in MetaMorpheus is 463 accessions, 10 charged, all `1+`. The ontology carries 85 charged terms of which 22 are negative, and not one of the 22 is reachable from a ptmlist entry in either repo. The 2021 `DatabaseTests/PSI-MOD.obo` snapshot gives the same answer as the 2014 XML, so this stays inert if the bundled ontology is refreshed.

One line in the new test is inert, though: `Assert.That(formalCharges.Values, Has.None.EqualTo(0))`. Neither fixture contains a `FormalCharge` of `0`, and `ParseFormalCharge` cannot return `0` regardless — a digit-free value throws inside `int.Parse` rather than yielding it — so that assertion passes with the fix, without it, and under the mutation above. `Assert.That(formalCharges.Values.Count(v => v < 0), Is.GreaterThan(0))` would bite, and is worth having given what the test is about: its fixture is a trimmed file (`DataFiles/Ontologies/PSI-MOD.obo.xml`, 85 of 2001 terms per PROVENANCE.md), and a future re-trim that dropped the negative terms would leave the rest of this test green against nothing. It mirrors what `TestPsiModLoading` already does with `anyNegativeValue` on the OBO side.

The bigger item is for later rather than here. There are still two independently written sign parsers for one ontology field in two assemblies — `ParseFormalCharge` in `Omics`, and `GetChargeString` plus a whole-line digit strip in `UsefulProteomicsDatabases.Loaders` — which is what produced this bug, and the OBO side is the looser of the two: `GetChargeString` takes the first `+` or `-` anywhere in the value and is safe only because that value is exactly `FormalCharge: "1-"`. Having the OBO overload call `ParseFormalCharge` would retire the class of bug. The cross-check you added covers it well enough for now, so I would not widen this PR for it.

Holding the checkmark only for the one-line assertion swap; the fix itself I am happy with, and I will approve as soon as it is pushed.

trishorts:
🤖 Done in 581d4813: swapped the inert `Has.None.EqualTo(0)` for `Assert.That(formalCharges.Values.Count(v => v < 0), Is.GreaterThan(0))`. Checked that it bites on its own: with `ParseFormalCharge` mutated to `return magnitude;` and the three negative-value `AreEqual` lines commented out, the test fails on the new line (`Expected: greater than 0 / But was: 0`). The branch is also up to date with master (merged 3 commits, no conflicts). Agreed on routing the OBO overload through `ParseFormalCharge` as a separate follow-up.

Issue #994, opened 2025-12-17
Create generic interfaces for PSMs, Peptides, and Proteins


Alexander-Sol:
This is mostly already done or in progress. This issue is designed to help organization

```mermaid
classDiagram
  class ISpectralMatch {
    <<interface>>
    +double[]? QuantitativeInformation
    +string FullFilePath
    +string FullSequence
    +string BaseSequence
    +int OneBasedScanNumber
    +double Score
    +IEnumerable~IBioPolymerWithSetMods~ GetIdentifiedBioPolymersWithSetMods()
  }
  
  class IBioPolymerWithSetMods {
    <<interface>>
    +Dictionary~ISampleInfo,double~? IntensitiesBySample
    +...
  }
  
  class IBioPolymerGroup {
    <<interface>>
    %% === Quantification ===
    +Dictionary~ISampleInfo,double~ IntensitiesBySample
  
    %% === Other Stuff ===
    +bool IsDecoy
    +bool IsContaminant
    +List~ISampleInfo~ SamplesForQuantification
    +HashSet~IBioPolymer~ BioPolymers
    +string BioPolymerGroupName
    +double BioPolymerGroupScore
    +HashSet~IBioPolymerWithSetMods~ AllBioPolymersWithSetMods
    +HashSet~IBioPolymerWithSetMods~ UniqueBioPolymersWithSetMods
    +HashSet~ISpectralMatch~ AllPsmsBelowOnePercentFDR
    +double QValue
    +double BestBioPolymerWithSetModsQValue
    +double BestBioPolymerWithSetModsScore
    +List~string~ ModsInfo
    +List~IBioPolymer~ ListOfBioPolymersOrderedByAccession
    +string GetTabSeparatedHeader()
    +void MergeWith(IBioPolymerGroup otherBioPolymerGroup)
    +IBioPolymerGroup ConstructSubsetBioPolymerGroup(string fullFilePath, List~SilacLabel~? silacLabels)
  
  
  }
  
  
  class ISampleInfo {
    <<interface>>

    +string FullFilePathWithExtension
    +string Condition
    +int BiologicalReplicate
    +int TechnicalReplicate
    +int Fraction
    +int? Plex
  }
  
  
  IBioPolymerGroup --> "0..*" ISpectralMatch : AllPsmsBelowOnePercentFDR
[...]

PR #1300, opened 2026-09-14
Build every sample column label in one place, and name an isobaric channel by its sample

trishorts:
> **Project of origin:** `QuantProject` (consumed by `sdrf`)

Milestone **R3** of the reporting tranche, on top of #1285 and #1296.

## The defect

A sample column label was built in three places, and they agreed only by coincidence:

- `SampleGroupBuilder` spelled an isobaric channel `{file}_{channel}` inline (`:119`) for the grouped protein table.
- `QuantificationWriter.SampleColumnLabel` got the same string from `IsobaricQuantSampleInfo.ToString()` for `PeptideQuantification.tsv` and `ProteinGroupQuantification.tsv`.
- The label-free branch spelled `{condition}_{biorep+1}` inline beside them (`:54`).

None of them could say which **sample** a channel held, because nothing in mzLib carried the name. MetaMorpheus's TMT design has a required `Sample Name` column, and `TmtExperimentalDesign.ToMzLibDesign` has to drop it, since `IsobaricQuantSampleInfo` had no field for it. So an 11-plex protein table names its columns after one file and eleven channel numbers, and the reader joins them to the design file by hand.

## The change

**`IsobaricQuantSampleInfo.SampleName`**, a `string?` init property.
- **Display only.** It stays out of `Equals`, `GetHashCode` and `UniqueIdentifier`, which are still keyed on file and channel. A sample name is free text, and two channels a design happens to name alike are still two measurements. A test pins that.
- **Not a constructor parameter**, so none of the 86 existing `new IsobaricQuantSampleInfo(` sites change.
- **Collapse keeps it only when every merged column shares it.** `CollapseStrategyBase.CollapsedSampleInfo` takes every other field from the group's first member, but groups are keyed on condition and the kept replicate dimensions, never on file, channel or sample. Collapsing biological replicates therefore merges different samples, and the merged column must not be named after the first of them. An unnamed member counts as a disagreement. MetaMorpheus runs `NoCollapse`, so its output is unaffected.

**`SampleGroupLabels.ForSample`** is the one function. `SampleGroupBuilder`'s isobaric and label-free branches and both of `QuantificationWriter.SampleColumnLabel`'s branches call it, so the grouped table and the matrices start every column from the same label. **That is where the agreement ends:** when two labels collide, each writer still makes them unique in its own way (see *Left alone*), so a collided channel can carry a different header in each file.

| Sample | Label |
|---|---|
| isobaric, the design names the sample | `{sample}_{file}_{channel}` |
| isobaric, unnamed | `{file}_{channel}`, **exactly as today** |
| label-free | the file name, or `{condition}_{biorep+1}`, as today |

**The file is in both isobaric forms.** The fractions of one plex share their sample and differ only by file. Without the file, six fractions all get the same name. They usually share a directory too, so `Disambiguate` has nothing to widen with and falls back to ordinals: `Patient7_126`, `Patient7_126_2`, … which no longer say which fraction is which.

`Identity`, `CountIdentity`, `CountLabel` and column keying are untouched.

## Why an unnamed channel does not fall back to condition and replicate

The plan's D1 specified `{condition}_{biorep}_{file}_{channel}` as a fallback, and the first commit here implemented it. **The second commit removes it, and the reason is adoptability:** that fallback renames every channel of every isobaric run the moment a caller upgrades, including callers that do not pass `SampleName` yet — which is all of them.

`integration` would have shown it. MetaMorpheus's TMT test design gives each channel a condition, but MetaMorpheus does not pass a name, so `MultiplexQuantificationTests.cs:124` and `:226-231` would have read `CondA_1_VA084TQ_6_126` where they assert `VA084TQ_6_126`.

As written, **nothing changes for a caller until it starts passing names.** This release is adoptable on its own, and MetaMorpheus opts in with its own PR: `SampleName = annotation?.SampleName` at `TmtExperimentalDesign.cs:527`, with its test expectations moving in the same diff. The fallback would rarely have fired there anyway, because MetaMorpheus's TMT design already requires a `Sample Name` column.

## Output

- **For every existing caller, no column changes.** No-design, label-free and unnamed-isobaric goldens hold. The one new refusal is a design that lists one sample twice (below). That covers a TMT channel and also a label-free file: `Add(f, sfi, sfi)` used to be accepted and now throws. No current producer builds either.
- **Only a channel whose design names its sample gets the new label, in both TMT tables.** The isobaric golden now holds one named and one unnamed channel: `Intensity_Patient7_goldenA_126` beside `Intensity_goldenA_127N`. Count columns (`SpectralCount_goldenA`) are still named by file alone.
- **Left alone:**
  - `IsobaricQuantSampleInfo.ToString()` no longer names any column.
  - `QuantificationWriter.UniqueColumnLabels` keeps its `_2`/`_3` suffixing. `SampleGroupLabels.Disambiguate` widens by directory first and its ordinals skip names already in use, so this is a different collision rule and a separate change. **Consequence:** colliding labels get different headers in the two files. Unnamed `run.raw` + `run.mzML` already do this on master, and you can't join the two tables by column name for those channels. No values move.

No MetaMorpheus production code parses these column names back. I checked for `Intensity_`/`SpectralCount_`/`CountOccupancy_` prefixes, `Split('_')` and `StartsWith` readers across EngineLayer, TaskLayer, GUI, GuiFunctions and CMD.

## Falsification

Each mechanism was broken on purpose and its named test went red, then green again once restored:

| Break | Test that goes red |
|---|---|
| drop the file from the isobaric label | `GetTabSeparatedHeader_Isobaric_SixFractionsOfOnePlexKeepDistinctColumns` (columns fall to `plex1_Patient7_126` plus ordinals) |
| point the matrix writer back at `ToString()` | `SampleColumnLabel_Isobaric_NamesTheChannelAsTheProteinTableDoes` |
| restore the condition-and-replicate fallback | `ForSample_IsobaricChannelWithNoSampleName_KeepsItsFileAndChannelLabel`, and the isobaric golden |
| stop copying `SampleName` through collapse | `CollapsingIsobaricSamples_PreservesChannelIdentity` |
| collapse takes the first member's name even when members disagree | `CollapsingSamplesThatDoNotShareAName_LeavesTheCollapsedColumnUnnamed` |
| `Add` stops applying the repeated-sample rule | `Add_RejectsOneChannelListedTwice_NamingBothSamples`, `FromSamples_RejectsOneChannelListedTwice` |
| the engine stops applying it | `Run_WithADesignThatRepeatsAChannel_ReturnsFailureNamingIt` |
| the rule keys on channel label alone, ignoring the file | `DescribeRepeatedSample_OneSampleOnTheSameChannelOfEveryPlex_IsNotARepeat` |
| the engine asks the rule one design key at a time | `Run_WithTwoDesignKeysHoldingOneChannel_ReturnsFailureNamingIt` |
| the engine stops skipping a null sample array, or a null dictionary | `ValidateEngine_WithNullsInTheDesign_ReportsNoRepeat` |
| the rule stops skipping null entries | `DescribeRepeatedSample_IgnoresNullEntries`, `ValidateEngine_WithNullsInTheDesign_ReportsNoRepeat` |
| `Add` checks for repeats before its file checks again | `Add_ReportsAnAlreadyAddedFileBeforeARepeatedSample` |
| the matrix writer labels an isobaric channel by `ToString()` again (end to end) | `Run_WithANamedDesign_NamesChannelsAlikeInTheMatricesAndTheGroupedTable` |
| the matrix writer's label-free branch asks `ForSample` without `labelFreeByFileName` | `SampleColumnLabel_UsesFileNameForLabelFreeSamples` |

**Two existing assertions were re-aimed, not just re-spelled.**
- `GetTabSeparatedHeader_Isobaric_CountColumnsAreOnePerFile` asserted `Does.Not.Contain("SpectralCount_fileA_126")`, which only means something if `fileA_126` is a label the header actually uses. Building that name from `ForSample` alone would not change anything, because its channels were unnamed and `ForSample` returns the same string. The test now names channel 126, so the two label forms differ, and first asserts that both channels' labels are present as intensity columns.
- `..._ChannelColumnsAreOrderedByReporterMz` compared two `IndexOf` results. A missing 127N column's `-1` satisfies "less than", so it now asserts presence first.

## A channel listed twice is now refused

Keeping `SampleName` out of equality is only safe if two samples cannot share a file and channel. `SampleExperimentalDesign.FromSamples` and MetaMorpheus's TMT design both refuse that shape, but a hand-built `IExperimentalDesign` could always pass it. Quantification matrices index columns by sample (`QuantMatrix`: `_colIndex[ColumnKeys[i]] = i`), so the second entry silently took the first one's column. With sample names in the labels, the merged column would also carry **one** of the two names over both values.

**Public SDRF has this shape.** I checked the canonical `bigbio/sdrf-annotated-datasets` corpus: 1,236 files, 59 of which name individual TMT/iTRAQ channels.
- **The one offender:** `PXD040455`, a TMT × SILAC design, lists a `_light` and a `_heavy` source name under the same reporter channel of the same file **551** times. No other document does.
- **Not reachable today:** MetaMorpheus's TMT design already refuses a channel described twice with different samples.
- **Reachable later:** an SDRF-to-design projection would hand it straight in.

**The rule is `SampleExperimentalDesign.DescribeRepeatedSample`, enforced in two places:**
- **`SampleExperimentalDesign.Add`** throws, which covers `FromSamples`, naming the channel and every sample name involved: `channel 127N of 'Chip1_K562HeLa_F2.raw' is listed 2 times, as 'Chip1_F2_TMT127N_light' and 'Chip1_F2_TMT127N_heavy'`.
- **`QuantificationEngine.ValidateEngine`** refuses the same thing from **any** `IExperimentalDesign` and returns a failure result, as its other checks do. `IExperimentalDesign` is the seam outside code implements; MetaMorpheus's `TmtMzLibExperimentalDesign` never goes through `SampleExperimentalDesign`.
  - It asks the rule of **every file's samples at once**, not one design key at a time. `CombinePeptideMatrices` merges all files' columns into one matrix, so a channel listed under two keys collides there just like a channel listed twice under one key. The engine reports that case as `channel 126 of 'shared.raw' is listed 2 times, as 'Pt1' and 'Pt3'`.
  - Null dictionaries, arrays and entries are skipped rather than thrown on.
- The rule applies to label-free samples too: `sample 'run7.raw' is listed 2 times`.
- `Add` applies it after its file-name and already-added checks, so an input that fails one of those keeps its old message.

**A bridge design is not a repeat.** `PXD008841` puts a sample named `pool` in 131N of every plex. Those channels are in different files, so they are different samples, and the rule says so.

## How the labels behave on SDRF

Measured on those same 59 channel-level documents:
- **Empty or reserved `source name`:** none, so every SDRF channel would get a named label.
- **Label collisions:** none, beyond the duplicate rows above.
- **Bridge design:** `PXD008841`'s 360 `pool` rows yield 360 distinct labels, because the file is in the label.
- **File extensions:** the file part is the stem, so a label is the same whether the design names the `.raw` or the search ran on the `.mzML`.
- **Cosmetic cost:** 9 documents already embed the channel in the source name (`230120_1_TMT126` → `230120_1_TMT126_230120_1_126`). Median label length is 57 characters, maximum 122.
- **Not parseable:** 23 documents put `_` in source names, so a label cannot be split back into its parts. Nothing parses it; an SDRF written alongside is the machine-readable key.
- **Not unique to its parts, for the same reason.** `S_1` in `run.raw` and `S` in `1_run.raw` both give `S_1_run_126`, and a channel named `plex1` in `x.raw` takes the label of the same unnamed channel of `plex1_x.raw`. Such labels are handled as collisions, see *Left alone*. Values stay on their own channels, because columns are keyed by path and channel, never by label. The corpus has no such collision, and `ForSample`'s remarks document it.

## Same on master, not addressed here

- **Same-named files in two directories lose intensities.** The design is keyed by file name, so matches from `C:\A\file1.raw` and `C:\B\file1.raw` share one entry, the combined matrix gets duplicate column keys, and one file's values are overwritten while the run reports success. `FromSamples` and MetaMorpheus refuse such a design, so only hand-built designs reach it. The fix belongs in `PivotByFile`.
- **Collapsed labels drop dotted conditions.** A collapsed column's synthetic path `{Condition}_Bio…_Collapsed` goes through `GetFileNameWithoutExtension`, so condition `0.125` becomes `0_127C`. `IsobaricQuantSampleInfo.ToString()` did the same on master.
- **Collapse can merge different channels.** Collapse groups on condition and the replicate dimensions it keeps, not on channel. That question belongs with R2 and #994.
- **A tab, CR or LF in a name breaks the header.** This applies to `SampleName`, `Condition` and `ChannelLabel` alike. Neither MetaMorpheus's reader nor SDRF (both TSV) can produce it. The fix belongs in the header writers, for every field.
## Conflicts with #1286 and #1287

Both touch `SampleGroupBuilder.cs` and `SampleGroupLabels.cs`, and both predate #1296, so this adds one more change for their rebase to carry: the isobaric and label-free label lines become calls to `ForSample`.

## Tests

Full mzLib suite, `Category!=ExternalService`: **6047 passed, 0 failed, 31 skipped**.

https://claude.ai/code/session_013YV5xjuyAbxvzeEe9EDAVZ


Alexander-Sol:
## Automated code review

**No substantive correctness defects.** A clean, well-tested, genuinely additive refactor. Verified the load-bearing invariants:
- **Identity correctness:** `IsobaricQuantSampleInfo` and `SpectraFileInfo` both override `Equals`/`GetHashCode` on file+channel and file+replicate-dims respectively (excluding `SampleName`), so `DescribeRepeatedSample`'s `GroupBy(sample => sample)` uses the intended semantics. Different fractions/plexes of one sample are not equal (different file path), so the engine's whole-design check produces no false positives on valid TMT/bridge designs. `SampleName` is correctly kept out of equality, and the new `Add`/engine guards close the resulting column-merge hole.
- **Additivity:** for unnamed isobaric channels `ForSample` returns `{file}_{channel}` exactly as `ToString()` did; the label-free branch's `{Condition}_{BiologicalReplicate+1}` is provably equal to the old `{conditionGroup.Key}_{bioRepGroup.Key+1}`. `QuantificationWriter.SampleColumnLabel` reproduces the old file-name behavior for label-free.
- **No hidden coupling to the old isobaric label string:** grouped-table column disambiguation keys on `SampleGroupResult.Identity` (still `{file}|{channel}`, unchanged); `Label`/`FilesInGroup`-key/`IntensitiesBySample`-key changes are display/handle-only and never parsed back. `CollapsedSampleInfo(members)` takes `members[0]` for every field but the name.

**Low — `SampleExperimentalDesign.cs` (`DescribeRepeatedSample`) via the engine path.** The order of names in the message (`... as 'Pt1' and 'Pt3'`) derives from `FileNameSampleInfoDictionary.Values` enumeration order, which `Dictionary<>` doesn't contractually guarantee. `Run_WithTwoDesignKeysHoldingOneChannel` asserts an exact ordered string, so that test is mildly fragile (works in practice because `Dictionary` preserves insertion order absent removals). Within `Add`, order is the caller's array order and deterministic.

By design (acknowledged in the description, not defects): colliding labels can get different headers in the two output files; labels containing `_` aren't round-trip-parseable.

Net: mergeable as-is; the only optional item is hardening that one order-dependent assertion. (The red `integration` CI check is unrelated — a MetaMorpheus RNA-digestion test `KeyNotFoundException: 'RNase PhyM'`, not this change.)

trishorts:
@Alexander-Sol thanks for the review. I took the Low at its source in `202f8102`, not just in the test: `DescribeRepeatedSample` now sorts the repeated channel's names ordinally, so the message no longer depends on the order a `Dictionary`'s values enumerate in. `Run_WithTwoDesignKeysHoldingOneChannel` keeps its exact assertion, which now holds by contract rather than by insertion order. A new test, `DescribeRepeatedSample_ListsTheNamesInOrdinalOrder_NotArrivalOrder`, feeds the names in reverse. It went red with the sort removed (`'Pt3' and 'Pt1'`) and green with it back. I added it to the table in the description.

On the two by-design items (headers that can differ between the two files when labels collide, and `_` in labels), I agree. Both are listed in the description and are out of scope here.

Alexander-Sol (approved):
Approved. See the detailed review comment above.

nbollis (approved):

PR #1301, opened 2026-09-15
RNase: Update cleave specificity and enable digestion termini to be tracked

### LINKED ISSUE #1302: DigestionMotif.Fits ignores a preventing residue before the cut ([X]|Y motifs)
`DigestionMotif.Fits` never applies a preventing residue that sits **before** the cut when the motif's `CutIndex == 0` (syntax `[X]|Y`).

**Where:** `mzLib/Omics/Digestion/DigestionMotif.cs:148`

```csharp
currentResidue = CutIndex != 0 ? sequence[location + m + n] : sequence[location - PreventingCleavage.Length + 1 + n];
```

`DigestionAgent` calls `motif.Fits(sequence, r)` with `r` = the index of the first inducing residue. For `[G]|U`, the inducing residue is the `U`, so the `CutIndex == 0` branch reads `sequence[r - 1 + 1 + 0] = sequence[r]`, which is the `U` itself, not the residue before it. The prevention check compares `G` against `U`, never matches, and the cut is always made. The bounds check on line 142 (`location - PreventingCleavage.Length + 1 + n < 0`) has the same `+ 1`.

**Repro:** found while reviewing #1301, which changes `RNase_MC1` to `[G]|U` (cleave 5′ of U, except GpU). The PR's own test `TestRnaseMC1_CleaveBeforeUridine_CurrentBehavior` pins the result: `AAGUAU` digests to `AAG, UA, U`, i.e. the GpU bond is cut.

**Likely fix:** drop the `+ 1` in both places (`location - PreventingCleavage.Length + n`), then flip that test to expect `AAGUA, U`. No shipped protease uses a before-the-cut prevention (`proteases.tsv` only uses `X[P]|`), so today only the RNase row is affected.

nbollis:
This pull request introduces several improvements and fixes related to RNA digestion logic, particularly focusing on accurate assignment and propagation of oligo termini, enhanced test coverage for digestion patterns (including non-specific and Cusativin digestion), and improved code clarity. The most important changes are summarized below.

I left in a PDF which has source information for any interested future developer. 

### Core Logic and API Improvements

* Refactored the `NucleolyticOligo` and `OligoWithSetMods` classes to expose `FivePrimeTerminus` and `ThreePrimeTerminus` as public properties (instead of protected fields), ensuring correct propagation and comparison of termini throughout the digestion pipeline. [[1]](diffhunk://#diff-9482b40091d2b3e83ca5512d4df969dbf9aa3dc3ac7b95c636d24b1d8d67097aL12-R21) [[2]](diffhunk://#diff-94d2742c3c1f7daa1e3cbe94051cfbcf7c17a8c27f1a34ca97e52a6ad7fe1a01L73-R90) [[3]](diffhunk://#diff-94d2742c3c1f7daa1e3cbe94051cfbcf7c17a8c27f1a34ca97e52a6ad7fe1a01L267-R268)
* Updated the `Rnase` class and all digestion routines to support multiple possible terminus remainders, allowing for more flexible and accurate modeling of digestion products. This includes updating the constructor and all relevant digestion methods to handle lists of possible termini. [[1]](diffhunk://#diff-63566148dddd78cae9f5cf91dde5996a337b12703c2cd45bb4112813944cfa17L11-R21) [[2]](diffhunk://#diff-63566148dddd78cae9f5cf91dde5996a337b12703c2cd45bb4112813944cfa17L67-R79) [[3]](diffhunk://#diff-63566148dddd78cae9f5cf91dde5996a337b12703c2cd45bb4112813944cfa17L96-R109) [[4]](diffhunk://#diff-63566148dddd78cae9f5cf91dde5996a337b12703c2cd45bb4112813944cfa17L117-R133) [[5]](diffhunk://#diff-63566148dddd78cae9f5cf91dde5996a337b12703c2cd45bb4112813944cfa17L140-R165) [[6]](diffhunk://#diff-63566148dddd78cae9f5cf91dde5996a337b12703c2cd45bb4112813944cfa17R202)

### Testing Enhancements

* Added comprehensive tests for non-specific (`X|`) digestion, verifying correct fragment generation, handling of missed cleavages, minimum length filtering, and correct assignment of termini for all fragments.
* Enhanced Cusativin and RNase 4 test coverage, including new tests for complex cleavage patterns, motif-specific prevention of cleavage, and validation that fragments with the same base sequence have consistent termini. [[1]](diffhunk://#diff-26b44a06a32c2a40827a8e77bb15b2bb5bbca21e4754db31f1e9282481e1b60bL45-R102) [[2]](diffhunk://#diff-26b44a06a32c2a40827a8e77bb15b2bb5bbca21e4754db31f1e9282481e1b60bL33-R43)
* Removed a redundant Cusativin test case from the test suite to avoid duplication and ensure test clarity.


mzLib/Transcriptomics/Digestion/rnases.tsv
```diff
@@ -77,21 +77,36 @@
...
-RNase PhyM	A|,U|	full
-RNase 4	U|A,U|G	full
-RNase U2	G|,A|	full
-Cusativin	C|	full
-RNase_MC1	|U	full
-colicin_E5	G|U	full
-singleN		SingleN
-singleC		SingleC
+RNase 1	G|,C|,A|,U|	full	H2O4P	O-3P-1	Mammalian RNase 1, slight pyrimidine preference
+RNase PhyM (>= 7M urea)	A|,U|	full	O3P		Up-N/Ap-N; almost exclusive and uniform under 7 M urea
```
trishorts, line 102: **Blocking.** Removing the `RNase PhyM` key breaks MetaMorpheus:

- The `integration` job fails on MM `Test/ParameterTest.cs:337` (`RnaseDictionary.Dictionary["RNase PhyM"]` → `KeyNotFoundException`).
- More importantly for users, MM resolves a saved digestion agent by name (`TaskLayer/MetaMorpheusTask.cs:94`, `.FromToml(tmlString => RnaseDictionary.Dictionary[tmlString.Value])`). Any task `.toml` saved with PhyM would stop loading. `FileSpecificParameters.cs:48` does the same `ContainsKey` lookup.

The cheapest fix is to keep this row's name as `RNase PhyM`. Its motif `A|,U|` is exactly the old definition, so saved files keep their meaning. Then add only the no-urea `A|,U|,G|` variant under a new name. If you'd rather keep both condition-specific names, it needs a paired MM PR that merges first, plus some alias for old tomls.

mzLib/Transcriptomics/Digestion/rnases.tsv
```diff
@@ -77,21 +77,36 @@
...
+RNase PhyM (< 7M urea)	A|,U|,G|	full	O3P		Up-N/Ap-N/Gp-N
+RNase 4	U|A,U|G	full	H2O4P,O3P	O-3P-1	U-A slightly preferred over U-G; ssRNA; mixed 3' products
+RNase U2	G|,A|	full	H2O4P	O-3P-1	Purines, A preferred over G
+Cusativin	C[C]|,U|A,X|U	full	H2O4P,O3P	O-3P-1	CpA 70-90%; CpG/CpU 80-100%; CpC resistant; mixed 3' products
```
trishorts, line 106: `X|U` cuts before *every* U, so it includes GpU. Your own `TestCusativin_CleavagePattern` shows it: `GUACUG` splits `G|U` at position 1. The validation PDF's Cusativin evidence lists CpA/CpG/CpU, UpA 70-90%, and UpU/ApU 14-30%, but no GpU.

If GpU isn't meant to be in there, `C[C]|,U|A,A|U,U|U` encodes exactly the reported sites. The test would then expect `GU, AC, UG`. If GpU is supported by the 2022 paper, a note in the Notes column would help.

mzLib/Transcriptomics/Digestion/rnases.tsv
```diff
@@ -77,21 +77,36 @@
...
+RNase_MC1	[G]|U	full	H2O4P,O3P	O-3P-1	ApU 70-100%; CpU 80-100%; UpU 45-60%; GpU 1.0-1.5%; mixed 3' products
```
trishorts, line 107: As your test comment already says, the `[G]` here is a no-op. For `CutIndex == 0`, `DigestionMotif.Fits` (`Omics/Digestion/DigestionMotif.cs:148`) reads `sequence[location]`, the U itself, rather than the preceding residue. So this row behaves identically to the old `|U` and still cuts GpU. It's not a regression, so it's not blocking. I filed the underlying bug as #1302; the fix is dropping the `+ 1` on lines 142 and 148.

Either fix it here (it's two characters, and `proteases.tsv` has no `[X]|Y` motifs that could shift) and flip `TestRnaseMC1_CleaveBeforeUridine_CurrentBehavior` to expect `AAGUA, U`, or leave it and point the test comment at #1302. That way the test pinning the wrong behavior gets revisited when the bug is fixed.

mzLib/Transcriptomics/Digestion/Rnase.cs
```diff
@@ -148,45 +159,46 @@ private IEnumerable<NucleolyticOligo> FullDigestion(NucleicAcid nucleicAcid, int
...
-                        cleavageIndexWithinTruncation++;
-                    }
 
-                    bool endPeptide = cleavageIndexWithinTruncation - missedCleavages - 1 >= 0 //if we're not going to go out of bounds (-1 to get in front of the end)
-                                      && oneBasedIndicesToCleaveAfter[cleavageIndexWithinTruncation - missedCleavages - 1] + 1 >= truncation.OneBasedBeginPosition //and it's not before the beginning
-                                      && truncation.OneBasedEndPosition.HasValue //and the proteolytic peptide even has an end
-                                      && !oneBasedIndicesToCleaveAfter.Contains(truncation.OneBasedEndPosition.Value) //and we haven't already cleaved here
-                                      && ValidLength(truncation.OneBasedEndPosition.Value - oneBasedIndicesToCleaveAfter[cleavageIndexWithinTruncation - missedCleavages - 1] + 1 - 1, minLength, maxLength); //and it's the correct size
-                    if (endPeptide)
+                    foreach (var (threePrimeTerminus, fivePrimeTerminus) in GetDigestedTermini(truncation.OneBasedBeginPosition,
```
trishorts: Pre-existing, but it's exactly the termini-propagation problem this PR is about, and the multiplicity now repeats it. The termini here are computed from the *truncation's* begin/end. But the `" start"` product ends at a cleavage site, and the `" end"` product begins after one.

I verified with a throwaway probe on this branch: `RNA("GUACUGAGUC")` with `TruncationProduct(3, 10)` under RNase T1 yields `3-6 ACUG 'trunc start'` with 3' = `HO` (the parent's original terminus). It should be `H2O4P`, so the product's mass is off by ~80 Da. The mirror case happens when a truncation starts at residue 1: the `" end"` product gets the original 5'.

Suggest calling `GetDigestedTermini` separately with each product's actual `(start, end)` inside the `if (startPeptide)` / `if (endPeptide)` blocks. That also hoists the two `while` loops out of the `foreach`, so they no longer re-run for each terminus combination. Happy for this to be a follow-up if you'd rather keep the PR tight.

mzLib/Test/Transcriptomics/TestRnase.cs
```diff
@@ -42,12 +49,205 @@ public void TestRnase4_DoesNotCleaveAfterUridineBeforePyrimidinesOrUridine()
...
+        {
+            var cusativin = RnaseDictionary.Dictionary["Cusativin"];
+
+            var products = cusativin.GetUnmodifiedOligos(new RNA("GCCUGA"), 0, 1, int.MaxValue)
+                .ToArray();
+
+            // C at pos 2 is followed by C at pos 3 → no cleavage there
+            // X|U fires after C (pos 3) → pos 3→4, C→U
+            // U|A is absent (no U followed by A)
+            // C[C]| fires after C at pos 3 is NOT followed by C - wait, pos3=C, pos4=U → C[C]| fires after pos 3
```
trishorts, line 98: nit: these three comment lines read as working notes ("wait, ... Actually ..."). Suggest replacing them with one line: `// C2 is followed by C3 → C[C]| prevented; X|U cuts C3|U4 → GCC, UGA`. The assertion could also be the exact `Is.EqualTo(new[] { "GCC", "UGA" })` rather than `Does.Not.Contain("GC")`, which is a weaker check.

mzLib/Test/Transcriptomics/TestRnase.cs
```diff
@@ -30,10 +30,17 @@ public void TestRnase4_CleavesAfterUridineOnlyBeforePurines()
...
 
-            Assert.That(products, Is.EqualTo(new[] { "AU", "GCU", "GA" }));
+            var distinctBaseSequences = products.Select(p => p.BaseSequence).Distinct().ToArray();
+            Assert.That(distinctBaseSequences, Is.EqualTo(new[] { "AU", "GCU", "GA" }));
+
+            var grouped = products.GroupBy(p => p.BaseSequence).ToDictionary(g => g.Key, g => g.Select(o => (o.ThreePrimeTerminus, o.FivePrimeTerminus)));
+            foreach (var group in grouped)
+            {
+                var distinctBaseSeqs = group.Value.Select(v => v).Distinct().ToArray();
+                Assert.That(distinctBaseSeqs.Length, Is.EqualTo(group.Value.Count()), $"Base sequence {group.Key} has multiple distinct termini: {string.Join(", ", distinctBaseSeqs.Select(v => $"({v.ThreePrimeTerminus}, {v.FivePrimeTerminus})"))}");
```
trishorts, line 42: nit: this checks that there are no *duplicate* `(3', 5')` pairs per base sequence, but the failure message says "has multiple distinct termini", which is the opposite. The PR description also calls it "consistent termini". Suggest "has duplicate terminus combinations", as in the MC1 test below. `Select(v => v)` is also a no-op. Same on line 61.

trishorts (commented):
Nice work. Letting the termini be list-valued is the right shape, and the tests are much stronger than what was there. I built the branch and all 449 `Transcriptomics` tests pass locally.

**One blocker:** renaming `RNase PhyM` breaks MetaMorpheus. That's why the `integration` check is red: MM's `TestFileSpecifcParameterOverwrite_WithRnaDigestionParams` throws `KeyNotFoundException: 'RNase PhyM'`. Details inline.

Everything else is non-blocking:
- Cusativin's `X|U` also cuts GpU.
- The MC1 `[G]|U` prevention never fires. It's a pre-existing `DigestionMotif` bug, filed as #1302.
- Truncation start/end products get the wrong termini. Also pre-existing, but it's the same area this PR is fixing.

On the PDF: I'd either update it or move it out of `Transcriptomics/Digestion/`. As committed, it describes an earlier state of this branch. It calls the table `rnases2.tsv` and says the parser doesn't read the terminus columns yet (E004). Its Part II lists Cusativin as `C|` and MC1 as `|U`, and its PhyM rows are named `(7 M urea)`/`(no urea)`. Its references point at `E:/agentplayground/...`, and it says human source verification is still required before sign-off. A future developer comparing it against `rnases.tsv` will find it disagrees on the rows it's meant to justify.

Small one: the `rnases.tsv` header says `blank = no enzymatic product or terminus not established`, but `ParseTerminusFormulas` turns a blank cell into the `O-3P-1`/`H2O4P` default. So the PhyM rows do get a 5'-OH. Maybe say "blank = Rnase default" so nobody reads it as "no terminus".

mzLib/Transcriptomics/Digestion/rnases.tsv
nbollis: I can fix this by changing the enzyme name in MetaMorpheus once we release mzlib

nbollis: Agreed that this is a cross-repo breaking change. We're keeping the new `(>= 7M urea)` / `(< 7M urea)` names here and will update MetaMorpheus to match once this mzLib release is out. The existing thread reply already notes that plan.

mzLib/Transcriptomics/Digestion/rnases.tsv
nbollis: You're right that `X|U` cuts before every U. That is intentional for this source representation — `X|U` encodes the UpU/ApU/CpU activity alongside `C[C]|` and `U|A`. I'm updating the validation PDF source to match; no code/test change needed here.

mzLib/Transcriptomics/Digestion/Rnase.cs
nbollis: Acknowledged. This is a pre-existing termini-propagation issue and is orthogonal to the RNase additions in this PR. We'd like to address it in a follow-up rather than expand this PR's scope.

mzLib/Transcriptomics/Digestion/rnases.tsv
nbollis: Fixed. Dropped the `+1` offset in `DigestionMotif.Fits` for `CutIndex == 0` preventing motifs (issue #1302). `RNase_MC1` now skips GU positions as intended, and the MC1 test expects `{AAGUA, U}`. Also updated `TestSeqCoverage.ReadCustomFile` for the corrected `|D[D]` behavior.

mzLib/Test/Transcriptomics/TestRnase.cs
nbollis: Cleaned up the working-notes comment and strengthened the `DoesNotCleaveBeforeCytidine` assertion to exact equality. (The Cusativin motif stayed as `X|U`, so the exact expected set is unchanged.)

mzLib/Test/Transcriptomics/TestRnase.cs
nbollis: resolved.

trishorts (approved):

Alexander-Sol (approved):

trishorts:
Heads-up on two consequences of renaming `RNase PhyM` to `RNase PhyM (>= 7M urea)`.

**1. mzLib CI is red on master and on every open PR.** MetaMorpheus's `ParameterTest.cs:337` looks the old name up by exact key (`RnaseDictionary.Dictionary["RNase PhyM"]`), so it throws `KeyNotFoundException` against mzLib master, and the `integration` job runs MetaMorpheus's tests against the mzLib under review. smith-chem-wisc/MetaMorpheus#2818 fixes the test by using `RNase T1`, whose name is stable in both the released package and master, which unblocks it without waiting for a release.

**2. Saved RNA task settings naming `RNase PhyM` no longer load.** MetaMorpheus resolves an RNase from a settings file through the same indexer, so those files now fail with the same exception. This is the protease story again: #1005 removed `semi-trypsin`, nothing in mzLib's CI noticed, and MetaMorpheus settings files (including the O-Pair Search paper's) stopped loading. MetaMorpheus#2812 is fixing that side now.

If it seems worth doing, the mzLib-side equivalents would be:
- resolve the old name to the row that replaced it (`RNase PhyM (>= 7M urea)` keeps the old `A|,U|` motifs), so old settings keep working;
- an RNase version of the name-pin test #1304 just added for proteases (`EmbeddedProteases_NoShippedProteaseNameIsRemovedOrRenamed`), which fails when a shipped name disappears and says what to do if the removal is deliberate.

Happy to open that if you would like, or leave it with you.

PR #1303, opened 2026-09-15
Return real semi-specific peptides for SearchModeType Semi, and fix the semi-specific digestion edge cases

trishorts:
## What changes

**Semi-specific digestion now returns the right peptides, however it is requested.** Searches that use it will change their answers: they get more semi-specific peptides, and correct missed-cleavage counts.

- **`SearchModeType = Semi` with `FragmentationTerminus.Both` (the default) returned the wrong peptides, silently.**
  - **What it returned:** only the C-terminal "seed" peptides that MetaMorpheus's non-specific engine trims after the search. Every branch of `SpeedySemiSpecificDigestion` tests `== N` and treats everything else as C.
  - **Example:** AMBP (2 missed cleavages, length ≥ 7) gave **38 peptides instead of its 1,726** semi-tryptic peptides.
  - **Who it hit:** Classic, Modern, Glyco and crosslink searches configured this way lost most semi-specific identifications without any error. After #1005 removed `semi-trypsin` (the protease the O-Pair Search paper's settings name for StcE + trypsin glycoproteomics), this was the only way left to ask for a semi-tryptic glyco search. In a real O-glyco search it dropped glyco PSMs from 348 to 11.
  - **Now:** `Protein.Digest` returns the semi-specific peptides for any terminus other than N or C.
- **A protease whose own specificity is Semi (`Protease.SemiProteolyticDigestion`) had three defects:**
  - For a protein that does not start with Met, residue 1 was still treated as a removable initiator Met, so semi peptides starting at residue 2 were dropped.
  - When a protein has no more internal cleavage sites than `MaxMissedCleavages`, every peptide starting after a removable Met was lost, and with `InitiatorMethionineBehavior.Cleave` the Met-retaining peptides were emitted anyway.
  - **Missed cleavages were reported as a residue distance, effectively the peptide length.** The fully tryptic `CCKDDK` reported 6, where fully specific digestion reports 1. That is not this protease's count, and not the non-specific convention of length − 1 either. MetaMorpheus feeds the value to PEP as `MissedCleavagesCount` and writes it to glyco output. Missed cleavages now come from the protease's own sites (`Protease.CountCleavageSitesBefore`), so a peptide reports the same count whichever digestion produced it.
- **The speedy seeds (`SpeedySemiSpecificDigestion`, Semi with terminus N or C) had two defects:**
  - Under `InitiatorMethionineBehavior.Cleave`, the C seeds near the protein start kept the Met (e.g. AMBP `[1-39]`), on ordinary proteins.
  - For proteins with few cleavage sites, no N seed started after a removable Met.

## What does not change

- No public signature changes. `ProteinDigestion` gains `SemiSpecificDigestion` and `WantsSemiSpecificSeeds`.
- Fully specific digestion is unchanged.
- Seed requests (Semi with terminus N or C) still return seeds, which MetaMorpheus's non-specific engine depends on. Their counts for Variable and Retain on ordinary proteins are unchanged, including the pinned 7/6 and 7/9.
- Peptide order for ordinary proteins is unchanged.

## How it is tested (test-first)

New `Test/ProteomicsTests/ProteolyticDigestion/SemiSpecificDigestionTests.cs`, 115 cases. **49 failed on master before any fix.**

- **The answer key.** A brute-force enumeration written from the definition of semi-specific digestion, sharing no code with mzLib's digestion. It is itself checked against fully specific digestion (all combinations) and a hand-worked example.
- **The matrix.** 12 proteins chosen for their edge cases (no initiator Met, adjacent sites, fewer sites than missed cleavages, no sites, real AMBP and PSGL-1 mucin sequences) × missed cleavages 0–3 × min length 1/7 × max length 25/unlimited × every initiator-Met behavior.
- **The Semi protease** matches the reference in peptides, Full/Semi labels and missed cleavages.
- **Seeds:** every semi peptide is reachable by trimming a seed from its specific terminus, and every seed is a legal peptide.
- **Semi + Both** matches the reference. The default terminus enumerates. The embedded trypsin gives 1,726 on AMBP. Variable oxidation matches the Semi protease. `keepOGlycopeptide` on a mucin keeps every Ser/Thr semi peptide.
- **Truncation products** only add peptides on their own boundary.

The fixture registers its own no-proline-rule trypsin, so it is unaffected by #1186's rename.

Three commits, each with its own tests: the Semi protease, the seeds, then the routing.

**Results:**
- Full mzLib suite: 6,153 passed, 0 failed.
- MetaMorpheus master built against a packed copy of this branch: 279 of 279 in its semi, non-specific, digestion, protease, TOML, glyco and search tests.

## Related

- **Stacked on this:** #1304 (draft) adds `DigestionParams.SpecificTermini`, one user-facing setting for fully specific, semi-specific (either terminus, or a required N or C terminus) and non-specific digestion, with `semi-trypsin` loading again as trypsin + semi-specific.
- **Downstream:** smith-chem-wisc/MetaMorpheus#2812 (draft) exposes it in every task window and settings file.
- **May conflict textually:** #1119 and #1290 (edit `ProteinDigestion.Digestion` next to the new method) and #1038 (makes `Protein.Digest` virtual).

https://claude.ai/code/session_01V1qU8gPh15GvhNYqTBdBxs


RayMSMS (dismissed):

nbollis (commented):
## Review summary

One approved major inline finding.

mzLib/Proteomics/ProteolyticDigestion/ProteinDigestion.cs
```diff
@@ -117,27 +139,51 @@ public IEnumerable<ProteolyticPeptide> SpeedySemiSpecificDigestion(Protein prote
                 int endIndex = nTerminusFragmentation ?
                     oneBasedIndicesToCleaveAfter[lastIndex] :
                     oneBasedIndicesToCleaveAfter[i];
+                bool startsAtProteinNTerminus = startIndex == 0;
 
-                int peptideLength = endIndex - startIndex;
-                if (peptideLength >= MinPeptideLength)
+                if (nTerminusFragmentation)
```
nbollis, line 146: [review-generic] C-terminal seed wrap-up retains a mandatory-cleaved initiator Met when Met is also a protease cleavage site

For a protein beginning with `M` and `InitiatorMethionineBehavior.Cleave`, `metMustBeRemoved` is true. In the C-seed path, line 157 advances past `M` only when residue 1 is not also a protease cleavage site. A protease that cleaves after `M` can therefore emit a C seed starting at residue 1; post-search trimming can then produce peptides retaining an initiator Met that digestion semantics require to be absent.

Suggestion: Advance past residue 1 whenever `startsAtProteinNTerminus && metMustBeRemoved`, regardless of `residueOneIsCleavageSite`. Add a C-seed regression using an M-cleaving protease and `InitiatorMethionineBehavior.Cleave`.

trishorts: Good catch, thank you. Confirmed and fixed in b7ceba18. Whether residue 1 is a cleavage site now only decides the extra residue-2 N seed; a C seed always starts after a Met that must be removed. Added a regression test that reruns the seed coverage and legality checks with a protease that cleaves after K, R or M on every test protein starting with Met. It failed on all 7 before the fix (C seeds such as [1-1], [1-2], [1-39] under Cleave) and passes now, along with the other 538 digestion tests.

nbollis (approved):

RayMSMS (approved):

PR #1304, opened 2026-09-15
Fix missing semi-specific peptides when the initiator Met is a cleavage site, and pin and document what SearchModeType and FragmentationTerminus return

trishorts:
Reworked after review (thanks @nbollis). The `SpecificTermini` enum, `ProteaseDictionary.TryResolveRemovedProtease`, the new non-specific digestion and the `Protein.Digest` rewiring are all gone. With #1303 merged, trypsin + `SearchModeType` Semi already returns real semi-specific peptides through the existing path, so this PR keeps only the bug it found and adds tests and docs for the settings that caused the confusion. Rebased onto current master (#1249).

## 1. Fix: missing semi-specific peptides when the removed initiator Met is a cleavage site

- **Symptom:** with a protease that cleaves after M (CNBr), `InitiatorMethionineBehavior.Cleave` and a protein starting with Met, semi-specific digestion **dropped** the peptides that end at the first window's C-terminal site with a ragged N-terminus. For `MPEPTIDEPEPTIDE` with one missed cleavage, 13 of 14 were missing. No error; the peptides were just gone.
- **Affected:** both ways of asking for semi-specific peptides: a Semi protease, and `SearchModeType` Semi with `FragmentationTerminus` Both. The seed path (N or C) was already right, and #1303 tests it for this case.
- **Cause:** in `Protease.SemiProteolyticDigestion`, the first window turns off its Met-cleaved start (residue 2 starts come from the next window when residue 1 is a site) and, under Cleave, its Met-retaining start, so it made nothing at all. Predates #1303.
- **Fix:** the first window adds exactly its ragged-start peptides in that case. Nothing is duplicated.
- **Test:** `SemiDigestion_ProteaseThatCleavesAfterTheInitiatorMet_ReturnsExactlyTheReferencePeptides`, both ways of asking × every Met-starting test protein × the settings matrix, against the brute-force reference (set, labels, missed cleavages). **Failed 12 of 14 on master.**

## 2. Tests that pin what each `SearchModeType` × `FragmentationTerminus` returns

New `SearchModeTypeDigestionTests.cs`. Its class remarks hold the whole table in one place:

| SearchModeType | FragmentationTerminus | `Protein.Digest` returns |
|---|---|---|
| Full | any | fully specific peptides |
| Semi | Both (default) | semi-specific peptides |
| Semi | N or C | seeds for the non-specific engine to trim, not peptides |
| None | N or C | singleN / singleC seeds |
| None | Both | singleC seeds, same as None + C |

- Each row checks **which method answers** it (`Digestion`, `SemiSpecificDigestion`, `SpeedySemiSpecificDigestion`, or the singleN/singleC protease) and **what the result means** (labels, specific termini, seeds fewer than peptides).
- Full + N/C and None + Both were untested. None + Both is pinned as it is today so any change to it is deliberate.
- Mutation check: restoring the #1303 routing bug fails the Semi rows.

## 3. A test that fails when a shipped protease name is removed or renamed

`EmbeddedProteases_NoShippedProteaseNameIsRemovedOrRenamed` reads the embedded `proteases.tsv` (not the dictionary, which other tests add to) and requires every currently shipped name to still be there. #1005 removed `semi-trypsin` with green CI, and MetaMorpheus settings files naming it stopped loading. The failure message says what to do if a removal is intended. Adding a protease needs no change. Mutation check: deleting the `CNBr_old` row fails it by name.

**Conflict note:** #1186 renames the `|P` proteases, so whichever of the two lands second updates `ShippedProteaseNames` (and #1186 already maps the old names).

## 4. Docs

- `CleavageSpecificity`: per-value docs, and the three roles the enum plays (an agent's specificity, the search mode, a product's FDR label).
- `DigestionParams.SearchModeType` / `FragmentationTerminus` / `SpecificProtease` / `Protease`: the table above, peptides vs seeds, and why only an engine that trims by precursor mass can use seeds. These replace the trailing `//` comments.
- `IDigestionParams`: short pointers to the same.
- `ProteinDigestion.SpeedySemiSpecificDigestion`: why a glyco search cannot use seeds. Also replaces an unresolvable `Protein.Digest` cref, which removes the CS1574/CS1580 warnings it raised.

No public API changes.

## Tests

- Semi-specific digestion fixture: 136 passed.
- New matrix + name tests: 13 passed.
- Full suite (`required.runsettings`): 6,257 passed, 0 failed, 32 skipped.

## Related

- **Consumer:** smith-chem-wisc/MetaMorpheus#2812 will be reworked to match: map `semi-trypsin` in settings files to trypsin + `SearchModeType` Semi when loading, and reject `SearchModeType` None outside the non-specific search.


RayMSMS:
Protease.GetSemiSpecificUnmodifiedPeptides and ProteinDigestion.SemiSpecificDigestion(Protein) look like dead code after this commit.

  - `GetSemiSpecificUnmodifiedPeptides (Protease.cs)` is a thin wrapper that just forwards to SemiProteolyticDigestion.
  - `ProteinDigestion.SemiSpecificDigestion `(ProteinDigestion.cs#L291) exists only to call that wrapper.
  - Before this commit, `Protein.Digest` called` ProteinDigestion.SemiSpecificDigestion` when `SearchModeType == Semi.`
  - This commit rewires `Protein.Digest` to call `SpecificTerminiDigestion instead`, which calls `Protease.GetPeptidesWithSpecificTermini `directly — and for S`pecificTermini.Either`, that method
    also calls `SemiProteolyticDigestion` directly, bypassing the wrapper entirely.

  I grepped the PR-head branch and found no remaining callers of either method, in production code or in `SemiSpecificDigestionTests.cs `/ `SpecificTerminiDigestionTests.cs`.

  Since the PR description already says this branch gets rebased to a single commit once #1303 merges — could `GetSemiSpecificUnmodifiedPeptides` and `ProteinDigestion.SemiSpecificDigestion`
  be dropped as part of that rebase, so they don't ship as unused code?

trishorts:
Good catch, agreed. Removed in 69a4b4a0 (folded into the SpecificTermini commit during the rebase onto master): `Protease.GetSemiSpecificUnmodifiedPeptides`, `ProteinDigestion.SemiSpecificDigestion`, and also `ProteinDigestion.WantsSemiSpecificSeeds`, which `IsSeedRequest` replaced and which likewise had no callers left. All three came in with #1303 and are in no release, so nothing external breaks; MetaMorpheus and ProteaseGuru don't call them either. The two doc comments that pointed at them now point at `SpecificTerminiDigestion` and `IsSeedRequest`. Digestion, SpecificTermini and Koina tests: 1,192 passed.

RayMSMS (dismissed):

mzLib/Omics/Digestion/SpecificTermini.cs
nbollis, line 1: We already define an Enum for this, please reuse that. Massive summary comment says this is all about digestion specificity, so the CleavageSpecificity enum should be reused.

mzLib/Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
```diff
@@ -450,6 +450,60 @@ public static bool TryGetProtease(string name, out Protease protease)
...
+        /// Proteases mzLib used to ship and has since removed, with the cleavage motifs and the specificity each name stood
+        /// for. Each is resolved to the currently shipped FULLY specific protease with exactly those motifs, plus that
+        /// specificity, so settings that name them keep working.
+        /// </summary>
+        /// <remarks>
+        /// <c>semi-trypsin</c> ("K|,R|", semi) was removed in #1005 (February 2026). It was the protease the O-Pair Search
+        /// paper's MetaMorpheus settings name for StcE + trypsin glycoproteomics. The replacement is found by motif, not by
+        /// name, because the name of the no-proline-rule trypsin is itself being changed (#1186).
+        /// </remarks>
+        private static readonly (string Name, string Motifs, SpecificTermini SpecificTermini)[] RemovedProteases =
```
nbollis: if we need semi-trypsin to do this, then it should be added as a protease otherwise metamorpheus won't be able to run. 

It seems like this new structure for parsing deleted proteases is only needed if the protease was deleted, adding it back should fix that and eliminate the need for this.

mzLib/Proteomics/Protein/Protein.cs
```diff
@@ -379,17 +379,13 @@ public IEnumerable<IBioPolymerWithSetMods> Digest(IDigestionParams digestionPara
...
-                searchModeType != CleavageSpecificity.Semi ? digestion.Digestion(this, topDownTruncationSearch)
-                : ProteinDigestion.WantsSemiSpecificSeeds(digestionParameters) ? digestion.SpeedySemiSpecificDigestion(this)
-                : digestion.SemiSpecificDigestion(this);
+            // DigestionParams.SpecificTermini decides which peptides are returned, in every search engine: fully specific,
+            // semi-specific (either terminus, or a required one) or non-specific. The one exception is MetaMorpheus's
+            // non-specific engine, which asks for fixed-terminus seed peptides it trims after the search (SearchModeType Semi
+            // or None with FragmentationTerminus N or C). See ProteinDigestion.SpecificTerminiDigestion. Before this, Semi
+            // with FragmentationTerminus Both returned C-terminal seeds and None with Both returned singleC ladders, and the
+            // Classic, Modern, Glyco and crosslink searches that asked for them lost peptides without any error.
+            IEnumerable<ProteolyticPeptide> unmodifiedPeptides = digestion.SpecificTerminiDigestion(this, topDownTruncationSearch);
```
nbollis: This circumvents the pre-existing digestion structure entirely

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
```diff
@@ -15,9 +16,22 @@ public DigestionParams() : this("trypsin")
...
             int maxModsForPeptides = 2, CleavageSpecificity searchModeType = CleavageSpecificity.Full, FragmentationTerminus fragmentationTerminus = FragmentationTerminus.Both,
-            bool generateUnlabeledProteinsForSilac = true, bool keepNGlycopeptide = false, bool keepOGlycopeptide = false)
+            bool generateUnlabeledProteinsForSilac = true, bool keepNGlycopeptide = false, bool keepOGlycopeptide = false,
+            SpecificTermini? specificTermini = null)
         {
-            Protease = ProteaseDictionary.Dictionary[protease];
+            // A protease mzLib no longer ships (e.g. "semi-trypsin", removed in #1005) still loads as the protease that
+            // replaced it plus the specificity its name implied, unless the caller says otherwise. A protease a user has
+            // defined under that name is found in the dictionary first and used as defined.
+            if (!ProteaseDictionary.Dictionary.TryGetValue(protease, out Protease resolvedProtease))
```
nbollis: i though the goal of removing semi-trypsin was to replace it with Trypsin where the cleavage specificity was set to semi?

nbollis (changes requested):
This does a TON of rewriting of our digestion and circumvents the original digestion path entirely. I am failing to see why the new enum is needed when we have an existing enum for digestion specificity. At first approximation, it appears like the issues this is fixing is could be due to bad parameters from the MetaMorpheus side.

trishorts:
Fair points, and thanks. With #1303 merged, trypsin + `SearchModeType` Semi already returns real semi-specific peptides through the existing path, which covers the case this PR was written for. I've cut it down:

- `SpecificTermini`, `TryResolveRemovedProtease`, the new non-specific digestion and the `Protein.Digest` rewiring are gone.
- What's left is the older bug it turned up (semi-specific peptides missing when the removed initiator Met is itself a cleavage site), fixed inside `SemiProteolyticDigestion`.
- Plus tests and docs for `SearchModeType` × `FragmentationTerminus`, since the "seeds vs peptides" meaning of N/C is what caused the original bug, and a test that fails if a shipped protease name is removed.
- Handling of old `semi-trypsin` settings moves to MetaMorpheus, mapping it to trypsin + Semi as #1005 intended.

The description is rewritten to match.

mzLib/Omics/Digestion/SpecificTermini.cs
trishorts: Agreed, removed. `Both`/`Either`/`Neither` were `Full`/`Semi`/`None`; the only values `CleavageSpecificity` lacked were the one-sided semi modes, and nothing needs those.

mzLib/Proteomics/ProteolyticDigestion/ProteaseDictionary.cs
trishorts: Agreed. We don't need `semi-trypsin` in mzLib: since #1303, trypsin (`K|,R|`) + `SearchModeType` Semi gives the same peptides. `TryResolveRemovedProtease` is removed; MetaMorpheus will map the old name when it reads settings files.

mzLib/Proteomics/Protein/Protein.cs
trishorts: Reverted; `Protein.Digest` keeps the existing branching. The only addition there is a comment.

mzLib/Proteomics/ProteolyticDigestion/DigestionParams.cs
trishorts: Yes, and since #1303 that works. The new setting is gone; this file now only gains doc comments on `SearchModeType` and `FragmentationTerminus`.

RayMSMS (approved):

nbollis (approved):

PR #1306, opened 2026-09-15
Read the spectrum ID and q-value that real mzIdentML writers produce

trishorts:
`MzidIdentifications` returns the wrong spectrum ID, or none, for most real mzIdentML files on PRIDE, and throws from `QValue` on Proteome Discoverer output. None of the published HUPO-PSI examples in `Test/DataFiles` exercise either path, so the suite could not see it. This adds five cut-down files downloaded from PRIDE Archive and fixes three defects in two accessors.

Found while grounding reading PRIDE-hosted identification files through Readers (the PRIDE client series: #1088, #1093, #1121, #1187, #1199). This PR is only the bug fix. The reader follows separately, on top of it.

## 1. `Ms2SpectrumID` returned null for MS-GF+ and Scaffold files

The method chooses between `spectrumID` and the spectrum title from `SpectraData[0].FileFormat`, and it matched that term by **display name only**. Writers carry the PSI-MS accession but do not always use its name:

| writer (PRIDE project) | FileFormat as written | accession | before | after |
|---|---|---|---|---|
| MS-GF+ (PXD078927) | `mzML file` | MS:1000584 | `null` | `controllerType=0 controllerNumber=1 scan=14316` |
| Scaffold (PXD000783) | `Mascot MGF file` | MS:1001062 | `null` | `Locus:1.1.1.3846.2 File:"…-F6.wiff"` |
| Proteome Discoverer (PXD019733) | `mzML format` | MS:1000584 | `scan=81409 file=454` | unchanged |

The format is now recognised by accession **or** by the existing name. That only widens the match, as long as a name and an accession never denote different formats, and no writer we found does that. Any file that matched before still takes the same branch.

## 2. `Ms2SpectrumID` returned a score threshold instead of the title

For MGF input the method read the result's **first** `cvParam` as the title. Scaffold happens to write the title first. Mascot Parser writes `Mascot:identity threshold` first:

```
Mascot Parser (PXD019591), SIR_26
  before: "17"
  after:  "16333: Scan 22014 (rt=62.2092) [\\qmcr.qmul.ac.uk\...\QE2_PharmacoDB_TP_Sample_16B.raw]"
```

This one returned a plausible-looking string with no error, which is why it is the worst of the three. The title is now selected by accession: MS:1000796, or the obsolete MS:1001416 that it replaced.

Only a result with **no** title term falls back to the first `cvParam`, so title-less files report what they reported before. The xiFDR fixture (PXD070193, 1.2.0) and the 1.3.0 HUPO-PSI fixture `noncovalently_assoc_1_3_0.mzid` take that path. A title term that is present but has no value, or an empty one, returns null. It does not take the fallback, which would hand back the threshold again.

## 3. `QValue` threw `ArgumentNullException` on Proteome Discoverer output

Proteome Discoverer scores each `SpectrumIdentificationItem` with a `userParam` and no `cvParam`, so the item's `cvParam` array deserializes as null. `QValue` called `.Where` on it. That throws for every PSM in the file, 861,250 of them in the published original. An item with no cvParams has no q-value, so it now returns the `-1` the method already uses for "absent".

The same applies to an MS:1002354 term with no value. `Convert.ToDouble((string)null)` is 0, which cannot be told apart from a genuine perfect q-value, so that also returns `-1` now.

## Fixtures

Five real files, each cut to its first two or three `SpectrumIdentificationResult`s plus the `DBSequence`, `Peptide` and `PeptideEvidence` entries those results reference. **Every kept line is byte-for-byte from the published file, line ending included.** Three sources are CRLF and two are LF, so the files are marked `-text` in `.gitattributes`. The only added text is the closing tags after the last kept result. The `ProteinDetectionList` is dropped, as are the `AnalysisCollection` elements that pointed at dropped content, so every `*_ref` resolves. `Test/DataFiles/PRIDE_MZID_PROVENANCE.md` records the source paths, SHA-256 checksums and the trim rule. 14–26 KB each.

| fixture | writer | mzIdentML | why it is here |
|---|---|---|---|
| `PXD078927_msgf_1_1_0.mzid` | MS-GF+ | 1.1.0 | "mzML file" name (defect 1); decoys |
| `PXD000783_scaffold_1_1_0.mzid` | Scaffold / Mascot | 1.1.0 | "Mascot MGF file" name (defect 1) |
| `PXD019591_mascotparser_1_1_0.mzid` | Mascot Parser | 1.1.0 | title not first (defect 2) |
| `PXD019733_proteomediscoverer_1_1_0.mzid` | Proteome Discoverer | 1.1.0 | userParam-only items (defect 3); exact-name control |
| `PXD070193_xifdr_1_2_0.mzid` | xiFDR | 1.2.0 | no title term, so it takes the fallback; results reference `SpectraData` index 5 of 9 |

All five also join the existing `EveryPsmIsReadable` sweep, with PSM and decoy counts read from the XML rather than off the reader. That sweep alone cannot catch defect 2, because `"17"` is not null. The targeted `Ms2SpectrumID` cases are what catch it.

The PRIDE files only reach the 1.1.0 and 1.2.0 arms, and every fix is repeated in all four version arms. Synthetic documents therefore drive each format and title case through all four namespaces.

## Verification

Every new assertion was run against unfixed master first and failed the predicted way: `null` for MS-GF+ and Scaffold, `"17"`/`"26"` for Mascot Parser, `ArgumentNullException` for Proteome Discoverer. The exact-name Proteome Discoverer case passed, as a control should.

The review round was also checked by mutation, each mutant rebuilt and run on its own. Every one turns the suite red:

| mutant | red |
|---|---|
| name-only format match in the 1.2.0 + 1.3.0 arms | 6 |
| first cvParam instead of the title lookup in the 1.2.0 + 1.3.0 arms | 2 |
| valueless title falls back to the first cvParam | 2 |
| obsolete MS:1001416 not recognised | 1 |
| valueless q-value read as 0 | 20 |

```
mzIdentML + mzML tests   237 passed, 0 failed   (net10.0-windows, current head)
full offline suite       6239 passed, 0 failed, 32 skipped (Test/required.runsettings, before the review round)
```

## Not in scope

- `Ms2SpectrumID` still reads only `SpectraData[0]`. The xiFDR fixture's results reference index 5 of its 9 `SpectraData` entries and get the right answer only because all nine are MGF. Proteome Discoverer's results do reference index 0 of its 30 entries, but its `spectrumID` values (`scan=N file=454`, `file=455`) point at two different RAW entries that this method does not resolve. The upcoming reader has to resolve both per result, and that is a new capability rather than a fix.
- `QValue` matches MS:1002354 exactly, so it misses its child terms, such as MS:1002054 `MS-GF:QValue`, which the MS-GF+ fixture carries on all 12 PSMs, and MS:1002723 and MS:1003125. `EveryPsmIsReadable` pins the current 0 and says so. Fixing that belongs in its own PR.
- The accessors' linear scans over `SequenceCollection` also stay as they are. The reader will index once rather than change these members.


nbollis (dismissed):

acesnik (commented):
🤖 All three defects here are the same shape — the reader trusted a display name or an array position where it should have trusted an accession — and the fixtures are what make that legible. Cutting five real writers' output down to its first few results is the right call; the published HUPO-PSI examples genuinely cannot reach any of these paths. I reproduced all three mutation claims: name-only matching turns 8 cases red with `But was: null`, an unconditional first-cvParam read gives `"17"`/`"26"` on Mascot Parser, and dropping the four `?.` guards gives `ArgumentNullException` from `Enumerable.Where` on every Proteome Discoverer PSM. The exact-name Proteome Discoverer case stays green, as a control should. Nothing here is vacuous.

The gap I'd want closed is that the accession lookups treat "term found but unusable" the same as "term not found", which in two places reinstates the bug being fixed. `SpectrumTitle`'s `?? all.Select(cv => cv.Value).FirstOrDefault()` fires on a null `.Value` whether the tuple came from a no-match `FirstOrDefault` or from a real `MS:1000796` with no `value` attribute, and `Convert.ToDouble((string)null, provider)` returns 0 rather than throwing. Measured against synthetic 1.1.0 documents:

```
MS:1000796 present, no value attribute  ->  "17"   (i.e. the MS:1001371 identity threshold, again)
MS:1000796 present, value=""            ->  ""
MS:1002354 present, no value attribute  ->  0      (QValue's most confident answer, not its -1)
```

`value` is optional on `AbstractParamType` in all four generated models (`mzIdentML1_1_0.cs:549` and siblings) and `PXD019591_mascotparser_1_1_0.mzid` already carries a valueless cvParam on a `SpectrumIdentificationItem` (`MS:1001363 peptide unique to one protein`), so the shape is real even though no fixture pairs it with these two accessions. The q-value one is the one I'd fix regardless: 0.0 is indistinguishable from a genuine perfect q-value. Related, and cheap while you're in there — the CV has an obsolete second `spectrum title`, `MS:1001416`, `replaced_by: MS:1000796` (`psi-ms.obo:9317-9322`); an older-CV file written with it currently falls through to the threshold as well (measured: `"17"`).

Second, half the diff is unpinned in the 1.2.0 and 1.3.0 arms. I reverted each fix in those two arms only and the suite stayed green both times:

```
IsFileFormat -> format.name == "..."  in dd120+dd130 only   ->  151 passed, 0 failed
SpectrumTitle() dropped from dd120+dd130 only               ->  151 passed, 0 failed
```

The null guard *is* pinned in all four arms, and `QValue_ItemWithNoCvParams_IsAbsentInEveryVersion` really does reach distinct arms — dispatch is by namespace, and each of the four lands on its own `ddNNN`. It's the accession matching that only the 1.1.0 fixtures and the 1.1.1 synthetic reach. `IsFileFormat` being one shared static is why patch coverage reads 100% while four call sites go unchecked. `Ms2SpectrumID_MzIdentML111_RecognisesTheFormatAndTitleByAccession` already takes a namespace-parameterised document, so crossing its three format cases with all four namespaces is a one-line change and a rename.

Worth its own PR rather than this one: the MS-GF+ fixture does report q-values and the reader can't see them. `MS:1002054 MS-GF:QValue` is `is_a: MS:1002354 ! PSM-level q-value` in the shipped obo, `PXD078927_msgf_1_1_0.mzid:142,153,165` carry it with values 0.0, 0.6588785 and 0.73832303, and `QValue` matches the parent accession exactly, so it returns -1 for all 12 PSMs and `EveryPsmIsReadable` now pins `withQValue = 0`. `MS:1002723` and `MS:1003125` are the other two children. Pinning current behaviour is right; the assertion just reads as "this file has no q-values" when it has twelve, so it's worth a word in the comment.

Four numbers in the body to fix while you're editing it, since that costs nothing: the Proteome Discoverer file has **30** `SpectraData`, not 24 (one `MS:1000584 mzML format` plus 29 `MS:1000563 Thermo RAW format`, ids 454–482). It's also not the file that demonstrates the `SpectraData[0]` shortcut — all three of its kept results reference `ID_MZML_FILE_with_spectra`, which *is* index 0; what's unresolved there is PD's own `file=454`/`file=455` inside the spectrumID, pointing at two different RAW entries. The fixture that actually reads the wrong element is xiFDR: its results reference `SpectraData` index **5** of 9, and it only gets the right answer because all nine are `MS:1001062`. That's the first fixture in the tree where that index is wrong, which strengthens rather than weakens your scope call. And "two existing 1.3.0 HUPO-PSI fixtures" is two test cases over one fixture — `noncovalently_assoc_1_3_0.mzid`; `multiple_spectra_per_id_1_3_0.mzid` is mzML and unaffected. "Matching on either one only widens the match" holds as long as an accession and a name never denote different formats, which no writer I can find does.

Minor: the cut left ten refs dangling — eight `spectrumIdentificationList_ref` in the xiFDR file (`:77,81,85,89,93,101,105,109`) and a `proteinDetectionList_ref` each in scaffold (`:56`) and mascotparser (`:122`). Inert for mzLib, which never resolves them, but it makes the files unusable as documents for anything that does. And since the doc comment claims these are byte-for-byte, they probably want the same treatment `.gitattributes:66-87` already gives `*.sdrf.tsv` and `*.obo` — `* text=auto` currently applies to them, so a Windows checkout hands back CRLF. The `Ontologies/PROVENANCE.md` pattern would also keep the PXD accessions and the trim rule findable later. Is xiFDR's absence from `EveryPsmIsReadable` deliberate? It's the one new fixture the sweep skips.

`MzidIdentificationsTests.cs:329-330` and `:350` now describe the old mechanism — that file's title is found by the `MS:1000796` lookup, not by being first.

Two caveats on how I checked this. `Test/Test.csproj` is `net10.0-windows` with a WPF project reference, so it builds but can't run on arm64; I ran `Test/MzIdentML/*.cs` through a throwaway `net10.0` harness, 151 cases. And nothing in mzLib or MetaMorpheus consumes `MzidIdentifications` outside tests — MetaMorpheus only writes mzIdentML — so none of this can move search output today.

trishorts:
🤖 Thanks, all of this reproduced. Addressed in 5b320b48 and b4102d15; the PR body is updated to match.

**A term that is present but unusable.** `SpectrumTitle` now finds the title term by index. If a term is present, its value is the answer, and a null or empty value returns null. Only a result with no title term at all takes the first-cvParam fallback. `MS:1001416` is recognised alongside `MS:1000796`. `QValue` goes through one `QValueOf` that treats a null or blank value as `-1`, so a valueless `MS:1002354` is absent, not 0. New tests cover a valueless title, an empty title, the obsolete title (1.1.0), and a valueless q-value in all four namespaces, plus a control that a valued q-value is still read. With the fixes reverted:

```
valueless title falls back to first cvParam   ->  2 red
obsolete MS:1001416 not recognised            ->  1 red
valueless q-value read as 0                   ->  20 red
```

**1.2.0 / 1.3.0 arms.** `Ms2SpectrumID_MzIdentML111_…` is now `Ms2SpectrumID_EveryVersion_RecognisesTheFormatAndTitleByAccession`, 3 formats × 4 namespaces. Your two reverts now go red:

```
IsFileFormat -> format.name == "..."  in dd120+dd130 only   ->  6 red
SpectrumTitle() dropped from dd120+dd130 only               ->  2 red
```

**MS-GF+ q-values.** Agreed that this belongs in its own PR. `EveryPsmIsReadable` now says the 0 pins current behaviour: the file carries `MS:1002054` on all 12 PSMs, and `QValue` matches only the parent accession. It's also listed under "Not in scope" in the body.

**Body numbers.** Corrected. PD has 30 `SpectraData` and its results all hit index 0; its unresolved part is `file=454`/`file=455`. xiFDR is now named as the file where `SpectraData[0]` is the wrong element (index 5 of 9). The 1.3.0 wording now names the single `noncovalently_assoc_1_3_0.mzid` fixture.

**Fixtures.**
- The dangling refs are gone. I dropped xiFDR's eight `SpectrumIdentification`s whose lists were cut and the `ProteinDetection` in Scaffold and Mascot Parser. A scan of every `*_ref` against every `id` now finds nothing unresolved in all five files.
- `-text` was more than cosmetic. Streaming the originals back from PRIDE showed that Scaffold, Mascot Parser and Proteome Discoverer are **CRLF** and MS-GF+ and xiFDR are LF, so the LF-normalized blobs made the "byte-for-byte" claim false for three files. They are now `mzLib/Test/DataFiles/PXD*.mzid -text` and stored with their source line endings.
- The same pass matched every non-blank fixture line against its source, except the added closing tags. It also confirmed each fixture keeps the source's first `SpectrumIdentificationList` and that list's first results.
- `Test/DataFiles/PRIDE_MZID_PROVENANCE.md` follows the `Ontologies/PROVENANCE.md` pattern. It records the source paths, SHA-256 of the files as served, line endings and the trim rule.

**xiFDR in the sweep.** It wasn't left out on purpose. It's in now, as `(4, 0, 0)`, and it passes.

**Stale comments** at `:329-330` and `:350` now describe the `MS:1000796` lookup.

Verified on net10.0-windows: mzIdentML + mzML filter 237 passed, 0 failed, with every mutant above rebuilt and run on its own.

PR #1307, opened 2026-09-15
Let SdrfBuilder write comment[label] bare

trishorts:
> **Project of origin:** `sdrf`

Adds `SdrfBuilderOptions.LabelForm`, so a caller can have `SdrfBuilder` write `comment[label]` bare
(`TMT127N`) instead of as a term (`NT=TMT127N;AC=PRIDE:0000519`).

**Counterpart: MetaMorpheus #2817** (draft), which writes one SDRF row per sample per channel for a
TMT search and will pass `LabelForm = Bare`. That is the caller this option exists for.

## Why a builder that authors CV terms would write a bare one

`SdrfBuilder`'s rule is that every value is written as a resolved term, and that rule stays for every
other column. The label is the exception, for two measured reasons:

- **The community writes it bare.** 44 of 51 channel-level TMT SDRFs in the 1,236-file
  `bigbio/sdrf-annotated-datasets` corpus do. `SdrfLabelForm.Bare`'s own doc comment already says
  "about six sevenths of the corpus".
- **quantms mishandles the accessioned form.** In `sdrf-pipelines` 0.1.6 a term-form label takes a
  branch in the OpenMS converter that keeps one label per file. It crashes with an `IndexError` on the
  second channel of every TMT file, and silently infers the wrong plex
  (`bigbio/sdrf-pipelines#344`). The bare branch works.

An SDRF that the main downstream pipeline cannot read defeats the point of writing one.

## The change

- **`LabelForm`**, typed as the existing public `SdrfLabelForm` enum (`Bare` / `Accessioned`) that
  `SdrfQuantAuditor` already reports with, so the writer and the auditor use one vocabulary.
- **Defaults to `Accessioned`.** Nothing changes for an existing caller.
- **Bare changes the spelling, not the resolution.** The label is still a resolved `CvParam`;
  `Bare` writes its `Name`.
- **A label with no name keeps its accessioned form**, so `Bare` never discards an accession.
- **A missing label behaves the same in either form**: it throws under `RequireSampleMetadata`, and
  writes `not available` otherwise.
- **A tab or newline in a bare label is refused**, as `SdrfCell.ToCell` refuses it in a term.

The production change is one option in `SdrfBuilderInputs.cs` and one private method in
`SdrfBuilder.cs`. No new type is added and nothing is newly made public.

## Tests

Added to `TestSdrfBuilder`:
- **Accessioned by default**, for two TMT channels of one file.
- **Bare writes the name alone, and only in `comment[label]`.** Organism and instrument stay terms.
  The document still validates, and `SdrfQuantAuditor` reads both cells as bare and the file as
  `ChannelLevel`.
- **A bare label with no name keeps `AC=`.**
- **A missing label** throws when strict and reads `not available` when lenient.
- **A tab in a bare label is refused.**

The bare test was seen **failing** with the option ignored.

Full suite, `Category!=ExternalService`, on `net10.0`: **6,235 passed / 0 failed / 32 skipped**. SDRF subset 126 / 0.


nbollis (approved):

Alexander-Sol (approved):
**Verdict: Approve** ✅

Adds a single option, `SdrfBuilderOptions.LabelForm` (`Bare` / `Accessioned`), reusing the existing public `SdrfLabelForm` enum, plus a private `LabelCell` helper that writes `comment[label]` as the resolved term's bare name (`TMT127N`) instead of the full CV term (`NT=TMT127N;AC=PRIDE:0000519`). Defaults to `Accessioned` (non-breaking), falls back to the term form whenever there is no name, and mirrors `SdrfCell.ToCell`'s separator refusal. Well-documented, and the motivation (a quantms/sdrf-pipelines 0.1.6 bug that crashes on the accessioned form) is credible and cited.

Verified as correct:
- **Non-breaking**: `LabelForm` initializes to `Accessioned` (`SdrfBuilderInputs.cs:155`), overriding the enum's zero value (`Bare = 0`), so existing callers are unchanged.
- **Fallbacks**: traced all four branches (null label / whitespace name → `Missing`; name-empty-but-accessioned → keeps `AC=`); each matches the docstring and tests.
- **Auditor round-trip**: bare output (`TMT126`, `TMT127N`) reads back as `Bare` (no `=`), the no-name case (`AC=PRIDE:0000519`) as `Accessioned` — matching the test assertions.

Minor observations, none blocking:

1. **Duplicated separator-refusal logic** (nit). The bare path re-implements the tab/newline guard already in `SdrfCell.ToCell`, including a near-identical error message (`SdrfBuilder.cs:394-398`). Defensible since the bare path deliberately bypasses `ToCell`, but the two copies can drift if the separator rule ever changes. A shared `SdrfCell.RefuseSeparators(string)` would keep them in lockstep.
2. **Bare name guarded against tab/newline but not `;` or `=`** (nit). A bare name containing `;` or a leading `NT=` would round-trip incorrectly (re-read as `Accessioned`). Not a regression — `SdrfCell.ToCell` has the same blind spot, and real channel names (`TMT126`, `label free sample`) never contain these — worth a one-line note only.

Staleness check (per the review ask): **not stale.** Branch merged `master` on 2026-09-17, merge-base equals master tip, 0 commits behind, `gh` reports `MERGEABLE`. It touches `SdrfBuilder.cs`/`SdrfBuilderInputs.cs`, which the sibling SDRF PRs (#1334 etc.) also edit, so a textual conflict is plausible when those merge — worth sequencing, but not a defect here.

Test assessment: good. Five tests cover accessioned-by-default, bare name-only (with a full validator + auditor round-trip proving other columns stay CV terms), bare-with-no-name keeping `AC=`, missing label in strict and lenient modes, and separator refusal; the bare test was confirmed red with the option ignored. Minor gaps (separator test only exercises `\t`; default test doesn't assert the `Accessioned`/`Bare` cell counts), none blocking.

<sub>Agent-assisted review requested via the `ready-for-agent` label.</sub>

PR #1308, opened 2026-09-16
Sequence-variant corpus: insertion tranche + keep mods on residues an anchored indel leaves unchanged

trishorts:
## Insertion tranche for the sequence-variant corpus, plus a mod-drop fix

Follows #1084 (L0 corpus), #1095 (L1 deletions) and #1098 (substitutions). Adds the
**insertion** half of L1 (12 new oracle nodes) and fixes the one bug they exposed. Every
`ExpectedForms` oracle is hand-derived from the domain model, **not captured from running
code**, so a node that goes red points at real behavior.

### The fix: `AdjustModificationIndices` dropped mods on residues a variant keeps

UniProt writes an insertion anchored on a kept residue: "insert AG after T4" is
`<original>T</original><variation>TAG</variation>` at position 4. `AdjustModificationIndices`
dropped every mod inside the variant's `[begin,end]`, so a phospho on T4 disappeared from the
variant even though T4 is still there, unchanged, at position 4.

The fix counts the unchanged prefix of original vs variant, then the unchanged suffix of what
remains, and keeps mods on those residues (prefix at the same position, suffix shifted by the
length change). The anchor can sit on either side (`T→TAG` keeps the leading T; `P→AGP` keeps
the trailing P). Real edits still drop the mod: `T→V`, `T→S`, a deleted residue, a
multi-residue substitution. The same rule covers the kept residue of an anchored deletion
(`TI→T`).

Without the fix, exactly **I01** and **D09** fail (in both test families); nothing else changes.

### New nodes

| Node | What it pins |
|------|--------------|
| **I00 / I01 / I02** | Insertion immediately **before / anchored on / after** a PTM residue. I00 shifts the phospho T4→6; I01 keeps it on the retained T4 (the fixed bug); I02 leaves it alone. |
| **I03 / I04** | Insertion at the protein **start** (`P1→AGP`, whole protein shifts +2) and **end** (`E7→EAG`, C-terminal extension). |
| **I05 / I06** | VCF insertion **passing / failing** the `minAlleleDepth` cutoff. Unlike the deletion/substitution twins these carry a phospho, so the `ApplyVariants` path's re-anchoring (T4→6) is checked too. |
| **I07** | Bottom-up: an inserted **K creates a trypsin cut site** (`A4→AK`: `PEPAKIDE` → `PEPAK` + `IDE`). |
| **I08** | An insertion that **carries a variant-borne mod** (`P3→PT`, phospho stored on the variant at the inserted T, subposition 4, one past the variant's own position). |
| **MI00** | Two insertions **flanking** a PTM; the phospho sits at T6 or T4 depending on the combination. |
| **MI01** | An insertion (`P1→PA`, +1) and a deletion (P3, −1) upstream of the PTM whose shifts **cancel**: the pair lands T back at 4 only if each step re-anchors in the frame the previous one left. |
| **D09** | Anchored **deletion** `TI→T` whose kept T keeps its phospho (deletion half of the I01 rule). |

### Harness change (test infra only)
- `BuildXml` labels VCF records for longer variants `inframe_insertion` instead of `frameshift_variant`. Descriptive only; nothing in the variant path reads the label.

### Result
- Corpus **86 / 86** green (43 nodes × `Corpus_Node` + `Corpus_Node_RoundTrip`).
- `DatabaseTests` + `Transcriptomics` + `*Variant*` tests **1188 / 1188**, no existing expectation changed.
- Based on current `master` (`c1d58c0b`).


acesnik (commented):
🤖 The diagnosis is the valuable part here: UniProt anchors an insertion on a residue it keeps, so `T→TAG` at position 4 is an edit that doesn't edit T4, and dropping every mod in `[begin,end]` discarded a phospho on a residue that never moved. Counting the unchanged prefix and then the unchanged suffix of what remains is the right shape, and bounding the suffix loop by `alternate.Length - keptPrefix` is what stops the two branches claiming the same residue — their destination ranges come out provably disjoint, so there's no duplicate-key throw lurking. I reran the mutation check: with only the `AdjustModificationIndices` hunk reverted, exactly I01 and D09 fail in both families, each missing precisely the phospho-carrying variant form. The new nodes aren't vacuous.

The fix does make one latent bug bite, and it changes produced databases, so I'd rather it were handled here than found later. `AdjustModificationIndices` stores the source protein's own `List<Modification>` by reference (`VariantApplication.cs:463`, `:467`), and the variant-mod merge below then does `modsAtPos.AddRange(kv.Value)` (`:479`), mutating that list in place. Until now the carried-over keys were always strictly outside the variant's applied span while variant-borne mods live inside it, so the two could never collide. The new kept-prefix branch admits exactly the anchor residue — which is where a variant-borne mod on an insertion sits. On `MPEPTIDE` with a phospho on T5 and `T5→TAG` carrying its own mod at key 5, the consensus protein's list at position 5 goes 1 → 2 → 3 → 4 over three `GetVariantBioPolymers()` calls; on master it stays at 1. With a second, unrelated `P2L` variant and a single call, the unvarianted consensus and the `P2L` isoform — which does not contain the insertion — both come back carrying the variant's modification:

```
P00003            MPEPTIDE    5:[Phospho, VariantBorne, VariantBorne]
P00003_T5TAG      MPEPTAGIDE  5:[Phospho, VariantBorne]
P00003_P2L        MLEPTIDE    5:[Phospho, VariantBorne]
P00003_P2L_T5TAG  MLEPTAGIDE  5:[Phospho, VariantBorne, VariantBorne]
```

A real substitution `T5→V` is unaffected on both arms, which is the control. `mods.Add(kv.Key, new List<Modification>(kv.Value))` in both re-base branches closes it.

Separately, the kept-suffix half of the rule isn't pinned by anything. Your own result is the evidence: if reverting the hunk fails only I01 and D09, and both of those are kept-prefix cases, then nothing in the 43 nodes discriminates on `keptSuffix`. I03 is the one right-anchored node (`P1→AGP`, keptSuffix=1), but its phospho sits downstream at T4 and takes the shift branch identically with or without the fix. `OP=T VAR=AGT POS=4` on `PEPTIDE` with `Phosphorylation@4` would do it — keptPrefix=0, keptSuffix=1, mod dropped by the old code and shifted 4→6 by the new.

That gap matters more than coverage, because two public helpers still use the old notion of "inside the variant" and MetaMorpheus calls both. `IsSequenceVariantModification` goes through `Includes`, plain `begin <= pos <= end` (`SequenceVariation.cs:304-307`), and the applied span for `T4→TAG` is `[4,6]` — so a mod on the residue this PR just established as unedited now reports as variant-borne. `MetaMorpheusTask.cs:895` uses exactly that to set `unappliedVariant`, and the database-mod loop below it matches on the tuple `(idx, mod, null)`, so in a written pruned database that mod moves off the protein's own list and onto the variant. `RestoreModificationIndex` (`VariantApplication.cs:101-106`) subtracts only for variations ending strictly before the position; for `P3→AGP` the applied span is `[3,5]` and the kept-suffix mod lands at 5, so `5 < 5` is false, it returns 5, and the consensus position is 3. That index is what `MetaMorpheusTask.cs:899` writes. Both of these I verified by reading the call sites — I did not run MetaMorpheus.

Last, and not for this PR: under the new rule a mod on the kept T of `T4→TAG` survives, but a TruncationProduct anchored on that same residue is dropped — `AdjustTruncationProductIndices` has no kept-residue notion, so a chain `(4,7)` falls through all three branches at `:387`, `:392` and `:406`. The header already parks processing as a later layer; I mention it only because this PR is what makes the two disagree on the same residue.

Otherwise it holds up: 86/86 at head, and no pre-existing expectation moved across the 1149 DatabaseTests/Transcriptomics/Variant tests I could compile on macOS (the 5 failures there are pre-existing path-separator and newline issues that fail identically on both arms).

trishorts:
@acesnik Thanks. All four points hold up, and the first three are fixed in c51b58c1 (rebased onto master `bb37c9c8`).

1. **Aliasing:** both re-base branches in `AdjustModificationIndices` now copy the list (`new List<Modification>(kv.Value)`), and so does the shift branch. Your `MPEPTIDE` repro (phospho on T5, `T5→TAG` with its own mod, plus an unrelated `P2L`) is now `SeqVar_AnchoredInsertionWithVariantBorneMod_DoesNotMutateSourceModLists`. The consensus and the `P2L` isoform carry only the phospho, and repeated `GetVariantBioPolymers()` calls no longer grow the list.
2. **Kept-suffix branch unpinned:** added corpus **I09**, which is your case: `PEPTIDE`, `Phosphorylation@4`, `OP=T VAR=AGT POS=4`. The phospho shifts 4→6 instead of being dropped. With master's `AdjustModificationIndices`, I01, D09 and I09 now all fail in both families, so each half of the rule is pinned.
3. **MetaMorpheus helpers:** `IsSequenceVariantModification` and `RestoreModificationIndex` now use the same kept-anchor rule as `AdjustModificationIndices`, via a shared `CountKeptFlanks`. `TestProteinVariantModMethods_KeptAnchorsOfAnchoredIndels` covers both of your examples: the kept T of `T4→TAG` is not variant-borne, and for `P3→AGP` the kept P at 5 restores to 3.
   - One existing expectation flipped: `TestProteinProperties` asserted that the mod on the kept P4 of `P4→PV` was variant-borne. Under the rule this PR establishes, it isn't, so I corrected the assertion.
   - Known limit: a variant-borne mod sitting *on* a kept anchor now classifies as the protein's own. These helpers only see an index, so they can't tell the two apart.
4. **TruncationProduct on a kept anchor:** agreed it's not for this PR. It's parked for the processing layer (L2), with your `(4,7)` example.

Local gate: 1204/1204 (DatabaseTests, Transcriptomics, variant corpus, `TestProteinProperties`). CI is 5/5 green on `c51b58c1`. I didn't run MetaMorpheus end to end, so the MetaMorpheus side rests on the call sites you cited.

nbollis (approved):

PR #1311, opened 2026-09-17
Add DissociationType.IsCid() and use it for the b1 skip

### LINKED ISSUE #1310: No DissociationType.IsCid() helper, so the CID/LowCID equivalence gets respelled in each repo
🤖 Raised out of MetaMorpheus#2717 at @trishorts's suggestion.

Low-resolution CID is still CID, so code that special-cases CID generally has to special-case LowCID with it. mzLib does that at `Proteomics/ProteolyticDigestion/PeptideWithSetModifications.cs:337`, skipping the b1 fragment:

```csharp
// skip first N-terminal fragment (b1, aDegree1, ...) for CID
if (r == 0 && (dissociationType == DissociationType.CID || dissociationType == DissociationType.LowCID))
```

MetaMorpheus#2717 has just added a second copy of the same equivalence, in `Crosslinker.IsCleavedBy`, to decide whether a CID-cleavable crosslinker also cleaves in a LowCID scan. Two repositories now each assert "CID or LowCID" in their own words, with nothing keeping them in step — including if another low-resolution CID-like type is ever added to the enum. An `IsCid()` extension next to the enum in `MassSpectrometry/Enums/DissociationType.cs` would give both one definition to call.

Worth being precise about the size of this, because that file mentions `LowCID` five times and only one is the equivalence. `:337` is the only place in mzLib that treats CID and LowCID as the same thing. The other four (`:323`, `:345`, `:407`, `:431`) are the opposite: LowCID-only branches for the star and degree ions that low-resolution CID produces and CID does not. An `IsCid()` helper must not absorb those. So this is one call site plus a name, not a sweep.

acesnik:
🤖 Closes #1310.

Low-resolution CID is still CID, so code that special-cases CID generally has to special-case LowCID with it. That equivalence was spelled out inline at `PeptideWithSetModifications.cs:337`, and since MetaMorpheus#2717 a second time in MetaMorpheus's `Crosslinker.IsCleavedBy`. This gives both one definition to call.

## What this changes

A `DissociationTypeExtensions.IsCid()` next to the enum in `MassSpectrometry/Enums/`, and the single call site:

```diff
- if (r == 0 && (dissociationType == DissociationType.CID || dissociationType == DissociationType.LowCID))
+ if (r == 0 && dissociationType.IsCid())
```

Nothing else. No behaviour change is intended — the helper is the same boolean expression, and the tests below check that over every member of the enum rather than on hand-picked cases.

## What IsCid deliberately excludes

Worth stating, because that file mentions `LowCID` five times and only one of them is the equivalence. `:337` is the only place in mzLib that treats CID and LowCID as the same thing. The other four (`:323`, `:345`, `:407`, `:431`) are the opposite — LowCID-only branches for the star and degree ions that low-resolution CID produces and CID does not. `IsCid()` must not absorb those, and does not.

HCD and ISCID are excluded too, which is the non-obvious part given that HCD is literally beam-type collision-induced dissociation. Neither shares the b1 behaviour this check guards, so sweeping them in would change results. The doc comment says so, so the next person does not have to rediscover it.

## Tests

`Test/MassSpectrometryTests/TestDissociationTypeExtensions.cs`:

- `IsCidIsTrueForExactlyCidAndLowCid` — enumerates every `DissociationType` and asserts the true set is exactly `{CID, LowCID}`, so adding an enum member cannot quietly widen or narrow the helper.
- `IsCidExcludesTheOtherCollisionInducedTypes` — HCD, ISCID, EThcD, AnyActivationType, Unknown.
- `IsCidMatchesTheInlineComparisonItReplaced` — asserts `d.IsCid()` equals the original inline expression for every enum member, which is what makes the call-site substitution provably behaviour-preserving rather than argued.

Verified on macOS/arm64 .NET 10. `Test.csproj` is `net10.0-windows` (it references `mzPlot`, which carries OxyPlot.Wpf), so it cannot host a run here; the tests were run through a throwaway `net10.0` harness instead. The three new tests pass.

The behaviour-preservation claim is measured, not argued. Running the suites that cover `PeptideWithSetModifications.Fragment` — `Omics/FragmentationTests`, `Omics/Modifications`, `ProteomicsTests/**` — gives **751 results, 750 passed, 1 failed** both before and after the change, with **zero** per-test outcome differences across 742 test methods. The single failure is pre-existing and unrelated: `TestPtmListLoader.TestReadingInvalidModifications` splits error text on a literal `\r\n`, so it fails on macOS line endings in both arms.

Two guards on that comparison, because "identical results" is exactly what a stale build also looks like. `Proteomics.dll` hashed `c6101543f1dc` on the changed arm, `967b6d01a7e2` on the baseline, and returned to `c6101543f1dc` on restore — so each arm genuinely recompiled. And the suite is demonstrably sensitive to this branch: mutating `IsCid` to `return false` fails 33 instead of 1, including `Test_CID_Fragmentation_No_Unmodified_B1_ions` and all 28 `Test_Fragment_ProductTypesWithAminoAcidSpecificities` cases.

## One thing a reviewer will reasonably ask

This does not deduplicate anything inside mzLib. `:337` was the only `CID || LowCID` conjunction in non-production-test source — every other `DissociationType.CID` occurrence is a dictionary key, a `switch` case, or a CV-term/reader mapping — so `IsCid()` ships with exactly one caller here. The point is to give the concept one name and one home so MetaMorpheus can share it instead of keeping a second copy; if you would rather not add a public API for a single internal caller until MetaMorpheus actually calls it, that is a fair objection and I would rather hear it now than after it is published.

## Related

MetaMorpheus's copy is `EngineLayer/CrosslinkSearch/Crosslinker.cs`, in `IsCleavedBy` (MetaMorpheus#2717). It cannot call this until MetaMorpheus takes a newer mzLib package, so that swap belongs in a MetaMorpheus PR after this ships rather than here.


trishorts (approved):
Looks good. I'm fine with adding the public helper while it has a single caller here: MetaMorpheus's `Crosslinker.IsCleavedBy` copy is a concrete second caller, and giving the concept one name is worth it. The exhaustive enum test is a good guard. mzLib has no existing `DissociationType` extension with this name (the only other extensions live in `Omics/Fragmentation/Oligo/DissociationTypeCollection.cs`), so nothing collides.

nbollis (approved):

PR #1312, opened 2026-09-17
Resolve terminal modifications by UNIMOD id to the terminal entry

trishorts:
`UnimodModificationLookup` returns the wrong modification for most terminal modifications identified by UNIMOD accession, with no warning. The cause is shared code in `ModificationLookupBase`, so `MzLibModificationLookup` and `GlobalModificationLookup` behave the same way. This PR fixes the identifier path there.

Found while grounding an mzIdentML reader for PRIDE files (the PRIDE client series). mzIdentML and ProForma both hand the lookup a UNIMOD accession, a position, usually the residue it sits on, and often a mass delta. In production today, `CanonicalModification.UnimodId` is set by `ProFormaSequenceParser`, so the live case is, for example, ProForma `[UNIMOD:1]-MPEPTIDE` converted to mzLib format.

## What it returned

Built by hand as `CanonicalModification.AtNTerminus` / `AtCTerminus` / `AtResidue` and resolved with `new UnimodModificationLookup()` against the shipped Unimod catalog:

| terminal modification | before | after |
|---|---|---|
| N-term UNIMOD:1, no residue, with mass | `Acetyl on T` (Anywhere) | `Acetyl on X` (N-terminal) |
| N-term UNIMOD:1, no residue, no mass | null | `Acetyl on X` (N-terminal) |
| N-term UNIMOD:1 on M | `Met->Hse on M` (Peptide C-terminal) | `Acetyl on X` (N-terminal) |
| N-term UNIMOD:1 on A | `Fluoro on A` | `Acetyl on X` (N-terminal) |
| N-term UNIMOD:1 on K | `Acetyl on K` (Anywhere) | `Acetyl on X` (N-terminal) |
| N-term UNIMOD:34 on A | `Methyl on X` (**Peptide C-terminal**) | `Methyl on X` (N-terminal) |
| C-term UNIMOD:2 on K | `Phospho on K` | `Amidated on X` (Peptide C-terminal) |
| C-term UNIMOD:2, no residue, no mass | null | `Amidated on X` (Peptide C-terminal) |
| C-term UNIMOD:34 on K | `Methyl on K` (Anywhere) | `Methyl on X` (Peptide C-terminal) |
| N-term UNIMOD:2 on K (amidation is C-terminal only) | `Phospho on K` | null |

Residue modifications already resolved correctly and still do (Phospho S/Y, Oxidation M, Carbamidomethyl C, Acetyl K, GG K, Deamidated N).

## Why

The hand-built cases turned up five separate errors on the identifier path, `ResolveInternal` → `GetPrimaryCandidates` → `SelectBestCandidate`:

1. **`FilterByUnimodId` matched the id as a substring**, `value.Contains("1")`. UNIMOD:1 started from entries for 1005 distinct ids, including Met->Hse (10) and Fluoro (127); UNIMOD:2 included Phospho (21).
2. **The position was never applied on this path.** `FilterByTerm` runs in one place, at the end of `ApplyCumulativeFilters`, after every earlier return. `SelectBestCandidate` prefers an exact residue over `X`, so a terminal modification lost to the side-chain entry with the same id (`Acetyl on K`).
3. **The opposite terminus was accepted**, for the same reason (`Methyl on X`, Peptide C-terminal, for an N-terminal methyl).
4. **List order broke ties.** With no residue, every Acetyl specificity shares one formula, so the first one in the catalog won (`Acetyl on T`).
5. **Terminal modifications with neither residue nor mass were unresolved.** The substring set had too many formulas to be "equivalent".

## The change

Everything is in `ModificationLookupBase`, private or already-protected members. No public API changes.

- `FilterByUnimodId` compares the whole id. `IsUnimodId` accepts `"35"` and `"UNIMOD:35"`, the two forms the existing tests use for `DatabaseReference` and `Accession`.
- `ResolveInternal` narrows the identifier's candidates to the position before selecting, in `NarrowToTerminus`:
  - An N- or C-terminal position keeps the entries restricted to that terminus (the existing `IsNTerminalRestriction` / `IsCTerminalRestriction`).
  - If there are none, it keeps the entries `FilterByTerm` allows there, which includes `Anywhere.`.
  - It never keeps the opposite terminus.
  - A residue position is left alone.
- **An id given with no residue, whose entries sit on different residues, is ambiguous and is not handed to `SelectBestCandidate`.**
  - That method's own last step says distinct residues are ambiguous. But its same-formula tie-break comes first, and with the exact candidate set it would have returned `Phospho on C` for `[UNIMOD:21]` with no residue.
  - `GlobalModificationLookupTests.TryResolve_WithUnimodIdWithoutResidue_ReturnsNull` pins null there. It used to pass only because the substring set was too mixed to tie.

### Alternatives considered

- **Exact id matching alone.** It fixes Met->Hse, Fluoro and Phospho, but not `Acetyl on K`, `Methyl on K`, or the opposite-terminus methyl.
- **Calling `FilterByTerm` before selecting.** It lets `Anywhere.` through, so `Acetyl on K` and `Acetyl on X` still compete and the residue preference still picks K. Terminal entries have to be preferred, with `Anywhere.` as the fallback.
- **Dropping `Anywhere.` at a terminus altogether.** That breaks `FilterByTerm_RespectsLocationRestrictions`, `MatchesTermRestriction_HandlesTerminalStrings` and `PeptideAndProteinConversion`, all of which rely on a side-chain modification being usable at a terminus. It would also make a terminal UNIMOD:21 on S unresolvable.
- **Fixing the tie-break inside `SelectBestCandidate`.** That method serves the name, formula and mass paths too, and `TryResolve_ResidueLessDeamidationMass_ResolvesToProteinModNotRnaInosine` depends on that tie-break resolving a residue-less mass. The ambiguity check is limited to the identifier path for that reason.
- **Sharing one Unimod-id parser.** `KoinaModelBase.TryGetUnimodId` and `CanonicalModification.WithResolvedModification` already parse ids exactly, but one is private in `PredictionClients`. Consolidating the three copies means editing those members, so it is left for its own PR.

## Tests

- **`ModificationLookupBaseTests`: 9 hand-built cases**, one per error, with `TestLookup` and fixed candidates:
  - whole-id matching;
  - the terminal entry preferred at N and C;
  - an exact residue kept among terminal entries (`Dimethyl on P`, N-terminal);
  - the `Anywhere.` fallback;
  - the opposite terminus never used;
  - a residue id that only contains the requested one;
  - a modification whose `LocationRestriction` is null (a derived type can leave it null);
  - the no-residue ambiguity.
- **`UnimodModificationLookupTests` (new): 34 cases against the real catalog**, covering 12 terminal accessions at N and C, the no-terminal-specificity fallback, the opposite-terminus null, and residue controls. Each case builds its own lookup, because results, including nulls, are cached per instance.

The first set of cases was run against unfixed master: 23 failed as described above. Two hand-built cases were added while fixing, and they pin the fix's own behaviour: the null restriction, and the no-residue ambiguity check. One real-catalog expectation was corrected: Unimod 214 has two N-terminal entries, and the existing isobaric filter keeps `iTRAQ-4plex on X`, the entry carrying diagnostic ions. Each part of the fix was then reverted on its own:

| reverted | red |
|---|---|
| whole-id match (back to substring) | 11 |
| terminal narrowing removed | 12 |
| narrowing keeps `Anywhere.` next to terminal entries | 9 |
| narrowing allows the opposite terminus | 2 |
| no-residue ambiguity check removed | 2 (including the existing `GlobalModificationLookupTests` case) |

```
SequenceConversion + ProForma + Koina + Modification tests   1300 passed, 0 failed   (rebased on 553576f4)
full offline suite (required.runsettings)                     6315 passed, 32 skipped, 4 failed   (on 2728c02c, before the rebase)
changed lines in ModificationLookupBase                       22 of 22 covered, all branches (Coverlet)
```

The 4 full-suite failures come from where I built, not from this change. `NuGetPackagingTests` (2) and `ParseComprehensiveVcfExamples` (2) search upward from the test output directory for `mzLib.nuspec` and a `.vcf` under `Test/`. My output directory was outside the repository, so they found neither.

## Not in scope

- **A residue position is not narrowed.** A residue-position UNIMOD:1 on M now resolves to `Acetyl on X` (N-terminal) where it used to give `Met->Hse on M`. The id is now right, but the entry is terminal-only. Excluding terminal entries at a residue would break the common ProForma spelling `Q[UNIMOD:28]PEPTIDE`, because Gln->pyro-Glu is Peptide N-terminal only in Unimod. That is a decision about ProForma input and deserves its own change.
- **The mass-only and name paths still apply the position only at the end.** They were not changed.
- `BuildCacheKey` does not include `UnimodId` or `MzLibId`, so two modifications with the same representation but different ids share a cache entry.

Related: #1039 adds `FdrBenchSequenceSerializer : UnimodSequenceSerializer` and terminal ground-truth cases. It reaches modifications through `MzLibId`, not `UnimodId`, and its terminal cases are X-motif with terminal restrictions, so it does not depend on this path and should be unaffected.


mzLib/Omics/SequenceConversion/BaseClasses/ModificationLookupBase.cs
```diff
@@ -210,8 +210,19 @@ protected virtual IEnumerable<Modification> GetPrimaryCandidates(CanonicalModifi
             var primaryList = primary?.ToList() ?? [];
             if (primaryList.Count > 0)
             {
+                // An identifier names the modification but not where it sits, so a terminal position
+                // is applied here: otherwise the residue preference below picks a side-chain entry
+                // (or one for the other terminus) over the terminal entry with the same identifier.
+                // Narrowing to nothing ends in ApplyCumulativeFilters, which returns null for no candidates.
+                primaryList = NarrowToTerminus(primaryList, context.Term);
```
nbollis, line 217: NarrowToTerminus prefers any explicitly N-/C-terminal candidates over all other candidates before SelectBestCandidate applies residue specificity. A known terminal residue can therefore discard the exact-residue Anywhere entry and select an X wildcard terminal entry instead. For example, Acetyl on X (N-terminal) plus Acetyl on K (Anywhere), queried for K at the N-terminus, retains only the terminal entry. Apply residue specificity before discarding Anywhere candidates, or keep the exact candidate in the comparison.

mzLib/Omics/SequenceConversion/BaseClasses/ModificationLookupBase.cs
```diff
@@ -210,8 +210,19 @@ protected virtual IEnumerable<Modification> GetPrimaryCandidates(CanonicalModifi
...
+
+                // With no residue given, an identifier whose entries sit on different residues is ambiguous.
+                // SelectBestCandidate would still pick one when they share a formula (Phospho on C, S, T, ...).
+                var residueUnknown = !context.TargetResidue.HasValue &&
+                                     primaryList.Select(m => m.Target?.Motif ?? "").Distinct().Count() > 1;
+
                 // We have primary candidates from specific identifiers - try to select best match
-                var uniqueFromPrimary = SelectBestCandidate(primaryList, context.TargetResidue);
```
nbollis, line 214: The residueUnknown check only prevents the direct SelectBestCandidate call. Control then falls through to ApplyCumulativeFilters; mass/formula filters can leave formula-equivalent candidates and SelectBestCandidate picks one specific motif. An identifier-only terminal query with no residue can return an arbitrary Phospho-on-residue entry when mass is present, while the same query without mass returns null. Carry residueUnknown through ApplyCumulativeFilters and reject the final result unless the remaining candidates have one distinct motif.

mzLib/Omics/SequenceConversion/BaseClasses/ModificationLookupBase.cs
```diff
@@ -210,8 +210,19 @@ protected virtual IEnumerable<Modification> GetPrimaryCandidates(CanonicalModifi
...
             {
+                // An identifier names the modification but not where it sits, so a terminal position
+                // is applied here: otherwise the residue preference below picks a side-chain entry
+                // (or one for the other terminus) over the terminal entry with the same identifier.
+                // Narrowing to nothing ends in ApplyCumulativeFilters, which returns null for no candidates.
+                primaryList = NarrowToTerminus(primaryList, context.Term);
+
+                // With no residue given, an identifier whose entries sit on different residues is ambiguous.
+                // SelectBestCandidate would still pick one when they share a formula (Phospho on C, S, T, ...).
+                var residueUnknown = !context.TargetResidue.HasValue &&
```
nbollis, line 221: The ambiguity check counts distinct Target motifs, so a wildcard X entry plus a residue-specific terminal entry is treated as ambiguous even though X is the catalog's explicit any-residue representation. UNIMOD:36 includes terminal entries such as Dimethyl on X and Dimethyl on P; a residue-less N-terminal lookup now returns null rather than the valid X entry. Treat X as the unambiguous fallback, or define and test the precedence explicitly.

mzLib/Omics/SequenceConversion/BaseClasses/ModificationLookupBase.cs
```diff
@@ -565,13 +599,32 @@ protected static bool IsMassRepresentation(string representation)
...
+    /// id is compared: a substring test would let UNIMOD:1 match 10, 21 and 127.
+    /// </summary>
+    private static bool IsUnimodId(string? value, int unimodId)
+    {
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            return false;
+        }
+
+        var id = value.Trim();
```
nbollis, line 621: IsUnimodId now requires the entire trimmed reference, after an optional UNIMOD: prefix, to be a decimal integer. DatabaseReference and Accession are free-form public-library inputs; values such as 'UNIMOD:35 (www.unimod.org)' or '35;36' previously matched by substring and now silently fail to resolve. Define and document the accepted grammar, or extract a properly delimited numeric id while avoiding collisions such as 1 matching 10 or 21.

nbollis (commented):
Read-only review of mzLib PR #1312. Four minor findings identified; no critical or major findings. No tests run.

trishorts:
Pushed `6fc7252` here in response to [nbollis's review on #1313](https://github.com/smith-chem-wisc/mzLib/pull/1313#discussion_r4041607328), since the code under discussion (`NarrowToTerminus`) lives in this PR and #1313 only carries it while stacked.

The objection was that the terminal narrowing does not work for RNA. It does: `Modification.ModificationLocation` accepts eight terminal `LocationRestriction` strings — the four protein/peptide forms plus `5'-terminal.`, `Oligo 5'-terminal.`, `3'-terminal.`, `Oligo 3'-terminal.` — and the substring test in `IsNTerminalRestriction`/`IsCTerminalRestriction` matches all eight. Nothing pinned it, so this commit adds twelve cases over the full set: prefer the terminal entry over a residue entry on each side, and never take the opposite terminus. Dropping the `5'`/`3'` arms reddens exactly the four oligo cases.

No production change. Lookup suite 89/89.

PR #1313, opened 2026-09-17
Read mzIdentML (.mzid, .mzid.gz) as Readers spectral matches

trishorts:
> **Project of origin:** `pride`

> **Stacked on #1312.** This branch contains #1312's commit, `Resolve terminal modifications by UNIMOD id to the terminal entry`, plus a merge of master. Review only the last commit here. Mod resolution depends on that fix, and the diff shrinks to this PR's own files once #1312 merges.

Readers can now read mzIdentML, plain (`.mzid`) or gzip-compressed (`.mzid.gz`, read straight from the stream), as `ISpectralMatch` records. The motivation is identification files downloaded from PRIDE Archive through `PrideArchiveClient` (#1088, #1093, #1121, #1187, #1199): for example, benchmarking a MetaMorpheus re-search against what the submitter reported. #1306 fixed the `MzidIdentifications` defects that real PRIDE files exposed, and this PR builds the reader on top of that.

```csharp
var file = new MzIdentMLResultFile(path);          // or FileReader.ReadResultFile(path)
foreach (var psm in file.Results.Where(r => r.Rank == 1 && r.PassThreshold))
    Use(psm.FileNameWithoutExtension, psm.OneBasedScanNumber, psm.FullSequence, psm.Accession, psm.QValue);
foreach (var skipped in file.SkippedMatches)       // what could not become a record, and why
    Log(skipped.SpectrumIdentificationItemId, skipped.Reason);
```

## What changes

**MzIdentML: `MzidIdentifications` gets two additions.** Nothing in mzLib or MetaMorpheus uses the class outside tests, so neither addition affects any caller.
- **`MzidIdentifications(Stream)`.** The existing version cascade tries each schema and re-opens its input for every attempt, which a `GZipStream` cannot do. The cascade moved **unchanged** into a private constructor that takes a stream opener:
  - the path constructor passes a `FileStream` opener;
  - the Stream constructor buffers the input once and passes a `MemoryStream` opener;
  - the caller owns the stream, and it is not disposed;
  - each arm, the legacy `.../1.1.0` namespace and the DTD rejection behave exactly as before, and the existing tests pin all of them.
- **`GetSpectrumMatches()`, plus `GetSpectraData()`.** These build indexes over `SequenceCollection` and `SpectraData` **once per call**, then yield every `SpectrumIdentificationItem` in **every** `SpectrumIdentificationList`, with its references resolved.
  - They return new version-neutral records, `MzidSpectrumMatch`, `MzidModification`, `MzidPeptideEvidence`, `MzidSpectraData` and `MzidUserParam`, with cvParams as `MzLibUtil.CvParam`.
  - They are written once per schema version, like the rest of the file; the four generated models share no types.
  - Reason: the existing accessors rescan `SequenceCollection` on every call. On a 1 GB Proteome Discoverer file from PRIDE that measured 178 ms per PSM, about 35 h for the file, while deserializing it took 9.5 s.
  - The old accessors keep reading `SpectrumIdentificationList[0]` only.

**Readers**
- **References:** `Readers.csproj` references `MzIdentML`. There is no cycle, and `mzLib.nuspec` already ships `MzIdentML.dll`.
- **File types:** `SupportedFileType.MzIdentML` (`.mzid`) and `SupportedFileType.MzIdentMLGz` (`.mzid.gz`) both map to `MzIdentMLResultFile`, whose `FileType` comes from the path, following the MsPathFinderT precedent.
  - `ParseFileType` gains a `.gz` arm, because `Path.GetExtension` sees only the last extension.
  - That arm throws `"Gz file type not supported"` for any other compressed file.
- **Software:** `Software.MzIdentML` names the format, as `MassSpecFile` does, since many engines write mzIdentML.
- **`MzIdentMLRecord : ISpectralMatch`:**
  - **`OneBasedScanNumber`** comes from the nativeID. `scan=N` gives N. `index=N`, a zero-based position in a peak list, gives N + 1, as `CasanovoMzTabFile` does, and the property documents that this is not an instrument scan. Anything else gives -1. `Mzml.nativeIdScanNumberParser` is reused for `scan=`.
  - **`FileNameWithoutExtension` / `SpectraFileLocation`** come from the SpectraData the result references. Proteome Discoverer's `scan=N file=K` names a different SpectraData by id (the RAW file) than the merged mzML its results reference, and the record reports the RAW file. Locations are split on either separator, because submitters write Windows paths.
  - **`AllModsOneIsNterminus`** comes from each modification's UNIMOD accession, resolved with `Omics.SequenceConversion.UnimodModificationLookup` from its position, the residue it sits on and its mass delta. mzIdentML location 0/1..n/n+1 maps to keys 1/2..n+1/n+2. **`FullSequence`** comes from `IBioPolymerWithSetMods.DetermineFullSequence`.
  - **`Accession`** is every protein the evidence names, de-duplicated and joined with `|`. **`IsDecoy`** is true only when every evidence entry is a decoy.
  - **Everything else in the item:** `Rank` and `PassThreshold`; `Scores` (the item's numeric cvParams and userParams, by name); `QValue` (MS:1002354 or one of its `is_a` children: MS-GF, MSPathFinder, ProSight); `ChargeState` and both m/z values; `SpectrumId`, `SpectrumTitle`; and the source `Match`.
- **`MzIdentMLResultFile.SkippedMatches`** lists the items that were not turned into records, and why:
  - crosslink items (MS:1002511);
  - a modification with no UNIMOD accession, or one the catalog does not resolve at that position;
  - a location outside the peptide, or two modifications at one position;
  - a substitution;
  - an unresolved peptide reference.

  Unreadable input throws `MzLibException` naming the file. `WriteResults` throws `NotSupportedException`.

## Decisions, and what was rejected

- **Stream constructor.** A second copy of the cascade for streams was rejected: two copies of the failure contract would drift apart. Picking the serializer by root namespace, without a buffer, was also rejected: it is a new dispatch path with different error messages. The buffered copy keeps the whole document in memory once, and the DTO graph costs more than that anyway (about 1.8x the document).
- **Every item is a record; there is no filter and no unified score.** mzIdentML files differ in what they keep. MS-GF+ has 9 of 12 items at rank > 1, Scaffold 3 of 3 below its threshold, Mascot 6 of 7. No two engines share a score term. Callers filter on `Rank`/`PassThreshold` and read `Scores`.
- **Skip and report, rather than throw or keep a partial record.**
  - One unknown modification should not make a whole PRIDE file unreadable.
  - A record with some of its modifications missing would carry a wrong `FullSequence`.
  - Crosslinks are left out rather than split into two linear records, which would misrepresent them.
- **`NotSupportedException` from `WriteResults`, not the house `NotImplementedException`.** Writing mzIdentML is out of scope, not unfinished. MetaMorpheus has its own writer. Copying the source file was rejected because it would silently ignore edits to the records.
- **Scan numbers are not parsed out of titles.** Mascot's title `16333: Scan 22014 (rt=…)` does carry the instrument scan. But every writer formats its titles differently, so `SpectrumTitle` is exposed for callers to parse, and `index=` stays a peak-list position.
- **Why #1312 is a prerequisite.** Before it, `UnimodModificationLookup` ignored the position when resolving by ID and matched IDs as substrings. #1312 measured an N-terminal acetyl on K resolving to `Acetyl on K`, and on M to `Met->Hse`. The reader passes the terminal residue, so without that fix Scaffold's N-terminal acetyl on `TAILER` would take the same path to the side-chain `Acetyl on T`.

## Tests

Offline, in the required job:
- **`TestMzIdentMLResultFile`** covers 12 mzid fixtures and one gzip copy. They span five PRIDE writers, MetaMorpheus and the HUPO-PSI examples, and schema 1.1.0 through 1.3.0.
  - For each file, record and skip counts equal the file's item count and crosslink count, both read out of the XML with a separate script rather than from the reader. `multiple_spectra_per_id_1_3_0` has three identification lists, which pins that all of them are read.
  - Per-writer checks:
    - **MS-GF+:** the Thermo nativeID, the child q-value, and a peptide shared between proteins.
    - **Proteome Discoverer:** `file=454` resolves to `1K_4D`.
    - **Mascot:** `index=16332` gives 16333, the title's query number.
    - **Scaffold:** a Windows location.
  - Resolved modifications: Phospho, Oxidation, Carbamidomethyl, Deamidated, N-terminal Acetyl with no `residues` attribute, and N-terminal pyro-Glu, each at the right key.
  - Synthetic documents cover every skip reason, the C-terminus (a methyl on a C-terminal K is the terminal entry, not `Methyl on K`), decoy and accession rules, score parsing, missing sections, the load-once guard, and the factory route.
- **`MzidIdentificationsTests`:**
  - the Stream constructor reads 5 fixtures exactly as the path constructor does, reads a non-seekable `GZipStream` and leaves it open, rejects null, and reads the legacy namespace;
  - `GetSpectrumMatches` and `GetSpectraData` are driven through **each** schema version's copy with a document that reaches every field, every unresolved reference, and every optional section.
- **`TestSupportedFileExtensions`:** rows for `.mzid`, `.mzID` and `.mzid.gz`, plus the new error message.

**Two new fixtures** are real PRIDE captures cut from the same downloads as #1306's Scaffold and Mascot Parser files. The rule is the first result carrying each modification kind; the source line endings are kept, and every kept line is checked against the download. `PXD078927_msgf_1_1_0.mzid.gz` is #1306's MS-GF+ fixture gzipped (marked `binary`). `PRIDE_MZID_PROVENANCE.md` records all of it.

**Live canary: `PrideMzIdentMLLiveTests`** (`[Category("ExternalService")]`, `[Category("Pride")]`).
- It pins one 8,563-byte file, `PXD000710/ma190_19_tandem_pproph.pep.mzid.gz` (2014), downloads it with `PrideArchiveClient`, and reads it from the `.gz`.
- It follows the skip-on-outage contract, so none of these can redden CI:
  - an outage or a stall (a 30 s `BodyStallTimeout`);
  - the overall 3-minute deadline;
  - an empty or non-gzip body;
  - a transfer that is not intact gzip;
  - manifest drift, reported as inconclusive.
- Only intact bytes that fail to read turn it red. Exact counts are asserted only while the file's SHA-256 still matches.
- It ran in **2 s**.

### Verification

```
reader + mzid + file-type tests    391 passed, 0 failed   (offline; the live canary excluded)
live canary                         passed (2 s)
changed lines, Coverlet             397 of 397 covered, every branch
full offline suite (required.runsettings)   6468 passed, 32 skipped, 4 failed
```

The 4 full-suite failures come from building outside the repository, not from this change. `NuGetPackagingTests` (2) and `ParseComprehensiveVcfExamples` (2) search upward from the test output directory for `mzLib.nuspec` and a `.vcf` under `Test/`, and my output directory was elsewhere. The same 4 failed the same way on #1312.

Each piece of logic was then broken on purpose and the suite re-run. Every mutant turned it red except one:

| mutant | red |
|---|---|
| `index=` not shifted to one-based | 5 |
| residue key = location | 4 |
| C-terminal key = length + 1 | 2 |
| N-terminal / C-terminal modification resolved as a residue | 1 / 1 |
| crosslinks, unresolved peptide, substitution, unresolved mod not skipped | 3 / 1 / 1 / 1 |
| duplicate location overwrites; location bounds off by one; negative location allowed | 1 / 1 / 1 |
| decoy when any evidence is a decoy | 1 |
| q-value only from the parent term; non-numeric q-value read as 0 | 1 / 1 |
| duplicate score name keeps the last; accessions not de-duplicated | 1 / 1 |
| PD `file=` not resolved; location not falling back to name | 1 / 1 |
| empty-file reload guard removed; parse errors not wrapped; `.gz` not decompressed | 1 / 3 / 3 (plus the live canary) |
| stream not buffered; only the first identification list read | 49 / 2 |
| no peptide via evidence (1.2.0); repeated id keeps the last entry | 1 / 1 |
| title ignores the obsolete accession; empty title not null | 1 / 1 |
| calculated m/z ignores its `Specified` flag | 2 |
| **file name keeps the directory** | **survives on Windows** |

The survivor is expected. `Path.GetFileName` already splits on `\` on Windows, which is where CI runs. The explicit split exists for Linux, where a submitter's `C:\…\run.raw` would otherwise come back as the whole path.

## Not in scope

- **Modifications named only by PSI-MOD, or by an MS term with no UNIMOD accession,** are skipped and reported. Every modification in the fixtures, and in the three PRIDE files scanned to pick them, carries a UNIMOD accession.
- **The accessors' `SpectrumIdentificationList[0]`-only behaviour and linear scans are unchanged.** The existing test pins the former.
- **mzTab and PRIDE XML are not read.** mzIdentML was the format the PRIDE survey pointed to (`design/IDFILES-GROUNDING.md` in the project notes).
- **Conflicts:** #1039, #1291 and #850 also append to `SupportedFileType` and its switches. Whichever merges second gets a small, mechanical conflict.

Touches existing types `MzidIdentifications`, `SupportedFileTypes`, `Software` and `Readers.csproj`.


mzLib/Omics/SequenceConversion/BaseClasses/ModificationLookupBase.cs
```diff
@@ -538,6 +549,29 @@ protected static bool MatchesTermRestriction(string? locationRestriction, Modifi
         };
     }
 
+    /// <summary>
+    /// For an N- or C-terminal position, keeps the candidates restricted to that terminus, or, when there
+    /// are none, the candidates allowed anywhere. Candidates for the opposite terminus are never kept.
+    /// A residue position, or no position, leaves the candidates unchanged.
+    /// </summary>
+    private List<Modification> NarrowToTerminus(List<Modification> candidates, ModificationPositionType? term)
+    {
```
nbollis, line 558: This does not work for RNA and does not cover all of our cases. I would use an extension method on the modification lookup type that checks the mod against the lookup type. Inspriation from DigestonProduct can be taken:

```    
private bool CanBeNTerminalOrFivePrime(Modification mod, int peptideLength)
       {
           return mod.LocationRestriction is "5'-terminal." or "Oligo 5'-terminal." or "N-terminal." or "Peptide N-terminal."
                  && ModificationLocalization.ModFits(mod, Parent.BaseSequence, 1, peptideLength, OneBasedStartResidue);
       }

       /// <summary>
       /// Determines if a modification can be applied to the C-terminal or 3' end of the peptide.
       /// </summary>
       /// <param name="mod">The modification to check.</param>
       /// <param name="peptideLength">The length of the peptide.</param>
       /// <returns>True if the modification can be applied to the C-terminal or 3' end; otherwise, false.</returns>
       private bool CanBeCTerminalOrThreePrime(Modification mod, int peptideLength)
       {
           return mod.LocationRestriction is "3'-terminal." or "Oligo 3'-terminal." or "C-terminal." or "Peptide C-terminal."
                  && ModificationLocalization.ModFits(mod, Parent.BaseSequence, peptideLength, peptideLength, OneBasedStartResidue + peptideLength - 1);
       }
```

nbollis (changes requested):

mzLib/Omics/SequenceConversion/BaseClasses/ModificationLookupBase.cs
trishorts: Thanks Nic — I checked this against the code rather than by eye, and RNA already works here. Details, plus what I did about it.

**It covers peptide *and* protein termini.** `IsNTerminalRestriction` is `Contains("N-terminal")`, which matches `"N-terminal."` (protein) and `"Peptide N-terminal."` (peptide); same on the C side. That was already exercised — `TryResolve_NTerminalUnimodId_PrefersTheTerminalModOverAResidueMatch` uses `"Peptide N-terminal."`, and `TryResolve_CTerminalUnimodId_…` puts `"N-terminal."` and `"Peptide C-terminal."` in the same lookup and requires the C-terminal one.

**It covers RNA too.** `Modification.ModificationLocation` accepts ten strings; the eight terminal ones are the four protein/peptide forms plus `5'-terminal.`, `Oligo 5'-terminal.`, `3'-terminal.`, `Oligo 3'-terminal.` — the same eight your `DigestionProduct` helpers enumerate. The substring test matches all of them (`Contains("5'-terminal")` catches the `Oligo` form as well). Nothing pinned that, though, which is a fair thing to object to, so `6fc7252` adds twelve cases over all eight strings: prefer the terminal entry over a residue entry on each side, and never take the opposite terminus. Removing the `5'`/`3'` arms from `IsNTerminalRestriction`/`IsCTerminalRestriction` reddens exactly the four oligo cases and leaves the other eight green, so the oligo support is load-bearing rather than incidental.

**`ModFits` can't be reached from this layer.** `ModFits(mod, sequence, digestionProductOneBasedIndex, digestionProductLength, bioPolymerOneBasedIndex)` needs the sequence and its length. `DigestionProduct` has `Parent.BaseSequence`; `ModificationLookupBase` resolves a `CanonicalModification`, which carries a position type, a residue index and a target residue and **no parent sequence or length at all**. So the localization half of your suggestion has nothing to pass. This is a catalog lookup — "which entry does this identifier name" — and localization stays where you have it.

**One thing worth flagging:** `IsNTerminalRestriction`, `IsCTerminalRestriction`, `MatchesTermRestriction` and `FilterByTerm` are all pre-existing on master — unchanged by this PR. `NarrowToTerminus` only calls them. So whatever the terminus vocabulary does or doesn't cover, this PR neither introduced nor changed it; it applies it on the identifier path, where the position was previously being ignored.

If what you're after is something different — a lookup that refuses to return an oligo mod to a peptide caller, i.e. keeping `5'-terminal.` and `N-terminal.` distinct rather than both mapping to `ModificationPositionType.NTerminus` — that's a change to the enum's contract (its own doc says N-terminus "or 5' end for nucleic acids") and reaches every caller, so I'd rather do it as its own PR than fold it in here. Say the word and I'll open one.

Tests: lookup suite 89/89, lookup + mzIdentML 365/365 on the merged branch.

nbollis (approved):

nbollis (approved):

nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve with nits** — correct, well-structured, and exceptionally well tested (mutation-tested). Reviewed against the real PR ref, not just the compacted diff.

**Summary:** Adds `MzIdentMLResultFile : ResultFile<MzIdentMLRecord>` reading mzIdentML (plain or gzip) into `ISpectralMatch` records via the Readers factory, with a buffer-once stream constructor on `MzidIdentifications` so the schema-version cascade can re-read a forward-only `GZipStream`. Unrepresentable items go to `SkippedMatches` rather than failing the file. Correctly wired into `SupportedFileType`/`Software`/`FileReader`.

**Findings (none blocking):**
1. *(minor, perf)* Plain `.mzid` is needlessly buffered whole into RAM — `LoadResults` always routes through the Stream constructor → `BufferOnce` copies the entire input to a `byte[]`. Buffering is only required for the non-seekable `GZipStream`; a plain `.mzid` is a seekable `FileStream`. On the 1 GB Proteome Discoverer file this PR targets, that adds a ~1 GB copy on top of the DTO graph. One-line fix: use the path constructor when the extension isn't `.gz`.
2. *(nit, maintainability)* `SpectrumMatches110/111/120/130` (~65 lines each) and `GetSpectraData` are copied 4× across generated schema types — consistent with existing per-version duplication, but a bug fixed in one arm must be hand-copied to three.
3. *(minor)* A modification with a null `Location` is skipped with the reason "modification location (none) is not on the peptide" — it's absent, not out of range; slightly misleading message.
4. *(minor, robustness)* `LoadResults` wraps only `InvalidOperationException`/`XmlException`/`InvalidDataException` into `MzLibException`; another type surfaced mid-read (e.g. `IOException`) would escape the "names the file" contract. No fixture hits it; low risk.
5. *(scope note)* The diff carries #1312's shared `ModificationLookupBase` change (exact Unimod-id test + `NarrowToTerminus`), which affects mod resolution well beyond mzid. It's the stacked prerequisite and is well covered, but mergers should be aware the reader's correctness depends on it and it touches other callers — merge ordering matters.

Verified correct: NRE safety via `NewCvParam` coalescing, `AllModsOneIsNterminus` key mapping with bounds + duplicate detection, scan-number parsing (`scan=` wins over `index=`), decoy/accession/QValue handling, gz-vs-plain stream disposal idempotency, and the lazy-load guard against reloading an all-skipped file.

PR #1314, opened 2026-09-17
Let ProForma terminal modifications past the Koina base-sequence check

trishorts:
> **Project of origin:** `proteoformSpectra`

`KoinaModelBase.TryCleanSequence` refuses every ProForma string with a terminal modification, before the converter ever sees it, and under `ReturnNull` it does so silently.

## Cause

The raw input is checked before parsing:

```csharp
var rawBase = BaseStripper.Replace(sequence, string.Empty);   // \[[^\]]+\]
if (!Regex.IsMatch(rawBase, AllowedAminoAcidPattern))          // ^[ACDEFGHIKLMNPQRSTVWY]+$
```

`BaseStripper` removes the bracketed modification but not ProForma's terminal separator. So `[UNIMOD:1]-MSKP…` becomes `-MSKP…`, which the pattern rejects. `HandleFailure` returns null, and no `SequenceWarning` is set, so a caller cannot tell this apart from a sequence that was never valid.

This affects every Koina model, because the check is in the base class, and it affects the commonest case: protein N-terminal acetylation. In one top-down project, 73 of 595 modified entries (12 %) were lost this way while the same strings predicted fine when sent to Koina directly. The server needs the terminal form, too: `[UNIMOD:1]-MSKP…` returns a full prediction, while `M[UNIMOD:1]SKP…` fails with `KeyError: 'M_1'`.

## Fix

A second regex, `TerminalModStripper` (`^(?:\[…\])+-|-(?:\[…\])+$`), removes terminal groups together with their separator before `BaseStripper` runs. Nothing else changes. The same pattern still checks the parsed base sequence, and a parsed base sequence never contains `-`.

A separator without a valid terminal group is still rejected before parsing: `-PEPTIDE`, `PEPTIDE-`, `PEP-[UNIMOD:1]TIDE`, and `[UNIMOD:1]-PEP*TIDE`.

## Tests

`KoinaModelBaseTests`:
- `TryCleanSequence_ProFormaTerminalModification_ReachesTheConverter`: N-terminal, C-terminal, both, and stacked N-terminal groups. All four cases **fail without the fix** and pass with it.
- `TryCleanSequence_TerminalSeparatorWithoutValidTerminalGroup_StillRejectedBeforeParsing`: the four malformed cases above never reach the parser.

`KoinaModelBaseTests` passes 19/19. All of `Test/KoinaTests` passes: 448 passed, 8 skipped, 0 failed.


mzLib/PredictionClients/Koina/AbstractClasses/KoinaModelBase.cs
```diff
@@ -12,6 +12,13 @@ public abstract class KoinaModelBase<TModelInput, TModelOutput>
 {
     protected static readonly Regex BaseStripper = new(@"\[[^\]]+\]", RegexOptions.Compiled);
 
+    /// <summary>
+    /// ProForma terminal modification groups, e.g. the "[UNIMOD:1]-" in "[UNIMOD:1]-MSKP" and the "-[UNIMOD:2]"
+    /// in "PEPTIDE-[UNIMOD:2]". They are removed together with their separator before the raw base-sequence
+    /// check, because <see cref="BaseStripper"/> removes only the brackets and would leave the "-" behind.
+    /// </summary>
+    protected static readonly Regex TerminalModStripper = new(@"^(?:\[[^\]]+\])+-|-(?:\[[^\]]+\])+$", RegexOptions.Compiled);
```
pcruzparri, line 20: **High** — The models only take mzLib sequences, so loosening this check can't make ProForma work

**Problem:** The input records take mzLib full sequences, and every Koina converter parses with `MzLibSequenceParser`. mzLib writes terminal mods as `[mod]PEPTIDE-[mod]`: no separator at the N-terminus and one group per terminus. This pattern follows ProForma instead (`-` at both ends, stacked groups), so the check now lets through strings the parser can't read, or reads wrong. With a real converter (`BuildAcceptAllConverter()`):

- Rejected one step later: `[UNIMOD:1]-PEPTIDE`, `[UNIMOD:1]-PEPTIDE-[UNIMOD:2]` and `[UNIMOD:1][UNIMOD:34]-PEPTIDE` return null with "C-terminal separator '-' at position 10 not followed by modification bracket." Even without a hyphen, `UNIMOD:N` isn't a name the mzLib lookup knows, so `PEPTIDE-[UNIMOD:2]` and `PEPM[UNIMOD:35]IDE` come back as unsupported modifications.
- Read wrong, with no warning: `PEPTIDE-[Unimod:Amidated on X][Unimod:Oxidation on E]` becomes `PEPTIDE[UNIMOD:35]-[UNIMOD:2]`, and `[Unimod:Acetyl on X]-[Unimod:Amidated on X]PEPTIDE` becomes `[UNIMOD:1]PEPTIDE-[UNIMOD:2]`. The check rejected both before.

The new tests show none of this because `FakeSequenceConverter` parses any string. The part that does work is mzLib's own C-terminal form: `PEPTIDE-[Unimod:Amidated on X]` used to fail the check as `PEPTIDE-` and now comes out as `PEPTIDE-[UNIMOD:2]`.

**Impact:** The `[UNIMOD:1]-MSKP…` entries from the description are still dropped, now with a warning that blames a C-terminal separator, and the tests report them as fixed. Two malformed shapes that used to be rejected are now predicted as a different peptide.

**Suggested fix:** I'd split this.

- Here: keep the C-terminal fix with a single group, `-\[[^\]]+\]$`, drop the N-terminal alternative, retitle, and test with a real converter.
- ProForma in its own PR:
  - Today's route: these models were built to take mzLib sequences, with other formats converted to mzLib before the input records are made. That already gets N-terminal acetyl through `TryCleanSequence`: `[Unimod:Acetyl on X]MSKP…` passes on models that allow UNIMOD:1 (those on `UnimodSequenceFormatSchema.Instance` write it without the hyphen).
  - Taking ProForma directly: the cleaner way is for the input to carry its own parser, and for the model to parse with it, paired with the model's own serializer. `ProFormaSequenceParser` paired that way already turns `[UNIMOD:1]-MSKPEPTIDE` into `[UNIMOD:1]-MSKPEPTIDE` on a model with the `-`/`-` schema.
  - Allowed mods: that change needs an explicit `AllowedUnimodIds` check on the parsed mods. The serializer only consults the allowed list for mods that arrive without a UNIMOD id, and ProForma's `UNIMOD:N` arrive with one, so with Prosit 2020 HCD's {4, 35}, `PEPS[UNIMOD:21]IDE` goes straight through.
  - The pre-check: a regex tied to one format shouldn't be the long-term validation, but it can't simply go. What it adds over the parsed check on line 211 is catching characters the parser skips: `PEP*TIDE` parses to `PEPTIDE` with only a warning, and `PEP TIDE` to `PEPTIDE` with none. The parser would need to reject those instead of skipping them first.

**Affected test scenarios:** `TryCleanSequence_ProFormaTerminalModification_ReachesTheConverter` passes only because parsing is faked; with `BuildAcceptAllConverter()` all four cases return null. It should use a real converter and cover `PEPTIDE-[Unimod:Amidated on X]` → `PEPTIDE-[UNIMOD:2]`. `TryCleanSequence_TerminalSeparatorWithoutValidTerminalGroup_StillRejectedBeforeParsing` should add the two misread strings above, and `[UNIMOD:1][UNIMOD:34]-PEPTIDE` moves there.

pcruzparri (commented):
Automated code review — 1 finding(s) at severity >= High, posted as inline comments.

mzLib/PredictionClients/Koina/AbstractClasses/KoinaModelBase.cs
trishorts: You're right, on every point. I reran your cases against `BuildAcceptAllConverter()` before answering and got your results back verbatim, down to the position in the error message:

| input | `apiSequence` | warning |
|---|---|---|
| `[UNIMOD:1]-PEPTIDE` | **null** | `C-terminal separator '-' at position 10 not followed by modification bracket.` |
| `[UNIMOD:1]-PEPTIDE-[UNIMOD:2]` | **null** | same |
| `[UNIMOD:1][UNIMOD:34]-PEPTIDE` | **null** | same, position 21 |
| `PEPTIDE-[UNIMOD:2]` | **null** | `unsupported modifications: UNIMOD:2 C-term` |
| `PEPTIDE-[Unimod:Amidated on X][Unimod:Oxidation on E]` | `PEPTIDE[UNIMOD:35]-[UNIMOD:2]` | **none** |
| `[Unimod:Acetyl on X]-[Unimod:Amidated on X]PEPTIDE` | `[UNIMOD:1]PEPTIDE-[UNIMOD:2]` | **none** |
| `PEPTIDE-[Unimod:Amidated on X]` | `PEPTIDE-[UNIMOD:2]` | none |

So the headline claim in my description was wrong: `[UNIMOD:1]-MSKP…` is still dropped, the failure just moved from the pre-check to the parser and arrives blaming a C-terminal separator. And the tests reported it fixed only because `FakeSequenceConverter.parse` ignores its argument. Trading two silently-misread shapes for a fix that isn't one is a bad trade, and I'd missed it.

I took your split.

**Here (b6a50db):** `TerminalModStripper` → `CTerminalModStripper`, `-\[[^\]]+\]$` — one group, C-terminus only, anchored. The N-terminal alternative is gone. Retitled to what it does. Both tests rebuilt on the real converter:

- `TryCleanSequence_MzLibCTerminalModification_SurvivesTheRawBaseSequenceCheck` asserts `PEPTIDE-[Unimod:Amidated on X]` → `PEPTIDE-[UNIMOD:2]` and `[Unimod:Acetyl on X]PEPTIDE-[Unimod:Amidated on X]` → `[UNIMOD:1]PEPTIDE-[UNIMOD:2]`.
- A third test pins your other point — that clearing the pre-check is not the same as being predicted: with `BuildConverter({35})`, `PEPTIDE-[Unimod:Amidated on X]` still returns null, but now with a warning that names the modification instead of being dropped with no warning at all.
- `…_RejectedBeforeParsing` gains `[UNIMOD:1]-PEPTIDE`, `[UNIMOD:1][UNIMOD:34]-PEPTIDE` and both misread strings, asserted never to reach the parser. That test keeps the fake converter on purpose, since "the parser was never called" is the assertion.

The three positive cases fail without the fix and pass with it; the eight rejection cases pass either way. `Test/KoinaTests`: 451 passed, 8 skipped, 0 failed.

The `<remarks>` on the regex now records why it stays narrow, with your two misread strings named, so the next person doesn't widen it for the same reason I did.

**ProForma in its own PR.** Your framing is the right one and better than what I was doing: the input carries its own parser, the model parses with it and serializes with its own, on a `-`/`-` schema — `UnimodSequenceFormatSchema` already takes `nTermSeparator` as a constructor argument, it's only `Instance` that's `""`. Plus the explicit `AllowedUnimodIds` check on the parsed mods; I hadn't noticed that the serializer only consults the allowed list for mods arriving without a UNIMOD id, so `PEPS[UNIMOD:21]IDE` goes straight through Prosit 2020 HCD's `{4, 35}` today. That's a separate bug and worth its own issue whatever happens to the ProForma work.

On the pre-check's long-term future: agreed it shouldn't be a format-specific regex, and agreed it can't just go. `PEP*TIDE` and `PEP TIDE` both return null from `TryCleanSequence` today only because this check catches them — the parser reaches `PEPTIDE` either way, with a warning for the first and nothing for the second. Making the parser reject unexpected characters instead of skipping them is a behaviour change in `SequenceParserBase` that reaches well past Koina, so I'd rather not fold it in here.

Worth stating for whoever picks up the ProForma PR: the motivating case is real on Koina's side. `[UNIMOD:1]-MSKP…` returns a full 174-ion prediction while `M[UNIMOD:1]SKP…` fails with `KeyError: 'M_1'` — the server keys terminal modifiers separately from residue modifications (3/3 on live calls). So the terminal form is required, not a preference. It just needs a route through mzLib that actually parses it.

Thanks for catching this — it would have shipped looking like a fix.

PR #1317, opened 2026-09-17
Opt-in: let a protease REQUIRE a modification at a subsite, and add the O-glycoproteases

trishorts:
> **Project of origin:** `cleavageRules`

> **Carries mzLib #1320.** While modelling a proline-aware StcE/trypsin double digest I found that
> `DigestionMotif.Fits` bounds-checked the wrong index, so every shipped `|P` protease made the exact
> cleavage it exists to forbid at sequence position 0 (`trypsin|P` digested `RPAAAAK` into `R` +
> `PAAAAK`). That is independent of glycans and is split out as **#1320** against `master`, with its own
> tests. The one-line change is also in this branch because the COMBO truth-set cases depend on it; it
> will drop out on rebase once #1320 lands.

> **DRAFT, and stacked.** Carries the commits of #1119 and #1290 (its parents, both also draft). Only the **top six** commits are its own: the truth set, slices 1-3, a correctness fix to where the glycan may come from, and slice 4 (the read-through). Un-draft and rebase to a single commit once its parents merge.

## The problem

mzLib cuts proteins into peptides and *then* decorates them with modifications, so no modification exists at the moment of cleavage. For trypsin that only loses you peptides. For an O-glycoprotease it is the whole rule: StcE does not cleave `S/T-X-S/T` — it cleaves it **only when the first S/T carries an α-O-GalNAc glycan**. An unglycosylated Ser/Thr is not a site at all.

`proteases.tsv` already admits this in a comment: the shipped StcE entries "OVER-DIGEST relative to the real enzyme -- their peptides are a superset of true StcE peptides", and OpeRATOR, IMPa and SmE were left out entirely because "modelling them needs glycosylation-aware digestion".

This is that mechanism. It is the mirror of #1119: that one drops a peptidoform the protease could not have produced **because of** a modification; this one drops a peptidoform it could not have produced **without** one.

## What lands

**`CleavageRequirement`** — a glycosylation class plus a Schechter–Berger **address**, `NonPrime(2, OLinked)` or `Prime(1, OLinked)`. An address rather than a flag because the required glycan is not in the same place for every enzyme:

| Enzyme | Motif | Required glycan |
|---|---|---|
| OpeRATOR/OgpA, IMPa, SmE, BT4244 | `X │ T/S*` | **P1′** |
| StcE | `T/S* X │ T/S` | **P2** (motif straddles the bond; P1′ is permitted, never required) |
| AM0627 | `T/S* │ T/S*` | both sides |
| ZmpC | `T/S* X X X │ X` | **P4** |

The subsite number is not capped at 1. Addressing uses `NonPrimeSubsite(n)`/`PrimeSubsite(n)` from #1290.

**Two stages, because one is not enough.** Dropping peptidoforms alone *annihilates* the digest: at `MaxMissedCleavages = 0` the read-through across an unjustified cut was never generated, so removing both fragments leaves nothing. So:

1. `DigestionAgent.FilterToFeasibleCleavageSites` removes impossible sites from the **site list** before enumeration. Spans, missed-cleavage counts and read-throughs then come out right **with no generation slack** — the read-through across a site that is not a site is just the ordinary peptide between the sites that remain.
2. `DigestionProduct.IsUnreachableWithoutRequiredModification` refines occupancy per peptidoform.

Each cut is answered by exactly one of the two products touching it — non-prime subsites live in the product *ending* at the cut, prime subsites in the one *starting* at it — so every internal cut is checked once, from the side that can see it.

**An optional, aliased `Cleavage Requirement` column**, read through `GetFieldValue`. A required column would break every custom `proteases.tsv` in the wild; a missing one yields no requirement and the file loads unchanged (there is a test using the old five-column format). Syntax reads as the literature writes a subsite: `P2:O-glycan`, `P1':O-glycan`. Malformed values **throw** — a silently dropped requirement would reinstate the over-digestion and look like a data problem rather than a typo.

**The enzymes.** StcE and StcE-trypsin gain `P2:O-glycan`; **OpeRATOR, IMPa and SmE are added** with `P1':O-glycan`. In a composite the requirement attaches only to motifs that have the named subsite, so StcE-trypsin carries it on its four StcE motifs and **not** on `K|` or `R|`.

`DigestionParams.RespectCleavagePromotingModifications`, **default false**, full-specificity only, concrete class only — `IDigestionParams` is untouched, so MetaMorpheus's two hand-written test doubles keep compiling.

## Evidence: a truth set, written first

`digestion-truth-set.tsv` is 43 hand-curated cases, each a **published substrate whose answer a paper states outright**, carrying its source DOI, an evidence grade, and whether the digest was observed or derived. It was committed **before** the implementation, and the numbers are the argument:

| | before | after |
|---|---|---|
| match | 6 | **25** |
| known gaps | 5 | 10 |
| enzyme not modelled | 29 | **5** |

Every remaining gap is a **named** missing capability, not an absence:

- **structure-level glycan** (STCE-02, OGPA-02/05/06) — the requirement resolves to a *class*, so it cannot tell α-O-GalNAc from β-O-GlcNAc, or Tn from core 1. Needs monosaccharide composition on `Modification`, which neither repo has.
- **FORBIDDEN-at-a-subsite** (IMPA-10, SME-02) — IMPa must not cleave *between* two adjacent glycosites. This models required-at-a-subsite only. Those two rows are the same substrate differing only by that rule.
- **P1 residue constraint** (IMPA-02) — a requirement names a modification, not "and P1 must not be Asp".
- **proline** (COMBO-01/02, TRYP-01) — unrelated to glycans; `StcE-trypsin` uses the proline-blind `K|,R|`. See #1186.

## Where the glycan may come from

Feasibility asks whether a satisfying modification could sit at the constrained residue, and it consults **both** sources: the modifications the database annotates on the protein, and the fixed or variable modifications the search configures. Consulting only the first was a real defect — a search supplying the O-glycan as a variable modification against an unannotated database, which is the ordinary way to do this, found no annotated glycosite, judged every site infeasible and returned the **undigested protein as its only product**. Feasibility from a configured modification is deliberately weak: one that fits anywhere makes every site feasible, so the filter stops discriminating and the per-peptidoform check does the work.

Relatedly, removing the initiator methionine is not a proteolytic cut, so a peptide starting at residue 2 of a Met-initiated sequence is no longer asked to justify its N-terminus with a glycan. It used to drop every initiator-cleaved form.

**There is deliberately no "nothing configured, go inert" escape**, and the absence is load-bearing. It looks like the natural mirror of #1119's `AnyConfiguredModificationCanBlockCleavage`, but the two corrections are not symmetric: an unconfigured *blocking* modification cannot remove a site the sequence really has, whereas an unsatisfiable *promoting* requirement means the protease genuinely has no site. Adding the gate makes a glycoprotease digest a bare protein and breaks the published unglycosylated controls (STCE-03, IMPA-12, OGPA-07). Returning the intact substrate is the enzymology. What it implies is a **precondition**, documented on the flag: a workflow that resolves the glycan *after* identification — a glyco search, where it arrives as a precursor-mass offset — must not set this flag, because digestion cannot see the glycan at all. Constraining that workflow means constraining localization, which is separate work.

## Slice 4: the read-through

Stage 2 used to remove a peptidoform whose cut was *feasible but unoccupied* and put nothing in its place. A variable glycan means some molecules carry it and some do not; the second population cannot be cut, so its intact peptide is a real product — and at `MaxMissedCleavages = 0` that span was never enumerated.

It now mirrors the blocking side: buy the span with generation slack, discount the sites that were never real, drop whatever is still over budget so the slack cannot leak.

**The slack is not `MaxMods`**, and that is the one place the mirror does not hold. A blocked site *carries* a modification and a peptidoform carries at most `MaxMods` of them, which bounds the blocking slack exactly. An unjustified site is the complement — a feasible site with **none** — and the modification budget says nothing about how many a peptide may span; a peptidoform with no glycan at all has every site it spans unjustified. The bound comes from the sequence instead: the most sites that can fall inside one peptide of at most `MaxPeptideLength` residues, from a sliding window over the site list. Exact, and cheap even when large, because the extra spans fail the length check before a peptide object is built.

```
StcE on RPPITQSSL, glycan at Thr5, MaxMissedCleavages = 0
  before   RPPIT*Q | SSL
  after    RPPIT*Q | SSL | RPPITQSSL      (read-through, reported missed cleavages = 0)
```

The glycosylated *intact* form still does not survive, which matters as much: with the glycan present StcE would have cut, so keeping it would make the correction a no-op that merely added peptides.

## Occupancy, and why the truth set grew a column

Slice 4 exposed that the corpus could not say what it meant. **STCE-01 and STCE-08 are the same substrate with the same glycosite and different published answers.** STCE-01 is a synthetic peptide, homogeneously glycosylated — there are no unglycosylated molecules and nothing to read through. STCE-08 deliberately models a mixed population. mzLib enumerates a localized modification both ways, which is right for a search and wrong for a synthetic substrate, so cases default to `occupancy = fixed` and compare only fully-occupied peptidoforms; STCE-08 declares `variable`. Those two rows are now the corpus's sharpest statement that occupancy, not sequence, decides the digest.

STCE-08's `known_gap` is cleared. Truth set: **26 match / 10 gaps** (was 25 / 11).

Also: a per-bond rule will over-generate on folded non-mucin glycoproteins. StcE's mucin selectivity is **not** a glycan-density rule — a single glycosylated motif on a free 9-mer is cleaved — but a consequence of conformation plus a separable binding module. Stated in the data file.

## Tests

11 end-to-end digestion tests, 17 parser/loader tests, and the 43-case truth set. **1439 pass across the digestion, protease, peptide, cleavage, truth-set, decoy and modification suites; 0 fail.** With the flag off, digestion is byte-for-byte what it was.

Naming, homes and conventions follow an `/oracle mzLib` pass against `smith/master @ 3755a0e8`.


trishorts:
### The known limitation now has an acceptance test, not just a paragraph

The missing read-through described under *Known limitation* was recorded in a code comment, this PR body and a commit message. None of those fails if it is ignored, so it is now **truth-set case STCE-08**:

```
STCE-08  StcE  GAP  RPPITQ|RPPITQSSL|SSL  ->  RPPITQ|SSL
```

A **variable** O-glycan at Thr5 of `RPPITQSSL` means the sample holds molecules that carry it and molecules that do not. StcE cuts the first population and cannot cut the second, so a correct digest holds three species — `RPPIT*Q`, `SSL`, and the intact `RPPITQSSL`. The intact form is missing today.

The site-list filter cannot catch this, and widening it would be the wrong fix: **the site is genuinely feasible, it is merely unoccupied in that peptidoform.**

It reports as a known gap rather than a failure, so CI stays green, but it appears in `TruthSet_Report` on every run. **Clearing its `known_gap` column is the definition of done** — and if it starts matching while still flagged, the test fails and says to clear it, so the fix cannot land unrecorded.

Fixing it means mirroring what #1119 does for blocking: generation slack in `ProteinDigestion.Digestion`, plus discounting those sites from the reported missed-cleavage count so the slack never leaks out. One thing does not carry over — #1119 bounds its slack with `MaxMods` because a peptidoform carries at most that many variable modifications, but here the unjustified sites are the ones **without** a modification, so `MaxMods` does not cap them. Deriving the right bound is the actual work.

PR #1318, opened 2026-09-18
Unify protein and RNA digestion through the shared DigestionAgent

nbollis:
## Motivation

Digestion previously duplicated the same search-mode logic across `Protease` (protein) and `Rnase` (RNA), owned a protein-only orchestrator (`ProteinDigestion`), and scattered named-agent tracking through `SpecificProtease` with no RNA equivalent. That made semi-specific and non-specific searches impossible to reason about, blocked RNA non-specific search support in MetaMorpheus, and forced downstream consumers to reach into protein-specific APIs.

## Summary of changes

- **`IDigestionParams`**: added `SpecificDigestionAgent`, the user-selected agent retained while an effective agent is resolved for the search mode.
- **`DigestionAgent`** (shared algorithm owner): centralized digestion algorithms now implemented once and shared by `Protease` and `Rnase`:
  - single-sided non-specific digestion;
  - top-down digestion;
  - full digestion;
  - speedy semi-specific seed digestion;
  - materialization hook `GetConcreteProducts` for domain-specific products.
- **`Protease`**: delegates every digestion mode to the shared `DigestionAgent`; protein-specific initiator-methionine behavior stays in `Protease`.
- **`Rnase`**: routes full, top-down, non-specific, and semi-specific seed digestion through the same shared algorithms; RNA termini and chemical formulas stay in `Rnase`.
- **`ProteinDigestion` removed**: `Protein.Digest` now calls `Protease.GetUnmodifiedPeptides` directly — one production digestion path.
- **Parameter fixes**: `RnaDigestionParams` keeps `SpecificRnase`, honors `SearchModeType` in clones, and maps agency selection for RNA termini.
- **Tests**: migration to the unified workflow plus regression coverage for RNA non-specific seeding and digestion-parameter cloning.
- **Docs**: added planning/integration notes.

## Testing notes

- Full mzLib solution build passes (20 projects).
- Focused digestion suites pass: semi-specific, search-mode routing, top-down, RNA digestion, and RNA non-specific seeding (311+ focused assertions).
- Existing public digestion APIs keep their behavioral contract; routing that used to be encoded in `ProteinDigestion` is now exercised through `Digestion`.

## Breaking changes

- Removal of `ProteinDigestion` class. Consumers should call `Protein.Digest` / `Protease.GetUnmodifiedPeptides` directly.
- `IDigestionParams` implementations must now provide `SpecificDigestionAgent`.


trishorts:
Reviewed this by hand at `fe1448d2` against `smith/master @ f6df7124` and `MetaMorpheus upstream/master @ 1760f4ee5`. **The refactor is sound** — two blockers, both already red in CI, both small fixes.

### What I checked and found correct

A unify-two-implementations PR is where silent behaviour changes hide, so I went looking for them specifically and didn't find any:

- **The `Retain`/`Cleave` → boolean hoist is behaviourally equivalent.** `Protease.cs:43-44` computes `retainMethionine` and `cleaveMethionine`; against the definitions still at `:88`/`:102`, `(i != 0 || retainFirstResidue)` is exactly `Retain(i, …)` and `(i == 0 && cleaveFirstResidue)` is exactly `Cleave(i, …)`.
- **Things moved, nothing was lost.** `SpeedySemiSpecificDigestion` → `DigestionAgent.cs:356`, now genuinely shared by `Protease.cs:67` *and* `Rnase.cs:50`; `TopDownDigestion` → `DigestionAgent.cs:195`, likewise; `SemiProteolyticDigestion` correctly *stays* protein-side at `Protease.cs:119`.
- **`WantsSemiSpecificSeeds` is dissolved, not dropped.** I nearly flagged this one. It was `fixedTerminusRequested || !proteaseCanBeSemiSpecific`; the `Semi` arm is only reachable when the protease's own `CleavageSpecificity` is `Semi`, so the second term is necessarily false there and the predicate reduces to the `when fragmentationTerminus is N or C` guard at `Protease.cs:66`. The other term is subsumed by the `None`/`SingleN`/`SingleC` arms. Equivalent.

### Blocker 1 — `IDigestionParams.SpecificDigestionAgent` needs a default implementation

`IDigestionParams.cs:20` declares it with no body. CI packs mzLib as 9.9.9, repoints **`Test/Test.csproj`**, and builds MetaMorpheus master, which has two hand-written doubles on the old surface. From the failing `integration` job of run 35377248012:

```
MetaMorpheus/Test/DigestionAgentNameTests.cs(128,50): error CS0535:
  'AgentlessDigestionParams' does not implement interface member 'IDigestionParams.SpecificDigestionAgent'
MetaMorpheus/Test/ParameterTest.cs(402,36): error CS0535:
  'BadDigestionParams' does not implement interface member 'IDigestionParams.SpecificDigestionAgent'
```

Suggested fix — a default interface member, so no existing implementor has to change:

```csharp
DigestionAgent SpecificDigestionAgent => DigestionAgent;
```

That default is also semantically right by your own doc comment: the two are the same object except when `SearchModeType` is `None`, and a double that never sets a search mode is exactly that case. `DigestionParams` and `RnaDigestionParams` already override it, so nothing in mzLib changes. A companion MetaMorpheus PR adding the member to both doubles would also work, but it couples the repos' merge order and leaves any out-of-tree implementor broken.

### Blocker 2 — RNA non-specific 5′ seeds are silently 3′ seeds

`RnaDigestionParams.cs:29-33`:

```csharp
Rnase = FragmentationTerminus == FragmentationTerminus.N ?   // protein vocabulary
   RnaseDictionary.Dictionary["singleN"] :
   RnaseDictionary.Dictionary["singleC"];
```

`FivePrime` and `ThreePrime` are **distinct** enum members, not aliases of `N`/`C`, so an RNA caller asking for 5′ never matches, falls to the `else`, and gets `singleC` — the 5′ half of the feature this PR exists to unblock. Your own new test catches it, and it's the single failure in the `build` job:

```
Failed NonSpecificParametersSelectEffectiveRnaseAndRetainSpecificRnase
  Expected: "singleN"  But was: "singleC"    (TestRnaseNonSpecificDigestion.cs:39)
```

`Rnase.cs:49` already gets this right a few files away:

```csharp
Rnase = FragmentationTerminus is FragmentationTerminus.FivePrime or FragmentationTerminus.N
    ? RnaseDictionary.Dictionary["singleN"]
    : RnaseDictionary.Dictionary["singleC"];
```

I grepped `DigestionAgent.cs`, `Rnase.cs`, `NucleicAcid.cs` and `RnaDigestionParams.cs` for the same protein-idiom-in-RNA-code slip; this is the only remaining instance.

### Minor — the description mentions a file that isn't in the diff

The summary lists **`DigestionParamsBase` (new)** holding shared agent-selection state and effective-agent resolution for both domains. There's no such file at `fe1448d2`, and both params classes still carry that state independently — which is *why* Blocker 2 can exist in one and not the other. Either it was dropped from the commit or the description is from an earlier plan. Not asking you to add it, but if it ever lands it would structurally prevent Blocker 2.

### Nit (unproven — flagging, not asserting)

`Protease.cs:43-44` evaluates `protein[0]` before the switch, where the old `GetUnmodifiedPeptides` was a pure dispatcher and each arm computed it internally. It's purposeful (`:45` needs it for `initialStartResidue`), but it means every arm now indexes residue 0, and `Protein` has no empty-sequence guard. **I could not show a zero-length protein is reachable from any real database path, and no test constructs one**, so this may be unreachable — mentioning it only because a guard is one line.

---

Separately, and selfishly: `SpecificDigestionAgent` on `IDigestionParams` is something I'd been blocked on for modification-aware digestion work, so thanks for closing it. Happy to take Blocker 1's default-member change as a PR against your branch if that's easier.

nbollis:
Thanks for the review. I have fixed the second issue with silent failure to set type properly. 

For the first blocking issue, I think we can resolve this simply by adding the new properties to the testing classes in MetaMorpheus. We do not need a base class, and a base class would still require a MetaMorpheus change to make integration tests pass.

trishorts (approved):

Alexander-Sol (approved):

PR #1319, opened 2026-09-18
Add RNA mode to FlashLFQ

Alexander-Sol:
## Summary

Makes FlashLFQ able to quantify RNA / oligonucleotides in addition to peptides.

- **`RnaMode` flag** on `FlashLfqParameters` (and threaded through the `FlashLfqEngine` convenience constructor as a trailing `rnaMode = false` parameter, so existing positional callers are unaffected).
- **RNA-aware theoretical isotope distributions.** When `RnaMode` is set, `CalculateTheoreticalIsotopeDistributions` builds the chemical formula from `Transcriptomics.RNA` (with a nucleotide-validity check) and uses a ribonucleotide averagine for the fallback, instead of the amino-acid path.
- **Shared averagine models.** `Averagine` and `OxyriboAveragine` now expose their per-residue average composition via `GetAverageChemicalFormula()` (a `Dictionary<char, double>`). FlashLFQ derives its averagine formula from that rather than hardcoding the composition inline. Peptide (non-RNA) numbers are unchanged — the amino-acid composition and scaling match the previous constants exactly.
- **Positive intensity for negative charge.** `IsotopicEnvelope` now divides the summed intensity by `Math.Abs(chargeState)`, so RNA's negative charge states (negative-mode ionization) yield positive intensities.

## Why

RNA ionizes in negative mode and is built from ribonucleotides, so the peptide-only assumptions (amino-acid averagine, amino-acid sequence→formula, and dividing intensity by a positive charge) produced wrong formulas and negative intensities. The `ToMz`/`ToMass` helpers already handle negative charge, so the isotopic-envelope intensity was the only remaining sign bug.

## Tests

New tests (all passing):
- `TestRnaModeTheoreticalIsotopeDistribution` — RNA-mode distribution is built from the ribonucleotide formula with the correct peak-finding mass, using a negative charge state.
- `TestRnaAveraginePathForUnparsableSequence` — the ribonucleotide averagine fallback handles a non-nucleotide sequence without throwing.
- `TestIsotopicEnvelopeNegativeChargeYieldsPositiveIntensity` — negative charge yields positive intensity.
- `TestAveragineComposition` — `Averagine`/`OxyriboAveragine` composition getters return the expected values and an independent copy.

The `Averagine`/`OxyriboAveragine` refactor is behavior-preserving: the existing deconvolution suites (including the full `DecoyAveragine` tests, which exercise the regenerated averagine tables) and the FlashLFQ suite pass unchanged.


mzLib/Test/FlashLFQ/TestFlashLFQ.cs
```diff
@@ -47,6 +47,69 @@ public static void TearDown()
             if (File.Exists(pepModelArtifact)) File.Delete(pepModelArtifact);
         }
 
+        [Test]
+        public static void TestRnaModeTheoreticalIsotopeDistribution()
```
trishorts, line 51: Non-blocking: these tests cover the theoretical-distribution step and the envelope constructor separately, but nothing sends a negative-charge RNA ID through `Run()` (the MS1 XIC, then `GetIsotopicEnvelopes`, then a quantified peak). A small end-to-end test with a synthetic negative-mode spectrum would lock in the claim that FlashLFQ can quantify RNA. It could be a follow-up.

mzLib/MassSpectrometry/Deconvolution/AverageResidue/Averagine.cs
```diff
@@ -14,29 +15,43 @@ public sealed class Averagine : AverageResidue
...
-        double averageO = 1.4773;
-        double averageN = 1.3577;
-        double averageS = 0.0417;
 
+    /// <summary>
+    /// Returns the average elemental composition of a single averagine unit as a map from element
+    /// symbol to average atom count (e.g. {'C', 4.9384}). Callers can scale this by a mass to derive
+    /// an approximate chemical formula for an arbitrary species.
+    /// </summary>
+    public Dictionary<char, double> GetAverageChemicalFormula() => new(AverageComposition);
```
trishorts, line 42: Non-blocking: `GetAverageChemicalFormula()` is defined separately on `Averagine` and `OxyriboAveragine`. If it were an abstract member on `AverageResidue`, FlashLFQ could hold an `AverageResidue` and drop the `RnaMode ? new OxyriboAveragine() : new Averagine()` branch. Any future averagine would also get the accessor automatically.

trishorts (dismissed):
Looks good. I built the branch and ran the FlashLFQ + Deconvolution suites locally: 758 passed, 0 failed. The averagine refactor keeps the old behavior (same elements, same order, same rounding). I checked the negative-charge path through the engine: `ToMz`/`ToMass` already use `Math.Abs`/`Math.Sign`, the `_chargeStates` range works for all-negative charges, and `IsotopicEnvelope` was the only place that divided by the charge. The two inline notes are non-blocking suggestions.

nbollis (approved):

RayMSMS (approved):

PR #1321, opened 2026-09-18
Sequence-variant corpus: range-encoded insertions and an insertion that slides a cut site

trishorts:
> **Project of origin:** `variants`

## Insertion nodes deferred from #1308: range encoding and a sliding cut site

@acesnik: three corpus nodes that were left over from #1308. **Test-only**, and they pass on master as it stands. As before, each `ExpectedForms` oracle is hand-derived from the domain model, **not captured from running code**.

### New nodes

| Node | What it pins |
|------|--------------|
| **I10** | Insertion encoded as a **BEGIN/END range** (`TI→TAGI` over [4,5]). Every earlier insertion used a single position, so begin == end. Here the variant is longer than its range. The phospho on the kept **left** anchor T4 stays at 4 (same forms as I01). |
| **I11** | Range insertion with the phospho on the kept **right** anchor (`PT→PAGT` over [3,4]). The modified residue is the range's END, which is what the kept-suffix branch counts back from. T and its phospho shift 4→6 (same forms as I00/I09). |
| **I12** | Bottom-up: an insertion upstream of an existing K **slides** the trypsin cut site. `E2→EAG` moves K from 4 to 6: consensus `PEPKIDE` → `PEPK` + `IDE`; variant `PEAGPKIDE` → `PEAGPK` + `IDE`. The site is neither created (I07/S08) nor destroyed (S07); it only moves. The unchanged downstream `IDE` appears once per proteoform because the harness digest is a per-proteoform union. |

### Not included: a mismatched range
`ApplySingleVariant` computes the replaced span from `OriginalSequence.Length`, while `OneBasedEndPosition` is used for pruning and naming. For a well-formed variant (I10/I11) the two agree. For a malformed one they don't: `<original>T</original><variation>TAG</variation>` declared over [4,5] is applied as `T4→TAG` today, and END=5 is ignored. Whether that input should be pruned as malformed or trusted by its original is a policy question, so no node pins it here yet.

### Result
- Corpus: 96/96 (48 nodes × `Corpus_Node` + `Corpus_Node_RoundTrip`)
- Local gate (`DatabaseTests|Transcriptomics|Variant|TestProteinProperties`): 1213/1213
- Branch is current with `master` (f6df7124, #1308)


nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve** — three correct, well-documented test nodes that add real coverage; no production changes.

**Summary:** Test-only PR adding three nodes to the sequence-variant corpus in `VariantCorpusTests.cs`: **I10** and **I11** encode an insertion as a multi-residue BEGIN/END range (original longer than one residue) with a PTM on the kept left (I10) / kept right (I11) anchor; **I12** is a bottom-up trypsin case where an upstream insertion slides an existing K cut site from position 4 to 6.

**Verification:** Each oracle was hand-traced through `ApplySingleVariant` / `AdjustModificationIndices` / `CountKeptFlanks` and trypsin digestion — all three are correct:
- **I10** (TI→TAGI over [4,5]): phospho at k=4 stays at 4 → `PEPTAGIDE`. ✓
- **I11** (PT→PAGT over [3,4]): phospho shifts +2 to position 6 → the T in `PEPAGTIDE`. ✓
- **I12** (E2→EAG, trypsin): K moves 4→6, next residue is I in both proteoforms → both cleave; per-proteoform union with no dedup correctly yields `IDE` twice. ✓

**Notes (nits):** I10/I11 reuse existing `ExpectedForms` but reach them through a genuinely distinct path (the 6-arg range constructor + a 2-residue `OriginalSequence`, exercising the kept-suffix branch the single-position nodes don't) — legitimate added coverage. The one latent production inconsistency the nodes skirt (`OriginalSequence.Length` vs `OneBasedEndPosition` for a malformed range) is explicitly and reasonably deferred by the PR. Style matches house conventions; `AssertCaseSelfConsistent` guards ExpectedCount/ExpectedForms drift.

PR #1324, opened 2026-09-19
Write an SDRF atomically, so a crash never leaves a truncated file

trishorts:
## What changes

**SdrfDocument.WriteResults now writes atomically.** A crash, kill or full disk partway through a write no longer leaves a truncated SDRF where the complete one should be.

- **Before:** the target was opened with `File.Create`, which truncates it in place. The existing validate-first and materialise-first ordering protected a file from a write that was *refused*, but nothing protected it from one that *failed*. A pipeline that treats "the file exists" as "the run is done" would accept the truncated file.
- **Now:** the document is written to a uniquely named sibling `<target>.<guid>.partial` and moved over the target with `File.Move(overwrite: true)`. That is a rename on one volume. The temp file is removed in `finally`, and cleanup never replaces the original exception.
- **Same pattern as** `ProteinDbRetriever` (the unique name means two writers of one path cannot collide on the temp file). `MslWriter` and `PrideArchiveClient` do the same inline.

## What does not change

Overwrite rather than throw on an existing file, no BOM, byte-identical round trips, and in-place writes (`WriteResults(FilePath)`). All the existing `TestSdrf` write tests pass unchanged.

## Tests

`TestSdrf`:
- `Write_OverAnExistingFile_ReplacesItWhole_AndLeavesNoTempFile`
- `Write_ThatFailsAtTheFinalMove_Throws_AndLeavesNoTempFile`: the target path is a folder, so the move fails; no `.partial` file survives, and the folder is untouched.

120/120 `TestSdrf*` pass locally.


nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve with nits** — the atomic-write pattern is correct, matches the repo's established convention, and is well tested for its threat model.

**Summary:** `SdrfDocument.WriteResults` previously wrote directly via `File.Create` (truncate-in-place), so a mid-write crash left a truncated SDRF that a "file exists = done" pipeline would accept. The PR writes to a unique sibling `<target>.<guid>.partial`, then promotes it with `File.Move(overwrite: true)` (a same-volume rename), cleaning up the temp file in a `finally` that swallows only `IOException`/`UnauthorizedAccessException`. Mirrors the existing `ProteinDbRetriever`/`MslWriter` patterns.

**Findings (all nits):**
1. Atomic *visibility*, not durability — there's no `Flush(true)`/`FlushFileBuffers` before the move, so on a true power-loss crash NTFS can commit the rename while the temp file's data blocks are still cached, yielding a present-but-truncated target after reboot. This fully covers the real threat (process crash/kill/full disk) but the title slightly overclaims for power loss. Consistent with every other writer in the repo (none fsync).
2. `File.Move` doesn't preserve the replaced file's ACLs/attributes (the promoted file inherits the temp file's); `File.Replace` would. Irrelevant for SDRF outputs and consistent with the sibling patterns.
3. Cleanup catch list matches `ProteinDbRetriever` exactly — fine.

Verified correct: overwrite handled (`overwrite: true`), byte round-trip preserved (`\r\n` + trailing newline), cross-volume is a non-issue (temp is always in the target's own directory), the writer is disposed before the move, and no temp file is left on either the success path or the validation-failure path. Two well-targeted tests (overwrite-replaces-whole, failed-final-move-cleanup).

**CI note:** the failing `integration` check is unrelated — a pre-existing downstream MetaMorpheus build break (`IDigestionParams.SpecificDigestionAgent` not implemented, CS0535), which #1330 fixes. `build`, `codecov`, `external-service-tests` all pass.

PR #1325, opened 2026-09-19
Decide per SDRF whether its sample half says anything

trishorts:
## What changes

**New: `SdrfSampleInformativeness.Assess(SdrfDocument)`, which decides for one SDRF whether its sample half says anything.**

A spec-valid SDRF is not the same as a useful one. Take a file generated from the data-file list: every sample cell a reserved word, one biological-replicate number throughout, and no factor column. It passes `SdrfValidator` and `SdrfDriftLint`. For building an experimental design, it is the same as having no SDRF.

`SdrfCoverage` can't make this call for one file:
- it does not know sample columns from assay columns, and assay columns are always filled, so a skeleton still looks half healthy;
- its `IsUninformative` flags any single-valued column, which inside one file means every organism column.

## The checks

Each check runs over the counts `SdrfCoverage.Measure` already produces, so the reserved-word rule and the once-per-row counting of repeated columns are the same:
1. **Factor value varies:** some `factor value[...]` column has two or more distinct real values. Every factor column is checked, not only the first.
2. **Sample described:** some `characteristics[...]` column other than organism and biological replicate holds a real value. Organism is excluded because a search can fill it without a human.
3. **Biological replicate varies.**

`SdrfSampleAssessment` returns the per-column counts behind each check, plus a `Verdict`:
- **Informative:** all three pass.
- **Skeleton:** none pass.
- **Partial:** anything else. A single-condition study is legitimately Partial, and whether that is good enough is the caller's decision.

## On the curated corpus

Over bigbio/sdrf-annotated-datasets @ `4f823dcd` (1,236 files): **300 Informative, 742 Partial, 194 Skeleton.** 193 of the 194 skeletons have no `sdrf annotation tool` column, so the tool name can't stand in for this check.

## Tests

`TestSdrfSampleInformativeness`: six unit tests, plus an `[Explicit]` corpus report that uses `MZLIB_SDRF_CORPUS` like the other corpus reports.

Public per D19. The named caller is a reanalysis pipeline that gates input SDRFs on this verdict through pyMzLib.


nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve with nits** — logic is correct for its stated design and cleanly reuses the existing, well-tested coverage counting.

**Summary:** Adds `SdrfSampleInformativeness.Assess(SdrfDocument)` (plus `SdrfSampleAssessment` and `SdrfSampleVerdict`). It reuses `SdrfCoverage.Measure` over a one-document collection, partitions per-column coverage into factor-value / sample-characteristic (excluding organism & biological replicate) / biological-replicate columns, and derives three booleans → a three-way verdict (Informative / Partial / Skeleton). Read-only, additive.

**Findings (all minor/nit):**
1. *(doc)* `Informative` requires biological-replicate variation AND a filled non-organism characteristic AND a varying factor. A legitimate n=1-per-condition study, or one that encodes condition only in `factor value[...]` without a mirrored `characteristics[...]` column, grades `Partial`. Not a bug (`Partial` defers to the caller), but the "can drive a design" enum doc overstates what `Informative` certifies — worth softening.
2. *(doc)* The class summary says "never throws on a bad document," three lines above an `ArgumentNullException` guard. Reword to "never throws on a *malformed* document."
3. *(test gap)* Two Partial branches are untested: (a) replicated-but-single-condition-and-described, and (b) a factor cell holding a *reserved word* failing to count as a distinct level — the load-bearing behavior, currently tested only on the characteristic path, not the factor path. Both are cheap to add.
4. *(nit)* Redundant explicit `LoadResults()` in `CorpusVerdictReport` (`Measure` lazy-loads anyway).

Verified correct: case-insensitive `factor value[`/`characteristics[` prefix matching; organism/bio-replicate exclusion; reserved-word + trim/case-insensitive counting inherited from `SdrfCoverage`; null-safe throughout (short/ragged rows, empty headers, empty column lists → `Skeleton`).

PR #1326, opened 2026-09-19
Read characteristics[age] cells into years, and refuse what needs a guess

trishorts:
## What changes

**New: `SdrfAge.TryParse(cell, out SdrfAge age)`, which reads a `characteristics[age]` cell into years.** Nothing in mzLib read SDRF ages before. It never throws.

`SdrfAge` carries:
- `Years`: the value, or a range's midpoint, or a bound;
- `MinYears` / `MaxYears`;
- two independent confidence facts:
  - `Precision` (`Exact`, `Range`, `LowerBound`, `UpperBound`): how much of the age the cell pins down;
  - `FollowsSpecification`: whether the cell uses the spec's grammar (`58Y`, `30Y6M`, `16W`, `40Y-85Y`, `>=90Y`) or words whose unit is still unambiguous (`3 year`, `6-8 weeks`, `4 hour`).

**What it refuses matters most.** A bare number has no unit, and the cell gives no way to recover one: `63` could be years or days in the same study. So `TryParse` returns false rather than assume years. Reserved words, templates filled with one (`not availableY`), free text, and out-of-order grammar (`6M30Y`) are refused the same way.

## On the curated corpus

Of 53,655 real `characteristics[age]` cells in bigbio/sdrf-annotated-datasets @ `4f823dcd`:

| Read as | Cells |
|---|---|
| exact, spec grammar | 44,321 |
| exact, words | 2,005 |
| range (spec / words) | 549 / 117 |
| lower bound (almost all the `>=90Y` de-identification cap) | 757 |
| **refused** | **5,906 (11.0%)** |

## Tests

`TestSdrfAge`: 35 cases, covering spec grammar, word forms, ranges, bounds and refusals, plus an `[Explicit]` corpus report that uses `MZLIB_SDRF_CORPUS`.

Public per D19. The named caller is a reanalysis pipeline that places datasets on an age axis through pyMzLib. Years only; developmental-stage binning waits until someone needs it.


nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve with nits** — does exactly what the title claims, correct against its contract, never throws, well tested for accepted and refused forms.

**Summary:** Adds standalone `SdrfAge.TryParse(cell, out SdrfAge)` parsing an SDRF `characteristics[age]` cell into a `years` value plus min/max bounds, a `Precision` enum (Exact/Range/LowerBound/UpperBound) and a `FollowsSpecification` flag. Central design: *refuse* (return false, never throw) anything ambiguous — bare numbers, reserved words, free text, out-of-order grammar — rather than guess a unit.

**Findings (none blocking):**
1. *(minor)* `FollowsSpecification=true` for ranges and bounds likely overstates SDRF-spec conformance: the official age encoding is the point form `nYnMnD`/`nW`, while hyphenated ranges and `>=`/`<` bounds aren't clearly part of that grammar. A caller filtering on `FollowsSpecification` to get spec-clean point ages still receives ranges/bounds. Consider documenting it as "unit written in spec tokens" rather than "cell is spec-conformant."
2. *(nit)* A degenerate `40Y-40Y` is classified `Precision.Range`, not `Exact` (strict `min > max` check). Values are correct, but a caller distinguishing point ages from cohorts by `Precision` would see a point age mislabeled.
3. *(nit)* Unicode inconsistency: the bound regex accepts `≥`/`≤`, but range parsing keys only on the ASCII hyphen — so an en-dash `40Y–85Y`, "40 to 85 years", `"3 year-old"`, and ISO-8601 `P40Y` are all silently refused. Consistent with the refuse-rather-than-guess philosophy, but these are plausible real-corpus forms (recall gap, not a bug).
4. *(nit)* Unit is inferred from a word's first letter (`word[0]`) — safe today, but silently mis-scales if the word list ever grows to share first letters (e.g. minutes/months). A lookup on the matched word would be more robust.
5. *(nit)* No sanity cap on absurdly large numeric input (can't throw, but is accepted). Test gaps: `FollowsSpecification` not asserted on bounds, no words-form bound, no degenerate-range case.

Verified correct: whole-string coverage rejects `6M30Y`/`30Y30Y`/`-5Y`/`1e3Y`/`1,000Y`; culture pinned to `InvariantCulture` everywhere (comma-decimal locales safe); `min>max` refused; `ToYears` covers every producible unit so the "never throws" promise holds.

**CI note:** the red `integration` check is an unrelated downstream MetaMorpheus build break (`IDigestionParams.SpecificDigestionAgent`, CS0535, fixed by #1330); `build`, `codecov`, `external-service-tests` pass.

PR #1327, opened 2026-09-19
Name the file a search read beside the file that was acquired

trishorts:
## What changes

**New: `SdrfAssay.SearchedDataFileName`, written to `comment[searched data file]` right after `comment[data file]`.**

A search often reads a derivative of the file that was acquired. MetaMorpheus, for example, calibrates `X.raw` into `X-calib.mzML` and searches that. The SDRF has to say both:
- `comment[data file]` is the **acquired** file. That is the spec's definition (REQUIRED, "Name of the raw file"; MS:1003083), all 5,157 values in the curated corpus are `.raw`, and it is the join key back to the deposition and to other reanalyses of the same data.
- The results **cannot be regenerated from that file** without re-running the transformation, so naming it alone hides a step. `comment[searched data file]` names what the search actually read.

**Behaviour**
- A row whose search read its acquired file names that file again in the new column. It does not leave the cell blank or write a reserved word, so the column always says whether a transformation happened.
- The column appears only when some row sets it, so existing callers get byte-identical documents.

## Not a specification term

The spec has nothing for this:
- `comment[associated data file]` means a sidecar needed to *read* the primary file, such as `.wiff.scan`.
- `comment[processed data file]` exists only in the metabolomics template.

Borrowing either would mislead. This is an extension column, and we intend to propose it upstream.

## Tests

`TestSdrfBuilder`:
- `ASearchedDerivativeIsNamedBesideTheAcquiredFile`: adjacency, both cells, the no-transformation row, and the validator.
- `NoSearchedDataFileColumnUnlessARowSetsOne`.

120/120 `TestSdrf*` pass locally.

**Consumer:** MetaMorpheus SDRF output (#2816), which today writes the `-calib` name into `comment[data file]` after a Calibrate → Search chain.


nbollis (approved):

Alexander-Sol (approved):
**Review verdict: Approve with nits** — correct, minimal, well-documented, preserves byte-identical output for existing callers, and is properly tested.

**Summary:** Adds opt-in `SdrfAssay.SearchedDataFileName`, emitted to a new `comment[searched data file]` column placed immediately after `comment[data file]`. When a search read a derivative (e.g. `X.raw` → `X-calib.mzML`), the new column names what was actually searched while `comment[data file]` keeps naming the acquired raw file (the deposition join key). The column is emitted only when at least one row sets the value; untransformed rows re-state their acquired file name rather than leaving the cell blank.

**Findings (none blocking):**
1. *(nit)* The searched-cell fallback writes `assay.DataFileName` raw, whereas the `comment[data file]` cell one line above routes through `Required(...)`. Failure scenario: the column exists (another row set it), this row's `DataFileName` is blank, and `RequireSampleMetadata=false` → `comment[data file]` becomes `"not available"` but `comment[searched data file]` becomes an empty cell (an `EmptyCell` warning + a two-cell mismatch). Narrow, since `DataFileName` is normally non-blank. Cheap fix: route the fallback through `Required(...)` too.
2. *(nit)* `SearchedDataFileName` is declared non-nullable and emits CS8618; the logic treats null as the normal "not set" state, so `string?` would be more honest (consistent with the rest of the record).
3. *(observation, not a defect)* Re-stating `DataFileName` in the searched column for untransformed rows is a deliberate, documented choice (disambiguates "not set" from "searched the raw file").

Verified correct: header and cells are both gated on a single document-wide bool, so rows stay aligned (no ragged-row risk); adjacency to `comment[data file]` holds; output is byte-identical when the column is unused; the column name is validator-clean; the round-trip needs no reader change (header-driven).

